PR #841, opened 2014-07-02
Select all projects without URL length limit to handle many projects and...

jenboi:
... paging


tarzanek:
#892  is trying to fix the same thing ...

Issue #609, opened 2013-05-22
cant pull revision of file or annotate if repo requires auth (Bugzilla #7830)


vladak:
status REOPENED severity *major* in component *webapp* for *---*
Reported in version *unspecified* on platform *i86pc/i386*
Assigned to: Trond Norbye

On 2009-03-31 14:43:29 +0000, John wrote:

> opengrok hangs when annotating a file, or when pulling a specific revision of the file if the backend repository uses a tls connection and requires authentication (https with basic auth)

On 2009-04-01 08:14:42 +0000, John wrote:

> a localized file:// type repository solved this problem.

On 2010-11-02 01:05:54 +0000, Tony Butt wrote:

> I second this - I had configured opengrok to scan 2 subversion repositories using remote access via http://localhost to try to isolate the interfaces to our repositories better.
> 
> I have history working using jdbc, but annotate or selection of a prior revision from the history would hang indefinitely.
> 
> Snooping the traffic using wireshark showed that authentication was being requested but not handled. There is a password stored for the tomcat user that should have worked, but apparently the username was not being passed through, or there was some other difficulty.

PR #882, opened 2015-02-02
#609: Allow specifying username and password when fetching SVN history

ralphmayr:
Allow users to specify the svn username and password used when fetching the SVN history via environment variables.


tarzanek:
nice
can I ask for one more favour?
Could you document this at least inside OpenGrok script header please?
(or if it's worthwhile to mention in README, please do so)

Otherwise I am ready to accept, you have an agreement if I remember correctly, so no other roadblocks

cheers
L

fwiw - long term this code should be gone and this should be part of project setup ...

ralphmayr:
No problem - just added some documentation in the script. Is that ok for you? (I'm not sure if we should describe that in some more detail maybe?)

tarzanek:
I think this is ok, thank you for the patch!

PR #883, opened 2015-02-10
Fix POSIX-compliance issue in OpenGrok

jethrogb:
POSIX sh/test uses a single equals sign for string comparison.


tarzanek:
thanks for this fix, I am not sure if you have OCA or not - http://www.oracle.com/technetwork/community/oca-486395.html
either you mail them per above process
or for such a small change I am fully OK with you saying you agree with OCA for this pull request :)

jethrogb:
No creativity was required to fix this bug, so the patch is not a creative work, and so there is no copyright on this patch.

tarzanek:
:-D so if that means yes, you agree, I will integrate this patch
thank you!
(note that ALL code and changes are under copyright ... lawyers "suck", I know)

PR #884, opened 2015-02-10
Intelli window

harryxp:


vladak:
It seems that the change have broken some tests:

```
Failed tests:

JFlexXrefTest.unterminatedHeredoc:346 expected:<..."/source/s?defs=cat"[]>cat</a> &lt;&lt; EO...> but was:<..."/source/s?defs=cat"[ onmouseover="onMouseOverSymbol('cat', 'undefined-in-file')"]>cat</a> &lt;&lt; EO...>

PhpXrefTest.basicTest:18 expected:<..."/source/s?defs=foo"[>foo</a> <a href="/source/s?defs=bar]">bar</a>> but was:<..."/source/s?defs=foo"[ onmouseover="onMouseOverSymbol('foo', 'undefined-in-file')">foo</a> <a href="/source/s?defs=bar" onmouseover="onMouseOverSymbol('bar', 'undefined-in-file')]">bar</a>>

PhpXrefTest.sampleTest:84 expected:<...f="/source/s?defs=a"[ onmouseover="onMouseOverSymbol('a', 'undefined-in-file')"]>a</a> <span class="...> but was:<...f="/source/s?defs=a"[]>a</a> <span class="...>
```

harryxp:
I'm trying to set up a development environment so I can fix the test cases.  I'm using eclipse and I've resolved all the compilation errors already.  However when I use JUnit a lot of the cases fail on `org.opensolaris.opengrok.util.TestRepository`.

For example when I run JUnit on `org.opensolaris.opengrok.analysis.JFlexXrefTest` it dies on line 75:

    repository.create(JFlexXrefTest.class.getResourceAsStream(
                "/org/opensolaris/opengrok/index/source.zip"))

where `repository` is a `TestRepository`.  I seem to miss stuff like `source.zip` as well.  Please advise.  Thanks!

tarzanek:
use netbeans/mvn, check travis yml file for package which are required for tests to pass and setup test repo
I don't know how much Eclipse will follow ant file, but mvn should pull all needed stuff (besides OS dependencies , so do check the yml, resp. I think it only does:
 - sudo apt-get install exuberant-ctags cvs git mercurial cssc bzr subversion )
hth
L

tarzanek:
with those OS dependencies the mvn/ant tasks that prepare the source.zip should kick in ... obviously this will work well on systems like s11 / ubuntu
on windows you're really on your own and need to setup ALL manually and override binary commands (using wrappers, or system properties env vars like e.g. -Dorg.opensolaris.opengrok.history.Mercurial=C:\Program Files\Mercurial\hg.exe" )

harryxp:
Hmm, I am on Debian jessie so I have access to apt-get.  Was trying netbeans this morning with no luck.  I encountered this http://comments.gmane.org/gmane.comp.java.ide.netbeans.user/183988 during the installation and despite the suggestion on this page the problem still persists.  Maybe I'll try netbeans on Mac later.  Anyway, this'll definitely cost me more time than expected.

tarzanek:
download nb from their page as zip
make sure to set in its conf file the jdkhome to Oracle jdk
start nb
open nb project
voila !
:)
xing the fingers
L

tarzanek:
but then you don't really need it
run ant test
or mvn to get tests running

from nb though you can get single test run and test debug out of box ...

harryxp:
Thanks for the tips.  I believe all the test failures are fixed.

Took me a while to set up netbeans on Debian.  With xmonad WM, it just starts a blank window with nothing in it.  _JAVA_AWT_WM_NONREPARENTING=1 fixed it.

tarzanek:
OK, even Travis is now happy
and now the hard part - which combos of OSes + browsers were tested?
I'd expect at least:
S11 + Firefox
Windows 7 (8, 10?) + Firefox / Chrome / Opera  (IE ?  Spartan?)
Ubuntu/Debian  + Firefox/Chrome
(so I know which not to test anymore :) )
if above work fine, I am OK with accepting this patch + I'd need you to get OCA ( http://www.oracle.com/technetwork/community/oca-486395.html ) signed from you (just print, sign, scan and send to that mail, you can ping me once you send the mail, so I can expedite the process) 
Thank you for outstanding work!

harryxp:
Thanks.  OCA signed and sent.  I added a tiny improvement.  Going to test some browsers.

tarzanek:
I have your OCA now
once you have the testing done I am happy to accept this, I think this might be the lead feature of next release!

harryxp:
@tarzanek Thanks!  I'm glad you guys are happy to accept it.  I did find some issues with IE under Windows 7 (haven't used IE for a long long time...) so let me see what I can do with that...

tarzanek:
only test with latest IE on win 7
it makes little sense to test old IEs

harryxp:
I've tested Chrome/Firefox on Debian, Chrome/Firefox/IE on Windows 7 and 8.1.  The only problem is IE on Windows 7.  The weird thing is that it works for IE 11 on Windows 8.1 but not on Windows 7.  My Windows 7 machine is extremely slow so I'd hope someone with a better environment to debug it.  It's all in utils.js.

I have no Solaris so I can't help with that, either.

Thanks!

tarzanek:
thank you, I will merge this and retest, if you will eventually figure it out please open a new pull request

web/utils.js
```diff
@@ -439,3 +439,173 @@ function checkEnter(event) {
...
+            highlightSymbol(symbol);
+        } else {
+            unhighlightSymbol(symbol);
+        }
+    }
+    return true;
+});
+
+function onMouseOverSymbol(symbol, symbolType) {
+    updateIntelliWindow(symbol, symbolType);
```
dobrou: Symbol highlighting feature is fantastic!
But I'm afraid nobody will ever discover this feature is hidden under 1,2 key press.

What about automatic highlighting symbol under cursor:
function onMouseOverSymbol(symbol, symbolType) {
    updateIntelliWindow(symbol, symbolType);
    
    //highlight symbol under cursor
    unhighlightAll();
    highlightSymbol(symbol);    
}

I tested this change and it works really nice.
It's not much annoying for user that don't care about this and it's very helpful for anyone trying to read the code in browser.

harryxp: Hi Martin,

I'm glad you like it.  I, too, have been thinking how to make users aware
of this feature :-)

I'm afraid that highlighting symbol under cursor automatically would be
distracting.  For example, if I want to highlight a symbol, now I need to
be very careful not to point to any other symbols on my way of moving my
cursor towards the symbol I care about; I also need to be careful when
moving my cursor away.  This can be hard if my target is buried in a
cluster of code.

To advertise this, we can add this to the feature list on the OpenGrok
website, to the release notes, and to the help page.

dobrou: 1) I use it now with OnMouseOver and I must say that problems you describe are true, but not very big deal. I didn't have any big problems highlighting exactly what I want. 
Sometimes it is just blinking too much.
Still it can be disturbing for user not interested in highlighting.

So let's improve it a bit:
1a) OnMouseOver could be improved with short delay - mouse have to stop for ~1s on symbol to highlight it. Highlight is removed as soon as another symbol is highlighted.
I think this solves all problems.

2) I propose this solution because all users will be able to notice the feature without reading any documentation.
Unfortunately I'm sure that from all users of OpenGrok server I installed, I'm the only one who have ever read help page/release notes/documentation.

dobrou: Just another idea how to automatically use symbol highlight - if user came to file view from search page - searched words could be highlighted immediately.
At least for basic search types (regex,proximity would be too complex.)

This would need adding words to href links from search page. Classes around search context/summarizer should be able to provide list of words from search query.

--
And thanks again for this really nice feature!
Please don't take this as pushing feature requests to you, I'd just like to share ideas. And hope that somebody coming around has power and time to make it reality :)

PR #888, opened 2015-02-16
add RSS feed autodetection

### LINKED ISSUE #885: rss feed should be discoverable
Only by a chance we found out that there are RSS feeds available e.g. by URLs like `http://grok/source/rss/opengrok-master/pom.xml` and generated by `rss.jsp`. This seems to be working for file history only (no directories).

The history page should provide RSS auto discovery as per http://www.rssboard.org/rss-autodiscovery

vladak:
fixes #885


vladak:
I ran the build on the latest cset with Jenkins (on Solaris) and everything passed.

tarzanek:
push!

PR #892, opened 2015-02-16
Search all exisitng projects by wildcard

dobrou:
Add possibility to search all existing projets just by url param
"...&project=*" instead of naming all projects one by one.

Useful for linking OpenGrok from external sites/apps, so you don't have to update current full list of projects.


vladak:
Somehow I am missing a UI element for this change. How would users know to use the link ?

dobrou:
This is intended to be used when linking from external apps, where it's impossible to construct url with current list of projects.
In OpenGrok UI, user can always select all current projects from projects list.

How would users know to use the link ?

If you think it will be useful, we can add one special line into projects list: 
<option value="*">(All Projects)</option>

Or we can implicitly use 'project=*' in case all existing projects are selected.
What do you think?

vladak:
The issue with these private variables is that they remain in source only and very likely you will be the only consumer of it unless it is documented somewhere.

The latter option sounds good to me.

tarzanek:
I vote for second option (so javascript magic) :)
(All Projects) will just be confusing and breaking existing design

dobrou:
I made mistake with branching with this pull request. I had to create new OpenGrok fork, with new branch with changes.
So I will close this pull request and I will create new one, once suggested improvements are ready.

Sorry for confusion.

Issue #679, opened 2013-10-28
regexp search only shows files, not line#s.


tad44:
Using 0.12-rc1.
When searching with regexp, the result only shows the files, not the line#s with the match.
It would be nice if it can show the line#s also.

Thanks for the very useful tool.

dobrou:
Hi,
with 0.12.1 version I don't see any line preview when I search with regex.
It would be nice to be able to use both features together.

And thanks for the very useful tool too!

PR #893, opened 2015-02-17
Show words in lines context for Regexp search

dobrou:
Simple implementation adding context lines with searched words for
Regexp search.
Java Regexp is used to find words on lines, so any differences to Lucene
Regexp are ignored.

This is easy solution, until there is something more robust 100% compatible with Lucene regexps. But for most cases this should work.

TODO: Add some unit tests, as soon as I will find right place for them in existing tests.

### DIFF
--- a/src/org/opensolaris/opengrok/search/context/QueryMatchers.java
+++ b/src/org/opensolaris/opengrok/search/context/QueryMatchers.java
@@ -34,6 +34,7 @@ import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.PhraseQuery;
 import org.apache.lucene.search.PrefixQuery;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.RegexpQuery;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.WildcardQuery;
 
@@ -89,9 +90,19 @@ public final class QueryMatchers {
             getTerm((TermQuery) query);
         } else if (query instanceof PrefixQuery) {
             getPrefix((PrefixQuery) query);
+        } else if (query instanceof RegexpQuery) {
+            getRegexp((RegexpQuery) query);
         }
     }
-
+    
+    private void getRegexp(RegexpQuery query) {
+        if (useTerm(query.getField())) {
+            String term = query.toString(query.getField());
+            term = term.substring(1, term.length()-1); //trim / from /regexp/
+            matchers.add( new RegexpMatcher(term, true) );
+        }
+    }
+    
     private void getBooleans(BooleanQuery query) {
         BooleanClause[] queryClauses = query.getClauses();
         for (int i = 0; i < queryClauses.length; i++) {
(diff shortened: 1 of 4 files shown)


dobrou:
This is fix for issue #679

tarzanek:
great stuff!
looks ok to me, I will try togive you a pointer to junit tests
also before accepting I need you to accept OCA - http://www.oracle.com/technetwork/community/oca-486395.html
ideally if you can print, sign and mail the scan back to the mail in the web page
alternatively I am ok with you saying here that you accept OCA for this patch (but since you have more patches in pipeline, signing OCA would be the best long term)

tarzanek:
so in class ContextTest are some examples on how you can test your matcher ... ev. check WildCardMatcherTest

vladak:
+1 for the unit test

src/org/opensolaris/opengrok/search/context/QueryMatchers.java
```diff
@@ -89,9 +90,19 @@ private void getTerms(Query query) {
             getTerm((TermQuery) query);
         } else if (query instanceof PrefixQuery) {
             getPrefix((PrefixQuery) query);
+        } else if (query instanceof RegexpQuery) {
```
dobrou: Similar 'GetTerm' can be found in Summarizer.java,
what exactly does Summarizer? Does it makes sense to add RegexpQuery there too?

tarzanek: Yes
it needs to be there as well afaik ... I will recheck and let you know

tarzanek: confirming, you need to patch Summarizer too - it creates the short descriptions in the listing of found docs
(or long ones? )
anyways, it needs to be there, but it will be similar to what you did now
(I think Summarizer was the reason why I didn't fix this bug 2 years ago ... I run out of time on it and then didn't return to that patch since :( )

dobrou: Actually only current change in QueryMatchers.java is enough to see matched parts of documents on search results page - that is what I understand you describe as 'descriptions in the listing of found docs'.
However similar change in Summarizer seems safe, whatever it really does. I'll fix it too and try to find out it's real purpose :)
Thank you for helping me with the fix!

tarzanek: well I wasn't sure if Summarizer is used for the short listing, or for the longer listing when there is an ALL link in results
but it IS definitely used and based on your observation it seems to be the long listing
(so test is to create file which matches on more than 10 lines, then you get an ALL link from search results page for that doc hit ... )

dobrou:
In latest submit:
1) unit tests added
2) added common file header to new files
3) no changes in Summarizer, because:
- words to highlight are generated directly from query, without any context from searched document. So there is no reasonable way to do this for regexp.
- tested on running OpenGrok, and I get properly highlighted keywords both on search page and under [all...] link.
4) OCA is in progress, but until it's processed, I declare here that I accept OCA for this patch.

5) I'd like to get also proximity search results in context. Best approach would be to switch whole context generator to use Lucene highligher API, as is mentioned in several TODOs.
But this would surely deserve own pull request...

tarzanek:
yes, switching to lucene highlighter would be cool
if you could do it, a lot of people will be very happy ! :)
(and we could drop summarizer and have whole code base on CDDL only so my dispute with lawyers will become significantly easier :) )

dobrou:
Years updated.
OCA sent today.

Thank you for your help with pull request
It's my pleasure to pay back to this great project at least with this small fix :)

tarzanek:
No, thank you for fixing stuff I didn't manage to find time for ;)
+ all the new goodies.
Merging now, once I get the OCA, I will ping you on the other pull req

PR #895, opened 2015-02-17
fix placement of the RSS XML logo

### LINKED ISSUE #887: RSS feed icon missing
The RSS capability should be more visible to the user - i.e. place a link to the history page saying 'RSS feed'. This should be accomplished by the existing `rssi` CSS style (and `rss.png`) but I do not see it anywhere on the page.

The page source has this:

```
<p class="rssbadge"><a href="/source/rss/mercurial/" title="RSS XML Feed of latest changes"><span id="rssi"></span></a></p>
```

but it seems to have zero effect, at least on FF.

vladak:
fixes #887

### DIFF
--- a/web/default/print.css
+++ b/web/default/print.css
@@ -473,6 +473,6 @@ a.xsr	{ /* subroutine */		color: #00f; font-weight: bold; }
 	vertical-align: middle;
 	background-image: url(./img/combined.png);
 	background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


tarzanek:
push! :)
(wasn't there a third style ? )

vladak:
The 3rd style is okay, I checked the offset in the `combined.png`.

PR #898, opened 2015-02-19
unify RSS title for directory and file

### LINKED ISSUE #891: RSS feed links should point to particular changeset in history view
When displaying RSS feed all the links lead to contents of the file which is being tracked. Ideally (if possible) the links should point to particular changeset in history view of the file. It would be nice to be able to display the actual diff directly (just like Github feed does) but that's not possible yet.

### LINKED ISSUE #896: unify RSS feed elements for directory and file
RSS feed `description` field differs for files and directories. For directory it contains list of files changed in given changeset, for file it contains path and changeset ID.

The changeset ID could be displayed in the `title` (maybe with first line of the comments) and the full comments (with newlines) in `description`.

vladak:
fill revision in DirectoryHistoryReader
fixes #896
fixes #891

### DIFF
--- a/web/rss.jsp
+++ b/web/rss.jsp
@@ -56,17 +56,6 @@ include file="pageconfig.jspf"
     String path = cfg.getPath();
     String dtag = cfg.getDefineTagsIndex();
     String ForwardedHost = request.getHeader("X-Forwarded-Host");
-    String requestURL = request.getRequestURL().toString();
-    // Play nice in proxy environment by using hostname from the original
-    // request to construct the URLs.
-    // Will not work well if the scheme or port is different for proxied server
-    // and original server. Unfortunately the X-Forwarded-Host does not seem to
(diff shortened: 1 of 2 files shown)


web/rss.jsp
```diff
@@ -56,17 +56,6 @@ include file="pageconfig.jspf"
     String path = cfg.getPath();
     String dtag = cfg.getDefineTagsIndex();
     String ForwardedHost = request.getHeader("X-Forwarded-Host");
-    String requestURL = request.getRequestURL().toString();
```
tarzanek: you're backing out your old push ... is this OK ? ;)

tarzanek:
if you tested it and it works, then shit it !
(perhaps junits might come handy to put guards to your code? ;) )

vladak:
funny typo, hopefully the change will not cause so much damage ;-)

PR #901, opened 2015-02-21
Patch highlight color for IE.

harryxp:
Comparing colors is different for each browser.


tarzanek:
Thank you!
Looks good, merging.

PR #903, opened 2015-03-07
Improve Perforce files history&annotate

dobrou:
Improve files version in Perforce repository:
Switch from '#file revision' to '@change number' as files version id.

Why?
'@change number' is version counter of whole repository. It's changelist ID, increased with each submit to repository.
'#file revision' - is version counter per file. Increased by 1 with every change on file. And is reset to 1 when file is moved/integrated!
(Description of #@ here http://www.perforce.com/perforce/r12.2/manuals/cmdref/o.fspecs.html)

That means with current #:
 - versions of one file are not always distinct
 - OG can show file version history like: 1,2,3,4,1,2 - (this completely ruins annotate feature on any moved/integrated file)
 - searching by changelist number does not work
 - directories use @ as version, so it is not consistent with files
 + the only advantage is that on annotate page, file versions are low numbers, so it's easy to compare them visualy 

After change to @:
 + file versions are distinct, so annotate is correct on moved/integrated files
 + searching for @ changelist id works
 + dirs/files have consistent versioning

----
Remaining issues (they are not introduced by this change):
1)
Getting version of moved/integrated file, before move/integration does not work.
Because old version is get from P4 on latest path, which is not valid before move/integration.
2)
I don't understand why, but storing history into DB with option '-D' still doesn't work.
It fails with:
WARNING: An error occured while creating cache for ... (PerforceRepository)
org.opensolaris.opengrok.history.HistoryException: java.sql.SQLIntegrityConstraintViolationException: The statement was aborted because it would have caused a duplicate key value in a unique or primary key constraint or unique index identified by 'CHANGESETS_REPO_REV_UNIQUE' defined on 'CHANGESETS'.
        at org.opensolaris.opengrok.history.JDBCHistoryCache.store(JDBCHistoryCache.java:677)
        at org.opensolaris.opengrok.history.Repository.createCache(Repository.java:328)
        at org.opensolaris.opengrok.history.HistoryGuru.createCache(HistoryGuru.java:496)
        at org.opensolaris.opengrok.history.HistoryGuru.ensureHistoryCacheExists(HistoryGuru.java:693)
        at org.opensolaris.opengrok.index.IndexDatabase.update(IndexDatabase.java:357)
        at org.opensolaris.opengrok.index.IndexDatabase$1.run(IndexDatabase.java:176)
I thought making files version distinct with this fix should help, but it's not enough.
I may create separate issue for this, if we can't solve it somehow easily as part of this pull request.

### DIFF
--- a/src/org/opensolaris/opengrok/history/PerforceHistoryParser.java
+++ b/src/org/opensolaris/opengrok/history/PerforceHistoryParser.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2015, Oracle and/or its affiliates. All rights reserved.
  */
 
 package org.opensolaris.opengrok.history;
@@ -30,7 +30,6 @@ import java.io.Reader;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -87,7 +86,7 @@ public class PerforceHistoryParser {
         ArrayList<String> cmd = new ArrayList<String>();
         cmd.add("p4");
         cmd.add("filelog");
-        cmd.add("-lt");
+        cmd.add("-lti");
         cmd.add(file.getName() + PerforceRepository.getRevisionCmd(rev));
         Executor executor = new Executor(cmd, file.getCanonicalFile().getParentFile());
         executor.exec();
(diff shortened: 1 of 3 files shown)


vladak:
Did the JavaDB history store fail before this change ? Maybe you can check the uniqueness of the history entries w.r.t. changeset IDs in the beginning of the `JDBCHistoryCache.java#store()` with some sort of assert.

tarzanek:
I merged this with https://github.com/OpenGrok/OpenGrok/commit/7aefe4b6f83081089f8a0e467b9fbca47d5b3e70

tarzanek:
@dobrou 
thank you very much for this, I merged it, we have your OCA on file
cheers and vdaka! :)

PR #904, opened 2015-03-07
Add Scopes feature for CSharp

dobrou:
I followed Scopes implementation for C/Java and added support for CSharp.

Lex parser without any modification should work in 99% cases.

Known issues:
- generic methods with type constraints are not recognized.
Eg. private T M4<T>(T p) where T : new()
Text 'where T' between ) : confuse the parser

--
We could easily add same support for Javascript, maybe more languages.
Do we want always copy the same .lex file (as it was done for C/Java), or should we share one base for all common languages?

--
I'd like to propose 2 simple patches to Scopes ui, if you and especially author of Scopes agree, I would:
- add at least word 'in ' before scope name in search context - for regular user it may not be clear what the 'name()' really means. 
- blue is nice, but too much visible compared to search results, 'darkgray' works good for me. Scopes are helpful, but search results are still more important.


vladak:
Unit test with some C# samples would be nice. Can you post some screenshorts for the UI change ?

vladak:
as for the Scope patches, ping @mrate .

mrate:
Personally I don't have problem with either of changes. I believe word "in" would not hurt and I don't have any opinion on color. I just "randomly" selected color that was significant and didn't look completely wrong :-) As @vladak said, it would be nice to see a screenshot.

Issue #902, opened 2015-03-07
Error when sorting search results (latest git build)


dobrou:
Regular search sorted by relevance works OK.
But when I click "Sort by: last modified time" or "Sort by: path", OG returns following error instead of search results:

java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: unexpected docvalues type NONE for field 'date' (expected=SORTED). Use UninvertingReader or index with docvalues.
    at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:526)
    at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:490)
    at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:313)
    at org.opensolaris.opengrok.web.SearchHelper.executeQuery(SearchHelper.java:340)

**Version: latest build from GIT (7.3.2015)**

It may be related with upgrade of Lucene to v5 according to [RE: Lucene 4.x -> 5 : IllegalStateException while sorting](http://mail-archives.apache.org/mod_mbox/lucene-java-user/201502.mbox/%3C022501d04f88$421a97d0$c64fc770$@thetaphi.de%3E)

dobrou:
http://lucene.apache.org/core/corenews.html
> FieldCache is gone (moved to a dedicated UninvertingReader in the misc module). This means when you intend to sort on a field, you should index that field using doc values, which is much faster and less heap consuming than FieldCache.

vladak:
Cannot reproduce it anymore on 0.13-rc10 so the fix worked, closing.

PR #905, opened 2015-03-07
Fix sorting by date/fullpath

dobrou:
Fix for #902 

It's my first touch on lucene, so I hope it's arranged correctly.
But sorting by date/path started to work again with this fix.


tarzanek:
thanks, good stuff, merging

tarzanek:
fwiw, it would be cool to have a junit test for this ...

tarzanek:
hmm, we have SortOrderTest , but it kinda needs a serious improvement :(

PR #906, opened 2015-03-08
Annotate - highlight recent file revisions lines

dobrou:
On annotate page - color file revisions numbers background, according to their age, so that recent revisions are highlighted more.

This way it's easy to visually distinguish new/old parts of file, without exactly reading revision ids.

Preview here:
opengrok-annotate-highlight

I will welcome any feedback, especially for css part. This could cooperate better with regular css styles in OpenGrok.

### DIFF
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -141,6 +141,7 @@ public final class HistoryGuru {
             }
             if (hist != null && ret != null) {
              Set<String> revs = ret.getRevisions();
+             int revsMatched = 0;
              // !!! cannot do this because of not matching rev ids (keys)
              // first is the most recent one, so we need the position of "rev"
              // until the end of the list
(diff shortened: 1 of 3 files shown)


src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -154,6 +155,8 @@ public Annotation annotate(File file, String rev) throws IOException {
                     ret.addDesc(short_rev, "changeset: " + he.getRevision() +
                         "\nsummary: " + he.getMessage() + "\nuser: " +
                         he.getAuthor() + "\ndate: " + he.getDate());
+                    ret.addFileVersion(short_rev, revs.size() - revsMatched); //history entries are coming from recent to older, file version should be from oldest to newer
```
vladak: pls keep the lines short

vladak:
Couple of comments:
  - the color should not be hardcoded, it should be defined in the CSS styles so that users can override it
  - I am torn w.r.t. hundreds of shades. An alternative would be to display just a couple of (tens ?) shades and the rest leave it at default color. This way it would be easier to identify changes to adjacent lines - say if some lines in the same neighborhood were changed by couple of recent changesets it would be hard to identify which lines were changed by which changeset since the shade difference would be negligible.

vladak:
Any updates ?

vladak:
Even if you are not going to make any changes, can you accept OCA pls ?

vladak:
pushed as c904440d32e181d0920989f48a1c903c329b4b36

dobrou:
No updates, unfortunately I didn't have much more free time to spend on this yet.
If somebody takes this as feature request including prototype, I could not have any complaints. Otherwise I will finish my work as soon I'll have some free evenings :)

Yes, my OCA was signed and delivered already.

vladak:
Don't worry, it is now sitting in master, @tulinkry will give it some tuning.

PR #908, opened 2015-03-10
Fix ctags Scala support.

harryxp:
Fix the escaping issue.


tarzanek:
thank you very much sir! :)

PR #910, opened 2015-03-18
Add Erlang analyzer

trondn:
This patch is a modification of the patches created by
Haitao Li <lihaitao@gmail.com> back in 2011 and made available
at https://bitbucket.org/liht/opengrok/branch/erlang-analyzer

The OCA is accepted


lht:
I accept and am OK with OCA for this patch.

tarzanek:
great, I guess we only miss some junits, but that can be done later ... accepting, since Travis passed
@trondn  @lht will one of you be available to write a junit for the analyzer so it's covered?

tarzanek:
hmm, I accepted an we miss copyrights, ok will add them later
...

tarzanek:
hmm, one more thing we forgot - add this to SearchHelper.java
into filedescriptions ... will fix with copyright ...

tarzanek:
see #912

tarzanek:
alt. I should rewrite in how we generate the filedescriptions - they should come from analyzers, this data should be isolated inside analyzer and not taken from helper class (but helper class should build it from the known analyzer list)

PR #912, opened 2015-03-19
Haskell support

harryxp:
Thanks!

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
+++ b/src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
@@ -18,7 +18,7 @@
  */
 
 /*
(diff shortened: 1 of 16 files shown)


src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
```diff
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2010, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2015, Oracle and/or its affiliates. All rights reserved.
```
tarzanek: 2010, 2015
in 2005 there was no perl analyzer, it was first public release with just C and Java

tarzanek:
Good stuff and thank you sir!
We again miss junit, but I am OK with that, to the gate with this patch! :)

PR #914, opened 2015-03-23
fallback to newer file if genre for older file is null

### LINKED ISSUE #832: diff of plaintext file between empty and non-empty version fails with 'Diffs for binary files cannot be displayed!'
Receiving this error "Diffs for binary files cannot be displayed! Files are test1(revision 5) and test1(revision 6)." when trying to compare the history of a config file.

File can be viewed on its own and error message does link to direct download of the file.

Centos 6
Opengrok-0.12.1 from source.
Tomcat 7
SVN

vladak:
fixes #832


tarzanek:
looks ok, push

Issue #915, opened 2015-03-24
Improved context (scope) functionality


mrate:
With integration of enhancement #897 "Add scope context to search results" a new functionality can be implemented:
- dynamically show name & signature of currently displayed function when scrolling through xref source page (function head and body can be encapsulated in div)
- add support for function folding/unfolding in xref source page (controls visibility of function body div)

tulinkry:
Already implemented

PR #916, opened 2015-03-24
Scopes v2

mrate:
Implementation of enhancement #915 Improved context (scope) functionality


tarzanek:
got a demo system in OWAN? ping me on IM where I can play with above

tarzanek:
looks lovely
small comments on UI though - I think I'd make the size smaller, ev. place it somewhere aside (to not consume height, but rather width)
but then I am no UX expert ... 
another thing is interning of signature - check on big .c files that it doesn't leak mem - ip.c , mpt.c, check ip.c in on10-patch gate
+ since this is UX change, test chrome, ff, ie, opera ideally on windows, ubuntu, solaris 11

tarzanek:
yes + check w3 validator , I probably have a local copy of w3 validator running on sun-n1, in case you want to validate internal sources (e.g. the ip.c )

tarzanek:
( http://validator.w3.org/ )

vladak:
+1 on smaller text

I'd probably move the scope to the left (so that it is right above the source code - avoids unnecessary eye movement) and maybe place it on some colored background.

mrate:
I did some UX changes, decreased font size, moved scope text to left and added some background color. I tested it on Ubuntu - Firefox, Chrome, Opera; Windows 7 - IE9, IE11, Firefox, Opera; OS-X - Safari. Everything looks good except IE*. In both Internet Explorer versions (9, 11) the code folding is broken and page layout is quite broken.

tarzanek:
IE is tricky, most probably there are some hacks how to make it work properly
we should at least get IE11 working somehow ...

mrate:
There was a problem with IE badly interpreting empty <span class="fold-icon" /> element, it should be fixed now. Tested on IE11 on Win7 and it looks and behaves exactly same as on Firefox / Opera / Chrome.

mrate:
Now also tested on Firefox on Solaris 12, seems ok :-)

tarzanek:
ok
this is the main reason I asked for w3c validator ...
so try to use the online one with code that is public

or if you have internal source indexed only I can start an internal w3c validator as a service
(though I would prefer you test with external validator and be careful to not leak any internal source ...)

validator would see the unclosed span and report it ...

mrate:
w3c seems to be (almost) valid, I added the possibility to turn off source code folding in configuration and made some small UI improvements.

It looks like onscroll div element was introduced in DOM level 2 and is not part of XHTML 1.0 (that's the only error in w3c)... I need to find out how to get over this one.

tarzanek:
we can eventually go to higher XHTML, though someone needs to asses the impact and what browsers will be ruled out

tarzanek:
@mrate care to finish this?
I don't have a prob to go to higher xhtml

mrate:
@tarzanek Sure.

tarzanek:
Thank you sir, merging!

PR #918, opened 2015-03-24
change behavior of tunable

vladak:
and also rename OPENGROK_REMOTE_REPOS variable in OpenGrok shell script.


tarzanek:
looks ok to me, please push

PR #919, opened 2015-03-24
fixes #913 - Haskell unit test

### LINKED ISSUE #913: add unit tests for Haskell
Haskell support was added (pull req #912), it needs some unit tests.

harryxp:
This was harder than I thought, mainly because Haskell identifiers can have single quotes in them.  It messed up Intelli Window JS code so I had to fix that.


harryxp:
The error is caused by HaskellXrefTest and HaskellSymbolTokenizerTest not being able to load org/opensolaris/opengrok/analysis/haskell/sample.hs.  I modeled them after the PHP counter parts.  The weird thing is that the pull request does contain sample.hs and all test cases pass on my machine.

After some research, I suspect I need to add <testResources> in pom.xml for Haskell.  Is that it?  Thanks.

tarzanek:
yes, correct
check opengrok-indexer/pom.xml

section 
 <testResource>
                <targetPath>org/opensolaris/opengrok/analysis/php/</targetPath>
                <directory>../test/org/opensolaris/opengrok/analysis/php/</directory>
                <excludes><exclude>*.java</exclude></excludes>
            </testResource>
I think you will need similar for your analyzer

tarzanek:
mergin, thnx

PR #920, opened 2015-03-25
make extended hangling of renamed files proper configuration option and ...

vladak:
...enable it by default


tarzanek:
push at will

PR #927, opened 2015-04-04
Fix #690 Navigate when line numbers are hidden.

### LINKED ISSUE #690: Navigate doesn't work without Line#
Using 0.12rc2.
Not sure if this is intentional, but, it seems navigate to function, etc. doesn't work when Line# is not displayed.

Thanks,

harryxp:
Disclosure: I only tested this on Linux/Chrome.

### DIFF
--- a/web/utils.js
+++ b/web/utils.js
@@ -1084,16 +1084,6 @@ function lnshow() {
     document.line_numbers_shown = 1;
 }
 
-function lnshow() {
-    $("a").each(
-        function() {
-            if (this.className == 'l-hide' || this.className == 'hl-hide') {
-                this.innerHTML = this.getAttribute("tmp");
-                this.className = this.className.substr(0, this.className.indexOf('-'));
-            }
-        });
-}
-
 /* ------ Highlighting ------ */
 /**
  * An expensive Highlighter:
--- a/web/utils.js
+++ b/web/utils.js
@@ -980,7 +980,7 @@ function get_sym_list_contents() {
         for (var j = 0; j < symbols.length; j++) {
             var symbol = symbols[j][0];
             var line = symbols[j][1];
-            contents += "<a href=\"#" + line + "\" class=\"" + class_name + "\">"
+            contents += "<a href=\"#" + line + "\" class=\"" + class_name + "\" onclick=\"lnshow(); return true;\">"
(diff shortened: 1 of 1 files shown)


web/utils.js
```diff
@@ -279,6 +279,16 @@ function lntoggle() {
         });
 }
 
+function lnshow() {
```
tarzanek: I think we had such function already ... care to check and dedup?

harryxp:
@tarzanek Sorry for the (long) delay.  I duplicated some code because otherwise it's kinda annoying - the state of whether line numbers are currently shown or hidden is embedded IN EACH LINE NUMBER LINK.   lntoggle() function simply toggles that state for each link.  However, to force the line numbers to show when an navigation link is clicked, we need to know the current state of line numbers.

I refactored the code to remove the duplication, but the code is actually longer and a flag is introduced...that's why I hesitated to do this.  But anyway it's still simple so I'm happy with either solution.

tarzanek:
looks ok, however I'd love to get a second pair of eyes to have a look, @tulinkry  can you please review too?

web/utils.js
```diff
@@ -263,18 +263,44 @@ function lsttoggle() {
...
+            if (this.className === 'l-hide' || this.className === 'hl-hide') {
+                this.className = this.className.substr(0, this.className.indexOf('-'));
+                this.innerHTML = this.getAttribute("tmp");
+            }
+        }
+    );
+    document.line_numbers_shown = 1;
+}
+
+function lnshow() {
```
tarzanek: 2x lnshow ? see line # 286

harryxp:
@tarzanek Oops.  Fixed.
I noticed that a new line-number-related issue has been introduced - if you jump to a definition using the navigation panel, causing a line number to be highlighted, then you close the line numbers using "Line#", the highlighted number is still there.

The behavior seems to be introduced by this commit:
22e51bd 2015-10-14 | Line highlighter [Kryštof Tulinger]

tarzanek:
thnx @harryxp 
@tulinkry  - seems you have a regression, can you please fix?

tarzanek:
@harryxp thnx, I will merge this
we should cut a RC so we can start catching such bugs as @tulinkry introduced ...

PR #931, opened 2015-04-07
jdk8 , fix ctag definitions to work on solaris too

tarzanek:
oh yeah ... and emma is dead :(
long live jacoco!

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -439,7 +439,7 @@ Solaris 11:
 
 Debian/Ubuntu:
 
-  # apt-get install sun-java8-javadb
+  # apt-get install libderby-java
 
 Other:
 
--- a/README.txt
+++ b/README.txt
@@ -35,9 +35,9 @@ Offical page of the project is on:
 2. Requirements
 ---------------
 
-    * Latest Java (At least 1.7, 1.8 is supported)
+    * Latest Java (At least 1.8)
       http://www.oracle.com/technetwork/java/
-    * A servlet container like Tomcat (7.x or later)
+    * A servlet container like Tomcat (8.x or later)
       supporting Servlet 2.4 and JSP 2.0
       http://tomcat.apache.org/
     * Exuberant Ctags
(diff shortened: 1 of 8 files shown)


README.txt
```diff
@@ -439,7 +439,7 @@ Solaris 11:
 
 Debian/Ubuntu:
 
-  # apt-get install sun-java6-javadb
+  # apt-get install sun-java8-javadb
```
kahatlen: I don't think there's any package with that name in Debian or Ubuntu. There seems to be a libderby-java package, though.

tarzanek: only in jessie - https://packages.debian.org/jessie/libderby-java
jessie will be out this week(next week) so I will probably use that?

vladak:
generally I don't like mixing unrelated things in single commit - here you have emma + regexps + 1.8.

tarzanek:
yeah, I also don't like it, but they have the connection
- currently jenkins is broken, because of regexps (my jenkins is on solaris 12, ctags break due to regexps not recognized \s - fyi I switched to ctags 5.9 from svn so I am on same as on my linux box)
jdk8 is needed because of jenkins too :)
and emma is not working in jdk8

so to get jenkins working, all above is needed :-D

README.txt
```diff
@@ -35,9 +35,9 @@ Offical page of the project is on:
 2. Requirements
 ---------------
 
-    * Latest Java (At least 1.7, 1.8 is supported)
+    * Latest Java (At least 1.8)
       http://www.oracle.com/technetwork/java/
-    * A servlet container like Tomcat (7.x or later)
+    * A servlet container like Tomcat (8.x or later)
```
vladak: my gripe with this is that Solaris 11.2+ does not even have Tomcat 7.

tarzanek: and it will never have it :)
but it will have tomcat 8, so we're good
fwiw , I used tomcat 7 and tomcat 8 from zip and all just worked

README.txt
kahatlen: Since there is no sun-java8-javadb package in any Debian/Ubuntu version, and sun-java6-javadb is so old it's not even in Wheezy, I'd say pointing to the one in Jessie is the best option. And it's available in Ubuntu 14.04, which is already released.

PR #935, opened 2015-05-10
split ignoredNames into list for files and directories

### LINKED ISSUE #508: IgnoredNames.java should have two lists -- one for files, one for directories (Bugzilla #18601)
status CAUSEKNOWN severity *major* in component *indexer* for *---*
Reported in version *unspecified* on platform *ANY/Generic*
Assigned to: Vladimir Kotal

On 2011-06-22 19:33:16 +0000, Doug McLaren wrote:

> In the course of my figuring out why opengrok 0.10 wouldn't index all my files, I found two different problems that caused the issue.
> 
> The simplest problem was that opengrok wasn't indexing anything in our directories called "tags".  (Combined with the other problem, it wasn't immediately apparent that anything in a directory called "tags" was skipped, but eventually it became clear.)
> 
> (The other problem will be the subject of another defect.)
> 
> In any event, the ctags program will create a file called "tags", so it makes sense to ignore these files, but there's no reason that I'm aware of to ignore a directory called tags.
> 
> Therefore, I suggest that IgnoredNames.java have two lists rather than one -- one list is a list of files to be ignored, and the other list is a list of directories to be ignored.  (There may also need to be some other code changes -- when the code actually checks to see if a file name should be ignored, it might have to tell the method that this is a file or it's a directory -- but that should be pretty obvious.)
> 
> For my own use, I'm simply recompiling IgnoredNames.java to not skip anything named "tags", but a proper fix is likely what I suggested above.
> 
> Also, some documentation on this list might be appropriate, or perhaps some way to customize the default list without recompiling.

On 2011-07-21 13:28:44 +0000, Lubos Kosco wrote:

> Trond :) 
> 
> me is affected too! 
> and me not like it
> 
> what will do ?
> 
> :-D
> 
> but seriously, this is a prob, I wasted 2 hours since I couldn't find my semantic tag sources which live in a directory called tags ;)

On 2012-02-21 21:30:40 +0000, Vladimir Kotal wrote:

> We will need to figure out what to do with the -i option since it is also generic.

vladak:
fixes #508

### DIFF
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -316,7 +316,7 @@ public final class HistoryGuru {
     }
 
     /**
-     * recursivelly search for repositories with a depth limit
+     * recursively search for repositories with a depth limit
      * @param files list of files to check if they contain a repo
      * @param repos list of found repos
      * @param ignoredNames what files to ignore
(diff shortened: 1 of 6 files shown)


tarzanek:
nice, very nice
please push!

PR #942, opened 2015-05-20
display table with project-repository mappings on front page

### LINKED ISSUE #939: display list of projects with repository pairing on front page
Front page is too empty and can be used to display list of projects mapped to repositories.

vladak:
fixes #939

Couple of things to note:
  - the table does not contain projects without repo (e.g. directories with just bunch of files)
    - I can change that if wanted
  - the table spans the whole page (see attached screenshot)
  - this is displayed by default if running with projects
    - can add config variable if desired

projects-repos

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Project.java
+++ b/src/org/opensolaris/opengrok/configuration/Project.java
@@ -123,7 +123,7 @@ public class Project {
      */
     public static Project getProject(String path) {
         Project ret = null;
-        String lpath=path;
+        String lpath = path;
         if (File.separatorChar != '/') {
             lpath = path.replace(File.separatorChar, '/');
         }
--- a/web/index.jsp
+++ b/web/index.jsp
@@ -52,7 +52,13 @@ include file="menu.jspf"
 
 include file="index_body.html"
 
-        %></div>
+        %>
+        <%@
+
+include file="repos.jspf"
+
+        %>
+        </div>
 <%
 }
 /* ---------------------- index.jsp end --------------------- */
--- a/web/index_body.html
+++ b/web/index_body.html
@@ -1,10 +1,10 @@
 <p>
-<b>OpenGrok - a "wicked fast" source browser!</b></p>
+<b>OpenGrok - a "wicked fast" source browser</b></p>
 
 <p>
-This is the front page of your <a 
-href="xref/">browsable and searchable source tree</a> (<a 
-href="xref/">xref/</a>).</p>
+This is the front page of your
+<a href="xref/">browsable</a> and searchable source tree.
+</p>
 
 <p>
 To setup and configure OpenGrok please refer to
@@ -17,8 +17,4 @@ You can replace this block of text (in index_body.html) with some more useful
 information about your source tree and its organization, with direct 
 links to key parts of code base.
 </p>
-<br/>
-<br/>
-<br/>
-<br/>
 <br/>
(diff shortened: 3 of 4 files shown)


tarzanek:
LGTM, please push!

ymnick:
After installing a new OpenGrok version I see the following columns headers but nothing gets displayed under it, is there a new configuration I am missing supporting this new table?

Project SCM type Parent (branch)

vladak:
Do you have projects enabled ?

ymnick:
Not sure I know how to answer. I have projects specified in C:\opengrok\configuration.xml like:

<void property="projects">
   <void method="add">
     <void property="description">
      <string>My Project</string>
     </void>
     <void property="path">
      <string>/My Project</string>
     </void>
    </object>
   </void>
...

vladak:
Weird. What version are you running ? (based on which changeset)

vladak:
Also, just to be sure - could you post the html source context of the table ?

vladak:
Plus, in your `configuration.xml`, I assume there is a repository matching `/My Project`.

ymnick:
I am running the local build forked from master a week ago with modified RuntimeEnvironment class to allow to run with Universal Ctags (identical to your change from yesterday). My repositories element in configuration.xml looks as follows, where p4_all is the source folder holding all the Perforce depots I index, 'My Project' being one of them:

\<void property="repositories">
   \<void method="add">
    \<object class="org.opensolaris.opengrok.history.RepositoryInfo">
     \<void property="directoryName">
      \<string>C:\p4_all\</string>
     \</void>
     \<void property="type">
      \<string>Perforce\</string>
     \</void>
    \</object>
   \</void>
  \</void>

vladak:
I can wrap the table printing code into the condition as you suggest, it would be safe thing to do. However, I wonder what is causing the map to be empty. My bet would be the `C:\` path somehow disrupts the assignments. It would be nice if you could debug this further as I do not have Windows environment here.

ymnick:
From \webapps\source\repos.jspf:

projects.size() = 110
env.getRepositories() = 1
r.getDirectoryName() = C:\p4_all
repoPath =

And from org.opensolaris.opengrok.configuration.RuntimeEnvironment.java:

canonicalPath = C:\p4_all
sourceRoot = C:\p4_all
canonicalPath.substring(sourceRoot.length() + 0) =

vladak:
If sourceRoot is indeed equal to canonicalPath in getPathRelativeToSourceRoot() then this is probably a strange configuration/setup. I'd assume that sourceRoot is always strict superset of any project/repo path, at least this is the common use case (I can imagine someone might want to clone a repository directly into source root, though). What does your configuration.xml say about the sourceRoot property ?

ymnick:
Not sure I understand why it's strange... I have Perforce repository with multiple depots and they all mapped to the corresponding subfolders under c:\p4_all:

 \<void property="sourceRoot">
   \<string>C:\p4_all\</string>
  \</void>

vladak:
It's probably just that I am used to the layout where all the projects (and hence repos) are nice and tidy under source root, each having it's own subdirectory. In this layout, the Perforce repo would be a subdirectory under source root and the depots sub-subdirectories.

I guess we need to explore the alternative layout more (or say it's not supported :-)).

ymnick:
I think I can see your point and if I would be using more than one repository it would probably make more sense to have it organized this way.

PR #946, opened 2015-05-26
escape opengrok jar file

### LINKED ISSUE #945: I should be able to install OpenGrok into a directory with spaces in the name.
When I explode OpenGrok into a directory with a space in the name (say, "Open Grok"), the indexer fails as follows:

    Loading the default instance configuration ...
    Error: Unable to access jarfile /Open

It shouldn't.

vladak:
fixes #945

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -129,8 +129,8 @@ else
 fi
 
 # TODO: Handle symlinks correctly (especially in ${0})
-SCRIPT_DIRECTORY="`dirname ${0}`"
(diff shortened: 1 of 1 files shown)


tarzanek:
please push, lgtm!
btw. there were some other script errors I have shellcheck running on sun-n1 so we can try to run it too
e.g. 
ssh sun-n1
cd <WS>
shellcheck OpenGrok

vladak:
thanks. feel free to file bugs for the remaining errors.

PR #951, opened 2015-05-28
Remove bashism in the OpenGrok script

kahatlen:
String equality should be tested with '=' and not with '=='. The
latter is an extension in bash. I'm seeing the following error on a
system where /bin/sh is dash:

/code/OpenGrok/OpenGrok: 717: [: Linux: unexpected operator


tarzanek:
lgtm, please push!

PR #959, opened 2015-07-01
READ_XML_CONF not set properly

gco:
-R $a was not valid, it should be referring to $READONLY_CFG_FILE

### DIFF
--- a/platform/solaris/smf/ogindexd
+++ b/platform/solaris/smf/ogindexd
@@ -53,7 +53,7 @@ read_config() {
 	fi
 
 	if [ -r "$READONLY_CFG_FILE" ]; then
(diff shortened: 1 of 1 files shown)


tarzanek:
looks good to me, before accepting I need either you sending OCA - http://www.oracle.com/technetwork/community/oca-486395.html over to Oracle
of for such a small patch the lawyers will have to be happy with you saying that you agree with OCA here
(e.g. "I agree with OCA for this patch" is enough ;) )

tarzanek:
OR if you're from Oracle from certain sparc team I have no problem accepting :)

tarzanek:
thank you !
merged now

PR #960, opened 2015-07-02
C++ identifiers within angle brackets should be treated as identifiers

abbeyj:
When indexing C++ source code, identifiers that are found within angle brackets are treated as filenames instead of identifiers.  For example, say you have the following source code:

    int main() {
        Wrapper<MyClass> x;
        return 0;
    }

When this gets processed, "MyClass" will link to "/source/s?path=MyClass".  Since MyClass is not a filename this search is not likely to return anything useful.  It should be linking to "defs" instead of "path".

The code was intended to handle things like "#include <filename.h>" but it activates in too many places.  The rule should be restricted so that it only applies following "#include".

My new code feels a bit hacky but I'm not familiar with JFlex and this was the best that I could come up with.  Suggestions for a better approach are welcome.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/c/CXref.lex
+++ b/src/org/opensolaris/opengrok/analysis/c/CXref.lex
@@ -30,6 +30,8 @@ import org.opensolaris.opengrok.analysis.JFlexXref;
 import java.io.IOException;
 import java.io.Writer;
 import java.io.Reader;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.opensolaris.opengrok.web.Util;
 
 %%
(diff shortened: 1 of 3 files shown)


tarzanek:
Do you think you could come up with junit tests for this problem?
I have also some concerns on how fast will the regexp matching do its job, but if the lexical rule is writen good enough, it shouldn't kick in unless needed ... 
so all in all it looks good to me, try to check the junit tests ...

I will also need to accept OCA from you - http://www.oracle.com/technetwork/community/oca-486395.html
Alternative is you accept OCA in this thread and say you accept it for this patch

tarzanek:
@abbeyj can you check my last comment and do the needful? I cannot do anything with this patch until then ...

abbeyj:
I'm still working on getting legal clearance.  I haven't forgotten but it is going to take a while longer.

tarzanek:
I got your OCA now, thank you.

What about some junit tests, do you think you can come up with some examples?
Have a look at existing tests, you can certainly reuse them.

abbeyj:
I added a simple test.  Please advise if you want more or have suggestions on writing a better one.

tarzanek:
test looks good, I will merge this soonish ...

tarzanek:
merged, thank you!

PR #963, opened 2015-07-09
Initial clojure support.

sanel:
Initial Clojure support.


tarzanek:
For new files copyright has to be only 2015
One file has still Sun micro copyright

Ctags regexps need tobe checked if they are valid
I am not sure about alnum, see my scala fixes, regexp has to be portable to solaris 11 ctags too

Other than this it looks ok, tests would be great I have some concerns on nested comments

Once copyright is fixed I am ok to accept

tarzanek:
for regexps, please check the comment in ctasg.java:
//Also note, that below ctags definitions HAVE to be in POSIX
            //otherwise the regexp will not work on some platforms
            //on Solaris regexp.h used is different than on linux (gnu regexp)            
            //http://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended

if you follow that wiki all will work on both solaris and linux, so basically don't use gnu, but always use POSIX

tarzanek:
fwiw - tests would uncover if definitions would be missing on some platforms (I run builds both on solaris 12 and debian jessie, travis runs them on ubuntu)

tarzanek:
fwiw, I quickly checked the regexps and they look OK
but ideal case would be to add a file with all the defs used for ctags regexps and run them in test on both solaris and linux ...

sanel:
I've been using these regexes on fairly large clojure project and they seems to capture all relevant fields. However, I'm interested, are there any more differences on Solaris ctags, beside usage of POSIX regexps?

I don't have Solaris here so I'll have to rely on those tests then. Will update them soon....

Also, can you drop me hash of commit for yours scala alnum fixes?

tarzanek:
Check history of ctags.java, last commit
There is no other difference, just solaris build of ctags links to posix regexp system lib

Btw. I think php and one more analyser has nice standalone tests ... so you can get inspired by them

sanel:
Great! Thanks :)

Also I noticed one tidbit that is following me for some time. OpenGrok will take search expression **my-func** and try to search something like **my AND func**, which isn't correct since it will find other non-related occurrences. The same can be applied for valid Clojure/Lisp functions, like **my-setter!** or **is-true?**.

Is there any way to tune Lucene to support these special chars in query or would you suggest different approach?

sanel:
Btw. is there official OpenGrok mailing list for development or using GitHub is the way? I find mailing lists much easier for dev discussion and collaboration.

vladak:
Oops, accidentally closed, reopening (this is what I get for accessing the web from smart phone)

The mailing lists are on https://github.com/OpenGrok/OpenGrok/wiki/Discussions

vladak:
As for the dash behavior, this is normal since opengrok tokenizes. To search for exact match enclose the term into double quotes.

tarzanek:
Improving default query would be cool
So it wouldn't tokenize but behave like human - divide words by whitespace
@kahatlen what do you think?

tarzanek:
@sanel do you want me to merge this branch? or you want to still work on it?

sanel:
Would you like me to resolve these conflicts first?

tarzanek:
Feel free to resolve them @sanel , it should be easy
If not I can do it later this week or through we

sanel:
I'm trying to solve these conflicts and noticed that in `SearchHelper.java`, *fileTypeDescription* got removed (https://github.com/OpenGrok/OpenGrok/commit/eb11fe3584b7b243fb0641da4ab2e157610bb767). Any reasons for this and is it safe to merge this file as is?

tarzanek:
@sanel I moved the fileTypeDescription to respective analyzer and it is getting inited as part of analyzer
the reason was to simplify creation of new analyzers and their isolation from the rest of the code in the same changeset
see e.g. private static final String name = "Shell script"; for ShAnalyzerFactory and respective constructor super(NAMES, PREFIXES, SUFFIXES, MAGICS, null, "text/plain", Genre.PLAIN, *name*);
hope this explains, fix should be easy, as per that changeset, just add the name and adjust constructor

tarzanek:
@sanel if you want I can merge this for you (I should mostly know all changes, so unless I hit something you redesigned I should be good to go)
just let me know, so we can expedite this ;)

tia
L

sanel:
Yes, please merge it :) I do have couple of things to ask/discuss, but I will ping you via mailing list.

tarzanek:
merged, let's see if travis will complain

tarzanek:
ok, I already see a prob - universal ctags only supports clj, no other extensions

tarzanek:
btw. @sanel care to write some junit tests ?
check php junits or master junits, it's good to have them to see if something breaks the language support

aletts:
Hey-

I'm running OpenGrok v0.12.1 rev 46014450c7a107b2f10690b17559958ddb79f960
.. along with Universal Ctags Development(57c7b08) from their OS X homebrew tap.

I'm a new user and am not seeing Clojure analysis and am not seeing Clojure in the file Type list in the OpenGrok web interface.  Is there something I need to do in order to enable the support?

sanel:
@tarzanek yup, no problem; I'll try to tackle them...

sanel:
Hi,

aletts <notifications@github.com> writes:
> Hey-
>
> I'm running OpenGrok v0.12.1 rev 46014450c7a107b2f10690b17559958ddb79f960
> .. along with Universal Ctags Development(57c7b08) from their OS X homebrew tap.
>
> I'm a new user and am not seeing Clojure analysis and am not seeing Clojure in the file Type list in the OpenGrok web interface.  Is there something I need to do in order to enable the support?

Is this the latest version from git? AFAIK, this got merged 12 days ago...

Best,
Sanel

aletts:
Hey, you somehow got to reply after I had deleted the post :)

I figured out my problem -- My opengrok.jar was built against older code, so I ended up building from lastest source to get it to work.

Now that it works, I've of course been skimming through the OpenGrok source to see if there's some way I can get it to treat '-' as a regular character, since that would make the code analysis vastly more convenient.. but I haven't been able to find that (and perhaps that would actually reside in Lucene code - I don't know).

sanel:
aletts <notifications@github.com> writes:
> Hey, you somehow got to reply after I had deleted the post :)

Noticed something weird was going on with github web ui ;)

> I figured out my problem -- My opengrok.jar was built against older code, so I ended up building from lastest source to get it to work.
>
> Now that it works, I've of course been skimming through the OpenGrok source to see if there's some way I can get it to treat '-' as a regular character, since that would make the code analysis vastly more convenient.. but I haven't been able to find that (and perhaps that would actually reside in Lucene code - I don't know).

For search? In theory scape char '\' should do that, but these things
are hidden deep inside Lucene tokenizer, that will probably require a
custom one.

Best,
Sanel

aletts:
The escape char hasn't been working for me unfortunately (results from using it don't make much sense, and using double-quotes instead yields far better results).

It has also crossed my mind that there may be some necessary changes (or also some possible workarounds to avoid the need to modify lucene code) in the webapp code.

PR #967, opened 2015-07-24
Add 'PH' suffix to Perl analyzer

ymnick:


tarzanek:
are you ok with http://www.oracle.com/technetwork/community/oca-486395.html ?
or if you signed it before and your name is on above list then I don't have any probs accepting this (if not, either sign, get confirmation from Oracle, or just say here you accept ;) )

ymnick:
I accept

PR #974, opened 2015-08-05
print table of project/repository mappings only if the map is non empty

vladak:


tarzanek:
lgtm please push/merge

vladak:
Thanks. Merged from the lake shore.

tarzanek:
Way to go ;)

PR #976, opened 2015-08-07
document partial reindex steps

### LINKED ISSUE #975: indexpart parameter of OpenGrok's script
Hi,
I am running OpenGrok on Linux. What is the indexpart parameter in the OpenGrok script?
Does it specify to index only some part of the source? How should one use it? The documentation is very lacking.

vladak:
fixes #975


tarzanek:
ship it!

PR #977, opened 2015-08-07
add JQuery tablesorter for sorting directory listing table

### LINKED ISSUE #831: allow sorting according to Date/Size/Name in file list view
Some JavaScript stanza can be added which would allow to click on the 'Name', 'Date', 'Size' column headers which would in turn change the sorting scheme (and ordering direction) on the fly (i.e. without refreshing the page).

This is handy in case the file listing is long/incomprehensible and the user wants to see the newest (most recently modified) files or biggest files in current directory.

vladak:
fixes #831

Not sure what to do about licensing. tablesorter is dual licenced - either MIT or GPL.

### DIFF
--- a/src/org/opensolaris/opengrok/web/DirectoryListing.java
+++ b/src/org/opensolaris/opengrok/web/DirectoryListing.java
@@ -120,8 +120,8 @@ public class DirectoryListing {
             }
         }
 
-        out.write("<table id=\"dirlist\">\n");
-        out.write("<thead>\n<tr><th/><th>Name</th><th></th><th>Date</th><th>Size</th>");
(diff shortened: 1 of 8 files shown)


web/jquery.tablesorter.js
```diff
@@ -0,0 +1,1031 @@
+/*
```
tarzanek: please include a version number inside javascript file name
this will terribly help when upgrading it, since it will break most of caches, if the version is ommited a lot of browser caches will still use old js
js caching is simply broken on browsers, versioning js filenames is a good way to fight this

tarzanek:
just include the version in file name (same way as we do for other js files)
other than that lgtm, so ship it once the file name is OK

vladak:
ok, renamed

PR #982, opened 2015-08-26
Add android aidl to "java" group

AsCatgz:


AsCatgz:
I accept OCA for this patch

tarzanek:
thank you sir, accepting as per http://developer.android.com/guide/components/aidl.html

PR #988, opened 2015-09-04
fix syntax highlight for C# verbatim strings

petrzelka:
CSharpXref doesn't work correctly for verbatim strings that end with a backslash. This is the fix.


tarzanek:
Hi 
lgtm
To accept however I need you to either agree here with OCA or sign it and mail a scan to Oracle
See http://www.oracle.com/technetwork/community/oca-486395.html
Once this is done or you will agree here I see no prob and will merge

Tia
L

tarzanek:
@petrzelka hello, I just need you to accept OCA here (write, that you're OK with OCA for this patch)
or fill in the paper work and send it to Oracle (this takes few days)
I am not able to merge this patch otherwise, if you have an OCA, just let me know with what name you registered.

petrzelka:
Hi Lubos,

Sorry for the delay.

I am OK with The Oracle Contributor Agreement for this patch.

If that's all you need, please go ahead and merge it in.

Thanks
Honza

tarzanek:
thank you sir!
merging and yes, it's enough for me for this patch, long term however you can go for the OCA process (it's just print, scan and mail the scan to Oracle)

Issue #948, opened 2015-05-26
lastModified value not updated in Configuration.getDateForLastIndexRun()


fregienj:
The home page shows the date for the last index run, but this date is not being updated on a re-index. It is only updated on a app server restart.

 It appears that the Configuration.getDateForLastIndexRun() method has a null check on 'lastModified' and only updates the 'lastModified' date if the previous value was null.  It looks like this null check should not be there.  I'm not sure what it buys you. You always should be getting the last modified date of the file, and not cache the original value.

tarzanek:
I suppose configuration being sent to webapp should make sure this is updated too
(upon reindex this config is sent to webapp)
and I would stay with caching, you don't need to poll this for every web request

scottk:
I have this issue as well, I created a pull request to allow users to manually update the referenced date without changing the original functionality

tarzanek:
@scottk @fregienj 
I updated the indexer that upon sending the configuration.xml if the index is refreshed, the timestamp will also be updated, also all "caching" stays
will that work for you?
Or @scottk how does your jenkins reindex job runs? Do you use sending of config back to webapp?

tarzanek:
(I will push soon, with 5.5.0 lucene update, but just wanted to figure out if it will fix your use case)

tarzanek:
this got fixed with  #1192

PR #991, opened 2015-09-08
Added method allowing user to manually update the lastModified variab…

scottk:
Currently, the "Indexes Created: xxxxx" section of the footer on the homepage only updates every time the entirety of opengrok is restarted. When using a third party (Jenkins jobs) to re-index, this date is not updated, as the opengrok service itself has not rebooted. 

Proposal - add method that allows user to manually update the lastModified variable in the Configuration object, to avoid restarting opengrok every time a third party re-index job is run. 

Open tickets concerning this problem: https://github.com/OpenGrok/OpenGrok/issues/948

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Configuration.java
+++ b/src/org/opensolaris/opengrok/configuration/Configuration.java
@@ -626,6 +626,16 @@ public final class Configuration {
     }
 
     /**
+     * Update the lastModified variable, in the case of running a third party indexing job
+     *
+     * @param d
+     */
+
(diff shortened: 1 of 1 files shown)


tarzanek:
@scottk how do you call this method?
btw. I was thinking more like the sending of config.xml should make sure the variables in webapp context get updated ...

tarzanek:
this might not be needed anymore since #1192 got merged ... @scottk  can you verify pls?

scottk:
@tarzanek Yes, no longer needed, i'll close it

Issue #590, opened 2013-05-22
display date of last modification in search results (Bugzilla #3549)


vladak:
status NEW severity *enhancement* in component *webapp* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Trond Norbye

On 2008-09-23 05:23:52 +0000, Vladimir Kotal wrote:

> It would be handy if date of last modification was displayed for every file listed in search results.
> 
> For example, it can be sitting on the same line as directory path and indented to the right.
> 
> This display mode should be configurable.

On 2008-09-23 09:00:29 +0000, Lubos Kosco wrote:

> By using and formating the lucene field "date" we can do this without big performance impact (reading the files from disk)
> 
> the code needs to be added after : 
> http://src.opensolaris.org/source/xref/opengrok/trunk/src/org/opensolaris/opengrok/search/Results.java# 83
> and the date from doc.get("date") just needs to be printed in useful manner (maybe lucene has such function already, or we need a new util function into http://src.opensolaris.org/source/xref/opengrok/trunk/src/org/opensolaris/opengrok/web/Util.java )
> - lucene we use is 2.2.0 , I couldn't find its javadoc, but 2.3.2 is here:  http://lucene.apache.org/java/2_3_2/api/core/index.html - so it might contain a function to convert the date already
> 
> This way we naturally will print only data from index which might be not synced with disc ... (but better than no datetime & without any noticeable performance impact, since we have the lucene document already in memory ;) )
> 
> Oh, when evaluating this bug I noticed
> http://src.opensolaris.org/source/xref/opengrok/trunk/web/search.jsp# 69
> should say sort = RELEVANCY;
> 
> otherwise you get in search results the "sorted by null" by default 
> (a small goodie, which could be fixed with this bug)
> 
> any takers for creation of a patch ? ;)

vladak:
Current Lucene version used is 5.3.0, quite a jump from 2008 :-)

PR #996, opened 2015-09-29
Issue 590

tulinkry:
I've finished some work on enhancement #590. I'm still not sure what the final look should be; I'm attaching an image of current form and waiting for reactions.

last_edited_screenshot

Reminding this placement from the issue:
"For example, it can be sitting on the same line as directory path and indented to the right"
- under directory path there is a list of files and the date is desired to be by an every file, so this is probably not correct, is it?

When we decide the final look I'll add a configuration option for the display mode.


src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -171,7 +175,15 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
                 out.write(rpathE);
                 out.write("\">");
                 out.write(rpath.substring(rpath.lastIndexOf('/') + 1)); // htmlize ???
-                out.write("</a></td><td><tt class=\"con\">");
+                out.write("</a>");
+                try {
+                    // insert last edited date if possible
+                    df = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT);
+                    out.write("<small class=\"edited\">(last edited ");
+                    out.write(df.format(DateTools.stringToDate(doc.get("date"))));
```
tarzanek: I have some doubts you will have unclosed <small tag if there will be an exception ...

tulinkry: I see the point. I'd change it to this:

```
try {
    df = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT);
    String dd = df.format(DateTools.stringToDate(doc.get("date")));
    out.write("<small class=\"edited\">(last edited ");
    out.write(dd);
    out.write(")</small>");
} catch ...
```

tarzanek:
lgtm, feel free to push, but perhaps ask @vladak or @kahatlen  for another ship it to be sure
(and feel free to share a link within owan so I can play a bit, got some concerns about this param you added)

vladak:
'last modified' would be better I think.

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -171,7 +175,18 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
...
+                out.write("</a>");
+                if(sh.lastEditedDisplayMode){
+                    try {
+                        // insert last edited date if possible
+                        df = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT);
+                        String dd = df.format(DateTools.stringToDate(doc.get("date")));
+                        out.write("<small class=\"edited\">(last edited ");
+                        out.write(dd);
+                        out.write(")</small>");    
+                    } catch (ParseException ex) {}
```
vladak: Should we log something here ?

tarzanek: it would make sense to add a log output, though it would popup in webapp container which are usually set to log only warning and above so not sure if it will help, but it will definitely help people who know how to setup loggers but lack info how to debug ...
so yes, log would be cool though not mandatory(I mean, if the <small won't show it means there was some exception, so you would have to increase the logger level or debug anyways, which from my POV is the same effort ;) )

tarzanek:
hmm, so I created a conflict(I just merged 2 other pull reqs) ... this will need to be resolved, ping me on oracle im if you'll have troubles with getting the merge done(or ask someone local on git, perhaps VladaM will be of use), we can open a vnc and a phone call and resolve this together worst case Krystof, ok?

tulinkry:
I merged everything, I hope it's fine like this

tarzanek:
looks good, @vladak I am going to merge, if there are other things let's solve them with new pull req

tarzanek:
I kinda "unmerged" this
so let's redo this thing after having your branch properly merged with upstream ...

PR #998, opened 2015-09-30
Add Golang support.

patricklundquist:
Adds support for Golang

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/golang/Consts.java
+++ b/src/org/opensolaris/opengrok/analysis/golang/Consts.java
@@ -1,0 +1,64 @@
+/*Patrick Lundquist
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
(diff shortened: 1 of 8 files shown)


tarzanek:
Hello Patrick, lovely, lgtm and very nice work, just one small thing before we can accept, I need you to either say here you're OK with OCA for this patch,
OR check  http://www.oracle.com/technetwork/community/oca-486395.html  and print, sign, scan and email back a signed copy to get registered with Oracle (they own trademark for OpenGrok)
thank you
Lubos

P.S. actually, there is one thing which could be added which is junit tests - add a sample go file into sources so tests will pick them up and recognize defs or full hits for go analyzer - but we can put it into next pull request (and I owe such request to our junits for scala)

src/org/opensolaris/opengrok/analysis/golang/Consts.java
```diff
@@ -0,0 +1,64 @@
+/*Patrick Lundquist
```
tarzanek: ok, so actually just one small prob, header has to be proper, please move the name to @author section, this line will break checkstyle ... ;)

patricklundquist:
Hi Lubos, I'm OK with OCA for this patch. 
Thanks!
Patrick

tarzanek:
thank you sir, merging!

PR #1000, opened 2015-10-02
Fix C# yyline mismatch issue.

shunlir:
This issue causes no html anchor generated at the definition position of
the symbol in C# xref source files, so can't navigate to that definition
location(in the same file) by a simple click.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/csharp/CSharpXref.lex
+++ b/src/org/opensolaris/opengrok/analysis/csharp/CSharpXref.lex
@@ -69,7 +69,7 @@ Number = (0[xX][0-9a-fA-F]+|[0-9]+\.[0-9]+|[0-9]+)(([eE][+-]?[0-9]+)?[ufdlUFDL]*
 
 {Identifier} {
     String id = yytext();
-    writeSymbol(id, Consts.kwd, yyline - 1);
+    writeSymbol(id, Consts.kwd, yyline);
 }
 
 "<" ({File}|{Path}) ">" {


tarzanek:
Got a sample test file? We can make a junit test out of this issue.

shunlir:
I checked all the other *Xref.lex under the sub-folders of  analysis, all use 'yyline' not 'yyline -1'
A simple C# source file can reproduce this issue, however, I'm not sure how to add a junit testcase easily as I'm not using an IDE.
The mismatch of 'yyline-1' with ctags line causes this code logic not executed; https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/analysis/JFlexXref.java#L378.

tarzanek:
Netbeans IDE is very easy to setup and ready to use with OpenGrok
I'd give it a go and feel free to ask, I'll try to help.
Ideally if you can provide a test case so we won't hit this issue next time.
Also before accepting this patch I need you to either check http://www.oracle.com/technetwork/community/oca-486395.html , print, sign and scan and mail back, or the easy route (for small patches usually), is just that you say that you agree with OCA for THIS patch.

tia
L

tarzanek:
looking at original c analyzer, that yyline -1 is definitely buggy, not sure why @haufelex split it this way ... perhaps digging through mail archives and asking him might get clarity, but until then, it's a bug

tarzanek:
@shunlir  I know this is a silly patch, but still I'd need just a sentence you're OK with OCA and I will merge ... thnx

shunlir:
I agree with the OCA, thanks.

tarzanek:
thank you sir, merging!

Issue #1002, opened 2015-10-02
Issue for Golang support


shunlir:
Hi @patricklundquist @tarzanek 
Thanks for your great job. I know OpenGrok added Universal ctags support at the end of Jul. 2015, and added support for Golang just now.
@patricklundquist, can you please add user-defined Golang mapping command options only when running Exuberant Ctags? Since Exuberant Ctags doesn't support Golang internally, however Universal ctags(https://github.com/universal-ctags/ctags) provides more powerful built-in support for Golang, if Universal ctags is used, the user-defined Golang mapping will take a higher priority, this is not desired.
The command line options you added for Golang mapping:
![111](https://cloud.githubusercontent.com/assets/14887081/10245640/01be2f5c-693c-11e5-8cbb-9dc0969ca487.png)
This is what Universal Ctags provides for Golang:
![image](https://cloud.githubusercontent.com/assets/14887081/10245676/4cc888f8-693c-11e5-901d-4d8f8a2a9677.png)

Maybe, this commit can help you to check the running Universal/Exuberant ctags: https://github.com/OpenGrok/OpenGrok/commit/3ea68c10d5745d361c86c32b59fe10e13157c251

patricklundquist:
Thanks Shunlir, I didn't know about the Universal ctags support. I'll look into this.

tarzanek:
Opengrok can detect universal ctags I added some basic check.
So this is doable  and makes sense, thnx for heads up

tarzanek:
this was fixed with #1005 , closing

PR #1005, opened 2015-10-05
Don't use user-defined Golang mapping command options with Universal ctags

patricklundquist:
Addresses issue #1002


tarzanek:
looks great, same Q regarding OCA - http://www.oracle.com/technetwork/community/oca-486395.html
if you accept state it, or get the registration in Oracle to avoid this silly questioning :(

tarzanek:
(I have to get the "accept  OCA" per patch, I cannot do it once for all, that's why the registration is there ... )

patricklundquist:
Thanks Lubos, I'll look into registering. 

I'm OK with OCA for this patch.

tarzanek:
thank you, in the gate now! :)

PR #1009, opened 2015-10-09
display last modified date in search results

### LINKED ISSUE #590: display date of last modification in search results (Bugzilla #3549)

tulinkry:
fixes #590

previous discussion in #996

### DIFF
--- a/src/org/opensolaris/opengrok/search/Results.java
+++ b/src/org/opensolaris/opengrok/search/Results.java
@@ -32,12 +32,15 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.Writer;
+import java.text.DateFormat;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.zip.GZIPInputStream;
 import org.apache.lucene.analysis.charfilter.HTMLStripCharFilter;
+import org.apache.lucene.document.DateTools;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.CorruptIndexException;
 import org.apache.lucene.index.IndexableField;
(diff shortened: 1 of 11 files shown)


tarzanek:
thank you sir, this looks good now!
hopefully no forced push anymore ;)

PR #1012, opened 2015-10-14
Introducing static comparator in IndexDatabase.java

### LINKED ISSUE #827: optimize sorting - comparator can be static
https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/index/IndexDatabase.java#L823

the comparator could be made static ...

tulinkry:
fixes #827


tulinkry:
Do you mean like this?

tarzanek:
thank you, merging

tulinkry:
We discussed with @vladak that the comparator is not really useful in this case, so perhaps another option is to remove it from the code completely. It looks like the rest of indexDown is not dependent on ordered file list. Only difference I've found out is that the log is not alphabetically ordered which is not that necessary. I've tested it also on travis with no failure.

What do you think?

PR #1017, opened 2015-10-27
Line highlighter

tulinkry:
I've implemented just a little feature which is similar to line selection on Github. By clicking on line number while holding shift key one can select range of lines. Those are appended to the url in the hash; so the url can be sent to another person.

Demo is available here:

http://virtual99.cz.oracle.com/lines/

Screenshot:

screenshot from 2015-10-27 18 40 49

Tested in IE9, Chrome 46.0 and Firefox 40.0.3.

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -16,6 +16,10 @@ div#src a.l:after, div#src a.hl:after {
 
 div#src a.l.selected:after, div#src a.hl.selected:after {
     content: "";
+/* highlight line number with anchor */
+div[id^='src'] a.l.target, div[id^='src'] a.hl.target {
+	background: orange;
+	color:yellow;
 }
 
 a:link {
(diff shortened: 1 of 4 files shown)


tarzanek:
thank you @tulinkry , resolve conflict, merge and I will merge to upstream

tulinkry:
Good to say that there is by default a tooltip which warns user that there is this new feature.
It can be turned off by setting the [default value](https://github.com/OpenGrok/OpenGrok/pull/1017/files#diff-96e3a065f54dc0d869951b891bf7c03bR614) for tooltip to false
```javascript
           // other code
            defaults: {
              // other options ...
              tooltip: true // set to false
            },
            // other code
```
or by calling the init function with tooltip set to false
```javascript
$(document).ready(function () {
    $.hash.init({ parent: "pre", tooltip: false})
}
```

tarzanek:
thank you, merging

PR #1020, opened 2015-10-30
Preserving whitespace on copy-paste

### LINKED ISSUE #540: copy-and-paste does not preserve whitespace (Bugzilla #19191)
status NEW severity *normal* in component *webapp* for *---*
Reported in version *unspecified* on platform *ANY/Generic*
Assigned to: Lubos Kosco

Original attachment names and IDs:
- _copy-paste.png_ (ID 4634)

On 2012-03-07 16:32:08 +0000, Vladimir Kotal wrote:

> When copy-pasting code from normal/annotate view the whitespace between columns is not preserved. See the screenshot for actual illustration.
> 
> I am pretty sure this worked in 0.10.

On 2012-03-07 16:37:33 +0000, Vladimir Kotal wrote:

> Created attachment 4634
> screenshot illustrating the problem when copy-pasting

On 2012-03-07 16:48:45 +0000, Erik Lorimor wrote:

> It might be useful to use the following fix
> 
> (I believe the element is to add a space just before the anchor, but depending on what is the exact location in which the space is missing please adjust the selector accordingly.)
> 
> a.r:before {
>     content: "\00a0";
> }
> 
> Please see the following article on Stack Overflow:
> 
> http://stackoverflow.com/questions/5467605/add-a-space-after-an-element-using-after

On 2012-03-07 17:23:10 +0000, Jens Elkner wrote:

> (In reply to comment # 0)
> > When copy-pasting code from normal/annotate view the whitespace between columns
> > is not preserved. See the screenshot for actual illustration.
> > 
> > I am pretty sure this worked in 0.10.
> 
> Yes, those "pseudo columns" are not whitspace padded anymore - CSS took over.
> Plan2fix: add a button, which lets the JS client pad on demand, since a single space between the "columns" isn't probably what you wish, right?
> (BTW: My plan is to fetch annotations per XMLHttpRequest in the background and wave them into them xref page, so wouldn't be a big overhead wrt. programming ;-))

On 2012-03-07 17:33:38 +0000, Vladimir Kotal wrote:

> (In reply to comment # 3)
> > (In reply to comment # 0)
> > > When copy-pasting code from normal/annotate view the whitespace between columns
> > > is not preserved. See the screenshot for actual illustration.
> > > 
> > > I am pretty sure this worked in 0.10.
> > 
> > Yes, those "pseudo columns" are not whitspace padded anymore - CSS took over.
> > Plan2fix: add a button, which lets the JS client pad on demand, since a single
> > space between the "columns" isn't probably what you wish, right?
> > (BTW: My plan is to fetch annotations per XMLHttpRequest in the background and
> > wave them into them xref page, so wouldn't be a big overhead wrt. programming
> > ;-))
> 
> Actually we would like the contents to be padded by default.

On 2012-03-07 17:49:29 +0000, Jens Elkner wrote:

> (In reply to comment # 4)
> 
> > Actually we would like the contents to be padded by default.
> 
> switchable anytime on demand per user prefs I should have add ...

On 2012-03-07 17:53:27 +0000, Vladimir Kotal wrote:

> (In reply to comment # 5)
> > (In reply to comment # 4)
> > 
> > > Actually we would like the contents to be padded by default.
> > 
> > switchable anytime on demand per user prefs I should have add ...
> 
> Sounds good.

On 2012-03-07 22:55:40 +0000, Charles LaBrec wrote:

> Here's some html that seems to give you a bit more control:
> 
> <html>
> <head>
> <style type="text/css">
> .bg {
>   background-color: #ccc;
>   padding: 0 4px;
>   margin: 0 2px;
> }
> 
> .space {
>   font-size: 0px;
> }
> 
> </style>
> </head>
> <body style="background-color: #fff">
> <span class="bg">aaaa</span><span class="space">&nbsp;</span><span class="bg">bbb</span><span class="space">&nbsp;</span><span>ccccc</span><br />
> </body>
> </html>
> 
> The &nbsp; completely disappears in Firefox and Chrome because it is in a span set to a 0px font size. (You can see this by removing/commenting the margin setting out. But it's there for conversion to text when copying and pasting into a text editor.

On 2012-03-08 10:05:17 +0000, Knut Anders Hatlen wrote:

> (In reply to comment # 3)
> > Plan2fix: add a button, which lets the JS client pad on demand, since a single
> > space between the "columns" isn't probably what you wish, right?
> 
> I'd say it's not worthwhile cluttering the interface with a button or some kind of configuration switch to choose between padded/unpadded annotation columns. Sounds like something that would confuse most users. Are there any downsides to always padding the columns?

On 2012-03-08 16:35:20 +0000, Vladimir Kotal wrote:

> (In reply to comment # 8)
> > (In reply to comment # 3)
> > > Plan2fix: add a button, which lets the JS client pad on demand, since a single
> > > space between the "columns" isn't probably what you wish, right?
> > 
> > I'd say it's not worthwhile cluttering the interface with a button or some kind
> > of configuration switch to choose between padded/unpadded annotation columns.
> > Sounds like something that would confuse most users. Are there any downsides to
> > always padding the columns?
> 
> I like the solution Charles came up with. It always pads and does not impact CSS/visual AFAIK.

On 2012-03-08 16:45:07 +0000, Lubos Kosco wrote:

> hope you guys know that such changes will impact xref cache ... hence making the old index from 0.11 needing a refresh (so I would probably skip this change from 0.11.1 hmm ? )

On 2012-03-08 16:53:50 +0000, Vladimir Kotal wrote:

> (In reply to comment # 10)
> > hope you guys know that such changes will impact xref cache ... hence making
> > the old index from 0.11 needing a refresh (so I would probably skip this change
> > from 0.11.1 hmm ? )
> 
> Will it if it's only done in CSS ? I'd really like to have this in 0.11.1.

On 2012-03-08 16:58:34 +0000, Lubos Kosco wrote:

> (In reply to comment # 11)
> > (In reply to comment # 10)
> > > hope you guys know that such changes will impact xref cache ... hence making
> > > the old index from 0.11 needing a refresh (so I would probably skip this change
> > > from 0.11.1 hmm ? )
> > 
> > Will it if it's only done in CSS ? I'd really like to have this in 0.11.1.
> 
> <span>&nbsp; is in html ... but if we can modify existing css to incorporate then it would work ... but not otherwise

On 2012-03-08 19:18:32 +0000, Vladimir Kotal wrote:

> (In reply to comment # 12)
> > (In reply to comment # 11)
> > > (In reply to comment # 10)
> > > > hope you guys know that such changes will impact xref cache ... hence making
> > > > the old index from 0.11 needing a refresh (so I would probably skip this change
> > > > from 0.11.1 hmm ? )
> > > 
> > > Will it if it's only done in CSS ? I'd really like to have this in 0.11.1.
> > 
> > <span>&nbsp; is in html ... but if we can modify existing css to incorporate
> > then it would work ... but not otherwise
> 
> If we combine Charles' (&nbsp with zero width) and Erik's solution (content in CSS) then it can be done without affecting xrefs.

On 2012-03-09 00:29:16 +0000, Charles LaBrec wrote:

> After some prototyping, it appears that using the CSS content property to add a space doesn't work to get that space into a text representation of the HTML. I also discovered that content added in this way ends up becoming part of the the selected tag, and not appended. Thus, if a "cell" has a background color, using this method adds extra space to one side.
> 
> If there's a requirement/desire to not change the current HTML, the other option is to use jQuery (or raw javascript if you'd rather) to add the space element.
> 
> Here is the above example done in this way:
> 
> <html>
> <head>
> <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.7.1/jquery.min.js"></script>
> 
> <script type="text/javascript">
> (function($) {
>   var spacer = $('<span class="space">&nbsp;</span>');
>   $(document).ready(function() {
>     $(".bg").each(function(i) {
>       $(this).after(spacer.clone());
>     });
>   });
> })(jQuery);
> </script>
> 
> <style type="text/css">
> .bg {
>   background-color: #ccc;
>   padding: 0 4px;
>   margin: 0 2px;
> }
> 
> .space {
>   font-size: 0px;
> }
> </style>
> </head>
> 
> <body style="background-color: #fff">
> <span class="bg">aaaa</span><span class="bg">bbb</span><span>ccccc</span><br />
> </body>
> </html>

On 2012-03-09 06:20:54 +0000, Charles LaBrec wrote:

> After posting my last comment, I realized that if the requirement is to have a solution that does not require changing the page's HTML, my solution is only viable if:
> 
> 1) jQuery is already being loaded. This isn't fatal, but you'd have to manipulate the DOM without it.
> 
> 2) The page already loads a script file into which you can put the code to manipulate the DOM. This is the critical piece. If it doesn't already load any JS under your control, then there's no way to inject the spaces, jQuery or no.

On 2012-03-09 14:17:17 +0000, Lubos Kosco wrote:

> 
> I have a bad feeling we will run into:
> 
> http://monaco.us.oracle.com/detail.jsf?cr=7141279
> for non oracle folks:
> https://bugzilla.mozilla.org/show_bug.cgi?id=723788
> 
> with jquery...
> 
> so http://www.youtube.com/watch?v=G6NNGVHrqho :-D
> (no pun intended, just good music and the feasible keyword)
> 
> (In reply to comment # 15)
> > After posting my last comment, I realized that if the requirement is to have a
> > solution that does not require changing the page's HTML, my solution is only
> > viable if:
> > 
> > 1) jQuery is already being loaded. This isn't fatal, but you'd have to
> > manipulate the DOM without it.
> > 
> > 2) The page already loads a script file into which you can put the code to
> > manipulate the DOM. This is the critical piece. If it doesn't already load any
> > JS under your control, then there's no way to inject the spaces, jQuery or no.

On 2012-03-13 15:22:12 +0000, Charles LaBrec wrote:

> Indeed, the jquery solution could run into browser performance issues on slower systems, of which many old Sun Ray servers could be classified as. So the correct solution depends on which is more important: performance, correct copy/paste as text, and leaving the original HTML unaltered. As is the usual mantra: "pick two"...
> 
> While the bug itself is not directly related to this issue, I found the following on firefox: https://bugzilla.mozilla.org/show_bug_cgi?id=116083 . The following comment explains why CSS styling (content property) did nothing:
> 
> > Currently, we copy what's in the dom, not what CSS is making us show on the
> screen.  We spent many days early on talking about how we might be able to copy
> taking CSS into account, but never came up with a design everyone could agree
> with (let along find time to implement it in a way that offered reasonable
> performance).
> 
> As an interesting side note, that bug is now over 10 years old...

On 2012-03-14 09:02:07 +0000, Vladimir Kotal wrote:

> (In reply to comment # 17)
> > Indeed, the jquery solution could run into browser performance issues on slower
> > systems, of which many old Sun Ray servers could be classified as. So the
> > correct solution depends on which is more important: performance, correct
> > copy/paste as text, and leaving the original HTML unaltered. As is the usual
> > mantra: "pick two"...
> 
> After having some additional discussions I am starting to think that the HTML code will have to be changed. Not sure if this is material for 0.11.2 or 0.12.

tulinkry:
fixes #540 

This deals with the problem of copy-paste whitespace between line number and the rest of the code.
It's pure javascript solution therefore it has no data overhead.

On Firefox and Chrome there is a space appended with css :after selector (but it's not copyable) and upon an user's selection this space is extended to ordinary space which is preserved in selected text.

Because of the natural form of files there is couple of limitations
 - Line gets selected from the beginning to the end (extending users selection)
 - Sometimes it wrongly guesses the first line (or last line) in selection resulting in more lines are selected
But for the usual use case - copying few lines of code to show them somewhere it's sufficient.

Solution is made effective. Lines processing and adding the whitespace is performed only upon the user's selection and only in the range of the selection.

Considering that indexed files shall not be extended with a space by default and considering the tradeoff between the efficiency and complexity; this solution looks reasonably ok for me.

IE preserves the css :after space, so there is no need for additional javascript (tested in IE 9 and 11)

Demo is here:
http://virtual99.cz.oracle.com/spaces/

Huge file  - can take a while to load -  [here](http://virtual99.cz.oracle.com/spaces/xref/sources/c/foobar.out) (150k lines)

### DIFF
--- a/web/default/print.css
+++ b/web/default/print.css
@@ -8,6 +8,19 @@ a {
 	text-decoration: none;
 }
 
+
+div#src a.l, div#src a.hl {
+    padding: 0;
+}
+
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -10,6 +10,14 @@ a {
 	text-decoration: none;
 }
 
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/offwhite/print.css
+++ b/web/offwhite/print.css
@@ -8,6 +8,18 @@ a {
 	text-decoration: none;
 }
 
+div#src a.l, div#src a.hl {
+    padding: 0;
+}
+
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/offwhite/style.css
+++ b/web/offwhite/style.css
@@ -10,6 +10,18 @@ a {
 	text-decoration: none;
 }
 
+div#src a.l, div#src a.hl {
+    padding: 0;
+}
+
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/polished/print.css
+++ b/web/polished/print.css
@@ -8,6 +8,18 @@ a {
 	text-decoration: none;
 }
 
+div#src a.l, div#src a.hl {
+    padding: 0;
+}
+
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/polished/style.css
+++ b/web/polished/style.css
@@ -10,6 +10,18 @@ a {
 	text-decoration: none;
 }
 
+div#src a.l, div#src a.hl {
+    padding: 0;
+}
+
+div#src a.l:after, div#src a.hl:after {
+    content: "\0020";
+}
+
+div#src a.l.selected:after, div#src a.hl.selected:after {
+    content: "";
+}
+
 a:link {
 	color: #2030A2;
 }
--- a/web/utils.js
+++ b/web/utils.js
@@ -21,6 +21,264 @@
  * Copyright (c) 2009, 2015, Oracle and/or its affiliates. All rights reserved.
  */
 
+(function(window, $) {
+   
+    var spaces = function () {
+        var inner = {
+            self: this,
+            initialized: false,
+            /**
+             * Mouse selection event
+             * - upon a user's selection triggers a select event
+             */
+            mouse: {
+                dragging: false,
+                init: function () {
+                    var that = this
+                    $(document).mousedown(function (e) {
+                       that.dragging = false
+                    }).mousemove(function(e){
+                        that.dragging = true
+                    }).mouseup(function(e){
+                        var wasDragging = that.dragging
+                        that.dragging = false
+                        if(wasDragging) {
+                            $(document).trigger("select");
+                        }
+                    }).dblclick(function(e){
+                        //$(document).trigger("select")
+                    });
+                },
+            },           
+            defaults: {
+                "selector": "a.l, a.hl",
+                "parent": "div#src pre",
+                "selectedClass": "selected",
+                "sourceContainer": "pre",
+            },
+            options: {},
+            indent: function($el){
+                return $el.each(function() {
+                    if(! $(this).is("." + inner.options.selectedClass))
+                        $(this).html($(this).html() + "&nbsp");
+                    $(this).addClass(inner.options.selectedClass)
+                });
+            },
+            /**
+             * @returns {Boolean} if client is IE
+             */
+            ie: function() {
+                var ua = window.navigator.userAgent;
+                if (ua.indexOf('MSIE ') > 0 || 
+                    ua.indexOf('Trident/') > 0 ||
+                    ua.indexOf('Edge/') > 0 )
+                    return true;
+                return false;
+            },            
+            getSelection: function () {
+                if (window.getSelection)
+                    return window.getSelection()
+                return null
+            },
+            /**
+
+             * Select closest element giben by options.selector to the actual
+             * element
+             * 
+             * @param {jQuery Object} $el actual element
+             * @param {boolean} next direction
+             * @param {boolean} last if it is the last element in array
+             * @param {int} depth max distance from the path between the actual
+             *                    element and the root element
+             * @returns {Object} of {element found, used given direction}
+             */
+            around: function ($el, next, last, depth) {
+              var slc = inner.options.selector
+              depth = depth || 10
+              next = next || false      
+              last = last || false
+
+                if($el.is(slc)) {
+                  return { "element": $el, "directionUsed": false };
+              }  
+
+              var $tmp = $el;
+              var $result = null
+              var parentDepth = 10
+              // scan every previous parent up to partentDepth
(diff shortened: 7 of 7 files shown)


tarzanek:
lgtm, please resolve conflict and merge, I will then merge the req
thank you @tulinkry !

tarzanek:
thank you, merging

PR #1025, opened 2015-11-10
Tomcat 8 support

gerbrand:
Minor changes to support tomcat 8 as well. Works with debian/ubuntu tomcat 8 package (currently 8.0.26-1)

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -8,6 +8,7 @@
 #   - Debian      (Linux)
 #
 # Supported Deployment Engines:
+#   - Tomcat 8
 #   - Tomcat 7
 #   - Tomcat 6
 #   - Glassfish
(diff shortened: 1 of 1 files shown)


tarzanek:
Hi @gerbrand looks lovely!
Before I can merge this I need you to either say you accept OCA here in this thread - check it:
http://www.oracle.com/technetwork/community/oca-486395.html
or you can print it, sign it, scan and send back to Oracle so we have your permission to use your code :)
Tia
L

tarzanek:
Hi, thank you, I just got your OCA signed
merging
just a side note, this patch is missing fixes for path for javadb detection, but I will fix that with my tomcat-8 support for solaris 11.3 (which is finally out), so it's ok to merge

Issue #480, opened 2013-05-22
grouping of projects (Bugzilla #17189)


vladak:
status FIXINPROGRESS severity *enhancement* in component *webapp* for *---*
Reported in version *unspecified* on platform *ANY/Generic*
Assigned to: Vladimir Kotal

Original attachment names and IDs:
- _project-groups.jpg_ (ID 4323)

On 2010-10-08 09:08:43 +0000, Lubos Kosco wrote:

> provide user configurable grouping such as on http://grok.czech.sun.com:8080/source/index.jsp

On 2010-10-08 09:09:53 +0000, Lubos Kosco wrote:

> Vladimir is working on this bug and in first iteration we will get a cli interface to create group definitions, which then can generate front page such as on grok.czech

On 2010-10-08 14:31:58 +0000, Vladimir Kotal wrote:

> Created attachment 4323
> screenshot of main page with project groups
> 
> This page with project groups was generated by crude shell script which traverses the directory with projects and uses simple configuration file which maps projects into group based on regular expression. The final version of the enhancement will be fully implemented within OpenGrok and managed by the OpenGrok script.

On 2011-01-05 08:36:02 +0000, Knut Anders Hatlen wrote:

> Resetting target milestone since this won't make it into 0.10.

vladak:
This change has several pre-requisites:
  - ability to display parent info for a project (#937)
  - ability to display branch info for repository
  - project listing on index page

vladak:
There are a few things for consideration:
  - how to specify the groups
    - the original shell script takes a config file in the form of `regexp whitespace name`. Initially I was thinking about CLI to feed this into OpenGrok configuration however this would be too convoluted. Rather, I'd like to create separate configuration file which can be read (simiarly to -R) and merged with runtime configuration. The config file would be Java beans XML representation (HashMap ?).
  - whether to introduce the ability to define subgroups

tarzanek:
I agree on both, they would be a good step forward to finally get full project support from grok ala other code search engines (which just consume SCM URL and name, ev. group/subgroup and generate the rest , I think code sight does it and I played with one more which name I forgot)

vladak:
This was implemented in cset d470e59c0405a31b7e5f194bd9b705e91b12bf0a

PR #1029, opened 2015-11-20
Implementation of project groupings

tulinkry:
I'm getting closer to #480, which I suppose is a prerequisite of the authorization layer.

Groups are taken from configuration file (example below) which is given to the jar run (-W option) and it is therefore propagated to the web application through the socket listener (then it lives in the main configuration file). Project can be in multiple groups.

A new class ProjectHelper is introduced to wrap operations around env.projects so that it can be easily changed later to support authorizator filtering of projects or groups. 

There are couple of screenshots:

1) Groups as a list on the main page. User can click on them and their content gets expanded.
screen
2) UI change of the selectbox to support top level (HTML restriction) groups
screen1
3) expanded groups
screen2

Working demo available [here](http://virtual99.cz.oracle.com/groups/).

Tested in IE 9, Firefox and Chrome.

Example of configuration:
```xml
<java version="1.8.0_65" class="java.beans.XMLDecoder">
	<void method="addGroup">
		<object class="org.opensolaris.opengrok.configuration.Group">
			<void property="name">
				<string>Haxtip</string>
			</void>
			<void property="pattern">
				<string>rfe.*|sql</string>
			</void>
			<void method="addGroup">
	 			<object class="org.opensolaris.opengrok.configuration.Group">
	 				<void property="name">
	 					<string>Git</string>
	 				</void>
	 				<void property="pattern">
	 					<string>haxtip|Replicator</string>
	 				</void>
			 		<void method="addGroup">
			 			<object class="org.opensolaris.opengrok.configuration.Group">
			 				<void property="name">
			 					<string>OpenGrok</string>
			 				</void>
			 				<void property="pattern">
			 					<string>metric|git|OpenGrok</string>
			 				</void>
			 			</object>
			 		</void>
	 			</object>
			</void>
		</object>
	</void>
	<void method="addGroup">
		<object class="org.opensolaris.opengrok.configuration.Group">
			<void property="name">
				<string>Python</string>
			</void>
			<void property="pattern">
				<string>Downloader</string>
			</void>
		</object>
	</void>
 </object>
</java>
```

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Configuration.java
+++ b/src/org/opensolaris/opengrok/configuration/Configuration.java
@@ -22,6 +22,7 @@
  */
 package org.opensolaris.opengrok.configuration;
 
+import java.beans.ExceptionListener;
 import java.beans.XMLDecoder;
 import java.beans.XMLEncoder;
 import java.io.BufferedInputStream;
@@ -41,9 +42,12 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.opensolaris.opengrok.history.RepositoryInfo;
@@ -77,6 +81,7 @@ public final class Configuration {
     private boolean historyCacheInDB;
 
     private List<Project> projects;
+    private Set<Group> groups;
     private String sourceRoot;
     private String dataRoot;
     private List<RepositoryInfo> repositories;
@@ -214,6 +219,7 @@ public final class Configuration {
         setHistoryCacheTime(30);
         setHistoryCacheInDB(false);
         setProjects(new ArrayList<Project>());
+        setGroups(new TreeSet<Group>());
         setRepositories(new ArrayList<RepositoryInfo>());
         setUrlPrefix("/source/s?");
         //setUrlPrefix("../s?"); // TODO generate relative search paths, get rid of -w <webapp> option to indexer !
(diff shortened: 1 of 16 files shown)


src/org/opensolaris/opengrok/configuration/Project.java
```diff
@@ -26,11 +26,20 @@
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.Comparator;
 
 /**
  * Placeholder for the information that builds up a project
  */
 public class Project {
+
```
tarzanek: if you properly implement equals and hashCode, there is no need for comparator ...

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -827,6 +838,25 @@ private static Configuration decodeObject(InputStream in) throws IOException {
         if (!(ret instanceof Configuration)) {
             throw new IOException("Not a valid config file");
         }
+        
+        for (Iterator<Group> it = ((Configuration)ret).getGroups().iterator(); it.hasNext();) {
```
tarzanek: for (Group g:getgroups() ) ...

if you use Netbeans, please notice the left side yellow/orange exclamation marks - hints - usually it should make you aware

Also this check is overengineering(not sure how did you come to such overly complex cycle in cycle ;) ... KISS!) - if you need a deduplicated list, use appropriate structure and make sure your object properly does hashCode and equals implementation

src/org/opensolaris/opengrok/configuration/Group.java
```diff
@@ -0,0 +1,198 @@
...
+     * names along all group's children's siblings.
+     *
+     * Same names do not violate this condition if they are in different
+     * branches. Same names do not violate this condition if they are in the
+     * same branch but they are not siblings
+     *
+     * @param g Group
+     * @return true if it satisfies conditions above
+     */
+    public static boolean isValid(Group g) {
```
tarzanek: I don't get what do we try to solve with this :)
But it also seems too complicated for me and that doesn't feel OK

tarzanek:
looks good, but I have some concerns about the design from code point of view + there is some design/restrictions, that slip my understanding.
Let's meet in the office and talk this out (since we have the luxury of meeting in person, I will skip comments here to save our time). Wed/Thu , check my internal calendar when I am free, I will jump in your office, ev. schedule a meeting and invite me & @vladak pls

tulinkry:
I modified the solution with native Java structures, demo isn't available now.

tulinkry:
I removed that unused import.

"I have a bad feeling document.ready is already called somewhere, but the old code might have been refactored, can you double verify no other jsp 9mast,head, ... list, ...) is calling .ready ?
Also I am not sure if jquery permits several .ready calls"

- ready() is already called twice (table sorter and the inteligence window init)
- ready() is a callback register, so it's not problem when we register multiple handlers for ready event - each of them is executed

There was a little problem when user defined no group. I fixed that so now it's compatible with both modes - using groups or without groups.

tarzanek:
merging now, so we have time to stabilize it

tarzanek:
thank you Krystof!
(keep them coming! :) )

Issue #970, opened 2015-07-27
junit tests and rename handling breaks with mercurial 3.4


tarzanek:
Get https://mercurial.selenic.com/release/mercurial-3.4.tar.gz
use this hg instead of your common working hg (seems to be 3.1.2 on travis ubuntu and on jessie)
run junit tests - they will break on renamed files, history won't contain proper log entries, since hg 3.4 doesn't give you proper output anymore

tarzanek:
(fwiw even 3.5-rc is affected)

tarzanek:
also note when Solaris 11.3 will be released it will ship this hg, so this bug is kinda hot

vladak:
Using the test repository from https://github.com/OpenGrok/OpenGrok/tree/master/testdata/repositories/mercurial (renamed `hg` to `.hg`) I see this:

hg 3.2.3:

```
$ hg log -f -r 'reverse(4:)' --template {rev}:{file_copies}\\n /var/tmp/mercurial/novel.txt
8:
7:bar.txt (foo.txt)novel.txt (book.txt)
6:
5:book.txt (plaintext.txt)
4:
```

hg 3.4:

```
$ hg log -f -r 'reverse(4:)' --template {rev}:{file_copies}\\n /var/tmp/mercurial/novel.txt 
8:
7:bar.txt (foo.txt)novel.txt (book.txt)
```

so 3.4 is missing 3 revisions.

vladak:
The problem seems to be in the revision specifier. If I omit the `-r` option for hg 3.4, then full list of revisions is displayed.

vladak:
Filed bug against Mercurial: https://bz.mercurial-scm.org/show_bug.cgi?id=4959

vladak:
AFAIK this bug is only problem for displaying diffs of revisions across renames, it should not affect history index.

cdugz:
I follow a bit mercurial, in 3.4 release notes: https://www.mercurial-scm.org/wiki/WhatsNew#Mercurial_3.4_.282015-05-01.29

``log: make -fr show complete history from the given revs (BC)``

flag BC means change sensitive towards Backwards Compatibility. There was also https://bz.mercurial-scm.org/show_bug.cgi?id=4921 and https://www.selenic.com/pipermail/mercurial/2015-October/048831.html. As said in 4921, technically it's a regression, but previous behavior (when both -r and -r supplied) is considered as misleading/dangerous.

vladak:
[...]

PR #1030, opened 2015-11-23
remove filtering of changesets when finding original name of renamed …

vladak:
…file in given revision

workaround for #970


tarzanek:
Lgtm feel free to push

Issue #1031, opened 2015-11-25
History page table expands, and does not wrap long words.


vbro:
In OpenGrok history page, revision comments with arbitrarily long strings and no white space cause the width of the history table to expand without limit.

This means that the user has to scroll arbitrarily to the right to see all the text in the commit comments. The comments column header is centered, along with the click-able link "(Show modified files >>>)". Users have to scroll arbitrarily to the right or do a browser search to find this link.

An example of a commit comment that triggers this behavior is the commit message for https://github.com/elastic/elasticsearch/commit/75b8075a53fbc676819a25c2431b1ca294dba0fb. This causes the width of the history table to expand to over 157,000 pixels. This issue can be reproduced by indexing the repository https://github.com/elastic/elasticsearch with OpenGrok and looking at the history for the entire project.

PR #1032, opened 2015-11-25
Fix for OpenGrok issue https://github.com/OpenGrok/OpenGrok/issues/1031

vbro:
Wrapping long strings in commit comments to prevent the history table width from expanding.

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -357,6 +357,19 @@ table#revisions {
         color: #777777;
 }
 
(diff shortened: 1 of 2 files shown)


tarzanek:
Can you check and fix other styles too pls?

vbro:
Done! The fix for Internet Explorer requires the CSS to be in the HTML.

tarzanek:
great, merging, thank you!

PR #1041, opened 2016-01-12
Make the Projects list box larger

abbeyj:
When there are many projects, scrolling through them all becomes difficult with such a small select box.  Dragging the slider by even one pixel can move the list more than 6 elements making it hard to scan for an item of interest and making scrolling very disorienting.

Making it a bit bigger in this case helps greatly.  For installations with a small number of projects the size of the box will not change so there should be no adverse impact on them.


tarzanek:
Thank you James, merging!

PR #1047, opened 2016-01-29
README updates to the tuning section for ULIMIT

dhireng:
I added a short note to README about ULIMIT


tarzanek:
good stuff, I expect you're OK with OCA so merging this in
thank you sir!

PR #1050, opened 2016-02-08
Analyze .tcc files as C++ source code

madars:
`.tcc` is a fairly standard extension for template implementations (e.g. used in libstdc++; see https://gcc.gnu.org/onlinedocs/libstdc++/latest-doxygen/files.html), so we should analyze `.tcc` files when working with C++ codebases.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/c/CxxAnalyzerFactory.java
+++ b/src/org/opensolaris/opengrok/analysis/c/CxxAnalyzerFactory.java
@@ -44,7 +44,8 @@ public class CxxAnalyzerFactory extends FileAnalyzerFactory {
         "HH",
         "CXX",
         "HXX",
-        "TXX"
+        "TXX",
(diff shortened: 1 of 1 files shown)


tarzanek:
https://github.com/OpenGrok/OpenGrok/blob/master/README.txt#L580
this can be used temporarily, tcc isn't really any standard (I even saw people using icc)
anyways, I will accept this (since the only clash is some time calculator program, which in 2016 is hopefully not used anymore ;) )

Just one formality - I need you to either sign and scan and email OCA to Oracle - http://www.oracle.com/technetwork/community/oca-486395.html
or just read it and say you accept it here, which is perfectly fine for such patch ;) .

madars:
Thanks for suggesting -A tcc:..., @tarzanek! I accept OCA for this contribution.

tarzanek:
thank you, merging

PR #1051, opened 2016-02-15
Fixing #927 regression

tulinkry:
As mentioned in #927, 
commit 22e51bd introduced a bug - not hiding a line number on ```lntoggle()```.

The commit above added a class 'target' to the highlighted links, but in the code there was a strong 
test for ```className === 'someotherclass'``` upon hiding/showing, which should be better replaced by ```$.fn.hasClass()```

Result tested in Mozilla and IE9.

### DIFF
--- a/web/utils.js
+++ b/web/utils.js
@@ -1060,25 +1060,34 @@ function lntoggle() {
 }
 
 function lnhide() {
-    $("a").each(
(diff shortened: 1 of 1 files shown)


tarzanek:
thank you sir!
merging

PR #1055, opened 2016-02-18
Don't  display subrepositories without a parent url

tulinkry:
In some cases with nested repositories it's not really helpful the information about subrepositories when they don't have their parent url (or branch). I'd suggest to skip those and if the information about parent url is available then provide the xref.

Current (note that the first two subrepositories in "android" are inside the .repo directory):
current screenshot

Proposed:
proposed screenshot

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -1007,7 +1007,7 @@ public final class PageConfig {
     public boolean resourceNotAvailable() {
         getIgnoredNames();
         return getResourcePath().equals("/") || ignoredNames.ignore(getPath())
-                || ignoredNames.ignore(resourceFile.getParentFile().getName())
+                || ignoredNames.ignore(resourceFile.getParentFile())
                 || ignoredNames.ignore(resourceFile);
     }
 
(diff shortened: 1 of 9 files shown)


src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -975,7 +975,7 @@ public String getResourcePath() {
     public boolean resourceNotAvailable() {
         getIgnoredNames();
         return getResourcePath().equals("/") || ignoredNames.ignore(getPath())
-                || ignoredNames.ignore(resourceFile.getParentFile().getName())
```
tarzanek: @vladak is this OK ?

tarzanek:
after @vladak approves, feel free to merge

src/org/opensolaris/opengrok/web/PageConfig.java
vladak: What this change does is that it converts the basename check to full path check. The basename is checked in the Filter class anyway so this expands the checking. The whole logic of resourceNotAvailable() is not very clear to me. @tulinkry can you explain the intention here ?

tulinkry:
After discussion we decided to add .repo to ignored dirs (the same like for .git, .svn and so on), which also solves the use case of broken .repo xrefs

src/org/opensolaris/opengrok/web/PageConfig.java
tulinkry: The intention is to distinguish between ignored file check and ignored dir check. In this case the parent file should be always a directory, which should be correctly identified by the ignoredNames class (and only the directory test should be performed). Using File object instead of the String basename solves this problem.

tarzanek:
@tulinkry is this PR still needed? or will you open a new one for ignoring of .repo ?

tarzanek:
thank you, merging

PR #1062, opened 2016-02-25
Move code out of utils.js

kahatlen:
utils.js has grown so that it no longer contains only utility functions. Some of the code that has been added to it, is run every time utils.js is loaded. This includes code that searches through the DOM of the page, which could be time-consuming when viewing large files in the xref view.

This pull request moves some code out of utils.js and into separate .js files that are only loaded from pages that need that functionality, with the aim to reduce the amount of JavaScript code that runs when loading xref pages.


tarzanek:
thank you sir!, merging!

Issue #1070, opened 2016-03-10
opengrok binary files


ssaadon:
Hi,
I want to have .lib to be managed on the same manner as .dll and .exe.
i.e, file content is not indexed, only it's location and the ability to download it.
I have tried to use "-i" option but it's ignoring the file completely which is bad for me.  
Is it possible?
Thanks.

vladak:
That's currently not possible, I'd call it fuzzy ignore :-) There might be a way how make the .lib files to be interpreted as certain type of files using ctags config.

tarzanek:
I think adding this extension to ignore/dummy analyzer, but not sure how many pieces from this idea are missing ;)

tarzanek:
hmm, I might not be that far from an actual fix, IgnorantAnalyzerFactory is in place
so we just need checking if -A can get to it, if not it would be a good pull request ...

tarzanek:
well, it might even work ...
-A dll:org.opensolaris.opengrok.analysis.data.IgnorantAnalyzerFactory
and dll-s will be suddenly ignored, but you will see them in tree (try it, I might have a bug in my thinking, I didn't try above!)

tarzanek:
I am actually wondering which from our analysers is using .lib
I thought it would be ignored - which analyser is picked?

ssaadon:
Unfortunately that didn't help.
I have tried it with: -A
.lib:org.opensolaris.opengrok.analysis.data.IgnorantAnalyzerFactory
Maybe I missed something? Do you have any other suggestion?

Thanks!

2016-03-10 19:39 GMT+02:00 Lubos Kosco <notifications@github.com>:

tarzanek:
Try lib without . please

ssaadon:
ERROR: You must specify: -A .extension:class or -A prefix.:class
       Ex: -A .foo:org.opensolaris.opengrok.analysis.c.CAnalyzer
           will use the C analyzer for all files ending with .foo
       Ex: -A bar.:org.opensolaris.opengrok.analysis.c.CAnalyzer
           will use the C analyzer for all files starting with bar.
       Ex: -A .c:-
           will disable the c-analyzer for for all files ending with .c


2016-03-10 23:29 GMT+02:00 Lubos Kosco <notifications@github.com>:

tarzanek:
so "-" works then ?

ssaadon:
It works, meaning that indexing process is running.
But it's still indexing the .lib files including their content.

[...]

PR #1075, opened 2016-03-15
Ignore LIB and PDB files.

moisei:
LIB and PDB files are binary files.
following discussion is at https://github.com/OpenGrok/OpenGrok/issues/1070#issuecomment-196912718


tarzanek:
This looks ok
I just need you to agree with oca http://www.oracle.com/technetwork/community/oca-486395.html#r

Or you did it already?

moisei:
I sent it to oracle  just now.

On Tue, Mar 15, 2016 at 9:16 PM, Lubos Kosco <notifications@github.com>
wrote:

> This looks ok
> I just need you to agree with oca
> http://www.oracle.com/technetwork/community/oca-486395.html#r
>
> Or you did it already?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/OpenGrok/OpenGrok/pull/1075#issuecomment-196981126>
>



-- 
Best Regards,
Moisei

tarzanek:
I just got your OCA
accepting, thank you!

Will you have a look at exposing the IgnoreAnalyzerFactory using -A ?

moisei:
Unfortunately, I am fully booked until Jun, not sure I could have a look at it before that.
Is there a way to validate my fix somehow? I've tried to install it on test platform but still I see a content of a lib file when click on it.

tarzanek:
ideal validation is to write a junit test, this way OpenGrok gets a guard so this won't be broken again ...
we run junit tests as part of every pull req, so no one will ever break your change, if you "wrap it" with a test

Issue #1076, opened 2016-03-20
collapsible revision messages on history page


vbro:
For long revision messages on the history page, allow the revision message to be collapsible with a "show more/show less" javascript link.
This allows for a more concise page and makes the "compare" radio buttons easily accessible for revisions with long messages.

tarzanek:
reopening, can you please address the comments from Knut in #1077  please?

tulinkry:
The comment about updating both styles has been already solved in #1080

vladak:
the other remaining comment from @kahatlen was filed as #1482 , thus closing.

PR #1077, opened 2016-03-20
Fix for https://github.com/OpenGrok/OpenGrok/issues/1076

vbro:
New feature to allow users to collapse and expand long revision
messages on the history page.

### DIFF
--- a/web/history.jsp
+++ b/web/history.jsp
@@ -279,8 +279,10 @@ revision2 = revision2 >= hist.getHistoryEntries().size() ? hist.getHistoryEntrie
                 %><%= author %><%
                 }
                 %></td>
-            <td><a name="<%= rev %>"></a><p><%
+            <td><a name="<%= rev %>"></a><%
+                int summaryLength = 200;
                 String cout = Util.htmlize(entry.getMessage());
+
                 if (bugPage != null && bugPage.length() > 0) {
                     cout = bugPattern.matcher(cout).replaceAll("<a href=\""
                         + bugPage + "$1\">$1</a>");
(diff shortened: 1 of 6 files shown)


web/history.jsp
```diff
@@ -279,8 +279,10 @@ revision2 = revision2 >= hist.getHistoryEntries().size() ? hist.getHistoryEntrie
                 %><%= author %><%
                 }
                 %></td>
-            <td><a name="<%= rev %>"></a><p><%
+            <td><a name="<%= rev %>"></a><%
+                int summaryLength = 200;
```
tarzanek: this should be probably a configuration variable, care to add it?
so people can at least set it using readonly xml config passed to indexer ?

vbro:
Done. Added a way to configure the threshold. Just curious, are there guidelines for configuration variable names?

tarzanek:
@vbro not really, but we have too many of them :) ideally we should clean them up
we have your OCA on file, so happily accepting, thank you sir!
fwiw - if you want to document this new variable, feel free to do so in readme, so it's easier for people to set up and don't ask unnecessary questions
(we should probably start a section on such variables that don't have indexer option, but just are taken from read only config in readme, we have quite a lot of them since 0.12 - @tulinkry  what do you think?)

vbro:
The environment variables are well documented in https://github.com/OpenGrok/OpenGrok/blob/master/OpenGrok. But not all the variables in Configuration.java are documented.
Perhaps the readme should have a section on how to setup/configure/customize the application using the read only xml and what all the different config variables mean. I can take a shot at starting such a section.

kahatlen:
Thanks for this improvement. A couple of suggestions for further improvements:

- Only the default style sheet is updated. Do we also need to change the polished and offwhite variants? And print.css?
- The code added to web/utils.js could probably be moved into the domReadyHistory() function. Then it is only executed in the history page.

tarzanek:
@vbro will you be able to patch the other stylesheets?
Or should we revert for the time being?
(sorry, I completely missed only default theme is patched :( )

vbro:
Yes. I missed that too. Please see https://github.com/OpenGrok/OpenGrok/pull/1080.
This includes patching offwhite and polished css files with css rules from another commit.
Would be a good idea to factor out the common css rules to a separate file and have default/offwhite/polished implement specific rules for those styles.

tarzanek:
@vbro yes, that is a good idea, if you understand css and can do it, it would be wonderful and would simplify new changes, tia!

kahatlen:
+1. It's a pain to have to remember to update all of them. (If it was up to me, I'd say having one style is enough. Then there's less to test when improving the UI.)

vladak:
+1 filed issue #1481 to track the non-default style removal.

PR #1080, opened 2016-03-30
minor css and js cleanup

vbro:
Moving toggle revision to domReadyHistory.
Copying css rules to offwhite and polised styles.


tarzanek:
thank you sir, merging! :)

PR #1081, opened 2016-04-04
change getProject() to use longest prefix matching

### LINKED ISSUE #1074: project matching can lead to index corruption
If there is already existing and indexed project named `foo` then using partial indexing of newly added project `foo-bar` will result of the data being added to index of project `foo`, thus causing index corruption.

This is consequence of insufficient project matching in getProject():

```
141        if (env.hasProjects()) {
142            for (Project proj : env.getProjects()) {
143                if (lpath.indexOf(proj.getPath()) == 0) {  !! prefix matching
144                    ret = proj;
145                }
146            }
147        }
148        return ret;
```

Clearly, the `indexOf()` will not do the job correctly. There needs to be exact match.

vladak:
fixes #1074

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Project.java
+++ b/src/org/opensolaris/opengrok/configuration/Project.java
@@ -18,14 +18,17 @@
  */
 
  /*
- * Copyright 2006, 2015, Sun Microsystems, Inc.  All rights reserved.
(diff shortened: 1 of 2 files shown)


tarzanek:
thank you sir, merging!

Issue #503, opened 2013-05-22
add authorization layer (Bugzilla #18242)


vladak:
status REOPENED severity *enhancement* in component *misc* for *---*
Reported in version *unspecified* on platform *ANY/Generic*
Assigned to: Lubos Kosco

On 2011-05-03 18:29:39 +0000,  wrote:

> Would like to see the ability to secure access to the source code that OpenGrok is presenting for browsing. 
> Need to add a way to redirect access to authentication via https.
> 
> This could require changes to the project, or updating of documentation on how to accomplish this.

On 2011-05-03 22:08:36 +0000, Darren J Moffat wrote:

> I'm not sure that doing authentication (and particularly authentication without a subsequent authorisation and audit system too) in the opengrok application is the correct way to solve this.
> 
> Why can't the hosting webserver/application server authentication and access control system be used for this ? What aren't they providing that requires this be pushed up into the opengrok app ?
> 
> I've always thought of opengrok as "web browser" cscope and there is no way we would add authentication into the UNIX cscope command.

On 2011-05-03 22:26:11 +0000, James C. McPherson wrote:

> Implementing closedgrok seems to be the wrong thing to do. Especially when
> we already have the infrastructure built in to apache / S1WS / etc to do 
> authentication against LDAP and LDAPS, which can be turned on really easily
> with a few lines in the server's configuration file.
> 
> Bruce - have a read of http://httpd.apache.org/docs/2.2/mod/mod_authnz_ldap.html
> for examples.

On 2011-05-04 08:47:58 +0000, Lubos Kosco wrote:

> I understand that the first phase should be leveraging existing pieces to do auth in opengrok - so tomcat auth(either through pam or other pluggable means) and ssl
> 
> so first phase should be a documentation page on how to achieve above
> 
> 
> Second phase is something which is already in place on opensolaris.org - restricting access per project
> for which the web application has to be adjusted slightly (to have users per project which have access) - opensolaris.org implementation is not that pluggable and I am curious how much can be leveraged and taken upstream ...
> 
> someone correct me if wrong
> L

On 2011-05-04 12:54:06 +0000, Lubos Kosco wrote:

> tomcat and ldap - http://blog.mc-thias.org/?title=tomcat_ldap_authentication&more=1&c=1&tb=1&pb=1
> 
> tomcat proxy(in case we will redirect the queries through some other machine or use different outer server name) - http://tomcat.apache.org/tomcat-6.0-doc/proxy-howto.html

[...]

PR #1083, opened 2016-04-05
authorizator framework

tulinkry:
Approaching #503, this should cover authorization for opengrok.

The core of this change is a plugin framework, which allows users to load their specific plugin to opengrok to perform the authorization.

It follows the pr #1029 - project groupings, so that the authorization can be adapted to restrict access to one project or even group of projects. 

_More detailed documentation (also for the groupings) will follow soon._
### Filter
All urls are covered through the ```AuthorizationFilter``` filter class; listings and output are mainly covered by an UI interface ```ProjectHelper``` which should be ONLY used for getting
filtered projects/groups.

### ClassLoader
The framework introduces a class loader also. Plugin directory can be configured in web.xml through the parameter ```<param-name>authorizationPluginDirectory</param-name>```. Default value is ```/var/opengrok/plugins```. Once upon deploy (or reload) the framework tries to load all classes contained in that directory as plugins. 
 - It accepts both .jar files and .class files. 
 - It's up to the admin to secure this directory from an untrusted access.

### Framework checks
The framework does a simple check. To allow a project/group ALL of the loaded plugins must agree that the project/group is allowed (conjunction behaviour).
If there is no plugin loaded, the project/group is allowed.

### Implementing plugin
For those who would like to implement their own plugin, there is an interface ```IAuthorizatorPlugin```, definining three methods (also described in the code):
 - ```reload()``` - called when the framework reloads plugins
 - ```isAllowed(HttpServletRequest, Project)``` - restricting the project to be visible for this request
 - ```isAllowed(HttpServletRequest, Group)``` - restricting the group to be visible for this request

Object of HttpServletRequest is available with all of its features - like session, attributes, getPrincipal() or getUser().

**NOTE: Allowing a group does not implicitly allow the underlying projects or groups. All such projects/groups must be enabled individually by the plugin.**

### WatchDog service
Also experimentally there is an available service for automate reloads of plugin directory when anything inside changes. One can enable this watch dog in web.xml with parameter ```enableAuthorizationWatchDog```. This should be used for development.


plugins/SamplePlugin.java
```diff
@@ -0,0 +1,26 @@
+import javax.servlet.http.HttpServletRequest;
+import org.opensolaris.opengrok.authorization.IAuthorizationPlugin;
+import org.opensolaris.opengrok.configuration.Group;
+import org.opensolaris.opengrok.configuration.Project;
+/**
+ * Sample authorization plugin.
+ * 
+ * Always just bypass all authorization requests.
+ */
+public class SamplePlugin implements IAuthorizationPlugin {
```
tarzanek: rename to SampleAuthorizationPlugin

plugins/SamplePlugin.java
```diff
@@ -0,0 +1,26 @@
+import javax.servlet.http.HttpServletRequest;
```
tarzanek: add proper CDDL header

build.xml
```diff
@@ -667,11 +669,29 @@ Copyright (c) 2005, 2015, Oracle and/or its affiliates. All rights reserved.
...
+        </available>
+        <condition property="plugins_missing">
+            <not>
+                <isset property="plugins_exists" />
+            </not>
+        </condition>
+    </target>
 
-    <target depends="create_repos" name="-post-compile-test">
+    <target depends="create_repos, create_plugins" name="-post-compile-test">
```
tarzanek: we need to make sure mvn works as well, it's not needed in first approximation, but when you will have a spare time have a look at pom.xml in indexer maven project dir opengrok-indexer

tarzanek: fwiw - the mvn build is used byt Travis ... (hence the motivation to keep it up to date ... we should probably move just to mvn, but we still have lots of unmigrated code in ant, e.g. release cycle, etc.)

src/org/opensolaris/opengrok/authorization/AuthorizationPluginClassLoader.java
```diff
@@ -0,0 +1,207 @@
...
+
+            if (jars == null) {
+                throw new IOException("Directory " + directory + " is not accessible");
+            }
+
+            for (File f : jars) {
+                try (JarFile jar = new JarFile(f)) {
+                    String filename = classname.replace('.', File.separatorChar) + ".class";
+                    JarEntry entry = (JarEntry) jar.getEntry(filename);
+                    //TODO incorect exception handling
```
tarzanek: ???

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -266,6 +266,7 @@ public Configuration() {
         setFetchHistoryWhenNotInCache(true);
         setHandleHistoryOfRenamedFiles(true);
         setRevisionMessageCollapseThreshold(200);
+        setPluginDirectory("/var/opengrok/plugins");
```
tarzanek: this has to use data root variable instead of hardcode

tarzanek:
I merged Vladas code first, please merge with upstream (ping me if you need help, but then github will guide you)

src/org/opensolaris/opengrok/authorization/AuthorizationPluginClassLoader.java
tulinkry: I forgot to remove this when I finished

src/org/opensolaris/opengrok/configuration/Configuration.java
tulinkry: I moved the logic with the default directory to getPluginDirectory

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1100,4 +1217,91 @@ public void run() {
...
+                            if (kind == ENTRY_CREATE) {
+                                reload = true;
+                            } else if (kind == ENTRY_DELETE) {
+                                reload = true;
+                            } else if (kind == ENTRY_MODIFY) {
+                                reload = true;
+                            }
+                        }
+                        if (reload) {
+                            Thread.sleep(2000); // experimental wait if file is being written right now
```
tarzanek: shouldn't we make this a param?

src/org/opensolaris/opengrok/web/ProjectHelper.java
```diff
@@ -116,114 +96,354 @@ private void generateTreeMap() throws IOException {
...
+        return cacheProjects("project_helper_grouped_repositories", all_repositories);
     }
 
+    /**
+     * @see #getProjects()
+     * @return filtered ungrouped projects
+     */
     public Set<Project> getUngroupedProjects() {
-        return projects;
+        return cacheProjects("project_helper_ungrouped_projects", projects);
```
tarzanek: if a string occurs 2x please make it a constant instead
do the same for the rest of the strings that occur 2x
(project_helper_ungrouped_projects, ... )

src/org/opensolaris/opengrok/web/ProjectHelper.java
```diff
@@ -116,114 +96,354 @@ private void generateTreeMap() throws IOException {
...
+    }
+
+    /**
+     * Checks if the project is a favourite project
+     *
+     * @param project project
+     * @return true if it is favourite
+     */
+    public boolean isFavourite(Project project) {
+        return cfg.getCookieVals("OpenGrokProject").contains(project.getDescription());
```
tarzanek: name of the cookie should also be some constant, if it's not, create one in appropriate file please - "OpenGrokProject"

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -87,6 +94,24 @@ public void contextInitialized(final ServletContextEvent servletContextEvent) {
...
+        } else {
+            if (env.getDataRootPath() == null) {
+                env.getConfiguration().setPluginDirectory(PLUGIN_DIRECTORY_DEFAULT);
+            } else {
+                env.getConfiguration().setPluginDirectory(env.getDataRootPath() + "/../" + PLUGIN_DIRECTORY_DEFAULT);
+            }
+            LOGGER.log(Level.INFO, "authorizationPluginDirectory is not set in web.xml. Default location will be used.");
+        }
+
+        String watchDog = context.getInitParameter("enableAuthorizationWatchDog");
```
tarzanek: I think the new params can also be constants
do we describe/document all the params in README or docs? (I think we describe some) Can we add there also these 2 new ones (and we perhaps lack the host:port for config listener)?

web/WEB-INF/web.xml
```diff
@@ -14,9 +14,41 @@
     <description>An address where OpenGrok can receive new configuration</description>
   </context-param>
 
+  <!--
```
tarzanek: I actually wonder if it is better to have this commented or uncommented.
Lots of folks have scripts that just come to web.xml and seds the values, which is easy.
However uncommenting and XML section is harder ... so I guess ... leave it uncommented for people to be able to hack this easily ...

web/WEB-INF/web.xml
```diff
@@ -14,9 +14,41 @@
...
+          Enable watching the plugin directory for changes in real time.
+          Suitable for development.
+      </description>
+  </context-param>
+  -->
+  
   <listener>
    <listener-class>org.opensolaris.opengrok.web.WebappListener</listener-class>
   </listener>
+
```
tarzanek: what is the impact on speed of the page loading?

tulinkry: The filter itself has almost no impact.

The implementation has:
 - checking for project is linear in size of projects
 - if there is such a project then it goes (number of plugins * number of projects)

Further impact depends on the plugin implementations. 
There can be:
 - another linear time for checking allowed projects
 - extra time for creating session if needed
 - another extra time for special calls of the plugin

The framework ensures that for a single request there will be no more than one call to a plugin isAllowed per single project/group (with attribute cache).

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: The documentation for both is described in #1087

src/org/opensolaris/opengrok/web/ProjectHelper.java
tulinkry: I extracted all of the strings in this class as constants. One can use them if this will have to change

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
tulinkry: I extracted this as a constant in the class context.
I wouldn't make it a param because it
 - should be enough always
 - is designed for development only; so anyone can change it while developing plugins

The motivation was just that the modify event is fired when the modification begins (for example compiling) and then framework can load an incomplete class).

web/WEB-INF/web.xml
tarzanek: This needs to be as static as possible. Avoid new object creation, cache everything that can be cached.
Get internally how many requests we get per minute on grok.cz @vladak has the logs. This should be the lowest target for scaling, nothing less.

tarzanek: Oh yes and ideally session should be reused for user, this already happens partially but I expect bugs ... the design is old.
If it will mean going up with servlet specs we should ... tomcat8 is lowest target for webapp.

tarzanek:
I'd say let's merge this and get our hands dirty ;)

PR #1085, opened 2016-04-07
More project prefix matching improvements

kahatlen:
This is a follow-up to the fix for #1074.

Project.getProject() currently creates a copy of the project list and
sorts it on every call. This is more work than necessary. A single
pass through the unsorted project list should be enough to find the
longest matching prefix.

The patch also fixes a bug which would cause getProject() to think
that the path /foof/bar would be in the project with path /foo, if
there was no project which had the path /foof.


tarzanek:
lgtm, please merge!

vladak:
Thanks for fixing the sloppy programming !

PR #1092, opened 2016-04-20
Scopes

mrate:
Moved scopes generation into Xref, removed all *ScopeParser.lex.


tarzanek:
lovely, merging, thank you sir!

PR #1093, opened 2016-04-22
fixing inconsistent behaviour of opengrok cookie

### LINKED ISSUE #1064: Unexpected behaviour of OpenGrokProject cookie
I discovered an unusual behavior of that cookie. A way to reproduce is:

While having three projects (android, old, git):
 - select two projects in multiselect box, press search and cookie becomes OpenGrokProject="android,git"
 - open the third one 'old' by an xref link and cookie becomes OpenGrokProject="old/git,old/android"
 - upon opening any other project (even the original two) with an xref link the cookie grows like OpenGrokProject="git/git/git/git/old/android,old/git" and so on.

Projects in form of old/git... are not recognized correctly for auto select and other features (because it expects comma separator). And also by a coincidence the cookie can grow to its maximum size.

This buggy behaviour seems to be caused by this [list.jsp#94](https://github.com/OpenGrok/OpenGrok/blob/master/web/list.jsp#L94) line where the slash '/' should be replaced with comma ',' because in [PageConfig](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/web/PageConfig.java#L678) the cookie is constructed similar way using comma as a glue.

tulinkry:
fixes #1064 
and may fix related issues

Solution is compatible with tomcat 9 (tested).

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -29,8 +29,10 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


tarzanek:
Looks good, let's merge

PR #1094, opened 2016-04-29
script wrapper for manipulating with groups

tulinkry:
This should help users to manipulate with group configuration if neccessary.

Script is located in tools/Groups and provides several commands
 - ```tools/Groups add``` - adding a group to configuration
 - ```tools/Groups delete``` - deleting a group from configuration
 - ```tools/Groups list``` - listing all available groups
 - ```tools/Groups match``` - match a project name against the groups

General section with usage is available under
```tools/Groups help```
Specific usage with arguments and options is under every command with ```--help``` switch
```tools/Group add --help```

Examples:

1. ```OPENGROK_INSTANCE_BASE=/var/opengrok tools/Groups list``` 

   Lists all groups in /var/opengrok/etc/configuration.xml

2.  ```READ_XML_CONFIGURATION=~/myconf.xml tools/Groups add name pattern```

   Adds a group to the configuration in ~/myconf.xml
   By default it prints the new xml configuration to the stdout. 
   Try options ```-o``` and ```-l``` to manipulate with  the output.

3.  ```tools/Groups add name pattern```

   Creates an empty configuration and inserts the group.

4.  ```tools/Groups add name pattern -i ~/myconf.xml -o ~/newconf.xml```

   You can even specify the input file with the ```-i``` switch and
   output file with ```-o``` switch.

4.  ```tools/Groups add name pattern -i ~/myconf.xml -d```

   ```-d``` performs the dry run - nothing gets changed.
   It just prints the current settings and the underlying command it would execute

**NOTE:**
The command  ```add ``` and  ```delete ``` comes also with ```-u``` option which provides a way to change the configuration IN PLACE. It is still very recommended to make a backup of the original configuration before making any changes with ```-u```, just in case.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Groups.java
+++ b/src/org/opensolaris/opengrok/configuration/Groups.java
@@ -1,0 +1,416 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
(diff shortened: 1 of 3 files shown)


tarzanek:
wondering if it wouldn't make sense to have all this inside OpenGrok script as a subcommand "groups"
but until then, merging
good job!
thank you sir
L

tulinkry:
The opengrok script is long now, so I decided to make this independent and transparent. But I can merge it with the main script.

Also I need to update the docs for groups :)

PR #1097, opened 2016-05-11
use infinite timeout for initial history index

### LINKED ISSUE #1095: use infinite timeout for initial history index generation
Generating history from scratch can be very time consuming process so it often happens that the executor is terminated before the extraction is finished. Those implementations of getHistoryLogExecutor() which take argument specifying history range can take this into account. If complete history is being retrieved then use infinite timeout, otherwise use default timeout.

vladak:
fixes #1095


tarzanek:
lovely, thank you sir!
merging and going to test ASAP

PR #1104, opened 2016-05-24
make parent/branch logging less aggressive

### LINKED ISSUE #1103: make parent/branch logging less aggressive
If branch and parent retrieval fails for repositories which use the Executor class to get the data (such as Mercurial), it might throw IOException which will be logged with SEVERE level. Additionally, the exec() method used will log another warning. This is too much, since the parent and branch are only used to generate the index page.

vladak:
fixes #1103


tarzanek:
Lgtm please push

Issue #1061, opened 2016-02-24
Selection of something else than code results in select-all-code


kosi2801:
Version: git-master as of 2016-02-21

When viewing the source of a file (e.g. after clicking a search result) it is not possible to select something different than lines in the code.
Trying to mark something with the mouse in the header of the page, e.g. "Cross Reference: xyz" or "xref: " restults in having all source itself been selected and not the intended portion. 
This makes it hard/impossible to e.g. select paths or filenames on the result page.

I recognized that this behavior is only present if line-numbers are displayed. When disabling line-numbers via "Line#" the selection of text on the page works as expected.

May possibly be related to issue #791 ?

tulinkry:
This is introduced wrongly in commit 62e437fefec11fd0bc1dd1091cca93b01db9063c  preserving whitespace on copy-paste fixes #540. I noticed that about week ago also and will try to fix it soon.

Problem is that there is some heuristic which should improve user selection but apparently it's not perfect. Quick solution is to disable the space plugin by commenting out it's init call here:
https://github.com/OpenGrok/OpenGrok/blob/master/web/utils.js#L788

PR #1107, opened 2016-06-01
disabling spaces js plugin

tulinkry:
I'm disabling the spaces plugin (copy - paste space problem) until the selection is improved. This can't occur in the next rc.

This temporarily solves the stopper #1061


tarzanek:
merging, thank you, @vladak will you cut a RC now?
(my tests with new universal ctags still run, I will have results in about 10h)

PR #1111, opened 2016-06-08
Added conditional Junit rule

fzakaria:
In order to remove lots of the warnings that exhibit if a developer attempts to run the tests
without some of the repository tools installed, the addition of JUnit rules are used.

If people aren't opposed to the PR I can add the remaining changes to annotate the remaining tests classes/methods.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@ Copyright (c) 2010, 2015, Oracle and/or its affiliates. All rights reserved.
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.7.4.201502262128</version>
+                <version>0.7.7.201606060606</version>
             </plugin>
 	    <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@ Copyright (c) 2010, 2015, Oracle and/or its affiliates. All rights reserved.
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.7.4</version>
+                <version>0.7.4.201502262128</version>
             </plugin>
 	    <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/README.txt
+++ b/README.txt
@@ -883,6 +883,24 @@ you should be able to connect your Github to Travis CI.
 OpenGroks Travis is here: https://travis-ci.org/OpenGrok/OpenGrok
 
 
+9.9 Maven
+------------------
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -150,7 +150,7 @@ Copyright (c) 2010, 2015, Oracle and/or its affiliates. All rights reserved.
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.7.4</version>
```
tarzanek: d'oh!
I forgot to move this up, we actually do support and run on jacoco 0.7.5 and looking at jacoco pages, we can just go to 0.7.7 (I will retest in our internal farm and let you know)

tarzanek:
looks lovely, I would love to merge immediately, 
one thing/blocker however - I need you to either sign, scan and email OCA to Oracle ( http://www.oracle.com/technetwork/community/oca-486395.html )
or say(write ;) ) here that you're OK with OCA for this patch

pom.xml
tarzanek: confirming, feel free to go to 0.7.7, works just fine

tarzanek:
fwiw - one more thing - I think we have much more tests that depend on mercurial
also - I'd love to have a switch to disable this conditional run - is there a way to disable those annotations easily ?

tarzanek:
re-disabling annotations - I can see javac has lots of -proc* options, but I think we should have an option to just disable the ones for ConditionalRun and leave others (whatever they may be)
- this should be documented well before we turn this on (I suppose for general acceptation ALL junit tests need to be passed, sometimes your code change can impact parts of junits you'd never dreamed off) - naturally travis will build all, but it has dependencies and I'd love to have a way to check if dependencies didn't change, ev. check them on a clean box (and if my test count decreases, I'd like to see which tests fail/have disabled themselves)

tarzanek:
the other thing is that our build process for junits depends on certain binaries (hg for sure, perhaps git too) to prepare the testing data from testdata folder ... so some dependencies will still be there ... but I agree these decorators will be put e.g. on sccs or cvs repos ;)

fzakaria:
I am OK with OCA for this patch

fzakaria:
@tarzanek I really appreciate the quick feedback.

1. I've `OK` the OCA for this patch 
2. I've added the system property switches
3. Updated the jacoca version

I have a few more changes/PR that I can submit but I'd like to keep each one small & tidy.

I will look into the creation of the `src.zip` created as part of the build script - perhaps a more idiomatic way through maven.

tarzanek:
If you have more patches then I suggest sending back a scan/photo of signed OCA to be fully ok with legal.
Note that we still use ant primarily but I do try to keep mvn updated(ant still does tasks that are harder/less flexible to do in mvn)

tarzanek:
Thank you @fzakaria merging ... I will add ant part tomorrow.

PR #1112, opened 2016-06-09
Addition of ClojureAnalyzerFactoryTest

fzakaria:
I noticed that when I was running the tests, it did not look like it was analyzing the Clojure code correctly.
I was seeing plenty of errors of the type `repetition-operator operand invalid` wth the **ctags** regex provided.

Grokking around, I found similar ctag regex that didn't result in the error and I wrote a test to verify that the definitions are correctly resolving.

I also added [Maven Wrapper](https://github.com/takari/maven-wrapper) because I think it is good practice for any project that uses Maven.

FWIW here is my version of ctags:
> ctags --version
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Sep 13 2015, 03:28:50
  Addresses: <dhiebert@users.sourceforge.net>, http://ctags.sourceforge.net
  Optional compiled features: +wildcards, +regex


tarzanek:
Looks ok but
We need to resolve the license headers to add to new files

Also the sample clj file - it is under cpl license, I need to figure out if it can be merged with cddl.
Other option is to find some other sample clj file under bsd/mit license or write something by yourself.

Also ideally get the oca to Oracle inboxes, I will ping the oca owner for opengrok to handle your signature quickly.

Ttyl
L

tarzanek:
Also the regexp needs to be posix compliant so ctags on Solaris or other posix system will do correct things

tarzanek:
And travis fails - https://travis-ci.org/OpenGrok/OpenGrok
since you added new file to sources the generic testing now picks up more definitions (10 instead of 7, you need to verify if it is true + if you change for some other file you need to check and retest it)

fwiw I use universal ctags - https://github.com/universal-ctags/ctags 
they are much better

tarzanek:
and I just recieved your OCA, thank you for that! :)
now we just miss the proper license headers in respective files

tarzanek:
fwiw - the skipped tests are nicely reported in jenkins test UI :)

src/org/opensolaris/opengrok/history/CVSRepository.java
```diff
@@ -67,6 +67,7 @@
             = Pattern.compile("([\\.\\d]+)\\W+\\((\\w+)");
 
     public CVSRepository() {
+        working = Boolean.FALSE;
```
fzakaria: this should not default to true (which RCS from it inherits does)

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
```diff
@@ -170,7 +170,7 @@ public void run() {
         });
         t.start();
         t.join();
-        assertEquals(new File(path).getCanonicalFile().getAbsolutePath(), instance.getDataRootPath());
+        assertEquals(new File(path), new File(instance.getDataRootPath()));
```
fzakaria: my /tmp folder was symlinked and had errors from canonical path. 
This test is more resilient.

fzakaria:
@tarzanek thanks for looking at it. Not sure why I didn't see the travis build earlier.
I tried to fix some errors however its tough to get a "perfect" build on my machine due to the myriad of settings that are configurable (derby not present etc..)

1. I replaced the clojure file with some home-grown sample
2. Fixed as many errors as I could that might have been on account of the inclusion of the new clojure file (i.e. Searcher)
3. Fixed some additional errors I had when running `mvn test`

Is the jenkins build private ? Travis doesn't show a test JUnit breakdown from the reports.
Also, one of the other developers said they would add `junit-force-all` to the Ant build - meaning that Jenkins should probably not be reporting skipped tests.

tarzanek:
@fzakaria sorry for travis, my fault, github introduced new permissions and I forgot to give part of them to travis, now travis should be seen again in pull reqs.
jenkins build is unfortunately private, but I will just add the junit-force-all to ant, for now it's ok like it is

src/org/opensolaris/opengrok/analysis/Ctags.java
```diff
@@ -148,20 +148,20 @@ private void initialize() throws IOException {
...
-                command.add("--regex-clojure=/\\([[:space:]]*defmacro[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/m,macro/");
-                command.add("--regex-clojure=/\\([[:space:]]*definline[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/i,inline/");
-                command.add("--regex-clojure=/\\([[:space:]]*defmulti[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/a,multimethod definition/");
-                command.add("--regex-clojure=/\\([[:space:]]*defmethod[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/b,multimethod instance/");
-                command.add("--regex-clojure=/\\([[:space:]]*defonce[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/c,definition (once)/");
-                command.add("--regex-clojure=/\\([[:space:]]*defstruct[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/s,struct/");
-                command.add("--regex-clojure=/\\([[:space:]]*intern[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/v,intern/");
-                command.add("--regex-clojure=/\\([[:space:]]*ns[[:space:]]+([-[[:alnum:]]*+!_:\\/.?]+)/\\1/n,namespace/");
+                command.add("--regex-clojure=/\\([ \\t]*create-ns[ \\t]+([-[:alnum:]*+!_:\\/.?]+)/\\1/n,namespace/");
```
tarzanek: please use :space: and :alnum: as defined in posix for regexps, don't use " \\t"

tarzanek: https://en.wikipedia.org/wiki/Regular_expression#Character_classes

tarzanek:
fwiw - all looks fine, I just don't like that you switch char classes to just few chars, that said clojure might be using the other whitespace so I might be wrong, but still, if possible please make sure the regexps are posix compliant, otherwise your code/ctag def will fail terribly on posix systems such as Solaris (ctags consume posix library there, so they behave slightly different than on linux where they use the gnu version of regexp library)

fzakaria:
@tarzanek updated the ctags regex to be POSIX compliant.
(test still passes, which I'm taking as a sign its working - hurray for tests?)

tarzanek:
thank you, merging, will recheck the tests on the private jenkins (which runs on solaris 12)

tarzanek:
fwiw - some change is still needed, since tests fail for me on Solaris, so I will just go ahead and commit the necessary posix changes

tarzanek:
ok, reason is actually that universal ctags has clojure parser which isn't as good as the gist we use
I filed https://github.com/universal-ctags/ctags/issues/988
let's see if they resolve it, until then I will just use clojure from opengroks defs

fzakaria:
awesome.

PR #1117, opened 2016-06-20
Changes to authorization framework

tulinkry:
I've come with some necessary changes to the framework design.

1. Method ```reload``` in ```IAuthorizationPlugin``` is split into ```load``` and ```unload```.
     - upon framework reload every plugin is destroyed and then instantiated again,
      so it is more load/unload than reload.
     - the ```unload``` method can be used to close any connection held by the plugin (db/ldap)
2. Shortcuts for manipulation with groups/projects
     - ```Group.getParents()```
     - ```Project.getGroups()```
     - ...
3. Whitelisting some opengrok useful classes - from logger and utils

Doc is changed as well.


tarzanek:
thank you, merging! :)

PR #1129, opened 2016-07-07
improving size of the projects select box

### LINKED ISSUE #1118: The project select box's size does not handle the groupings
The size of the project select box is approximated from the number of projects (something like Math.min(15, project.count). This does not work with the groupings design (shown on the screenshot)

screenshot from 2016-06-20 18 35 14

The group "admins" contains 5-6 other projects but the select box shrinks to two lines.

tulinkry:
fixes #1118 

With the groupings it's necessary to compute the number of projects before it's actually rendered.


tarzanek:
thank you
merging

PR #1130, opened 2016-07-07
adding scopes css

### LINKED ISSUE #1108: css print styles are broken with the scopes
Line numbers are spread in the content like this: (using default/print.css)

screenshot from 2016-06-01 17 46 33

Looks like css print styles don't know about scopes and fail to display it

tulinkry:
fixes #1108 

print styles were forgotten with the new scopes css styles


tarzanek:
thank you sir!
merging

PR #1131, opened 2016-07-07
sample authorization plugin does not implement the interface

tulinkry:
The SampleAuthorizationPlugin did not implement the IAuthorizationPlugin interface. This was introduced by #1117 but forgotten by travis.

Also:
Regading the authorization framework we have two new wikis:
- https://github.com/OpenGrok/OpenGrok/wiki/OpenGrok-Authorization
- https://github.com/OpenGrok/OpenGrok/wiki/OpenGrok-Groupings

Containing fairly similar things like in doc directory but with .md desing.

### DIFF
--- a/plugins/SampleAuthorizationPlugin.java
+++ b/plugins/SampleAuthorizationPlugin.java
@@ -32,8 +32,13 @@ import org.opensolaris.opengrok.configuration.Project;
 public class SampleAuthorizationPlugin implements IAuthorizationPlugin {
 
     @Override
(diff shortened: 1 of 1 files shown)


tarzanek:
so shouldn't the sample plugin be built somehow or tested?
that it implements all the ifaces?

tarzanek:
I will merge for now, but I don't like the fact travis doesn't test this build, let's include the build of sample plugin in travis/mvn

PR #1133, opened 2016-07-12
adding sample plugins to maven build

tulinkry:
mentioned in #1131

I'm not sure if this is the correct way but I tested it with netbeans and jenkins builds

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -417,7 +417,7 @@ Copyright (c) 2005, 2016, Oracle and/or its affiliates. All rights reserved.
         dest="lib/${servlet-api.jar}" verbose="true" usetimestamp="true"/>
     </target>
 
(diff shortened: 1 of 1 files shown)


tarzanek:
lgtm, travis looks ok, which is the only thing needed
thank you sir!

tarzanek:
AH, you just added it to ant build
maven build happens in pom.xml
anyways I am going to accept it 

fwiw - check OpenGrokMaven job in internal jenkins on how to use maven with OpenGrok

tarzanek:
so the commit name is a bit misleading, but I guess we will just let it go
please look at pom.xml and adjust so this test happens in mvn (travis is based on mvn) too

tulinkry:
I don't see it in the travis build, I will add it

Issue #794, opened 2014-03-11
generate history index for renamed files in git repo separately


vladak:
Just like it was done for Mercurial in #22 (and bunch of follow-up changesets) it should possible to follow complete history of renamed files in git (and provide diffs between revisions of a file across renames).

caminada:
I have a partial fix for this issue (the easy part...)

It's not yet implemented in the Indexer, aka creating of historycache at indexing time is not yet working (workaround: use option -k "*" to prevent cache building at all)

The second part has to be implemented in GitHistoryParser, but it's not as easy as in MercurialHistoryParser because git does not track renamed file with a tag like "file_copies:"
It will probably be very time consuming to iterate trough the full history of *all* files to recognize path changes for cache building...

tulinkry:
Implemented in #1345

caminada:
yeah! thank you @tulinkry 🥇

PR #1135, opened 2016-07-14
Show history in git beyond renames

smehrbrodt:
Using `git log --follow`


vladak:
Not sure this is as easy as adding --follow. Can you check what happens if you try to diff 2 revisions across renames in the history view ? Also see #794.

caminada:
Hi @vladak @smehrbrodt

Adding "--follow" is really not enough. I have a patch here for #794 that I was working on last year or so. It's currently against 0.12 - I'll try to convert it for master and make a separate pull request.

vladak:
Okay, that's ggod to hear. Clising this one.

PR #1139, opened 2016-07-19
adding sample plugins to maven build

tulinkry:
once again and better
#1133 added sample plugins to ant build
this PR should hopefully add the plugins directory into maven build and track it with travis-ci

(https://travis-ci.org/OpenGrok/OpenGrok/builds/145835366#L3581)


plugins/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.opensolaris.opengrok</groupId>
+    <artifactId>plugins</artifactId>
+    <packaging>jar</packaging>
+    <version>0.13-SNAPSHOT</version>
```
tarzanek: does this version need to be hardcoded? can it be inherited? or project.version used?

plugins/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <name>OpenGrok Sample Plugins</name>
+
+    <parent>
+        <groupId>org.opensolaris.opengrok</groupId>
+        <artifactId>opengrok</artifactId>
+        <version>0.13-SNAPSHOT</version>
```
tarzanek: dtto as above

plugins/pom.xml
tulinkry: This version can be ommited which means it gets inherited. However I unified the style with the other pom files in other submodules where this is always present.

plugins/pom.xml
tulinkry: I dug a bit but it seems this cannot be replace by an expression. I've found some workarounds but they don't work in every case.

As above, I got inspired by the other pom files in the opengrok (opengrok-indexer, opengrok-web) where this is not solved.

tarzanek:
will merge for now, however I'd love to have the version just in one pom

Issue #1136, opened 2016-07-18
hide previous search results upon new search


vladak:
just like most search engines do

vladak:
This was fixed in PR #1140 cset dab33d94273a0b00747cffdb96bb7a4cb30ecf8c

PR #1140, opened 2016-07-19
hiding previous results after search click

tulinkry:
implements #1136 

Tested in Firefox 39.0 and Explorer 9.0. However this should work everywhere. 

internal demo: http://virtual99.cz.oracle.com/clear/


tarzanek:
thank you, merging, we need jslint, we have too much js code ...

PR #1141, opened 2016-07-19
Allow "@" in checkout paths

caminada:
- allow to checkout repositories with paths containing "@" (e.g.  sources/OpenGrok@master, sources/OpenGrok@0.12-stable)

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -198,14 +198,13 @@ public final class PageConfig {
          * The code below extracts file path and revision from the URI.
          */
         for (int i = 1; i <= 2; i++) {
-            String[] tmp = null;
             String p = req.getParameter("r" + i);
             if (p != null) {
-                tmp = p.split("@");
-            }
-            if (tmp != null && tmp.length == 2) {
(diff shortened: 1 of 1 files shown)


caminada:
Diffs don't work if the checkout path contains "@". With the above patch it's possible to have repository aths (and project names in the web GUI) like sources/OpenGrok@master or sources/OpenGrok@0.12-stable

tarzanek:
thank you, looks good
I need you to agree with OCA here for this patch(since this diff is small), or print/sign/scan/mail it to Oracle (and future patches will be accepted automagically) - http://www.oracle.com/technetwork/community/oca-486395.html

caminada:
Ok, I've sent the signed OVA but didn't hear anything from them yet.

tarzanek:
yeah, it takes a day or two, I will know it asap, since the person responsible always pings me or vladak first thing, let's hope this person doesn't do vacation :)

tarzanek:
hi @caminada 
they pinged me that the pdf that came in was broken (unreadable, last part of it missing, hence they couldn't open it :( )
can you please resend?

tarzanek:
i just recieved the oca
thank you, merging

PR #1144, opened 2016-07-21
Fix semantics of -H flag and disabling history generation

joelagnel:
OpenGrok generates history regardless of what OPENGROK_GENERATE_HISTORY is set
to, for local repositories.  This variable currently only goes into effect for
remote repositories.  Change its semantics such that if -H is not passed, then
history generation and caching is disabled altogether for all repositories.
This is useful when indexing large local repositories such as a local mirror of
the AOSP sources worth 70GB. History generation in this case results in the
indexing never completing.

We also fix the semantics of OPENGROK_GENERATE_HISTORY in the wrapper script,
such that if its set to "off", then we don't pass -H at all to OpenGrok.

Signed-off-by: Joel Fernandes <agnel.joel@gmail.com>

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -268,6 +268,11 @@ DefaultInstanceConfiguration()
         GENERATE_HISTORY="-r ${OPENGROK_GENERATE_HISTORY}"
     fi
 
+    if [ "$OPENGROK_GENERATE_HISTORY" != "off" ]
+    then
+        GENERATE_HISTORY="${GENERATE_HISTORY} -H"
+    fi
+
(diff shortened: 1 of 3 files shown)


tarzanek:
lgtm, however before I can merge I need you to either accept OCA here for this patch, or print, sign, scan and mail it back to Oracle so for next patches we won't have this requirement
http://www.oracle.com/technetwork/community/oca-486395.html

joelagnel:
Thanks. I accept OCA for this patch.

tarzanek:
Thank you sir! Merging

PR #1154, opened 2016-08-01
adding more complex plugin example 

tulinkry:
As described on [wiki](https://github.com/OpenGrok/OpenGrok/wiki/OpenGrok-Authorization#http-basic-tutorial) I added a more interesting example of a plugin containing group discovery.

Also mentioned in the tutorial that this plugin **does not cache anything** and traverses through the group structure every time it is asked for a decision - which is not desirable for production mode.

This plugin can be used as an inspiration for other user defined plugins.


tarzanek:
thank you, merging

PR #1156, opened 2016-08-03
fixing clear button

tulinkry:
The clear button wasn't working at all after moving to jquery 3.1.0. Obviously it couldn't select the [type=text] without having the inputs as text.

Moreover recalling #1136 I forgot to hide the suggestions - fixed now.


tarzanek:
dekuju moc, dobrou noc! :)

PR #1160, opened 2016-08-15
new fold icon when group of projects is collapsed

tulinkry:
In order to support better graphical intuition when viewing the main page with groups I added fold icons indicating which group is collapsed and which is not.

Screenshots:
screenshot from 2016-08-15 14-55-38
screenshot from 2016-08-15 14-56-29

It is also slightly responsible.

Tested in IE 9, Chrome 52, Firefox 48.

Internal demo available [here](http://virtual99.cz.oracle.com/arrows/)


tarzanek:
thank you sir, merging!

PR #1161, opened 2016-08-15
Multiselect

tulinkry:
New UI enhancement for the old multiselect.

This uses a code from https://github.com/pbauerochse/searchable-option-list which is a little bit edited to fulfil the requirements and to behave like the old one.

JS and CSS files are prepared minified for production but the normal version is also prepared for development.

Screenshots:

 - Selected projects are displayed on the right side of the selectbox.
screenshot from 2016-08-15 17-42-33

 - Selection can be done with mouse, keyboad. It is possible to live search for projects and select them with keyboard. Also old hacks like double click on project, double click on group, enter pressing for the first selected project are preserved.
screenshot from 2016-08-15 17-42-43

 - Upper limit for number of displayed projects is 30 now and can be easily configured. Also this is a point of discussion, currently 30 projects selected looks like this:
screenshot from 2016-08-15 17-42-54
 - The display is close to be responsible - it behaves like a three column layout. On medium displays the visualization of the selected items can be a little confusing because it is displayed behind the selectbox but is fully functional (user can close the selectbox)
screenshot from 2016-08-15 17-54-06

On my machine the latency when selecting ~300 projects is about 1s which is fair.

Tested in IE9, Chrome 52, Firefox 48.

Internal demo [here](http://virtual99.cz.oracle.com/multiselect/)


tarzanek:
@tulinkry can you resolve the conflicts ? 
thnx
L

tarzanek:
(other than the conflicts I am OK with this change, it looks nice)

tulinkry:
resolved

tarzanek:
thank you sir, merging!

naseer:
When I first open the page, all my projects are selected by default. How can I get it to select the last searched project? Also, is there an easy way to disable this feature without reverting these commits?

vladak:
@naseer is this with recent master branch ?

tulinkry:
The projects are stored in a cookie, try to investigate that cookie or delete your browser cookies (`ctrl + shift + delete` on firefox) to start with a clean sheet. This works no matter this multiselect plugin, it's just a javascript soup around the original project picker.

If you still want to disable this behaviour (or any other recent javascript plays in OpenGrok) then comment out its init line in `utils.js`. In this case:
```javascript
    $('#project').searchableOptionList(searchableOptionListOptions);
```

naseer:
@tulinkry Thanks - I missed emails from this thread in github - this is on a recent master build (around 12 days old)

I tried clearing the cookies - it doesn't appear to work in Chrome. I was looking for a cleaner way(configurable switch) to go back to the old behaviour - since I have only a few projects - it used to be a one click action to select the project and then search. Now it is a minimum of two clicks - clear selection, select the project from the dropdown, then search.

tulinkry:
Reverting to the old project picker is made simpler as described in #1255. That preserves the original click elements.

PR #1145, opened 2016-07-21
jquery 3.1.0

tulinkry:
I upgraded the version of jquery in OpenGrok. I used that jquery migrate script for going up to 3.1.0.

**Migration caused that the depending javascripts also needed to be updated**
- tablesorter -> old => used its up-to-date fork
- tooltip (annotations) -> old => replaced by tooltip in jquery-ui

**Moreover I organized the js code in a special js subdirectory so it is more clean.**

There are couple of js and css files now, maybe we should consider to minify them for the production.

**There are only two deprecated problems, however they are not in our code.**
From my point of view this is ok, we can just wait when the two modules will release their next fixed version. Tablesorter seems very active, I believe that jquery-ui is as well.
- One is in the new version of tablesorter
```
"JQMIGRATE: jQuery.fn.unbind() is deprecated" jquery-migrate-3.0.0.js:62:4
console.trace(): jquery-migrate-3.0.0.js:64
migrateWarn() jquery-migrate-3.0.0.js:64
.unbind() jquery-migrate-3.0.0.js:423
a.tablesorter.bindEvents() jquery.tablesorter-2.26.6.min.js:2
a.tablesorter.setup() jquery.tablesorter-2.26.6.min.js:2
a.fn.tablesorter/<() jquery.tablesorter-2.26.6.min.js:3
.each() jquery-3.1.0.min.js:2
r.prototype.each() jquery-3.1.0.min.js:2
a.fn.tablesorter() jquery.tablesorter-2.26.6.min.js:3
<anonymous> utils.js:463
g/</j() jquery-3.1.0.min.js:2
g/</k<() jquery-3.1.0.min.js:2

"JQMIGRATE: jQuery.fn.bind() is deprecated" jquery-migrate-3.0.0.js:62:4
console.trace(): jquery-migrate-3.0.0.js:64
migrateWarn() jquery-migrate-3.0.0.js:64
.bind() jquery-migrate-3.0.0.js:419
a.tablesorter.bindEvents() jquery.tablesorter-2.26.6.min.js:2
a.tablesorter.setup() jquery.tablesorter-2.26.6.min.js:2
a.fn.tablesorter/<() jquery.tablesorter-2.26.6.min.js:3
.each() jquery-3.1.0.min.js:2
r.prototype.each()
```
- Second is in the new version of jquery-ui
```
"JQMIGRATE: jQuery.unique is deprecated, use jQuery.uniqueSort" jquery-migrate-3.0.0.js:62:4
[...]

PR #1171, opened 2016-08-19
new intelligence window

tulinkry:
The intelligence window is under new jquery coat. That should make the further maintenance and usability easier.

It comes with some new features, mainly
 - the window is by default draggable 
    - this can be turned off with:
    ```
    $.intelligenceWindow.init({draggable: false});
    ```

 - previous/next links for jumping among the symbols in a file

The doc is updated for the intelligence window and also for older feature - diff jumper.

Internal demo [here](http://virtual99.cz.oracle.com/intelligence-window/)


tarzanek:
I like this
also I think it could be used for onMouse over changeset logs when doing annotate ... (to get a better popup window than default hover)

tarzanek:
+ please fix the junit, it expects a bit different xref

tulinkry:
> please fix the junit, it expects a bit different xref

It should be ready now

> also I think it could be used for onMouse over changeset logs when doing annotate ... (to get a better popup window than default hover)

With the update of jquery to 3.1.0 (#1145) I added a new version of tooltip from jquery-ui which seems fine to me right now. Maybe it could be enough to just play with the options of that tooltip

tarzanek:
thnx, merging

PR #1174, opened 2016-08-25
removing regression of eee3266

### LINKED ISSUE #1173: viewing file in specific revision is broken
When viewing file in specific revision (file -> annotate -> click on specific revision) it displays wrong information about the lines. A way to reproduce is to index the opengrok source code and search for some file, for example: `src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java`, and display some specific revision.

### Annotate
screenshot from 2016-08-25 12-59-16

### Clicking on revision `6c5588de` - current result
screenshot from 2016-08-25 13-00-54

### Clicking on revision `6c5588de` - expected result
screenshot from 2016-08-25 12-59-32

tulinkry:
fixes #1173

During this I've discovered [GitRepository.java#L272-L318](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/history/GitRepository.java#L272-L318), which is the part of the code which inserts different annotations for the specific lines when the revision number is wrong. The git is run with `-C` option searching for lines in the same commit which also got changed. Problem is like in #1173 that it inserts cruelly out of the way results into the file.

**NOTE: the problem happens still when the revision number is wrong (i. e. when user change the parameter `r` in url to whatever)**

It's hard to convert this to unit test - the lines are always eventually found even when they're wrong - so can anyone explain me the intention here?

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -642,7 +642,7 @@ public final class PageConfig {
         }
         getRequestedRevision();
         try {
-            annotation = HistoryGuru.getInstance().annotate(resourceFile, rev.isEmpty() ? null : rev.substring(2));
+            annotation = HistoryGuru.getInstance().annotate(resourceFile, rev.isEmpty() ? null : rev);
         } catch (IOException e) {
             LOGGER.log(Level.WARNING, "Failed to get annotations: ", e);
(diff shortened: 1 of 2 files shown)


tarzanek:
well I don't know the intention either ...
@trondn ping! we need your help here ...
I will merge this and if @trondn will clarify we can always improve your strict fix

PR #1186, opened 2016-09-16
reuse IndexReader objects for multi-project searches

### LINKED ISSUE #1132: consider reusing IndexSearcher objects
In issue #1116 there was a suggestion to find out if the initial part of `searchMultiDatabase()` where `FSDirectory.open` is called and `MultiReader` is instantiated could be called just once / after each reindex. For webapp, the place for potential modification is `SearchHelper#prepareExec()`.

If this is doable, this should speed up multi-project searches with high number of projects significantly (assuming file system cache is warmed up).

vladak:
This fixes #1132 and speeds up multi-project search by doing the right thing w.r.t. IndexSearcher object reuse as stressed in just every book about Lucene.

There are 2 problems with this change:
  - the `destroy()` method should be called after each SearchHelper or SearchEngine is used for searching, ideally in `try-finally` block. However, this is not easy to do, namely in `search.jsp` where the code is interspersed with `include` blocks etc. It is important to call `destroy` to make sure all `IndexSearcher` objects are properly returned to their corresponding `SearcherManager` objects. I briefly considered to use Finalizer however all references dissuade everyone from using it so I chickened out.
  - the relationship between IndexSearcher/SeracherManager is a bit clumsy. The SearcherManager's are kept in a project map in RuntimeEnvironment and the consumers (SearchHelper or SearchEngine) have to keep their own IndexSearcher map for multi-project searches so that they can return the IndexSearcher objects in `destroy()`. Also, invalidating SearcherManager objects when corresponding project disappears after configuration change is not done to keep things simple. As a result, the map grows unbounded.


vladak:
Another problem is partial reindex - usually it is done from the `OpenGrok` shell script with environment variable `OPENGROK_WEBAPP_CFGADDR=none` so the webapp has no way of knowing that refresh is needed. This will be taken care of after some time (more precisely no later than after `getIndexRefreshPeriod()` timeout) thanks to the `indexReopenThread` background thread periodically calling `maybeRefresh()` on all SearcherManager objects. I filed #1187 to track this.

test/org/opensolaris/opengrok/web/SearchHelperTest.java
```diff
@@ -18,18 +18,159 @@
...
+import org.opensolaris.opengrok.index.IndexerTest;
+import org.opensolaris.opengrok.search.QueryBuilder;
+import org.opensolaris.opengrok.util.TestRepository;
 
 /**
  * Unit tests for the {@code SearchHelper} class.
  */
 public class SearchHelperTest {
+    TestRepository repository;
+    private final String ctagsProperty = "org.opensolaris.opengrok.analysis.Ctags";
```
tarzanek: dup?
can we take it from a final variable from code instead of redeclaring?

test/org/opensolaris/opengrok/web/SearchHelperTest.java
```diff
@@ -18,18 +18,159 @@
...
+    public static void setUpClass() throws Exception {
+    }
+
+    @AfterClass
+    public static void tearDownClass() throws Exception {
+    }
+
+    @Before
+    public void setUp() throws IOException {
+        repository = new TestRepository();
```
tarzanek: we also duplicate this on quite a few places, dedup might be nice (but for this dup not in this changeset, just for future)

tarzanek:
ship it!
let's get our hands on it and let's get dirty! :)

vladak:
I tried to address the `destroy()` issue as suggested by @tulinkry by adding a hook into `requestDestroyed()` and by wrapping the search engine in JSONSearchServlet into try/finally block.

test/org/opensolaris/opengrok/web/SearchHelperTest.java
vladak: filed #1198 to track this

PR #1194, opened 2016-09-27
fixing auto focus on content area

tulinkry:
After the update to jquery 3.1.0 there is a visible focus on the content area of the page.

It's the line between the search bar and the content.
screenshot from 2016-09-27 11-40-57

For me it's not necessary and it doesn't look nice, so that's why this diff removes it. I hope there is not some special action connected to this.

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -425,9 +425,8 @@ input.submit { /* start search button , clear button */
     left: 0;
     right: 0;
     overflow: auto;
-    padding: 1ex; /* should be the same as body's margin */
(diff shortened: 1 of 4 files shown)


tarzanek:
this is purely for people who scroll with spacebar ...
without this focus you cannot scroll with keyboard ...

tulinkry:
Ok. I'll find a different way around this.

tulinkry:
This leaves the old functionality (maybe added more because the space navigation wasn't working for me after f5)

Tested: recent Chrome, recent Firefox, IE9

tarzanek:
lgtm, thank you sir!

PR #1196, opened 2016-09-30
using --date option when retrieving git history

tulinkry:
fixes #1193

Using `--date=rfc` option for `git log` command overwrites user configuration and always retrieves the dates in the rfc2822 format.

Covered with a test case.

PR #1197, opened 2016-09-30
allowing more date formats for svn date parsing

tulinkry:
Like in #1196 for git repositories, this extends the number of valid date formats for svn repositories. Which should prevent unlikely events of slightly different date format in the repository history.


tarzanek:
thank you sir, lgtm, let's get it into test

PR #1195, opened 2016-09-27
messaging addition

tulinkry:
The configuration listener seems unemployed most of the time so I added a messaging feature which listens on the same socket.

As a first shot it is capable of two things:

1.  send "normal" and "abort" messages to the web application
2.  display some useful information about projects

The information are displayed:

1. On the main page just above the project list - for general messages (tag 'main')
2. In the directory listing just above the listing - for project specific messages (tag $projectName)
3. In the xref/history/... page just above the content - for project specific messages (tag $projectName)
4. After the search there is a (!) mark indicating that there are messages in this project

Currently supported messages:

1. **normal**
    - can display information on the page
    - can be general (with tag 'main' - default) or project specific (tag with project name)
    - the color can be controlled with the css class
2. **abort**
    - can erase some messages from the system
    - basically it erases all messages with at least one tag from the tags of this abort message

**The messages can't enforce any kind of restrictions**, they just display/delete some user content - they are useful to notify the clients that index/reindex or anything else is in progress and so some inconsistencies can occur.

**All messages disappear with webapp deploy. They are current state of the runtime environment of the OpenGrok**

Related to #1187. This is now possible with just implementing `Message` subclass with desired effect.

It comes with a java main file to automatically send the message to the application and also with a shell wrapper to make it even more easier. The shell wrapper is located under `tools`. 

This shell wrapper only needs to find your `opengrok.jar` java file so it can run the `Messages.java` main class in it. This is possible by setting `OPENGROK_DISTRIBUTION_BASE` environment variable which points to the OpenGrok distribution directory (which contains `dist/opengrok.jar`). Other options are available through set of switches.

Usage from the command line is simple:
```bash
$ tools/Messages --help # see usage and defaults
$ tools/Messages "Hello" # send a message with text "Hello"
$ tools/Messages -e "+1 min" "Hello" # set expiration to 30 minutes from now
$ tools/Messages -e "+30 min" "Hello" # set expiration to 30 minutes from now
$ tools/Messages -c "error" -e "+30 min" "Hello" # display the message as error
$ tools/Messages -c "warning" -e "+30 min" "Hello"  # display the message as warning
$ tools/Messages -c "warning" -e "+30 min" --type abort "Hello" # send abort message
```

Here come some screenshots:

1. General message
screenshot from 2016-10-06 16-45-56
2. Warning in project picker
screenshot from 2016-10-06 16-46-09
3. All warnings (!) has mouseover table with the messages
screenshot from 2016-10-06 16-46-44
4. Searching
[...]

PR #1203, opened 2016-10-06
package Groups script

vladak:


tulinkry:
Yes, I noticed that like month ago. The same goes for the upcoming message script in #1195 .

tarzanek:
lgtm please merge

PR #1229, opened 2016-10-12
include Groups and Messages scripts

### LINKED ISSUE #1225: Messages script missing from release build
I just downloaded the latest release candidate and would like to try out the new messaging functionality.

Walked into a problem since I'm running the release build and the new Messages script (`tools/Messages`) is not included in the release build (`opengrok-0.13-rc3.tar.gz`). It would be nice if it was copied over to the `bin` folder.

vladak:
fixes #1225

Also fixes permissions for the OpenGrok script.


tarzanek:
thank you sir!

PR #1230, opened 2016-10-12
store reference to SearcherManager in IndexSearcher super-class

### LINKED ISSUE #1207: Reusing IndexSearcher - possible enhancement
Reusing searchers as introduced in #1186 has some drawbacks:
1. The map of SearcherManagers grows indefinitely (however not really fast)
2. Once project is deleted and restored again, it can't be searched without webapp restart

These problems are cause by the multithreaded application and that the configuration change can arrive at anytime.

I suggest considering following solution which should theoretically work:
- subclass IndexSearcher and add get+set methods for an instance of SearcherManager
- adapt the ThreadPoolSearcherFactory so it returns the new subclass
- when acquiring new IndexSearcher from the manager, cast it to the subclass and inject the manager
- when releasing do similarly and release the searcher

This technique should not have the problem with the unbounded map - the closed instance of SearcherManager can be removed from the map while every acquired IndexSearcher already holds the instance of SearcherManager where it should return itself. Also it should solve the problem with deleted+restored projects as SearcherManager for deleted project could be immediately removed from the map.

vladak:
fixes #1207

It seems this works. Tested by repeating multiple searches (with positive and none hits), interleaved with reindex.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/SuperIndexSearcher.java
+++ b/src/org/opensolaris/opengrok/configuration/SuperIndexSearcher.java
@@ -22,23 +22,33 @@
   */
 package org.opensolaris.opengrok.configuration;
 
-import java.io.IOException;
+import java.util.concurrent.ExecutorService;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.search.IndexSearcher;
-import org.apache.lucene.search.SearcherFactory;
+import org.apache.lucene.search.SearcherManager;
 
 /**
- * Factory for producing IndexSearcher objects.
- * This is used inside getIndexSearcher() to produce new SearcherManager objects
- * to make sure the searcher threads are constrained to single thread pool.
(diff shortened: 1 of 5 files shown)


src/org/opensolaris/opengrok/configuration/SuperIndexSearcher.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.opensolaris.opengrok.configuration;
+
+import java.util.concurrent.ExecutorService;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.SearcherManager;
+
+/**
+ * Wrapper class over IndexSearcher which keeps SearcherManager around so
```
tarzanek: the pattern you wanted to use here is singleton, so this might be more optimized, let's leave it like this for now ...

tarzanek:
fix license and merge please

PR #1236, opened 2016-10-13
fixing autolocation of opengrok.jar

tulinkry:
The `OPENGROK_DISTRIBUTION_BASE` was set incorrectly to different directory and also it was not possible to overwrite `OPENGROK_JAR`.


tarzanek:
this will break running OpenGrok script from developers workspace, no ?

tarzanek:
AH, I am blind, this is groups file ... ok, good, merging :)

PR #1237, opened 2016-10-14
improving performance of project picker filter

tulinkry:
Fixed by replacing the recursion with simple iteration and all not necessary calls to jquery are replaced with some kind of counter.

Also a timeout for the search is introduced, so that the search does not fire after every key press immediately (for example when user is typing more letters). By default to 300ms.

The speedup is nice, tested on ~1000 repositories and 200 groups. The input was `595` for which there is only one result in one group.
Old setup:
```
searching: timer started
searching: 876.03ms
searching: timer started
searching: 1065.39ms
searching: timer started
searching: 1067.71ms
```
New setup (without the timeout):
```
searching: timer started
searching: 53.42ms
searching: timer started
searching: 150.53ms
searching: timer started
searching: 163.55ms
```


vladak:
Tested on the same setup as this was reported in the issue, it is subjectively significantly faster so that it can be considered generally usable.

tarzanek:
@tulinkry can you pls resolve conflicts?

tulinkry:
resolved

vladak:
Could you try contributing at least some of these changes to upstream ? This would make maintenance easier.

PR #1239, opened 2016-10-14
Directory listing when projects are disabled no longer displays nullptr exception

tulinkry:
Reported in #1223, caused by cset 8389245

PR #1240, opened 2016-10-14
let user to change if the projects are enabled or disabled

### LINKED ISSUE #1223: Failure to index root files of a single project directory
opengrok 0.12.6 seems to ignore the root project files e.g.  base directory, and instead only indexes their directories.

For the base directory files to be indexed you need to create a directory that contains your project and point the indexer at it.

tulinkry:
after #1239 is in upstream this fixes #1223

Introducing new env variable to control if user wants to have projects disabled or enabled. By default it's turned on.

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -54,6 +54,9 @@
 #                                                 directory
 #                                   local	- for local repos only
 #                                   uionly	- enabled for UI only
+#   - OPENGROK_ENABLE_PROJECTS    Enable projects (set it to true or false)
+#                                 Every directory in SRC_ROOT is
+#                                   considered a separate project
 #   - OPENGROK_SCAN_REPOS         Disable Scan for repositories (*)
 #   - OPENGROK_SCAN_DEPTH         how deep should scanning for repos go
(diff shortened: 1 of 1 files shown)


OpenGrok
```diff
@@ -254,7 +257,12 @@ DefaultInstanceConfiguration()
 
     # OPTIONAL: Enable Projects
     #           (Every directory in SRC_ROOT is considered a separate project)
-    ENABLE_PROJECTS="-P"
+    OPENGROK_ENABLE_PROJECTS="${OPENGROK_ENABLE_PROJECTS:-true}"
+    case $OPENGROK_ENABLE_PROJECTS in
+        true) ENABLE_PROJECTS="-P" ;;
+        false) ENABLE_PROJECTS="" ;;
+        *) ENABLE_PROJECTS="-P" ;;
```
vladak: This would go against the concept of least surprise I think, would leave it disabled if neither true or false.

tarzanek: I would leave it like it is, -p was default since opensolaris ...

tarzanek:
thnx, merging

PR #1242, opened 2016-10-17
Messages options

tulinkry:
1. Messages script now has both long and short options
2. Tag "main" is displayed whenever browsing a project (but not on the homepage)


tarzanek:
thnx

Issue #1241, opened 2016-10-16
0.13-rc3: Double click on a project from “In Project(s)” will redirect to wrong url and cause Error: File not found!


amotzte:
When pressing on project “projectname”
Expected: page will redirect to /source/xref/projectname
Observed: page redirects to /source/undefined/projectname which shows “error File not found!”

Also if there is a message for that project (use the new Messages feature) it will append a “!”  to the end of the url which will also redirect to “error File not found!” if xref will be fixed

It seems to be related to a cookie named "OpenGrokProject" somehow. If this cookie is deleted the  problem is not seen (until pressing again on the project which recreating the cookie again)

vladak:
Defn'tly a regression, thus marking as stopper.

amotzte:
Thanks!

amotzte:
@tarzanek  it seems that there is still an issue when double click on a project that has a Message on it:
**Reproduce** 
Add a Message to some project
Double click on the project in the "In Project(s)" window 
**Expected** 
Redirect to source/xref/\<project\>
**Observed**
Redirect to source/xref/\<project\>**!** which causes _"Error: File not found!"_

tarzanek:
@amotzte can you repaste this into new issue please?
We'll try to look at it ... tia
L

PR #1243, opened 2016-10-17
fixing regression for project picker - dblclick not working

tulinkry:
Reported in #1241, originally caused by #1195 ([affected diff](https://github.com/OpenGrok/OpenGrok/commit/838924562130977ca1a3d3839d146fcda39ea1af#diff-ff384b0a1615edf5a637ae6a69d748a7R438)) which after all produced invalid javascript on the main page causing the variable `document.xrefPath` being unset which is crucial for the dblclick and other redirect features of the project picker.

Because the original variable `document.projects` was not used on the main page anymore, I removed it.

Now the project picker knows about the exclamation marks and uses a different technique to obtain the project name.


tarzanek:
thnx, merging

Issue #1098, opened 2016-05-12
grey bar in directory listing should be converted to floating window


vladak:
On testing instance running 42cc096fd002a6eab6be082271e28d4853acd318 I see a grey bar in directory listing which should not really be there.
![grey-bar](https://cloud.githubusercontent.com/assets/2934284/15210029/9bcb87f0-1834-11e6-98fc-e6d533cb17af.png)

mrate:
I did some improvements to the scopes UI and this is one of the issues it addresses. Pull request is ready.

tarzanek:
confirming @mrate last pull requests #1100 and #1101 fixed this, closing

vladak:
As of 8c13ddaaf5d79ec5a234c256cb1a5ddaeceb7545, I am still seeing the grey bar, reopening.

vladak:
fixed by cset 0c79c99dd8a7d8db88ecaa19f41282cb01f6ff0b

PR #1248, opened 2016-10-21
Scopes window

tulinkry:
Refactoring the window plugin - now it's easy to create a window with less code.

Moving scopes into separated window instead of the grey panel.

Try internal demo [here](http://virtual99.cz.oracle.com/scopes/xref/OpenGrok/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java) - experiment with navigate also.

screenshot from 2016-10-21 12-28-20


vladak:
Looks nicer than before.

tulinkry:
Also fixes #1098

tarzanek:
thank you, merging

Issue #1247, opened 2016-10-21
Double click on a project that has a Message on it will redirect to "Error: File not found!" page


amotzte:
**Reproduce** 
Add a Message to some project
Double click on the project in the "In Project(s)" window 
**Expected** 
Redirect to source/xref/<project>
**Observed**
Redirect to source/xref/<project>**!** which causes "Error: File not found!"

PR #1249, opened 2016-10-21
Minifying the project picker javascript

tulinkry:
The #1247 has already been fixed however while resolving some conflicts the js file was forgotten to be minified - the web uses only the minified version.


vladak:
single long line diffs are hard to read.  Would be nice to see a diff of formatted code.

tulinkry:
There is currently no change in the code. This is the minified version of `searchable-option-list-2.0.2.js` which already sits in the master.

The changes were commited in #1243 and in #1237.

tarzanek:
merging, thnx

Issue #1228, opened 2016-10-11
The search box line highlight color doesn't match the rest of the grok interface


nerakhon:
![screen shot 2016-10-11 at 15 39 15](https://cloud.githubusercontent.com/assets/376883/19272472/f75bea86-8fc8-11e6-811c-738d3e213d21.png)

the blueish highlight totally doesn't matech the rest of the page

tulinkry:
Closed by #1252

Issue #1226, opened 2016-10-11
The new homepage is less readable


nerakhon:
Several issues:

A) The new homepage design uses too large fonts for the sections.

B) Also some default expansions of short sections would IMO make sense.
For sections with 1-3/4 children it is a waste of time to have to go and expand.

C) The show all button is in contrary very small and "hurts eyes"

D) For expansion everywhere else we use + sign shouldn't we use the same on homepage?

Including screenshot with marked issues

![screen shot 2016-10-11 at 15 28 00](https://cloud.githubusercontent.com/assets/376883/19272183/e1d4498e-8fc7-11e6-9d16-db0f26ad93cd.png)

nerakhon:
E) The home page is incorrectly aligned as well the top starts right on the left margin, the sections seem to be centered.

F) Overall the font sizes aren't really harmonized and could use some help, good example is the search field for picking project.

![screen shot 2016-10-11 at 15 37 19](https://cloud.githubusercontent.com/assets/376883/19272366/9da4c828-8fc8-11e6-95e3-a7ef63e14d0c.png)

tulinkry:
Closed by #1252

PR #1252, opened 2016-10-21
ui changes

tulinkry:
This should address #1226

> A) The new homepage design uses too large fonts for the sections.
> B) Also some default expansions of short sections would IMO make sense.
> For sections with 1-3/4 children it is a waste of time to have to go and expand.
> C) The show all button is in contrary very small and "hurts eyes"
> D) For expansion everywhere else we use + sign shouldn't we use the same on homepage?
> E) The home page is incorrectly aligned as well the top starts right on the left margin, the sections seem to be centered.
> F) Overall the font sizes aren't really harmonized and could use some help, good example is the search field for picking project.

A) I put them down a little bit
B) Added for < 5. However as it's visible in the demo, for many groups it isn't that much practical.
C) Show button is the same size like all other buttons on the main page
D) Replaced
E) I don't see/understand that
F) I made the text input font bigger while the select options smaller

 and #1228.

Replaced by the color of a button. I'm not sure what other I could use - grey (is already the background), white (all over the page), orange (too much aggresive)

[Demo](http://virtual99.cz.oracle.com/ui/) with many groups and projects.


tarzanek:
@tulinkry  can you please resolve the conflict?

tulinkry:
Resolved.
Again I'd like to remind that I didn't get this point:
> E) The home page is incorrectly aligned as well the top starts right on the left margin, the sections seem to be centered.

tarzanek:
@nerakhon Honzo? E) ?

nerakhon:
Put the page on at least 24 inch LCD and check alignment of sections. I will provide a screenshot on Monday if necessary.

tulinkry:
If it was this, then the commit has been updated
screenshot from 2016-10-31 11-58-10
to this:
screenshot from 2016-10-31 11-58-37

nerakhon:
screen shot 2016-10-31 at 12 56 05

tulinkry:
Aligned to the right - opened:
screenshot from 2016-10-31 16-18-51
Collapsed:
screenshot from 2016-10-31 16-19-14
Without groups:
screenshot from 2016-10-31 16-19-54

nerakhon:
one last nit, did you consider making the right text box the same height as the left ones ot vice versa?

tarzanek:
well actually lgtm, can I merge @nerakhon ? :)

nerakhon:
I would ask for the fix of the textbox sizes afterwards I am keen to give +1.

tulinkry:
Now at the same size
screenshot from 2016-11-03 15-10-31

tarzanek:
:) thnx guys, merging

Issue #995, opened 2015-09-28
Render README.md files in place when browsing directories


m-renaud:
When browsing directories in xref/ the README.md file will be displayed under the directory listing and it would be nice if the markdown was actually rendered.

tarzanek:
@m-renaud  is there some easy to bundle (java) library to render the file?
ideally if that lib would be supported (so not abandoned) and it's size would be reasonable
I plan to integrate tika, I'll have a look if tika can do something like this already ... (but please shoot alternatives)

tulinkry:
I've found two:
- https://github.com/rjeschke/txtmark
- https://code.google.com/p/markdown4j/ (which is fork of the previous one)

tulinkry:
However like I did it in other project recently, it is possible with javascript. So it could be easy to embed a javascript (which would load only when the file is *.md) and to display that file in the browser.

tulinkry:
closed with #1253

PR #1253, opened 2016-10-21
allowing markdown readmes to be rendered as html

tulinkry:
As suggested in #995 this is the implementation which uses javascript converter. I'm not sure if it is really useful, but it's not bad.

Examples (internal):
[example 1](http://virtual99.cz.oracle.com/markdown/xref/application/)
[example 2](http://virtual99.cz.oracle.com/markdown/xref/caching)
[example 3](http://virtual99.cz.oracle.com/markdown/xref/latte)


vladak:
I'd put the readme contents into a frame or at least use some light background - like github does.

tulinkry:
Check the examples again :)

vladak:
Looks better except the long lines crawl out of the grey frame which looks terrible.

vladak:
Also, just for completeness: I am a bit worried that this is XSS prone.

tulinkry:
And I would like to point out [this fact](https://github.com/showdownjs/showdown#xss-vulnerability) that processing markdown is not XSS safe!

tulinkry:
Ok so as mentioned in the [code](https://github.com/OpenGrok/OpenGrok/pull/1253/files#diff-d117286e3441fbba4b026d0dc13a2cd6R1114), the converter behaviour can be completely disabled by commenting out the init call and it falls back to the old behaviour (just prints the readme source code).

tarzanek:
so is there a way to fix/sanitize  the possible xss ?

tulinkry:
As described [here](https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-(and-how-to-mitigate-it)) I added a xss filter after the markdown process.

However the problem was more difficult - not to render the html in the page, so I decided to take it through ajax - the markdown readmes are loaded through ajax call and then processed, filtered and rendered. Maybe this is too complicated and we should just render the readmes as they are.

I used [this xss filter](https://github.com/leizongmin/js-xss) which seems ok (but I haven't looked into much details).

tarzanek:
version missing for xss.min.js

tarzanek:
(make the version part of the file name, this will help with caching later on)

tulinkry:
> Looks better except the long lines crawl out of the grey frame which looks terrible.
Can't reproduce this

tarzanek:
@vladak I am OK with merge, if you're OK, then please merge! :)

vladak:
Still need to see if the issue with long lines reported earlier persists.

vladak:
It seems to be resolved, merging.

PR #1255, opened 2016-10-24
allowing users to disable the new project picker and use the old one

tulinkry:
If anyone likes the old behaviour then he can disable the new project picker by commenting out its init call in `utils.js`.

Select/DeselectAll and invert should then work like expected. Support for selecting particular groups is not migrated.

```javascript
/**
 * Initialize the new project picker
 */
    init_searchable_option_list()
```

### DIFF
--- a/web/js/utils.js
+++ b/web/js/utils.js
@@ -1528,15 +1528,29 @@ function togglerevs() {
 }
 
 function selectAllProjects() {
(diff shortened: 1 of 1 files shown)


tarzanek:
thnx, merging

naseer:
Thanks!

PR #1256, opened 2016-10-24
Fixing copy-paste problem when copying scope context

### LINKED ISSUE #1232: copy-n-paste over function with scope does not preserve newline
When I copy-n-paste over the function with the minus icon, the newline is not preserved in the text.

tulinkry:
This should fix #1232. It works for me on java and C files. Can anyone else please verify it?

The idea is clear - `display: block` automatically does a new line break so `display: inline` leaves the responsibility to the xref text where the newline is already in the source code after each line.


tarzanek:
thnx, merging

PR #1257, opened 2016-10-24
hiding also the scope space on line toggle

### LINKED ISSUE #1224: White space prefix remain after line numbers are turned off
When I switch of line prefix, white space delimiter which was between line num and code remains which complicate copy out of code.

tulinkry:
fixes #1224

We should anyway consider to replace the `&nbsp;` by something else, like a fix-sized element, since ui positioning is not a job of html anymore.


tarzanek:
merging, thnx

PR #1260, opened 2016-10-27
adding a possibility to send a response about message acceptance

tulinkry:
This is one option.

The other option is to use `ObjectInput/OutputStream` which decodes the whole object (and does not need the EOF solution). The problem with this is that the xml decoder does not stop reading after the root element is closed - that's why I introduced this artificial EOF.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -30,6 +30,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.net.InetAddress;
 import java.net.ServerSocket;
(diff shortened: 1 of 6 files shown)


vladak:
Is there a test case that would cover at least the `MESSAGE_LIMIT` case ?

vladak:
Also, why did you prefer extending the FilterInputStream/FilterOutputStream classes as opposed to decoding the whole object ?

tulinkry:
I added the test case. I'm not sure if that's the way to do it, anyway it works.

I prefer extending the FilterInputStream/FilterOutputStream because that keeps the configuration protocol on xml basis (which is what we currently use).

As I read somewhere, the other option uses binary format (and therefore the output/input stream knows exactly where the object ends) and is, in a sense, the same as the xml format. The xml format is preferred when using Java Beans and also is human readable.

So I decided to stay with the old OpenGrok scheme.

tarzanek:
I think this can be merged ... merging ;)

PR #1262, opened 2016-11-01
Add current indexed version to repos table.

amotzte:
## Why needed 
Currently the repos table on home page already shows some details about the available projects. I found it very helpful to also show what is the current **indexed version** for each project. This way you can immediately know how much up to date the project is.

As a start I added the information for git SCM only but I prepared the infrastructure to add for other SCM (much like the “branch” column).

## Here is the result:
screen shot 2016-11-11 at 1 20 15 am

### DIFF
--- a/src/org/opensolaris/opengrok/history/Repository.java
+++ b/src/org/opensolaris/opengrok/history/Repository.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.history;
 
@@ -363,6 +363,18 @@ public abstract class Repository extends RepositoryInfo {
      * Determine branch of this repository.
      */
     abstract String determineBranch() throws IOException;
+
+    /**
+     * Determine and return the current version of the repository.
+     *
+     * This operation is consider "heavy" so this function should not be
+     * called on every web request.
+     *
+     * @return the version
+     */
+    String determineCurrentVersion() throws IOException {
+        return null;
+    }
 
     /**
      * Returns true if this repository supports sub repositories (a.k.a.
--- a/src/org/opensolaris/opengrok/history/RepositoryInfo.java
+++ b/src/org/opensolaris/opengrok/history/RepositoryInfo.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.history;
 
@@ -42,6 +42,7 @@ public class RepositoryInfo implements Serializable {
     protected String datePattern;
     protected String parent;
     protected String branch;
+    protected String currentVersion;
 
     /**
      * Empty constructor to support serialization.
@@ -58,6 +59,7 @@ public class RepositoryInfo implements Serializable {
         this.datePattern = orig.datePattern;
         this.parent = orig.parent;
         this.branch = orig.branch;
+        this.currentVersion = orig.currentVersion;
     }
 
     /**
@@ -166,5 +168,13 @@ public class RepositoryInfo implements Serializable {
 
     public void setBranch(String branch) {
         this.branch = branch;
+    }
+
+    public String getCurrentVersion() {
+        return currentVersion;
+    }
+
+    public void setCurrentVersion(String currentVersion) {
+        this.currentVersion = currentVersion;
     }
 }
--- a/web/repos.jspf
+++ b/web/repos.jspf
@@ -131,6 +131,7 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                     <td><b>Mirror</b></td>
                     <td><b>SCM type</b></td>
                     <td><b>Parent (branch)</b></td>
+                    <td><b>Current version</b></td>
                     </tr>
                     </thead>
                     <tbody>
@@ -175,8 +176,36 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                             if (branch == null) {
                                 branch = "N/A";
                             }
+                            String currentVersion = ri.getCurrentVersion();
+                            if (currentVersion == null) {
+                                currentVersion = "N/A";
+                            }
                             %><td><%= Util.htmlize(type) %></td><%
                             %><td><%= Util.htmlize(parent) %> (<%= Util.htmlize(branch) %>)</td><%
+                            %><td><%
+
+                            // Current index collapse threshold minimum of 10
+                            int summaryLength = Math.max(10, cfg.getCurrentIndexedCollapseThreshold());
+                            String cout = Util.htmlize(currentVersion);
+
+                            boolean showSummary = false;
+                            String coutSummary = currentVersion;
+                            if (coutSummary.length() > summaryLength) {
+                                showSummary = true;
+                                coutSummary = coutSummary.substring(0, summaryLength - 1);
+                                coutSummary = Util.htmlize(coutSummary);
+                            }
+                            if (showSummary) {
+                                %>
+                                <span class="rev-message-summary"><%= coutSummary %></span>
+                                <span class="rev-message-full rev-message-hidden"><%= cout %></span>
+                                <span  data-toggle-state="less"><a class="rev-toggle-a rev-message-toggle "  href="#">show more ... </a></span>
+                                <%
+                            }
+                            else {
+                                 %><span class="rev-message-full"><%= cout %></span><%
+                            }
+                            %></td><%
                             %></tr><%
                             cnt++;
                         }
@@ -224,11 +253,12 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                     <div class="panel-body<% if (groups.size() > 0) {%>-accordion<% } %> <% if (pHelper.hasUngroupedFavourite()) { %> favourite<% } %>"
                         <% if (pHelper.hasUngroupedFavourite()) { %>data-accordion-visible="true"<% } %>>
                         <table>
-                            <thead>
+                        <thead>
                         <tr>
                         <td><b>Mirror</b></td>
                         <td><b>SCM type</b></td>
                         <td><b>Parent (branch)</b></td>
+                        <td><b>Current version</b></td>
                         </tr>
                             </thead>
                             <tbody>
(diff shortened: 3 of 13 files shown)


tarzanek:
can you fix tests @amotzte ? thnx (check travis - https://travis-ci.org/OpenGrok/OpenGrok/builds/172456917 )
+ we will need to sort out OCA - either check http://www.oracle.com/technetwork/community/oca-486395.html , print, sign , scan, mail back to oracle
or I am ok if you say you agree with OCA for this patch in this thread

tarzanek:
also - and this is a prereq - we will need to cache this somehow to history folder, I don't think it is good to query SCM for every web request

amotzte:
Hi @tarzanek as far as I understand, all of the table information (Including this new field) is cached. The determineCurrentVersion is being called in only 2 cases: 
1. When server starts 
2. After index 

The only function that is being called for every request is getCurrentVersion. This function returns the value from a local variable. Am I missing something?

web/repos.jspf
```diff
@@ -263,8 +264,13 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                                 if (branch == null) {
                                     branch = "N/A";
                                 }
+                                String currentVersion = ri.getCurrentVersion();
```
tulinkry: There are two places in `repos.jspf` where a line about repository is being printed to the output. So this should be handled in both cases.

src/org/opensolaris/opengrok/history/AccuRevRepository.java
```diff
@@ -357,4 +357,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
+    @Override
+    String determineCurrentVersion() {
+        return null;
+    }
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/BazaarRepository.java
```diff
@@ -369,4 +369,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/ClearCaseRepository.java
```diff
@@ -418,4 +418,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/MercurialRepository.java
```diff
@@ -681,4 +681,9 @@ String determineParent() throws IOException {
 
         return executor.getOutputString().trim();
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/PerforceRepository.java
```diff
@@ -237,4 +237,9 @@ public static String getRevisionCmd(String rev) {
         }
         return "@" + rev;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/RCSRepository.java
```diff
@@ -164,4 +164,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/RazorRepository.java
```diff
@@ -349,4 +349,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/RepoRepository.java
```diff
@@ -129,4 +129,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/Repository.java
```diff
@@ -364,6 +364,8 @@ final void createCache(HistoryCache cache, String sinceRevision)
      */
     abstract String determineBranch() throws IOException;
 
+    abstract String determineCurrentVersion() throws IOException;
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/RepositoryInfo.java
```diff
@@ -42,6 +42,7 @@
     protected String datePattern;
     protected String parent;
     protected String branch;
+    protected String currentVersion;
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/SCCSRepository.java
```diff
@@ -292,4 +292,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/SSCMRepository.java
```diff
@@ -400,4 +400,9 @@ String determineParent() throws IOException {
     String determineBranch() {
         return null;
     }
+
```
tulinkry: Wrong copyright date.

src/org/opensolaris/opengrok/history/Repository.java
tulinkry: I wonder why it's made abstract when you return `null` in all repositories except `GitRepository`. Also this is about the other `determineBranch` and `determineParent`. @tarzanek

amotzte: Good point... I followed the determineBranch function (without thinking about it) but you are absolutely right...

src/org/opensolaris/opengrok/history/RepositoryInfo.java
amotzte: Fixed

web/repos.jspf
amotzte: Fixed

web/repos.jspf
```diff
@@ -171,8 +172,13 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                             if (branch == null) {
                                 branch = "N/A";
                             }
+                            String currentVersion = ri.getCurrentVersion();
+                            if (currentVersion == null) {
+                                currentVersion = "N/A";
+                            }
                             %><td><%= Util.htmlize(type) %></td><%
                             %><td><%= Util.htmlize(parent) %> (<%= Util.htmlize(branch) %>)</td><%
+                            %><td><%= Util.htmlize(currentVersion) %> </td><%
```
tulinkry: When the commit message is quite long (which sometimes happen) the table looks ugly then - I'd rather see it on a single line.
I'd suggest something like displaying the date, changeset, author and first like 20-30 characters from the message (maybe with an option to show the whole message like it is in history view)

screenshot from 2016-11-03 15-36-27

tulinkry: My output does not match your output. Especially I don't like the enclosing \'.

screenshot from 2016-11-03 15-36-27

amotzte: Fixed

amotzte: screen shot 2016-11-03 at 11 29 31 pm
screen shot 2016-11-03 at 11 29 49 pm

tarzanek:
hmm, I was more like thinking about the situation where the history is not stored in javadb and just uses plain files - will it cache the data even in this case? ;)

amotzte:
@tarzanek  All field (branch, parent scm...) in the "homepage" table are stored in projects configuration configuration.xml. Here is an example:


```xml
 <void method="add">
    <object class="org.opensolaris.opengrok.history.RepositoryInfo">
     <void property="branch">
      <string>master</string>
     </void>
     <void property="currentVersion"> <!--   <-- this is the new addition --> 
      <string>2016-09-04 17:00: 638bcfb mregev Merge pull request #2 from SPVSS-IH-UI/ui-server-ha</string>
     </void>
     <void property="datePattern">
      <string>EE, d MMM yyyy HH:mm:ss Z</string>
     </void>
     <void property="directoryName">
      <string>/var/opengrok/src/vgw_tests_and_deployment/ih-ui-server</string>
     </void>
     <void property="parent">
      <string>git@github3.cisco.com:SPVSS-IH-UI/ih-ui-server.git</string>
     </void>
     <void property="type">
      <string>git</string>
     </void>
    </object>
   </void>
```  

So to my understanding all of these fields are cached/stored (in RuntimeEnvironment?) so we will not be executing git command for each web request in the same way that are are not going it now for the rest of these fields .

web/repos.jspf
```diff
@@ -263,8 +293,37 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                                 if (branch == null) {
                                     branch = "N/A";
                                 }
+                                String currentVersion = ri.getCurrentVersion();
+                                if (currentVersion == null) {
+                                    currentVersion = "N/A";
+                                }
                                 %><td><%= Util.htmlize(type) %></td><%
                                 %><td><%= Util.htmlize(parent) %> (<%= Util.htmlize(branch) %>)</td><%
+                                %><td><%
```
tulinkry: Now it looks like this:
screenshot from 2016-11-04 12-24-41

Which has not really necessary margin, can you avoid this?

amotzte: We have to set the table column to be at fixed size so in case user choose to open the "show more..." it will not change the other columns size (which feel annoying in the eye..). Currently they are divided with default size (each column  get the exact some size).  I guess I can reduce the size of "SCM type" to something smaller but still fixed and give more space to "Parent (branch)" and "Current version". But then each column has a different size which some say looks less neat. What do you think?

tarzanek:
@amotzte thanks & sorry for stupid Q, this explains, I was under the impression this needs to be stored in history folder per every project, if we serialize it down to config. xml then great :)
thank you
L

web/repos.jspf
amotzte: @tulinkry, I set only "Current version" to be fixed so the rest can adjust according to their size.  I also fixed the extra margins. See new screenshots.

tulinkry: I'd like to stay with the one line like this:
screenshot from 2016-11-10 16-51-57

web/repos.jspf
```diff
@@ -125,12 +125,17 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                 <%
                 if (pHelper.getRepositories(group).size() > 0 ) {
                 %>
-                    <table>
+                    <table >
+                    <col/>
```
tulinkry: Avoid using `<col>` as the [attributes are mostly obsolet in html5](http://www.w3schools.com/TAgs/tag_col.asp)

Extract this into the css files (`default/style.css`, `polished/style.css`, `offwhite/style.css`). Do not forget to change all of them. You can set a `class` on the column or select it with some other more complicated selector.

web/repos.jspf
```diff
@@ -125,12 +125,17 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                 <%
                 if (pHelper.getRepositories(group).size() > 0 ) {
                 %>
-                    <table>
+                    <table >
```
tulinkry: extra space?

web/repos.jspf
```diff
@@ -175,8 +180,36 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
...
+                            if (coutSummary.length() > summaryLength) {
+                                showSummary = true;
+                                coutSummary = coutSummary.substring(0, summaryLength - 1);
+                                coutSummary = Util.htmlize(coutSummary);
+                            }
+                            if (showSummary) {
+                                %>
+                                <span class="rev-message-summary"><%= coutSummary %></span>
+                                <span class="rev-message-full rev-message-hidden"><%= cout %></span>
+                                <div  data-toggle-state="less"><a class="rev-toggle-a rev-message-toggle "  href="#">show more ... </a></div>
```
tulinkry: As it is in my screenshot...I'd make this `<span>` instead of `<div>`.

web/repos.jspf
```diff
@@ -223,12 +256,17 @@ Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
                     <% } %>
                     <div class="panel-body<% if (groups.size() > 0) {%>-accordion<% } %> <% if (pHelper.hasUngroupedFavourite()) { %> favourite<% } %>"
                         <% if (pHelper.hasUngroupedFavourite()) { %>data-accordion-visible="true"<% } %>>
-                        <table>
-                            <thead>
+                        <table/>
+                        <col/>
```
tulinkry: The same problem like above about the css.

web/repos.jspf
tulinkry: screenshot from 2016-11-10 16-51-57

web/repos.jspf
amotzte: Fixed

tulinkry:
I'm fine

tulinkry:
@vegetablemao You might want to try the following snippet:
```java
String d = "Fri, 14 May 2010 23:24:24 +0200";
String one = "EE, d MMM yyyy HH:mm:ss Z";
String two = "d MMM yyyy HH:mm:ss Z";

SimpleDateFormat format = new SimpleDateFormat(one, Locale.getDefault());

try {
    Date date = format.parse(d);
    System.out.println("Successfully parsed " + date);
} catch (ParseException ex) {
    System.out.println("Parsing failed");
}
```
as this is the way we parse the git log dates. If it fails, try to change the line where the format is created to
```java
SimpleDateFormat format = new SimpleDateFormat(one);
```
to see if it helps.

tulinkry:
@vegetablemao support for old git is already in the trunk (#1314)
and you may want to try the current head (with #1336) to get more information about the
```text
GitHistoryParser.process: Failed to parse author date: AuthorDate: Fri, 14 May 2010 23:24:24 +0200
```
as I added more verbose information to that exception.

PR #1268, opened 2016-11-10
limiting the width of the scopes window

### LINKED ISSUE #1267: scopes window is sometimes too wide
For functions with significant number of arguments the scopes window grows too wide enough so that it covers the source code. Maybe it should be limited to half width of the screen, be transparent, wrap the line somehow etc.

tulinkry:
fixes #1267


tarzanek:
thnx merging

PR #1269, opened 2016-11-10
render scope signature as html

tulinkry:
Changes the scope view from:
`foo( ..., List&lt;String&gt; subFiles, List&lt;String&gt; repositories, List&lt;String&gt; zapCache, boolean listRepoPaths)`
to:
`foo( ..., List<String> subFiles, List<String> repositories, List<String> zapCache, boolean listRepoPaths)`


tarzanek:
thnx

PR #1281, opened 2016-11-14
fixing revision number in annotation

tulinkry:
Observable for example on OpenGrok file `web/default/style.css`. Find revision `1412e1e9` (image attached) and click on it - that takes you into that revision. However the revision number then shows 16/15 which is not correct. I think it's because there are some revisions which don't have information about author (that's different story - they're not available in `git log` in this case):

Before:
screenshot from 2016-11-14 14-42-08

After:
screenshot from 2016-11-14 14-41-08


tulinkry:
> that's different story - they're not available in git log in this case

For this particular case this is solved with adding `--follow` switch to the git log command when it's run on single file. I didn't test it generally.

PR #1282, opened 2016-11-16
removing extra newline in C/C++ files after one line comment

### LINKED ISSUE #1163: Bad End of Line appears in HTML for C and C++ index files
Starting at or around 0.13RC2 C and C++ files that were indexed have an End Of Line appear in the HTML file whenever a "//" is present.

I have attached an image to better visualize the problem.

This only happens with comments of type //

grokproblem

tulinkry:
fixes #1163

I'm not a pro in lex files. However the removed line added a `{EOL}` into the output and after that the `startNewLine();` added a second one. It looks fine to me to trash the trailing whitespace and EOL as it's done also [in general EOL](https://github.com/OpenGrok/OpenGrok/blob/65a0ddded9e4776c836b5c683f3c33f750e0546d/src/org/opensolaris/opengrok/analysis/c/CXref.lex#L128-L132) later and basically it's not valuable for anything.

Requires reindex.


tarzanek:
thank you
it's quite funny but both c / c++ lexers for xref should be in sync, but they aren't
it seems we will need to verify their outputs, since they (at least the C one) doesn't seem sane to me

PR #1284, opened 2016-11-16
adding current version info into mercurial repository

tulinkry:
following #1262 for mercurial repos

### DIFF
--- a/src/org/opensolaris/opengrok/history/MercurialRepository.java
+++ b/src/org/opensolaris/opengrok/history/MercurialRepository.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2006, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2006, 2016, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.history;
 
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/history/MercurialRepository.java
```diff
@@ -681,4 +681,43 @@ String determineParent() throws IOException {
...
+        cmd.add("1");
+        cmd.add("--template");
+        cmd.add("{date|isodate}: {node|short} {author} {desc|strip}");
+
+        ProcessBuilder pb = new ProcessBuilder(cmd);
+        pb.directory(directory);
+        Process process = pb.start();
+
+        try {
+            process.waitFor(15, TimeUnit.SECONDS);
```
tarzanek: again , I though we deal with timeouts somehow globally, but @vladak knows this code better, otherwise seems ok, merging

tulinkry: maybe it's better to use it like [here](https://github.com/tulinkry/OpenGrok/blob/0cb5b1bf01ac7ec40123298024fc9b1f25d3e70c/src/org/opensolaris/opengrok/history/MercurialRepository.java#L669-L682) - since we use just a single line from the output

PR #1285, opened 2016-11-16
adding waitfor for the current branch process

tulinkry:
fixed process timeout & exit code

### DIFF
--- a/src/org/opensolaris/opengrok/history/GitRepository.java
+++ b/src/org/opensolaris/opengrok/history/GitRepository.java
@@ -41,6 +41,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/history/GitRepository.java
```diff
@@ -675,15 +676,27 @@ String determineCurrentVersion() throws IOException {
...
         pb.directory(directory);
-        Process process;
-        process = pb.start();
+        Process process = pb.start();
 
-        try (BufferedReader in = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
-            if ((line = in.readLine()) != null) {
-                line = line.trim();
+        try {
+            process.waitFor(15, TimeUnit.SECONDS);
```
tarzanek: I have merged, but I have some doubts that this will be a problem for long first inits of git repo, no?
will retest and worst case open a bug against this

vladak:
I dont like the hardcoded timeout either.

tulinkry:
maybe it's better to use it like [here](https://github.com/tulinkry/OpenGrok/blob/0cb5b1bf01ac7ec40123298024fc9b1f25d3e70c/src/org/opensolaris/opengrok/history/MercurialRepository.java#L669-L682) - since we use just a single line from the output. So that I don't reinvent the wheel

I'll change it

PR #1287, opened 2016-11-21
using Executor class for determining the current version

tulinkry:
It's without timeout like it's done in [MercurialRepository#determineParent()](https://github.com/tulinkry/OpenGrok/blob/ad09d029c88af0dcb7492df5e3ed9f66086b417b/src/org/opensolaris/opengrok/history/MercurialRepository.java#L668)


tarzanek:
thank you ;)

PR #1292, opened 2016-11-25
converting navigate window into the new window plugin

### LINKED ISSUE #1289: Convert the navigation window with the new window plugin
Also maybe design it in a way that it don't need to insert `function get_sym_list(){ ...` this in the html.

tulinkry:
fixes #1289 

Just a new look for the old navigation window. Automatically moves down when scopes are visible (but that isn't new neither).

screenshot from 2016-11-25 17-39-24
screenshot from 2016-11-25 17-39-53


tarzanek:
thnx, merging

PR #1293, opened 2016-11-26
bump required Java and Tomcat versions for gh-pages

boidolr:
Update required Java version to 1.8 and Tomcat to 8.x, same as given in the requirements file.


tarzanek:
thank you, merging

PR #1305, opened 2016-12-30
Display empty navigation window if there are no definitions

### LINKED ISSUE #1302: Navigation window does not open
I noticed while testing something else that the navigation window (navigate) does not open sometimes.
Fails with:

```
ReferenceError: get_sym_list is not defined            [Learn More]utils.js:1607:5
```

tulinkry:
1. detecting if xref has no definitions
fixes #1302
2. auto resizing the window on the first load


tarzanek:
thnx, merging

PR #1306, opened 2016-12-30
correct handling of the real EOF

tulinkry:
The artificial EOF was not designed correctly and it didn't bother to check the real EOF.


tarzanek:
now I would like to see tests for this ... (or bug report ? )
I thought EOF is eof ;)
so I don't get what the poet wanted to say ... care to explain?

tulinkry:
The problem is explained here #1260.

I will cover it with tests.

tarzanek:
ah, ok the bug explains, merging

PR #1309, opened 2016-12-30
fixing mouse location for scrolled page

tulinkry:
On a long page the message window position was wrongly determined, ignoring the actuall scrolled height.

I verified the intelligence window works as well.


tarzanek:
thnx, merging

Issue #1310, opened 2017-01-01
something wrong with git log


warjiang:
when i use OpenGrok to generate index ,something woring happened.
![image](https://cloud.githubusercontent.com/assets/6478745/21582277/a7d8c58e-d08b-11e6-9a3e-360761d5997c.png)

tulinkry:
This doesn't harm anything. It's trying to get the value which is displayed on the main page in the "Current version" column, you probably get "N/A" in that column.

Try to read [this comment](https://github.com/OpenGrok/OpenGrok/pull/1262#commitcomment-20273958) which may solve the problem.

warjiang:
:thumbsup: version of current git is 1.9.1. When i try to upgrade git to 2.11.0 and execute the generating index script，no warning.Thanks!

tulinkry:
This is now fixed in upstream with #1314. It should support git < 2.6

PR #1314, opened 2017-01-05
Supporting older git < 2.6 which does not have --date:format option

tulinkry:
Apparently our feature from #1262 is only supported in git2.6+ (as discovered [here](https://github.com/OpenGrok/OpenGrok/pull/1262#commitcomment-20273958)).

This patch should support older git versions by converting the date from rfc, without using the `--date=format:` option.

Also this was reported in #1310 which gave me the clue that not everybody uses upgraded git.


tarzanek:
lgtm

PR #1318, opened 2017-01-06
Messages are group aware

### LINKED ISSUE #1212: Messages could be Groups aware
It would be nice if a message could be applied to whole project group.

tulinkry:
fixes #1212

The special group "other" (projects without any group) is not handled. Only on the main page there is the exclamation mark next to the section "Others".

When viewing single project there are all messages which are connected to the project - main messages, project messages and group messages:
screenshot from 2017-01-06 17-38-35

On the main page, the exclamation mark is there for every group.
screenshot from 2017-01-06 17-42-36

### DIFF
--- a/src/org/opensolaris/opengrok/search/Results.java
+++ b/src/org/opensolaris/opengrok/search/Results.java
@@ -18,7 +18,7 @@
  */
 
 /*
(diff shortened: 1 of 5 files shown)


tarzanek:
nice, thank you, merging

PR #1320, opened 2017-01-12
adding versions into js scripts

tulinkry:
Solution for the problem mentioned in [comment for jquery update](https://github.com/OpenGrok/OpenGrok/pull/1145#issuecomment-234251318).

We've already met this problem with some new features added to opengrok ui.


tarzanek:
thnx, merging

PR #1324, opened 2017-01-16
throw exceptions on parsing errors so that file that cause it can be …

### LINKED ISSUE #1323: history parser exceptions lack file info
Often I see great number of exceptions like this in indexer logs:

```
Jan 14, 2017 12:04:33 AM org.opensolaris.opengrok.history.SCCSHistoryParser initFields
WARNING: An error occured while parsing date
java.text.ParseException: Unparseable date: ":0/02/07 15:25:57"
        at java.text.DateFormat.parse(DateFormat.java:366)
        at org.opensolaris.opengrok.history.SCCSHistoryParser.initFields(SCCSHistoryParser.java:130)
        at org.opensolaris.opengrok.history.SCCSHistoryParser.getRevision(SCCSHistoryParser.java:149)
        at org.opensolaris.opengrok.history.SCCSHistoryParser.parseFile(SCCSHistoryParser.java:90)
        at org.opensolaris.opengrok.history.SCCSHistoryParser.parse(SCCSHistoryParser.java:68)
        at org.opensolaris.opengrok.history.SCCSRepository.getHistory(SCCSRepository.java:261)
        at org.opensolaris.opengrok.history.FileHistoryCache.get(FileHistoryCache.java:499)
```

The problem is that they are impossible to pair with actual file that caused it. Seems like bunch of history parsers exhibit the pattern of catching the `ParseException` and just logging it.

vladak:
…identified

fixes #1323

### DIFF
--- a/src/org/opensolaris/opengrok/history/AccuRevHistoryParser.java
+++ b/src/org/opensolaris/opengrok/history/AccuRevHistoryParser.java
@@ -158,7 +158,11 @@ public class AccuRevHistoryParser implements Executor.StreamHandler {
                     date = df.parse(data[2]);
                     entry.setDate(date);
                 } catch (ParseException pe) {
-                    LOGGER.log(Level.WARNING, "Could not parse date: " + line, pe);
+                    //
+                    // Overriding processStream() thus need to comply with the
+                    // set of exceptions it can throw.
+                    //
+                    throw new IOException("Could not parse date: " + line, pe);
                 }
 
             } else if (line.startsWith("  #")) {  // found comment
--- a/src/org/opensolaris/opengrok/history/AccuRevHistoryParser.java
+++ b/src/org/opensolaris/opengrok/history/AccuRevHistoryParser.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2010, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.history;
 
--- a/src/org/opensolaris/opengrok/history/ClearCaseHistoryParser.java
+++ b/src/org/opensolaris/opengrok/history/ClearCaseHistoryParser.java
@@ -78,8 +78,7 @@ class ClearCaseHistoryParser implements Executor.StreamHandler {
         String s;
         HistoryEntry entry = null;
         while ((s = in.readLine()) != null) {
-            if (!"create version".equals(s) &&
-                    !"create directory version".equals(s)) {
+            if (!"create version".equals(s) && !"create directory version".equals(s)) {
                 // skip this history entry
                 while ((s = in.readLine()) != null) {
                     if (".".equals(s)) {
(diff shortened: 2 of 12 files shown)


vladak:
I was thinking to make the file information available in each history parser however:
  - most of the parsers can parse directories as well as files
  - the instantiation often looks like this:
    `return new SCCSHistoryParser().parse(file, this);`
    so there is no file info contained per se in the object

vladak:
Need @tulinkry to look at this since I modifies his SVN date parsing test case.

vladak:
The trouble with history parsers in general is that some of them were hiding problems and reporting them only to log (with WARNING). I think if there is a problem parsing history, HistoryException should be always raised and the caller should react accordingly.

That said, the passing of exceptions is a bit messy. There are some 3 kinds of history parsers:
  - simply read file from disk and interpret its contents (SCCS)
    - non-problematic, can manipulate/throw exceptions freely
  - execute external command and parse its output (Mercurial)
    - usually constrained by `processStream()` interface which can only throw IOException so if there is problem e.g. with parsing the date which normally results in ParseException, such has to be wrapped inside IOException
  - ditto plus interpret the output as XML (Subversion)
    - same set of problems as previous kind plus the XML parser can throw SAXException

test/org/opensolaris/opengrok/history/SubversionHistoryParserTest.java
```diff
@@ -202,16 +204,13 @@ public void testDateFormats() {
                 HistoryEntry e = result.getHistoryEntries().get(0);
                 assertEquals(revId, e.getRevision());
                 assertEquals(author, e.getAuthor());
-
-                if (expectedException) {
-                    assertNull(e.getDate());
-                } else {
-                    assertEquals(new SimpleDateFormat(format).parse(date), e.getDate());
-                }
+                assertEquals(new SimpleDateFormat(format).parse(date), e.getDate());
```
tulinkry: This seems reasonable for me

test/org/opensolaris/opengrok/history/SubversionHistoryParserTest.java
```diff
@@ -67,7 +67,9 @@ public void tearDown() {
      */
     @Test
     public void parseEmpty() throws Exception {
-        History result = instance.parse("");
```
tulinkry: Bad copyright

src/org/opensolaris/opengrok/history/GitRepository.java
```diff
@@ -178,10 +178,10 @@ Executor getHistoryLogExecutor(final File file, String sinceRevision)
     }
 
     /**
-     * Formatter for rfc 2822 which allows (as rfc) the optional day at the
```
tulinkry: Bad copyright

src/org/opensolaris/opengrok/history/AccuRevHistoryParser.java
```diff
@@ -158,7 +158,11 @@ public void processStream(InputStream input) throws IOException {
                     date = df.parse(data[2]);
                     entry.setDate(date);
                 } catch (ParseException pe) {
-                    LOGGER.log(Level.WARNING, "Could not parse date: " + line, pe);
```
tulinkry: Bad copyright

src/org/opensolaris/opengrok/history/ClearCaseHistoryParser.java
```diff
@@ -78,8 +78,7 @@ public void processStream(InputStream input) throws IOException {
         String s;
         HistoryEntry entry = null;
         while ((s = in.readLine()) != null) {
-            if (!"create version".equals(s) &&
```
tulinkry: There is a missing copyright, one should be added

vladak:
fixed the copyrights

tulinkry:
I will modify the tests again anyway.
Seems ok to me now

PR #1330, opened 2017-01-17
Header cleanup

vladak:
makes the UI a bit more tidy and uniform across different view

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -254,7 +254,7 @@ label {
     margin: 0 0 0.5ex 0;
 }
 
-#pagetitle { /* short verbal summary/description of the shown content */
+#dtag { /* short verbal summary/description of the shown content */
 }
 
 #filename { /* "Cross Reference: ..." */
@@ -265,13 +265,14 @@ label {
 }
 
 #Masthead {
+    font-family: monospace; /* ala <kbd> */
     clear: left;
 }
 
 #Masthead tt { /* slashes in the xref: line */
 }
 
-#sbar, #bar { /* full search and default navbar */
+#sbar, #bar { /* default navbar */
     border-top: 4px solid #ffc726;
     margin: 0;
     padding-top: 1ex;
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -267,6 +267,7 @@ label {
 #Masthead {
     font-family: monospace; /* ala <kbd> */
     clear: left;
+    margin-bottom: 1ex;
 }
 
 #Masthead tt { /* slashes in the xref: line */
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -685,6 +685,10 @@ table#dirlist { /* the "Name" column */
     width: 3em;
 }
 
+span.simplified-path {
+    color: #888;
+}
+
 /* file display */
 div[id^='src'] {
 }
--- a/web/history.jsp
+++ b/web/history.jsp
@@ -151,7 +151,6 @@ revision2 = revision2 >= hist.getHistoryEntries().size() ? hist.getHistoryEntrie
         + uriEncodedName %>"><%= path %></a>
     (Results <b> <%= start + 1 %> - <%= thispage + start
             %></b> of <b><%= totalHits %></b>)
-    <p class="slider"><%= slider %></p>
     </caption>
    
     <thead>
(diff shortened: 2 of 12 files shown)


vladak:
index and history view:

hist
index

vladak:
In history view, the slider is now displayed only at the bottom.

web/index.jsp
```diff
@@ -41,7 +41,7 @@ include file="httpheader.jspf"
 include file="pageheader.jspf"
 
             %></div>
-            <div id="Masthead"></div>
```
tulinkry: Copyright

web/history.jsp
```diff
@@ -22,7 +22,10 @@ Copyright (c) 2005, 2016, Oracle and/or its affiliates. All rights reserved.
 
 Portions Copyright 2011 Jens Elkner.
 
---%><%@page import="
```
tulinkry: Copyright

web/search.jsp
```diff
@@ -94,27 +94,30 @@ include file="pageheader.jspf"
 
 %>
         </div>
-        <div id="Masthead"></div>
```
tulinkry: Copyright

web/minisearch.jspf
```diff
@@ -0,0 +1,112 @@
...
+
+When distributing Covered Code, include this CDDL HEADER in each
+file and include the License file at LICENSE.txt.
+If applicable, add the following below this CDDL HEADER, with the
+fields enclosed by brackets "[]" replaced with your own identifying
+information: Portions Copyright [yyyy] [name of copyright owner]
+
+CDDL HEADER END
+
+Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: When it's a new file I always add only the actual year. At this point I'm afraid to ask if that's the way to do it

web/default/style.css
```diff
@@ -265,13 +265,14 @@ label {
 }
 
 #Masthead {
+    font-family: monospace; /* ala <kbd> */
```
tulinkry: It's not quite visible on the screenshot however I'd add a margin/padding for the bottom part of the info text. There is a little space at the top of it and so there should be a little space under it. I mean it's almost touching the orange line.

screenshot from 2017-01-18 14-33-23

web/pageheader.jspf
```diff
@@ -16,14 +16,13 @@ information: Portions Copyright [yyyy] [name of copyright owner]
...
 
-Copyright (c) 2007, 2010, Oracle and/or its affiliates. All rights reserved.
+Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
 
 --%><% 
 /* ---------------------- pageheader.jspf start --------------------- */
 {
     cfg = PageConfig.get(request);
 %>
-<a href="<%= request.getContextPath() %>/"><span id="MastheadLogo"></span></a>
```
tulinkry: I actually use this logo to return to the homepage all the time. But that's just a matter of habits.

tulinkry:
Also when viewing history of a file, the page title is (null). That's not expected.

screenshot from 2017-01-18 14-41-38

tulinkry:
The search field in history view is not working. It doesn't submit

screenshot from 2017-01-18 14-44-46

web/index.jsp
vladak: fixed

web/history.jsp
vladak: dto

web/minisearch.jspf
vladak: well, it's not exactly new since the content is basically copied from mast.jsp. I don't know the exact rules either so it's probably safer to leave the old year there.

web/pageheader.jspf
vladak: The alternative would be to replace the 'Home' link with a small logo, however the concept of clicking on OpenGrok logo leading to the home of the webapp is a bit strange as one could expect to lead to the project page on Github. So leaving it for next time.

web/default/style.css
vladak: added bottom margin

vladak:
fixed the history.jsp title.

vladak:
the non-working search was caused by missing opening `<form>` tag, good catch !

tulinkry:
Seems ok to me

PR #1337, opened 2017-01-20
adding a link to toggle scopes window in xref

### LINKED ISSUE #1333: make Scopes window 1st class citizen
When the Scopes window is closed, there is no way to open it again using UI elements. Possibly it can be toggled similarly to 'Navigate' and 'Line #'.

tulinkry:
fixes #1333


tulinkry:
Conflicts resolved

PR #1342, opened 2017-01-23
make index_body.html an include like {header,footer}_include

### LINKED ISSUE #533: mv index_body.html to DATA_ROOT/body_include (Bugzilla #19174)
status NEW severity _minor_ in component _webapp_ for _---_
Reported in version _unspecified_ on platform _ANY/Generic_
Assigned to: Lubos Kosco

Original attachment names and IDs:
- _bodyinc.patch_ (ID 4630)

On 2012-02-15 06:09:41 +0000, Jens Elkner wrote:

> Created attachment 4630
> patch
> 
> same as for {header,footer}_include

vladak:
fixes #533 

One of Jens'es forgotten commits. I think it makes sense for all include files to follow common scheme.


vladak:
By default, there will no longer be the message urging admins to perform customization. I like it better to be clear by default.

tarzanek:
will you add body_include default back ?
I think it could be populated the same way, with the info that admin can change this in body_include file ...
e.g. OpenGrok script could create such file when first run on new data dir?

tarzanek:
why I want it back?
Simply because it makes people look at how the code works and kinda opens the first door to get them to contribute back ... this is an open source project, we need people to play with the code ...

vladak:
I removed it on purpose - to have the default to be as clean as possible.

vladak:
I don't believe the presence of default index insert will make people to contribute back more.

tarzanek:
then merge please, let's see what will happen

Issue #517, opened 2013-05-22
multisearch breaks the whole search when one of the indexes is missing or has problems (Bugzilla #18974)


vladak:
status NEW severity _minor_ in component _webapp_ for _---_
Reported in version _unspecified_ on platform _ANY/Generic_
Assigned to: Lubos Kosco

On 2011-09-27 12:21:20 +0000, Lubos Kosco wrote:

> The whole multisearch should continue if one of the projects has problems with its indexes
> 
> PSARC folder had corrupted index on grok.cz and the search just broke instead of continuing
> - problem was that the indexreader upon attach couldn't recognize the directory - at this stage we should have just skipped this index directory instead of breaking the whole search
>   (not sure if lucene supports this though :( )

PR #1347, opened 2017-01-26
set HTTP error codes for error conditions

### LINKED ISSUE #17: "Index database(s) not found." should not be `200 OK`
When the index database is temporarily gone, OpenGrok returns `"Index database(s) not found."` (https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/web/SearchHelper.java) with a `200 OK` HTTP status code.

This breaks down HTTP accelerators (and probably various web-crawlers, too), which cache error pages momentarily (seconds), but non-error pages for an extended period of time (minutes or hours), where such OpenGrok error pages will be cached extensively, instead of being expired promptly.  Plus web-crawlers might update its index, instead of ignoring the result, and trying back later.

With `"Index database(s) not found."`, a non-200 http status code should be returned instead, probably `503 Service Unavailable`.

vladak:
fixes #17

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -18,7 +18,7 @@
  */
 
  /*
-  * Copyright (c) 2006, 2016, Oracle and/or its affiliates. All rights reserved.
+  * Copyright (c) 2006, 2017, Oracle and/or its affiliates. All rights reserved.
   */
 package org.opensolaris.opengrok.configuration;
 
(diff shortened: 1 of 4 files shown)


tulinkry:
This is ok. However I'm not sure if this doesn't go against the problem in #517.

vladak:
I think the multisearch issue is effectively about hiding the failure so it should not clash.

vladak:
Actually yes, there is a conflict. However the fix for #517 would have to radically change this area anyway (need to return `MultiReader` and at the same time error condition about project(s) with invalid indexes) so making the current code behaving correctly is more important I think.

tulinkry:
Ok

PR #1349, opened 2017-01-27
using a timeout for scopes update

tulinkry:
For a large file (.c file 8000 lines) the performance of the scopes window almost disables the user to scroll causing the browser to freeze.

This introduces a timeout so the callback is not triggered after every scroll event.


vladak:
Care to share the actual diff ? The versiobed renames make the change hard to comprehend.

Also, why is the scopes slow in the first place ? Is it just about the callback to be frequently called or is there some room for improvement elsewhere ?

tulinkry:
https://gist.github.com/tulinkry/f266ef587b95903b3da576d1680370fe/revisions

As I tried it the whole file is laggy, the screen freezes. I did a firefox profiler and everything was hung for tens of minutes. Eventually I got at least some answers - these calls to `hasClass` and `closest` in this function. Maybe it's actually the `getElementFromPoint` which, as I suppose, is the problem here for a file witch 8000 lines.
My outcome is that the event fires so often (which surely does) and for a long file this feels laggy. This timeout preserves the callback to be processed each time the scroll event is called.

vladak:
8k LOC and longer files are pretty common, this should be a stopper. Anyway,  merging now.

PR #1372, opened 2017-01-31
add list of projects to the title of the search page

### LINKED ISSUE #1365: add list of searched projects to the title of search results page
As noted in `PageConfig#getSearchTitle()`, the list of projects can be added to the title. Perhaps good idea would be to trim them (and append `...`) if the list is too long (say 10 projects).

vladak:
fixes #1365

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2011, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2017, Oracle and/or its affiliates. All rights reserved.
  * Portions copyright (c) 2011 Jens Elkner.
  */
 package org.opensolaris.opengrok.web;
@@ -45,6 +45,7 @@ import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import javax.servlet.ServletRequest;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
@@ -1340,6 +1341,14 @@ public final class PageConfig {
         return file.getName();
     }
 
+    private String addTitleDelimiter(String title) {
+        if (!title.isEmpty()) {
+            return title + ", ";
+        }
+
+        return title;
+    }
+
     /**
      * The search page title string should progressively reflect the search terms
      * so that if only small portion of the string is seen, it describes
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1340,6 +1340,14 @@ public String getShortPath(String path) {
         return file.getName();
     }
 
+    private String addTitleDelimiter(String title) {
```
tulinkry: copyright

tulinkry:
Looks good

src/org/opensolaris/opengrok/web/PageConfig.java
vladak: fixed

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1353,31 +1361,35 @@ public String getSearchTitle() {
...
         }
 
-        // TODO: possibly add projects too
+        if (req.getParameterValues(QueryBuilder.PROJECT) != null && req.getParameterValues(QueryBuilder.PROJECT).length != 0) {
+            if (!title.isEmpty()) {
+                title += " ";
+            }
+            title += "in projects: ";
+            String projects[] = req.getParameterValues(QueryBuilder.PROJECT);
+            for (int i = 0; i < projects.length; i++) {
```
tulinkry: Sorry that I didn't see this during the first review. In Java8 I think there is a possibility to make this shorter. Something like:
```java
Arrays.asList(a).stream().collect(Collectors.joining(","))
```

vladak: good suggestion, thanks !

tulinkry: I'm not sure but I suspect that raw "," is without the space between the elements

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1353,31 +1362,30 @@ public String getSearchTitle() {
...
         }
 
-        // TODO: possibly add projects too
+        if (req.getParameterValues(QueryBuilder.PROJECT) != null && req.getParameterValues(QueryBuilder.PROJECT).length != 0) {
+            if (!title.isEmpty()) {
+                title += " ";
+            }
+            title += "in projects: ";
+            String projects[] = req.getParameterValues(QueryBuilder.PROJECT);
+            title += Arrays.asList(projects).stream().collect(Collectors.joining(","));
```
tulinkry: I'm not sure but I suspect that the raw "," is without the space between the elements

vladak: yep, I realized that too and decided it looks better that way.

Issue #1360, opened 2017-01-30
footer displayed above page content for certain xrefs


vladak:
In 0.13-rc8 indexing file called foo.s with these contents:

```
/*
 *	 <- foo
 */
```

the page footer is placed above the contents:

![foo](https://cloud.githubusercontent.com/assets/2934284/22445721/0e981b98-e749-11e6-888e-79b687c4f78d.png)

This seems to be happening only with Universal ctags, not Exuberant ctags.

Notably, there is a folding icon on the line with `foo`. It seems that the arrow somehow triggers this behavior.

The badly placed footer is just a consequence of incorrectly formatted HTML it seems:

![html](https://cloud.githubusercontent.com/assets/2934284/22445791/61f10778-e749-11e6-8090-d6805fdf2bed.png)

vladak:
This happens not only for files with the arrow but also other files, my theory is this happens for all files where ctags detects functions.

tulinkry:
https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/analysis/JFlexXref.java#L339-L343 Maybe?

vladak:
Oh yes, I suspected bad scope termination.

mrate:
I'll investigate it, feel free to assign this issue to me. It seems like a good candidate for a test suite file.

vladak:
@mrate thanks. I fail to see the opening part of the `div` elemnt in the code.

PR #1382, opened 2017-02-02
Scopes

mrate:
Fix for #1360


src/org/opensolaris/opengrok/analysis/JFlexXref.java
```diff
@@ -308,7 +308,7 @@ public void write(Writer out) throws IOException {
 
         // terminate scopes
```
tulinkry: copyright 2017

vladak:
So, after this change are all the tags in the page closed ?

vladak:
Also, some sort of unit test would be nice.

vladak:
Verified that the change fixes the footer problem and HTML code comes out clean (at least FF does not mark any tags red anymore)

PR #1385, opened 2017-02-06
fix encoding/decoding of IgnoredNames objects

### LINKED ISSUE #1383: ignoredNames serialization problem causes indexing of hidden directories/files
When upgrading from 0.12 to 0.13-rc8, I noticed partial indexing is harvesting all the files under the `.hg` directories of Mercurial repositories. In fact, it was harvesting all the files under directories which would normally be ignored like `SCCS`, etc. This caused huge increase of indexing times.

Tracking down on what is happening I discovered this `ignoredNames` object dump from in `configuration.xml` as written by the indexer from 0.13-rc8 when run with `-W /var/opengrok/etc/configuration.xml -i bar.txt`:

```
  <void id="IgnoredNames0" property="ignoredNames">
   <void id="ArrayList0" property="items">
    <void index="26">
     <string>bar.txt</string>
    </void>
    <void index="27">
     <string>SCCS</string>
    </void>
    <void index="28">
     <string>CVS</string>
    </void>
    <void index="29">
     <string>CVSROOT</string>
    </void>
    <void index="30">
     <string>RCS</string>
    </void>
    <void index="31">
     <string>Codemgr_wsdata</string>
    </void>
    <void index="32">
     <string>deleted_files</string>
    </void>
    <void index="33">
     <string>.svn</string>
    </void>
    <void index="34">
     <string>.repo</string>
    </void>
    <void index="35">
     <string>.git</string>
    </void>
    <void index="36">
     <string>.hg</string>
    </void>
    <void index="37">
     <string>.razor</string>
    </void>
    <void method="add">
     <string>.bzr</string>
    </void>
   </void>
   <void property="items">
    <object idref="ArrayList0"/>
   </void>
  </void>
```

This does not look correct. Even if this could somehow work with all the index numbers, the default elements (i.e. anything but `bar.txt`) should not be present as they are not meant to be configurable.

In 0.12, i.e. prior to the changes done for #508 , the section would have looked like this:

```
  <void id="IgnoredNames0" property="ignoredNames">
   <void property="items">
    <void method="add">
     <string>bar.txt</string>
    </void>
   </void>
  </void>
```

so there is something wrong with serialization of `ignoredNames`. The class has 2 subclasses to provide functionality for files and directories.

There is a `getItems()`  method which looks like this:

```
44      public List<String> getItems() {
45          List<String> twoLists = new ArrayList<>();
46          twoLists.addAll(ignoredFiles.getItems());
47          twoLists.addAll(ignoredDirs.getItems());
48          return twoLists;
49      }
```

This method is used when `env.writeConfiguration()` is called from Indexer's `main()` so whatever it spits out has consequences on what will end up in `configuration.xml`.

Now, the problem with ignored files/directories seems to be happening for some reason only with partial indexing.

vladak:
fixes #1383

### DIFF
--- a/src/org/opensolaris/opengrok/index/IgnoredFiles.java
+++ b/src/org/opensolaris/opengrok/index/IgnoredFiles.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2007, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.index;
 
@@ -72,7 +72,7 @@ public final class IgnoredFiles extends Filter {
      * @return true if this file should be ignored, false otherwise
      */
     public boolean ignore(File file) {
-        return match(file) && file.isFile();
+        return file.isFile() && match(file);
     }
 
     /**
--- a/src/org/opensolaris/opengrok/index/IgnoredNames.java
+++ b/src/org/opensolaris/opengrok/index/IgnoredNames.java
@@ -18,11 +18,12 @@
  */
 
 /*
- * Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.index;
 
 import java.io.File;
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -31,10 +32,11 @@ import java.util.List;
  *
  * @author Vladimir Kotal
  */
-public class IgnoredNames {
+public class IgnoredNames implements Serializable {
+    private static final long serialVersionUID = 1L;
 
-    private final IgnoredFiles ignoredFiles;
-    private final IgnoredDirs ignoredDirs;
+    private IgnoredFiles ignoredFiles;
+    private IgnoredDirs ignoredDirs;
 
     public IgnoredNames() {
         ignoredFiles = new IgnoredFiles();
@@ -48,6 +50,13 @@ public class IgnoredNames {
         return twoLists;
     }
 
+    public void setItems(List<String> item) {
+        clear();
+        for (String s : item) {
+            add(s);
+        }
+    }
+
     public void add(String pattern) {
         if (pattern.startsWith("f:")) {
             ignoredFiles.add(pattern.substring(2));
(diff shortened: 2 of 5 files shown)


src/org/opensolaris/opengrok/index/IgnoredFiles.java
```diff
@@ -18,11 +18,12 @@
  */
 
 /*
- * Copyright (c) 2007, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.index;
 
 import java.io.File;
+import java.util.List;
```
tulinkry: Why this import?

src/org/opensolaris/opengrok/index/IgnoredNames.java
```diff
@@ -31,10 +32,10 @@
  *
  * @author Vladimir Kotal
  */
-public class IgnoredNames {
+public class IgnoredNames implements Serializable {
```
tulinkry: Shouldn't serializable object have some kind of `long` ID? Like [here](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/history/GitRepository.java#L58) in `GitRepository`.

src/org/opensolaris/opengrok/index/IgnoredNames.java
```diff
@@ -48,6 +49,13 @@ public IgnoredNames() {
         return twoLists;
     }
 
+    public void setItems(List<String> item) {
```
tulinkry: You just moved the method or?

src/org/opensolaris/opengrok/index/IgnoredNames.java
tarzanek: http://www.javapractices.com/topic/TopicAction.do?Id=45

src/org/opensolaris/opengrok/index/IgnoredFiles.java
vladak: removed

src/org/opensolaris/opengrok/index/IgnoredNames.java
vladak: yep, moved

src/org/opensolaris/opengrok/index/IgnoredNames.java
vladak: added serial version ID

tulinkry:
Ok

Issue #1271, opened 2016-11-11
consider EOL'ing JDBC


vladak:
The JDBC backend for storing history index data does not scale (at least with JavaDB) and is probably not much used, we should consider EOL'ing it.

vladak:
Another reason is JavaDB does not scale with renamed files. Recently support for renamed files in Git was added so this will be even more apparent

vladak:
Also, having centralized DB to store history would stand in way to #779

Issue #984, opened 2015-08-31
Warnings logged during index process


bond-:
I see warnings logged during index process. Unsure how I should handle them.

Is it possible to include a proper message as to what needs to be done during this error scenario.

```
11:13:35 WARNING: cannot write latest cached revision to file: null
11:13:35 WARNING: An error occured while creating cache for /path/to/project (GitRepository)
java.lang.NullPointerException
	at org.opensolaris.opengrok.history.FileHistoryCache.storeLatestCachedRevision(FileHistoryCache.java:595)
	at org.opensolaris.opengrok.history.FileHistoryCache.finishStore(FileHistoryCache.java:306)
	at org.opensolaris.opengrok.history.FileHistoryCache.store(FileHistoryCache.java:403)
	at org.opensolaris.opengrok.history.Repository.createCache(Repository.java:327)
	at org.opensolaris.opengrok.history.HistoryGuru.createCache(HistoryGuru.java:495)
	at org.opensolaris.opengrok.history.HistoryGuru.access$000(HistoryGuru.java:53)
	at org.opensolaris.opengrok.history.HistoryGuru$1.run(HistoryGuru.java:545)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

PR #1386, opened 2017-02-06
History generation respect ignored files

### LINKED ISSUE #1297: WARNING: Cannot write latest cached revision to file
I made directory `/var/opengrok/src` `/var/opengrok/data` `/var/opengrok/etc` and put code into `src`. All worked well until I `index` a new source directory. Then I got the WARNING:
```
WARNING: Cannot write latest cached revision to file xxx(the name of source directroy)
java.io.FileNotFoundException: /var/opengrok/data/historycache/OpenGroklatestRev (No such file or directory)
....
```
Is a bug? If so, how could I fix it?
Thanks:)

### LINKED ISSUE #1315: bug generating history for git repos with symlink to itself
I have noticed with the new Opengrok-0.13-rc5 a bug generating history for some git repos. It seems to be happening for git repos that have a symlink in them. here is a example error:

```
Dec 29, 2016 6:18:11 PM org.opensolaris.opengrok.history.HistoryGuru createCache
WARNING: An error occured while creating cache for mydir (GitRepository)
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
        at java.lang.String.substring(String.java:1931)
        at org.opensolaris.opengrok.history.FileHistoryCache.isRenamedFile(FileH
istoryCache.java:141)
        at org.opensolaris.opengrok.history.FileHistoryCache.store(FileHistoryCa
che.java:394)
        at org.opensolaris.opengrok.history.Repository.createCache(Repository.ja
va:337)
        at org.opensolaris.opengrok.history.HistoryGuru.createCache(HistoryGuru.
java:509)
        at org.opensolaris.opengrok.history.HistoryGuru.access$000(HistoryGuru.j
ava:54)
        at org.opensolaris.opengrok.history.HistoryGuru$1.run(HistoryGuru.java:5
60)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:51
1)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.
java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor
.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

### LINKED ISSUE #1351: The history generation does not respect ignored files/dirs and symlinks
Discovered while solving #1315.

Reproducible with steps:
```bash
mkdir repo
cd repo
git init
mkdir foo
touch main.c
git add .
git commit -m "commit"
```

Then run the indexer with an `-i main.c` switch to ignore such file. Investigating the data directory, there is a record for main.c in the historycache.

 - hitting http://localhost:8080/source/history/repo/main.c works (as unexpected) and shows the commit
 - hitting http://localhost:8080/source/xref/repo/main.c doesn't (as expected)

I believe the same works for `acceptedSymlinks` as the code lacks this code as well.

tulinkry:
fixes #1351
fixes #1315
hopefully fixes #1297


vladak:
Any significant performance degradations ?

vladak:
Not directly relevant however it would be nice to have some statistics about ignored/accepted file/dir counts (especially in the light of recent bug in `IgnoredNames` #1383) as it might provide quick sanity check at the end of indexing.

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,236 @@
...
+     *
+     * @param project if the file is relevant to some project (may be null)
+     * @param file the file to check
+     * @return true if the file should be included, false otherwise
+     */
+    public static boolean accept(Project project, File file) {
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        if (!env.getIncludedNames().isEmpty()
+                && // the filter should not affect directory names
+                (!(file.isDirectory() || env.getIncludedNames().match(file)))) {
```
vladak: this is a bit hard to read, maybe expand this to && ? Also, the comment should say why - I assume this is because of the directory check down below.

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param parent parent directory of the file
+     * @param file the file to check
+     * @return true if the file should be included, false otherwise
+     */
+    public static boolean accept(Project project, File parent, File file) {
+        try {
+            File f1 = parent.getCanonicalFile();
+            File f2 = file.getCanonicalFile();
+            if (f2.equals(f1)) {
+                LOGGER.log(Level.INFO, "Skipping links to itself...: {0} {1}",
```
vladak: these messages should be self-explanatory (i.e. make it clear what are the strings)

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param project a project (may be null)
+     * @param path the path to a file
+     * @return true if the file is local to the current repository
+     */
+    public static boolean isLocal(Project project, String path) {
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        String srcRoot = env.getSourceRootPath();
+
+        if (path.startsWith(srcRoot)) {
+            if (env.hasProjects()) {
```
vladak: could do the project != null check here to avoid the substring operation

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,236 @@
...
+        // Always accept local symlinks
+        if (isLocal(project, canonicalPath)) {
+            return true;
+        }
+
+        for (String allowedSymlink : env.getAllowedSymlinks()) {
+            if (absolutePath.startsWith(allowedSymlink)) {
+                String allowedTarget = new File(allowedSymlink).getCanonicalPath();
+                if (canonicalPath.startsWith(allowedTarget)
+                        && absolutePath.substring(allowedSymlink.length()).equals(canonicalPath.substring(allowedTarget.length()))) {
```
vladak: it'd be nice to have the logic explained in a comment

src/org/opensolaris/opengrok/util/AcceptHelper.java
tulinkry: I've expanded this. The question why is for @trondn because I'm quite missing the purpose of the included names here.

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,236 @@
...
+                return false;
+            }
+        } catch (IOException exp) {
+            LOGGER.log(Level.WARNING, "Failed to resolve name: {0}",
+                    absolutePath);
+            LOGGER.log(Level.FINE, "Stack Trace: ", exp);
+        }
+
+        if (file.isDirectory()) {
+            // always acceptFile directories so that their files can be examined
```
tulinkry: Refactor typo

src/org/opensolaris/opengrok/util/AcceptHelper.java
tulinkry: I added more info

src/org/opensolaris/opengrok/util/AcceptHelper.java
tulinkry: I added a comment

src/org/opensolaris/opengrok/util/AcceptHelper.java
tulinkry: Not really here, because then if the `project == null` then the return value is `true` which is not the original intention. I moved the code around to avoid the substring operation.

tulinkry:
I filed #1388 for the statistics addon which would be a non-trivial change in this PR.

tulinkry:
Performance:

Of course that the performance must go down due to this because we do more work on the file system checking the files. I tried to improve it so it does not that many operations but still some are neccessary to check it. A future improvement is to collect all files which are accepted once and use it later in the index phase (this is history phase) - now it's done twice (once per phase).

With this in mind when I index the OpenGrok source code there is no significant slow down when compared to the master branch without this PR.

tulinkry:
Possibly fixes also #984

tulinkry:
@vladak I made the changes; can you review please?

vladak:
It'd probably be prudent to address this in JDBC history cache however given there are some plans to EOF it (#1271) it's not probably necessary assuming the EOL will happen sooner than later.

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,237 @@
...
+     *
+     * @param file the file to check
+     * @return true if the file should be included, false otherwise
+     */
+    public static boolean accept(File file) {
+        return accept((Project) null, file);
+    }
+
+    /**
+     * Check if I should accept this file into the database
```
vladak: it'd nice to have some more info about the function, e.g. mention how exactly project is relevant

src/org/opensolaris/opengrok/util/AcceptHelper.java
```diff
@@ -0,0 +1,237 @@
...
+     * @param parent parent directory of the file
+     * @param file the file to check
+     * @return true if the file should be included, false otherwise
+     */
+    public static boolean accept(File parent, File file) {
+        return accept(null, parent, file);
+    }
+
+    /**
+     * Check if this file should be accepted.
```
vladak: same here, state how parent is used in the method

vladak:
Also, `HistoryGuru#addRepositories()` should be addressed too - the traversal scans for the repositories inside SCM metadata directories which slows down the process unnecessarily and incurs more I/O.

vladak:
Any plan to refresh this ?

vladak:
This was superseded by #3137.

PR #1392, opened 2017-02-07
defining the type of cfg (PageConfig)

### LINKED ISSUE #1356: Source code annotation culumns are too thin
grok
Source code annotation columns are too thin. Changeset number is trimmed (maybe is overwritten by next column) and author name overlap into code.  FF 45.7 (chrome works better, but still not optimal)

tulinkry:
fixes #1356

This was some kind of race condition (that's what I think). The part under the diff to write the `<style>` into the page config sometimes didn't get executed. It's still in our code that the `cfg` variable is defined as a private property of the compiled JSP which must be a race condition if there is a heavy traffic and threads start to use this variable. **Therefore I suggest the next fix which is to get rid of this private variable.**

This fix makes the `cfg` local variable which shadows the member variable.

I discovered that this behaviour only happens under heavy traffic (or heavy load of the machine + heavy traffic). I tested the localhost with apache benchmark:

```bash
ab -n 1000 -c 10 -H 'X-FROM: 007' http://localhost:8080/source/xref/opengrok-master/src/org/opensolaris/opengrok/index/IndexDatabase.java?a=true
```

The results of successfully stored `<style>` data was:
 - 478 before the fix (out 1000 requests)
 - 10000 after the fix (out of 1000 requests)

### DIFF
--- a/web/list.jsp
+++ b/web/list.jsp
@@ -18,7 +18,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 
 CDDL HEADER END
 
-Copyright (c) 2005, 2016, Oracle and/or its affiliates. All rights reserved.
+Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
 Portions Copyright 2011 Jens Elkner.
 
 --%>
@@ -51,7 +51,7 @@ org.opensolaris.opengrok.web.DirectoryListing"
     if (request.getCharacterEncoding() == null) {
         request.setCharacterEncoding("UTF-8");
     }
-    cfg = PageConfig.get(request);
+    PageConfig cfg = PageConfig.get(request);
     Annotation annotation = cfg.getAnnotation();
     if (annotation != null) {
         int r = annotation.getWidestRevision();


tulinkry:
Also it's kinda hard to make a unit test case for this

vladak:
It's interesting that most of jsp's follow this kind of pattern. I wonder if this could happen elsewhere too.

tulinkry:
Yes It's everywhere. We need to refactor to use only local scope variables

vladak:
Perhaps change them all while you're at it ?

I still wonder what is the exact mechanism of the race. Is it possible that all threads handling requests share the same cfg variable ?

tulinkry:
Yes I believe that this compiled jsp class (containing all of the code) has only one instance in the container (which looks reasonable that each request does not create a new one) and gets executed in parallel for the incoming requests and thus the private variable is shared across the threads. That's what comes to my mind.

> Perhaps change them all while you're at it ?

I tried yesterday but for some files it's quite complicated (as they share the scopes and some variables through the includes) so I decided to put it in another PR.

tulinkry:
> has only one instance in the container

This is true, the compiled jsp is eventually used as a servlet which indeed has only one instance in tomcat serving the `doGet` or `doPost` methods.

PR #1400, opened 2017-02-09
turn configuration update into Message

### LINKED ISSUE #1355: allow to update just the configuration in webapp
When thinking about how to update Group configuration in readonly config and refresh the webapp without reindexing, I found there is no easy way. It should really be possible to run something like `OpenGrok updateconfig` and the webapp would use the new config.

vladak:
fixes #1355

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -98,6 +98,7 @@ import static java.nio.file.FileVisitResult.CONTINUE;
 import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
 import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
 import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static org.opensolaris.opengrok.configuration.Configuration.makeXMLStringAsConfiguration;
 
 /**
  * The RuntimeEnvironment class is used as a placeholder for the current
@@ -1144,11 +1145,15 @@ public final class RuntimeEnvironment {
      * @param port the port to use on the host
      * @throws IOException if an error occurs
      */
-    public void writeConfiguration(InetAddress host, int port) throws IOException {
-        try (Socket sock = new Socket(host, port);
-                XMLEncoder e = new XMLEncoder(sock.getOutputStream())) {
-            e.writeObject(threadConfig.get());
-        }
+    public void writeConfiguration(String host, int port) throws IOException {
+        Message m = Message.createMessage("config");
+        m.addTag("reindex");
+        m.setText(configuration.getXMLRepresentationAsString());
+        m.write(host, port);
+    }
+
+    public void writeConfiguration(InetAddress hostAddr, int port) throws IOException {
+        writeConfiguration(hostAddr.getHostAddress(), port);
     }
 
     protected void writeConfiguration() throws IOException {
@@ -1280,8 +1285,8 @@ public final class RuntimeEnvironment {
     }
 
     /**
-     * Add a message to the application Also schedules a expirationTimer to
-     * remove this message after its expiration.
+     * Add a message to the application.
+     * Also schedules a expiration timer to remove this message after its expiration.
      *
      * @param m the message
      */
@@ -1505,8 +1510,43 @@ public final class RuntimeEnvironment {
     private Thread configurationListenerThread;
 
     /**
-     * Start a thread to listen on a socket to receive new configurations to
-     * use.
+     * Set configuration from a message. The message could have come from
+     * the Indexer (in which case some extra work is needed) or is it just
+     * a request to set new configuration in place.
+     *
+     * @param m message containing the configuration
+     * @param reindex is the message result of reindex
+     */
+    public void applyConfig(Message m, boolean reindex) {
+        Configuration config;
+        try {
+            config = makeXMLStringAsConfiguration(m.getText());
+        } catch (IOException ex) {
+            LOGGER.log(Level.WARNING, "Configuration decoding failed" + ex);
+            return;
+        }
+
+        setConfiguration(config);
+        LOGGER.log(Level.INFO, "Configuration updated: {0}",
+                configuration.getSourceRoot());
+
+        if (reindex) {
+            // We are assuming that each update of configuration
+            // means reindex. If dedicated thread is introduced
+            // in the future solely for the purpose of getting
+            // the event of reindex, the 2 calls below should
+            // be moved there.
+            refreshSearcherManagerMap();
+            maybeRefreshIndexSearchers();
+            // Force timestamp to update itself upon new config arrival.
+            config.refreshDateForLastIndexRun();
+        }
+    }
+
+    /**
+     * Start a thread to listen on a socket to receive new messages.
+     * The messages can contain various commands for the webapp, including
+     * upload of new configuration.
      *
      * @param endpoint The socket address to listen on
      * @return true if the endpoint was available (and the thread was started)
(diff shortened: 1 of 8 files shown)


src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -100,6 +103,14 @@ public static Message createMessage(String type) {
         return null;
     }
 
+    public Configuration getConfiguration() {
```
tulinkry: I'd rather introduce a field called `file` containing the content. Configuration has no meaning for any other type of message however we can use appending a file content in the future somehow.

Actually you can change it so it uses the `text` field for a file content and use `tags` for signalizing the reindex without the need to change the abstract type.

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -43,6 +43,7 @@
 import java.util.TreeSet;
```
tulinkry: copyright

src/org/opensolaris/opengrok/configuration/messages/Messages.java
```diff
@@ -126,18 +133,33 @@ public static void main(String[] argv) {
...
         Message m = Message.createMessage(type);
 
         if (m == null) {
             System.err.println("Uknown message type " + type);
             b_usage();
-            System.exit(4);
+            System.exit(1);
+        }
+
+        if (type.compareTo("config") == 0) {
```
tulinkry: This is against the OOP design in this case. Better introduce some new method like `validate` for the message and return boolean there. This is also related to #1403.

I can change it in the future

src/org/opensolaris/opengrok/configuration/messages/Messages.java
```diff
@@ -28,6 +28,9 @@
 import java.util.ArrayList;
```
tulinkry: copygirh

test/org/opensolaris/opengrok/configuration/messages/MessagesTest.java
```diff
@@ -133,15 +133,15 @@ public void testMessageSendSuccess() {
     @Test
```
tulinkry: copyright

test/org/opensolaris/opengrok/configuration/messages/MessagesTest.java
```diff
@@ -133,15 +133,15 @@ public void testMessageSendSuccess() {
...
     public void testMessageSendWrongHost() {
         RuntimeEnvironment.getInstance().setMessageLimit(100);
-        Assert.assertEquals(5, invokeMain("localhost", portNum + 2));
+        Assert.assertEquals(1, invokeMain("localhost", portNum + 2));
```
tulinkry: why 1?

tools/Messages
```diff
@@ -325,42 +330,50 @@ do
...
-[ -z "$MESSAGE_EXPIRE" ] && echo "The expire must be set" && Usage && exit 5
+if [ $MESSAGE_TYPE != "config" ]; then
+    [ -z "$MESSAGE_TEXT" ] && echo "The text of the message must be specified" && Usage && exit 3
+    [ -z "$MESSAGE_EXPIRE" ] && echo "The expire must be set" && Usage && exit 5
 
-if [ -z "$MESSAGE_EXPIRE_TIMESTAMP" ]
-then
-    MESSAGE_EXPIRE_TIMESTAMP=`"$DATE" --date="$MESSAGE_EXPIRE" +"%s"`
-    if [ $? -ne 0 ]
+    if [ -z "$MESSAGE_EXPIRE_TIMESTAMP" ]
```
tulinkry: I'm not sure about this what happens when the timestamp is not set for the config message? Because it's still present in the call to the java class.

test/org/opensolaris/opengrok/configuration/messages/MessagesTest.java
vladak: it's the changed exit code. `Messages` produce error messages to `stderr` so I saw no extra value in the distinct error codes.

test/org/opensolaris/opengrok/configuration/messages/MessagesTest.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/messages/Messages.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/messages/Message.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/messages/Message.java
vladak: good idea, changed

src/org/opensolaris/opengrok/configuration/messages/Messages.java
vladak: moved the load to setTextFromFile() in Message at the cost of not checking the usage fully in Messages

tools/Messages
vladak: `./tools/Messages -v -n config /var/opengrok/etc/configuration.xml` results in:

`
/usr/jdk/latest/bin/java -Xmx2048m -classpath :/export/home/vkotal/opengrok-git-vladak/tools/../dist/opengrok.jar org.opensolaris.opengrok.configuration.messages.Messages -m config -c info -g load -s localhost -p 2424 -f /var/opengrok/etc/configuration.xml --
`

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1144,11 +1144,19 @@ public void writeConfiguration(File file) throws IOException {
...
      */
-    public void writeConfiguration(InetAddress host, int port) throws IOException {
-        try (Socket sock = new Socket(host, port);
-                XMLEncoder e = new XMLEncoder(sock.getOutputStream())) {
-            e.writeObject(threadConfig.get());
-        }
+    public void writeConfiguration(String host, int port) throws IOException {
+        Message m = Message.createMessage("config");
+        m.addTag("reindex");
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
tulinkry: https://github.com/OpenGrok/OpenGrok/blob/11fb9582c1764812abdd20b783adb6af7a823033/src/org/opensolaris/opengrok/configuration/Configuration.java#L980-L984

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1505,8 +1513,39 @@ public void stopConfigurationListenerThread() {
...
-     * use.
+     * Set configuration from a message. The message could have come from
+     * the Indexer (in which case some extra work is needed) or is it just
+     * a request to set new configuration in place.
+     *
+     * @param m message containing the configuration
+     * @param reindex is the message result of reindex
+     */
+    public void applyConfig(Message m, boolean reindex) {
+        XMLDecoder d = new XMLDecoder(new ByteArrayInputStream(m.getText().getBytes()));
```
tulinkry: https://github.com/OpenGrok/OpenGrok/blob/11fb9582c1764812abdd20b783adb6af7a823033/src/org/opensolaris/opengrok/configuration/Configuration.java#L998-L1003

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -0,0 +1,25 @@
+/*
```
tulinkry: The whole copyright notice is missing here

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -0,0 +1,25 @@
...
+import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
+
+/**
+ *
+ * @author Vladimir Kotal
+ */
+public class ConfigMessage extends Message {
+    @Override
+    public byte[] apply(RuntimeEnvironment env) throws IOException {
+        if (hasTag("reindex")) {
```
tulinkry: Possibly shorter to just write
```java
env.applyConfig(this, hasTag("reindex"));
```
but this is ok

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -108,6 +111,16 @@ public void setText(String text) {
         this.text = text;
     }
 
+    public void setTextFromFile(String filepath) {
+        try {
+            byte[] encoded = Files.readAllBytes(Paths.get(filepath));
+            this.text = new String(encoded);
+        } catch (IOException ex) {
+            System.err.println("Cannot read '" + filepath + "': " + ex);
+            System.exit(1);
```
tulinkry: Do not use system.exit, this is not a main program. Throw an exception or return a boolean

tools/Messages
tulinkry: ok fine

test/org/opensolaris/opengrok/configuration/messages/MessagesTest.java
tulinkry: Oh yes that's the exit code. I was confused

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
vladak: added

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
vladak: changed

src/org/opensolaris/opengrok/configuration/messages/Message.java
vladak: good catch, bad refactor, fixed.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: fixed

PR #1408, opened 2017-02-13
always terminate the timer thread

### LINKED ISSUE #1407: Incorrectly terminating timer thread
I know this is not a big problem however I think in some circumstances the timer thread in [Executor#151](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/util/Executor.java#L151) is not stopped correctly.

I sometimes see messages in my tomcat log:
```log
13-Feb-2017 15:26:03.839 WARNING [http-nio-8080-exec-3] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [source] appears to have started a thread named [Timer-3] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 java.lang.Object.wait(Native Method)
 java.lang.Object.wait(Object.java:502)
 java.util.TimerThread.mainLoop(Timer.java:526)
 java.util.TimerThread.run(Timer.java:505)
13-Feb-2017 15:26:03.840 WARNING [http-nio-8080-exec-3] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [source] appears to have started a thread named [Timer-1] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 java.lang.Object.wait(Native Method)
 java.lang.Object.wait(Object.java:502)
 java.util.TimerThread.mainLoop(Timer.java:526)
 java.util.TimerThread.run(Timer.java:505)
```

It's not really deterministic; sometimes it occurs sometimes it doesn't. As you can see the messages don't contain any useful information but I still suspect this timer for long processes. This also was happening before we add the timer thread for messages around #1195.

tulinkry:
possibly fixes #1407 

This correctly terminates the timer thread. Hope to see that it will fix #1407

### DIFF
--- a/src/org/opensolaris/opengrok/util/Executor.java
+++ b/src/org/opensolaris/opengrok/util/Executor.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 
 package org.opensolaris.opengrok.util;
@@ -37,7 +37,6 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/util/Executor.java
```diff
@@ -148,7 +147,12 @@ public int exec(final boolean reportExceptions, StreamHandler handler) {
         ProcessBuilder processBuilder = new ProcessBuilder(cmdList);
         final String cmd_str = processBuilder.command().toString();
         final String dir_str;
-        Timer t = new Timer();
+        Timer t = null;
+
+        if (this.timeout != 0) {
+            // invoking the constructor starts the background thread
+            t = new Timer();
```
vladak: could you move it further down - possibly under the `if this.timeout != 0` check ?

tulinkry: Moved

PR #1412, opened 2017-02-14
Remove file/revision ambiguity from git history log command

DaveRandom:
Same fix for same issue that is avoided by https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/history/GitRepository.java#L165


vladak:
Hi Dave, I know this is minor change however we need you to get OCA (Oracle Contributor Agreement) before we can merge this in.

DaveRandom:
I have sent the signed document via email as per the process described [here](http://www.oracle.com/technetwork/community/oca-486395.html).

Is there anything else I need to do?

vladak:
Great, accepting this now.

Issue #1398, opened 2017-02-09
null ptr exception in getSearchTitle()


vladak:
With 0.13-rc9 the following exception is seen:

```
java.lang.NullPointerException
        at org.opensolaris.opengrok.web.PageConfig.getSearchTitle(PageConfig.java:1361)
        at org.apache.jsp.search_jsp._jspService(search_jsp.java:264)
        at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:70)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:729)
        at org.apache.jasper.servlet.JspServletWrapper.service(JspServletWrapper.java:443)
        at org.apache.jasper.servlet.JspServlet.serviceJspFile(JspServlet.java:385)
        at org.apache.jasper.servlet.JspServlet.service(JspServlet.java:329)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:729)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:230)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:165)
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:192)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:165)
        at org.opensolaris.opengrok.web.StatisticsFilter.doFilter(StatisticsFilter.java:55)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:192)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:165)
        at org.opensolaris.opengrok.web.AuthorizationFilter.doFilter(AuthorizationFilter.java:66)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:192)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:165)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:199)
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:108)
[...]

PR #1415, opened 2017-02-15
removing member (global) variable cfg

tulinkry:
- most of the changes are just about introducing local `cfg` variable.
- on same places it was necessary to store the "context" into the request
- the most difficult was the `history.jsp` where there was an `if` statement from the top to the bottom - please check it

This completely solves the problem in #1392


web/history.jsp
```diff
@@ -86,90 +65,149 @@ include file="httpheader.jspf"
...
+                    } else if (label == labelEnd && i < totalHits) {
+                        buf.append("&gt;&gt;");
+                    } else {
+                        buf.append(label);
+                    }
+                    buf.append("</a>");
+                }
+                label++;
+            }
+            slider = buf.toString();
```
vladak: seems like a candidate for getting this into method in `PageConfig` ?

tulinkry: Quite yes. It is already three times in the code

vladak:
I am still wondering why the `private PageConfig cfg` in `pageconfig.jspf` did not work.

web/history.jsp
```diff
@@ -86,90 +65,149 @@ include file="httpheader.jspf"
...
+            request.setAttribute("history.jsp-slider", slider);
+        } else {
+            // set the max index to max or last
+            thispage = totalHits - start;
+        }
+%>
         </div>
         <div id="Masthead">History log of 
         <%= Util.breadcrumbPath(context + Prefix.XREF_P, path,'/',"",true,cfg.isDir()) %>
```
vladak: what about this `cfg` ?

vladak:
In `history.jsp` there are bunch of `cfg` references (`cfg.isDir()` calls on L234, L272, L285 + `cfg.getRevisionMessageCollapseThreshold()` on L359) where I am not sure about `cfg` origin.

vladak:
Other than the comments above this looks good.

vladak:
Just an idea for `history.jsp`: try to get it compiled into Java code and then search for all `cfg` occurrences.

web/history.jsp
tulinkry: In scope from line 95. The scope ends on 161.

tulinkry:
> In history.jsp there are bunch of cfg references (cfg.isDir() calls on L234, L272, L285 + cfg.getRevisionMessageCollapseThreshold() on L359) where I am not sure about cfg origin.

The local scope for this `cfg` is 173-434

tulinkry:
> Just an idea for history.jsp: try to get it compiled into Java code and then search for all cfg occurrences.

That's what I was doing the whole time to fix the missing member variable.

tulinkry:
> I am still wondering why the private PageConfig cfg in pageconfig.jspf did not work.

[As you found in the `PageConfig`](https://github.com/OpenGrok/OpenGrok/issues/1398#issuecomment-280021861) the page config must not be shared among requests. And this member variable **was** shared among parallel requests because the jsp is turned into a simple servlet and tomcat holds only one instance of a servlet to serve all incoming requests.

web/history.jsp
tulinkry: I extracted the slider into utils

vladak:
Looks good, pls merge.

Issue #1403, opened 2017-02-10
Messages -n abort should not require text


vladak:
`/usr/opengrok/bin/Messages -n abort -t foo` exits with error:

```
The text of the message must be specified
```

This does not make sense to me since there is text to be deleted.

tulinkry:
I know that this could be solved with simple if statement. However the original design was to be independent on the exact message type and therefore I decided to keep the text required. Perhaps we could make it optional for all messages and act upon in the specific message?

vladak:
I'd make text optional.

PR #1416, opened 2017-02-15
validate method for validating the specific message type

tulinkry:
- oop design for validating custom types
- basic test coverage
- filling the documentation with the new config type

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/messages/Message.java
+++ b/src/org/opensolaris/opengrok/configuration/messages/Message.java
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
-/*
+ /*
  * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.configuration.messages;
@@ -45,7 +45,6 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import org.opensolaris.opengrok.configuration.Configuration;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.util.XmlEofOutputStream;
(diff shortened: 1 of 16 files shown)


tulinkry:
And fixes #1403

src/org/opensolaris/opengrok/configuration/messages/StatsMessage.java
```diff
@@ -43,4 +46,18 @@
         }
         return Util.statisticToJson(env.getStatistics()).toJSONString().getBytes();
     }
+
+    @Override
+    public void validate() throws Exception {
+        if (getText() == null) {
+            throw new Exception("The message must contain a text.");
+        }
+        List<String> allowed = Arrays.asList(new String[]{"get", "reload", "clean"});
```
vladak: these should probably be defined only once in the file

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -84,6 +83,19 @@
     public abstract byte[] apply(RuntimeEnvironment env) throws Exception;
 
     /**
+     * Validate the current message and throw an exception if the message is not
+     * valid. Further implementation can override this method to get the message
+     * into a state they would expect.
+     *
+     * @throws Exception
+     */
+    public void validate() throws Exception {
```
vladak: what about `className` ?

vladak:
What about performing validation in `apply()` ?

tulinkry:
> What about performing validation in apply() ?

good idea. fixed

src/org/opensolaris/opengrok/configuration/messages/StatsMessage.java
tulinkry: fixed

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: Class name has only meaning for normal message so I don't validate it. That's a bad original design for now.

vladak:
Looks good, squash and merge.

PR #1417, opened 2017-02-15
hiding empty groups

### LINKED ISSUE #1389: Empty project groups should be hidden
On the OpenGrok front page, the empty project groups should be hidden (by default).
hide_empty_groups

tulinkry:
fixes #1389

### DIFF
--- a/web/repos.jspf
+++ b/web/repos.jspf
@@ -126,7 +126,9 @@ Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
                         </h4>
                     </div>
                         <div class="panel-body-accordion<% if (pHelper.hasFavourite(group)) { %> favourite<% } %>"
-                        <% if (pHelper.hasFavourite(group) || pHelper.getRepositories(group).size() < 5) { %>data-accordion-visible="true"<% } %>>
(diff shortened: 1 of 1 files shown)


web/repos.jspf
```diff
@@ -126,7 +126,9 @@ Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
                         </h4>
                     </div>
                         <div class="panel-body-accordion<% if (pHelper.hasFavourite(group)) { %> favourite<% } %>"
-                        <% if (pHelper.hasFavourite(group) || pHelper.getRepositories(group).size() < 5) { %>data-accordion-visible="true"<% } %>>
+                        <% if (pHelper.hasFavourite(group) ||
+                                (pHelper.getRepositories(group).size() > 0 && pHelper.getRepositories(group).size() < 5)) {
```
vladak: not related to this however finally I see the reason for the behavior where some groups are expanded and some not. Time for a tunable ?

tulinkry: #1420

tulinkry: And it's written in a code comment that it depends on two things
1. user's cookie
2. this threshold

PR #1420, opened 2017-02-16
making the group collapse threshold a tunable

tulinkry:


tulinkry:
I updated the wiki with the threshold

PR #1423, opened 2017-02-17
refresh SearcherManagers after partial reindex

### LINKED ISSUE #1187: signal to the webapp once partial reindex is done
Once the configuration listener protocol is extended so that allows more than just Configuration object to be passed, the protocol should be further extended so that partial reindex notifies the webapp to call maybeRefresh() on **specific** SearcherManager objects. This is a follow-up to the changes for #1132.

vladak:
fixes #1187

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -1172,6 +1172,26 @@ public final class RuntimeEnvironment {
     }
 
     /**
+     * Send message to webapp to refresh SearcherManagers for given projects.
+     * This is used for partial reindex.
+     *
+     * @param subFiles list of directories to refresh corresponding SearcherManagers
+     * @param host the host address to receive the configuration
+     * @param port the port to use on the host
+     * @throws IOException if an error occurs
+     */
+    public void signalTorefreshSearcherManagers(List<String> subFiles, String host, int port) throws IOException {
+        Message m = Message.createMessage("refresh");
+        m.setText("refresh");
+        for (String proj : subFiles) {
+            // subFile entries start with path separator so get basename
+            // to convert them to project names.
+            m.addTag(new File(proj).getName());
+        }
+        m.write(host, port);
+    }
+
+    /**
      * Generate a TreeMap of projects with corresponding repository information.
      *
      * Project with some repository information is considered as a repository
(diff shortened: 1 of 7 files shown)


vladak:
Not entirely sure how to test this.

This would probably work however is a bit heavyweight:
  - disable periodic refreshing of searcherManagers
  - full index of a repo
  - start webapp listener
  - perform search (so that searcherManager is created)
    - either the whole webapp needs to be running or use the index files directly
  - modify the repo
  - partial index
  - perform search for the changes done with the most recent modification

src/org/opensolaris/opengrok/configuration/messages/RefreshMessage.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Message for refreshing SearcherManagers in the webapp after partial reindex.
+ * The list of projects is carried in tags.
+ *
+ * @author Vladimir Kotal
+ */
+public class RefreshMessage extends Message {
+    @Override
+    public byte[] apply(RuntimeEnvironment env) throws IOException {
```
tulinkry: Did you get this to compile with merged master? `apply` now is marked as `final`. The method you should override is `applyMessage`

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1172,6 +1172,26 @@ protected void writeConfiguration() throws IOException {
...
+     * This is used for partial reindex.
+     *
+     * @param subFiles list of directories to refresh corresponding SearcherManagers
+     * @param host the host address to receive the configuration
+     * @param port the port to use on the host
+     * @throws IOException if an error occurs
+     */
+    public void signalTorefreshSearcherManagers(List<String> subFiles, String host, int port) throws IOException {
+        Message m = Message.createMessage("refresh");
+        m.setText("refresh");
```
tulinkry: Is the text necessary? Add a validation for text field? And possibly create a test for this message testing the validation.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1172,6 +1172,26 @@ protected void writeConfiguration() throws IOException {
...
+        for (String proj : subFiles) {
+            // subFile entries start with path separator so get basename
+            // to convert them to project names.
+            m.addTag(new File(proj).getName());
```
tulinkry: No need to check if the file actually exists or is a project?

src/org/opensolaris/opengrok/configuration/messages/RefreshMessage.java
vladak: merging now

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: the text is not needed, will add simple validation+test

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: For correctness sake, this is checked in `maybeRefreshIndexSearchers()`. Not sure it is worth adding the reply back to the `Messages` tool that no such project exists.

tulinkry: Ok I see. I don't think there should be a reply

src/org/opensolaris/opengrok/configuration/messages/RefreshMessage.java
```diff
@@ -0,0 +1,47 @@
+/*
```
tulinkry: I don't know why it disappeared anyway write a documentation note about this new type to:
1. Messages.java (usage)
2. README.txt (message types)

I can update the wiki by myself

vladak: added, although the message is sort of private.

PR #1425, opened 2017-02-17
escaping titles in linkify()

tulinkry:
+ removing non existent import

```java
-<%@page import="org.jcp.xml.dsig.internal.dom.Utils"%>
```

I have no idea how that came there.

### DIFF
--- a/src/org/opensolaris/opengrok/web/Util.java
+++ b/src/org/opensolaris/opengrok/web/Util.java
@@ -448,7 +448,7 @@ public final class Util {
                     sb.append("&#39;");
                     break;
                 case '"':
-                    sb.append("&#39;");
+                    sb.append("&quot;");
                     break; // \\\
                 case '&':
                     sb.append("&amp;");
@@ -1368,7 +1368,7 @@ public final class Util {
                 return String.format("<a href=\"%s\"%s title=\"Link to %s\">%s</a>",
                         constructed.toString(),
                         newTab ? " target=\"_blank\"" : "",
-                        Util.htmlize(url),
+                        Util.encode(url),
                         Util.htmlize(url)
                 );
             } catch (MalformedURLException | URISyntaxException ex) {
--- a/src/org/opensolaris/opengrok/web/Util.java
+++ b/src/org/opensolaris/opengrok/web/Util.java
@@ -434,48 +434,67 @@ public final class Util {
 
     /**
      * Converts different html special characters into their encodings used in
-     * html. Currently used only for tooltips of annotation revision number view
+     * html.
      *
      * @param s input text
      * @return encoded text for use in &lt;a title=""&gt; tag
      */
     public static String encode(String s) {
-        StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < s.length(); i++) {
-            char c = s.charAt(i);
-            switch (c) {
-                case '\'':
-                    sb.append("&#39;");
-                    break;
-                case '"':
-                    sb.append("&quot;");
-                    break; // \\\
-                case '&':
-                    sb.append("&amp;");
-                    break;
-                case '>':
-                    sb.append("&gt;");
-                    break;
-                case '<':
-                    sb.append("&lt;");
-                    break;
-                case ' ':
-                    sb.append("&nbsp;");
-                    break;
-                case '\t':
-                    sb.append("&nbsp;&nbsp;&nbsp;&nbsp;");
-                    break;
-                case '\n':
-                    sb.append("&lt;br/&gt;");
-                    break;
-                case '\r':
-                    break;
-                default:
-                    sb.append(c);
-                    break;
-            }
+        StringBuilder sb = new StringBuilder((int) Math.max(10, s.length() * 1.5));
+        try {
+            encode(s, sb);
+        } catch (IOException ex) {
+            // IOException cannot happen when the destination is a
+            // StringBuilder. Wrap in an AssertionError so that callers
+            // don't have to check for an IOException that should never
+            // happen.
+            throw new AssertionError("StringBuilder threw IOException", ex);
         }
         return sb.toString();
+    }
+
+    /**
+     * Converts different html special characters into their encodings used in
+     * html.
+     *
+     * @param s input text
+     * @param dest appendable destination for appending the encoded characters
+     * @throws java.io.IOException
+     */
+    public static void encode(String s, Appendable dest) throws IOException {
+        for (int i = 0; i < s.length(); i++) {
+            char c = s.charAt(i);
+            if (c > 127 || c == '"' || c == '<' || c == '>' || c == '&' || c == '\'') {
+                // special html characters
+                dest.append("&#").append("" + (int) c).append(";");
+            } else if (c == ' ') {
+                // non breaking space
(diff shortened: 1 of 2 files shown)


vladak:
That particular import bit me today when trying to deploy the webapp.

tulinkry:
I guess NB were autosuggesting some import and I accidentally hit that. However I wonder why it didn't fail yesterday.

tulinkry:
I created another PR #1430 for the unused import to get rid of it immediately

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -1294,7 +1294,7 @@ public static String linkify(String url, boolean newTab) {
                 return String.format("<a href=\"%s\"%s title=\"Link to %s\">%s</a>",
                         constructed.toString(),
                         newTab ? " target=\"_blank\"" : "",
-                        Util.htmlize(url),
+                        Util.encode(url),
```
vladak: The comment next to `Util.encode()` says `Currently used only for tooltips of annotation revision number view` so it should be augmented.

vladak: while you're at it, `encode()` should probably use `Appandable` like `htmlize()` and say that it actually appends the converted char.

tulinkry: I added the appendable parameter

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -433,48 +433,67 @@ public static String readableSize(long num) {
...
+                dest.append("&lt;br/&gt;");
+            } else {
+                dest.append(c);
             }
         }
-        return sb.toString();
+    }
+
+    /**
+     * Encode url
```
vladak: nit: URL ?

tulinkry: I changed that to capital URL...I'm not sure what did you mean?

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -433,48 +433,67 @@ public static String readableSize(long num) {
...
      * Converts different html special characters into their encodings used in
-     * html. Currently used only for tooltips of annotation revision number view
+     * html.
      *
      * @param s input text
      * @return encoded text for use in &lt;a title=""&gt; tag
      */
     public static String encode(String s) {
-        StringBuilder sb = new StringBuilder();
+        StringBuilder sb = new StringBuilder((int) Math.max(10, s.length() * 1.5));
```
vladak: could you explain the formula in the comment perhaps ?

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -433,48 +433,67 @@ public static String readableSize(long num) {
...
+            // StringBuilder. Wrap in an AssertionError so that callers
+            // don't have to check for an IOException that should never
+            // happen.
+            throw new AssertionError("StringBuilder threw IOException", ex);
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Converts different html special characters into their encodings used in
```
vladak: HTML

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -433,48 +433,67 @@ public static String readableSize(long num) {
...
+    /**
+     * Encode URL
+     *
+     * @param url string URL
+     * @return the encoded URL
+     * @throws URISyntaxException
+     * @throws MalformedURLException
+     */
+    public static String encodeURL(String url) throws URISyntaxException, MalformedURLException {
+        URL old = new URL(url);
```
vladak: the variable naming is a bit confusing here. What about url -> urlstr, old -> url ?

test/org/opensolaris/opengrok/web/UtilTest.java
```diff
@@ -256,6 +258,19 @@ public void diffline() {
     }
 
     @Test
+    public void testEncode() {
+        String[][] tests = new String[][]{
```
vladak: make sure all the if/else if cases in encode() are covered, e.g. `'\t'` is missing.

test/org/opensolaris/opengrok/web/UtilTest.java
```diff
@@ -343,12 +358,19 @@ public void testLinkify() {
         assertTrue(Util.linkify("smtp://example.com/OpenGrok/OpenGrok").equals("smtp://example.com/OpenGrok/OpenGrok"));
         assertTrue(Util.linkify("just some crazy text").equals("just some crazy text"));
 
-        // escaping tests
+        // escaping url
+        assertTrue(Util.linkify("http://www.example.com/\"quotation\"/else")
+                .contains("href=\"" + Util.encodeURL("http://www.example.com/\"quotation\"/else") + "\""));
```
vladak: use something more precise than contains() ? e.g. compareTo()/contentEquals() ?

test/org/opensolaris/opengrok/web/UtilTest.java
```diff
@@ -343,12 +358,19 @@ public void testLinkify() {
...
+        assertTrue(Util.linkify("https://example.com/><\"")
+                .contains("href=\"" + Util.encodeURL("https://example.com/><\"") + "\""));
+        assertTrue(Util.linkify("http://www.example.com?param=1&param2=2&param3=\"quoted>\"")
+                .contains("href=\"" + Util.encodeURL("http://www.example.com?param=1&param2=2&param3=\"quoted>\"") + "\""));
+        // escaping titles
         assertTrue(Util.linkify("http://www.example.com/\"quotation\"/else")
-                .contains("href=\"http://www.example.com/%22quotation%22/else\""));
+                .matches(".*title=\".*?" + Util.encode("http://www.example.com/\"quotation\"/else") + ".*\".*"));
```
vladak: ditto, exact comparison ?

tulinkry: I added `contains` as it is with the `href` attribute with the exact content "Link to 'url'". I don't think it's the best solution to have it like the whole content compassion as explained above.

test/org/opensolaris/opengrok/web/UtilTest.java
tulinkry: These are tests only for the url in href. If I was to do the tests for the whole content the test would look quite complicated (because the encode has to be called three times).

Also this doesn't prevent from future updates of the `linkify` method to contain some additional parameters.

test/org/opensolaris/opengrok/web/UtilTest.java
tulinkry: updated

src/org/opensolaris/opengrok/web/Util.java
tulinkry: updated

src/org/opensolaris/opengrok/web/Util.java
tulinkry: changed

src/org/opensolaris/opengrok/web/Util.java
tulinkry: added a comment

PR #1430, opened 2017-02-17
removing unused and non-existent import

tulinkry:
Just to accept this faster to have a non-failing build of webapp:

removing non existent import

```java
-<%@page import="org.jcp.xml.dsig.internal.dom.Utils"%>
```

I have no idea how that came there.


tarzanek:
@tulinkry Thank you sir! :-D

PR #1431, opened 2017-02-18
Webapp won't start if config dir / file doesn't exist

ozankiraz:


src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -786,7 +786,9 @@ public void setCurrentIndexedCollapseThreshold(int currentIndexedCollapseThresho
     public Date getDateForLastIndexRun() {
         if (lastModified == null) {
             File timestamp = new File(getDataRoot(), "timestamp");
-            lastModified = new Date(timestamp.lastModified());
+            if(timestamp.exists()){
```
ozankiraz: Without file existence check, date will be set as Jan 01 1970, while it should be null, this causes bogus behaviour

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -65,15 +65,17 @@
...
+               
         if (config == null) {
             LOGGER.severe("CONFIGURATION section missing in web.xml");
         } else {
             try {
-                env.readConfiguration(new File(config));
+                env.readConfiguration(configFile);
             } catch (IOException ex) {
-                LOGGER.log(Level.WARNING, "OpenGrok Configuration error. Failed to read config file: ", ex);
+                LOGGER.log(Level.WARNING, "OpenGrok Configuration error. Failed to read config file: {0}", ex.toString());
```
ozankiraz: Log levels are Warning for missing config, so exception is not thrown, but message is written to logger

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -99,9 +101,12 @@ public void contextInitialized(final ServletContextEvent servletContextEvent) {
         }
 
         try {
-            RuntimeEnvironment.getInstance().loadStatistics();
+            RuntimeEnvironment runtimeEnvironment = RuntimeEnvironment.getInstance();
+            if (runtimeEnvironment.getConfiguration().getDataRoot() != null) {
```
ozankiraz: null value check could be done in loadStatistics() method, but breaks StatsMessageTest.java => testInvalidReload, since this test expects exception to be thrown

web/index.jsp
```diff
@@ -29,6 +29,10 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
+    String sourceRootPath = cfg.getSourceRootPath();
+    if(sourceRootPath == null || sourceRootPath.isEmpty()){
+        throw new java.io.FileNotFoundException("Configuration File Not Found");
```
ozankiraz: Throw exception here, so error handler can catch it

vladak:
Looks quite reasonably.

On the delivery side, you need to sync up with OoenGrok:master branch and ideally have just a single commit in the PR.

ozankiraz:
Hi @vladak ,
I'll try to get it sorted, I'm not experienced with Github, so didn't realise it.

DaveRandom:
@ozankiraz hints for what you need to do:

- Add another remote to your local clone: `git remote add upstream https://github.com/OpenGrok/OpenGrok` (this will also work with `git@github.com:OpenGrok/OpenGrok.git` if you prefer to work over SSH)
- Sync your local changes with the upstream remote: `git pull --rebase upstream master`. After doing this you may need to resolve conflicts, git will instruct you on what it expects you to do. By the looks of if you shouldn't currently have any issues with this.
- "Squash" your commits into a single commit: `git rebase -i HEAD~14`. This will start an interactive rebase, it will launch a text editor to enable you to complete it, more info [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html).
- Once you have done this you will need to perform a forced push (`git push -f`) to the `master` branch of your `origin` remote to update this pull request.

I HIGHLY RECOMMEND that you practice this by creating a new local branch based on your current master (`git checkout -b practice-branch`) since this can potentially irrevocably destroy your history if you get it wrong. DO NOT perform any `git push -f` commands until you are 100% certain it will do the right thing.

Good luck

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: I don't understand this? This suppress the stack trace which might be helpful

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: Do this in load statistics and fix the test where it should test if the file actually exists (provide some nonsense path).

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -99,9 +101,12 @@ public void contextInitialized(final ServletContextEvent servletContextEvent) {
...
+                runtimeEnvironment.loadStatistics();
+            }
         } catch (IOException ex) {
-            LOGGER.log(Level.INFO, "Could not load statistics from a file.", ex);
+            LOGGER.log(Level.INFO, "Could not load statistics from a file.", ex.toString());
```
tulinkry: I don't understand this? This suppress the stack trace which might be helpful

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -65,15 +65,17 @@
     public void contextInitialized(final ServletContextEvent servletContextEvent) {
         ServletContext context = servletContextEvent.getServletContext();
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
-
+
```
tulinkry: extra whitespace

web/foot.jspf
```diff
@@ -35,12 +36,14 @@ org.opensolaris.opengrok.web.Prefix"
...
     PageConfig cfg = PageConfig.get(request);
-
+    Date dateForLastIndexRun = cfg.getEnv().getDateForLastIndexRun();
         %>
     <div id="footer">
 <p><a href="http://opengrok.github.com/OpenGrok/"
  title="Served by OpenGrok"><span id="fti"></span></a></p>
-<p>Indexes created <%= cfg.getEnv().getDateForLastIndexRun() %></p>
+    <% if(dateForLastIndexRun != null) { %>
+    <p>Indexes created <%= cfg.getEnv().getDateForLastIndexRun() %></p>
```
tulinkry: You retrieve the variable twice. Use the value you retrieved earlier.

web/index.jsp
tulinkry: What if the file is not readable?

ozankiraz:
I'll send new one

PR #1433, opened 2017-02-20
Msg getconf

vladak:


src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -33,16 +33,36 @@
...
 
     @Override
     public void validate() throws Exception {
-        if (getText() == null) {
-            throw new Exception("The message must contain a text.");
+        if (hasTag("setconf")) {
+            if (getText() == null) {
+                throw new Exception("The setconf message must contain a text.");
+            }
+        } else if (hasTag("getconf")) {
```
tulinkry: How about making this default? (if there is no tag at all append a "getconf" tag like [here](https://github.com/tulinkry/OpenGrok/blob/43d8e39cdcb82e4d3241ffb8be3af6a3adf111d3/src/org/opensolaris/opengrok/configuration/messages/NormalMessage.java#L46-L48))
So that `./Messages -n config` automaticaly gets the config without needing the `getconf` parameter.

vladak: Somehow I am not thrilled by the idea of the app second guessing users' intentions.

tulinkry: Ok. Let's merge

Issue #1357, opened 2017-01-30
short URL for all projects search wanted


ilovemaui:
so i recently enabled projects in my opengrok instance. but do a search, you need to specify all of them (the url). so perhaps we can add an enhancement/keyword that will let you search all projects.. for example..
i have proj1 proj2 proj3 proj4 proj5
so I have to do a search like this:
https://myopengrok/source/search?q=foobar&defs=&refs=&path=&hist=&type=&project=proj1&project=proj2&project=proj3
&project=dpkg&project=proj4&project=proj5

if would be nice if we could have a single keyword that would include all of them. thanks.
fyi, this should also be added to command line tool.

tulinkry:
In the web interface you can hit the "select all" button above the project picker. The selection is stored in a cookie afterwards and then it's selected automatically. It kinda solves the problem in the web interface however there is no special keyword.

vladak:
So this is asking for something like this:
```
https://myopengrok/source/search?q=foobar&defs=&refs=&path=&hist=&type=&project=ALL
```

vladak:
Related to #1048 in a sense that too many projects make the URL long enough to hit limits.

ilovemaui:
perhaps something like that.. before, I didn't use projects at all, and we just searched everything anyhow. I am currently hardcoding the list, but this enhancement would make that unnecessary.

tulinkry:
@vladak I'd suggest a different parameter than project as the project name can be arbitrary. Something like this:
```plain
https://myopengrok/source/search?q=foobar&defs=&refs=&path=&hist=&type=&search_all=true
```

ilovemaui:
additionally, and i don't know if it related. but when you open a browser for the 1st time, there are no projects selects. While it does remember once you put in the projects you want. is there any way to default to include all projects the first time a browser is open?

tulinkry:
No there is not

ilovemaui:
ok. thanks. I wasn't sure if this could be a useful enhancement

tulinkry:
> ok. thanks. I wasn't sure if this could be a useful enhancement

By the way this was already suggested a long time ago (#28) and it could be an useful enhancement

adamscybot:
This is needed. I have 100's of projects and so when I want to share the link with another developer its prohibitively long.

tulinkry:
Also kinda duplicate of #563

dhireng:
I updated the code to include a search all projects button. The question is how do i submit a push?

[...]

Issue #28, opened 2013-04-01
Cannot select multiple default projects (only one path accepted)


cnst:
It doesn't seem like it's possible for the administrator to pre-select multiple default projects, for the situation where no project is provided by the user-agent through the cookies.

The `-p` option seems to require a path, e.g. `-p "/OpenBSD"`, and it would seem that providing multiple comma-separated values does not result in any desired effects.

This is what the documentation says:

> -p /path/to/default/project
>        This is the path to the project that should be selected by default in the web application(when no other project set either in cookie or in parameter). You should strip off the source root.

PR #1437, opened 2017-02-20
using multiple default projects

tulinkry:
Implements enhancement proposed in #1357 and also back in old days in #28

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Configuration.java
+++ b/src/org/opensolaris/opengrok/configuration/Configuration.java
@@ -92,11 +92,11 @@ public final class Configuration {
     private String urlPrefix;
     private boolean generateHtml;
     /**
-     * Default project will be used, when no project is selected and no project
+     * Default projects will be used, when no project is selected and no project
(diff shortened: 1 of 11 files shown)


vladak:
pls resolve the conflicts. I will take a look afterwards.

tulinkry:
Resolved

vladak:
Couple of notes:
  - I wonder if this should be reflected in the `OpenGrok` shell script ?
  - pls also change the usage CommandLineOptions.java
  - some test+documentation for the `__all__` special value ?

tulinkry:
> pls also change the usage CommandLineOptions.java

changed

> some test+documentation for the __all__ special value ?

added a test case + documented the `__all__` in the usage

tulinkry:
> I wonder if this should be reflected in the OpenGrok shell script ?

I don't see that the wrapper support default project. This could be part of another enhancements in different PR.

vladak:
okay then. pls file a new enhancement for the `OpenGrok` shell script.

PR #1438, opened 2017-02-21
adding query string into revision url

### LINKED ISSUE #1427: Revision links are not working
screenshot from 2017-02-17 17-41-14

These links don't work often because the history is now paginated

tulinkry:
fixes #1427

### DIFF
--- a/web/history.jsp
+++ b/web/history.jsp
@@ -246,7 +246,7 @@ document.domReady.push(function() {domReadyHistory();});
                     if (entry.isActive()) {
                         String rp = uriEncodedName;
             %>
-            <td><a href="<%= context + Prefix.HIST_L + rp %>#<%= rev %>"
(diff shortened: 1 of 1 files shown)


web/history.jsp
```diff
@@ -246,7 +246,7 @@ document.domReady.push(function() {domReadyHistory();});
                     if (entry.isActive()) {
                         String rp = uriEncodedName;
             %>
-            <td><a href="<%= context + Prefix.HIST_L + rp %>#<%= rev %>"
+            <td><a href="<%= request.getRequestURL().append('?').append(request.getQueryString()) %>#<%= rev %>"
```
vladak: The `#` and `rev` can be appended as well. That said, the cascade of appends is a bit hard to read.

tulinkry: Changed that. Also there was a problem when the query string was `null`.

PR #1440, opened 2017-02-21
Web app start fix when config file is missing

ozankiraz:


ozankiraz:
Same fix from earlier, as single commit, I should work on my git skills :)

no_config_error_fix

@DaveRandom Thanks a lot for detailed explanation, I tried to apply steps, but I failed :( We're using Accurev as SCM, so this is something new for me, I need some time and practice. I decided to go with fresh start as noob person.
Actually the reason I started working on Opengrok repository was to send fixes for accurev repository.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -371,12 +371,20 @@ public void setSourceRoot(String sourceRoot) {
...
         String sourceRoot = getSourceRootPath();
+        
+        if(sourceRoot == null){
+            throw new FileNotFoundException("Source Root Not Found");
+        }
+        
         if (canonicalPath.startsWith(sourceRoot)) {
             return canonicalPath.substring(sourceRoot.length() + stripCount);
         }
+
```
ozankiraz: Lets keep these spaces here :) I'll be more careful next time I promise

web/index.jsp
```diff
@@ -29,6 +29,10 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
+    String sourceRootPath = cfg.getSourceRootPath();
```
ozankiraz: I thought checking sourceRootPath value should be enough, since there is only one way to make 100% sure to check if file is readable is to read it in try catch block ( http://stackoverflow.com/questions/11901382/how-to-check-if-a-file-is-readable) which seems like overkill

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -786,7 +786,9 @@ public void setCurrentIndexedCollapseThreshold(int currentIndexedCollapseThresho
     public Date getDateForLastIndexRun() {
         if (lastModified == null) {
             File timestamp = new File(getDataRoot(), "timestamp");
-            lastModified = new Date(timestamp.lastModified());
+            if(timestamp.exists()){
```
ozankiraz: to fix bogus behavior of Index Created Date info displayed as Jan 01 1970

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1135,7 +1143,7 @@ public int getGroupsCollapseThreshold() {
      * @param file the file to read
      * @throws IOException if an error occurs
```
ozankiraz: Didn't make changes in the comment, as FileNotFoundException is IOException
https://docs.oracle.com/javase/7/docs/api/java/io/FileNotFoundException.html?is-external=true

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -371,12 +371,20 @@ public void setSourceRoot(String sourceRoot) {
      * @throws FileNotFoundException If the file is not relative to source root
      * @return Path relative to source root
      */
-    public String getPathRelativeToSourceRoot(File file, int stripCount) throws IOException {
+    public String getPathRelativeToSourceRoot(File file, int stripCount) throws IOException, FileNotFoundException {
```
tulinkry: FileNotFoundException is a subclass of IOException

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1135,7 +1143,7 @@ public int getGroupsCollapseThreshold() {
      * @param file the file to read
      * @throws IOException if an error occurs
      */
-    public void readConfiguration(File file) throws IOException {
+    public void readConfiguration(File file) throws IOException, FileNotFoundException {
```
tulinkry: FileNotFoundException is a subclass of IOException

web/foot.jspf
```diff
@@ -35,12 +36,15 @@ org.opensolaris.opengrok.web.Prefix"
...
     PageConfig cfg = PageConfig.get(request);
-
+    Date dateForLastIndexRun = cfg.getEnv().getDateForLastIndexRun();
+    
         %>
     <div id="footer">
 <p><a href="http://opengrok.github.com/OpenGrok/"
  title="Served by OpenGrok"><span id="fti"></span></a></p>
+<% if(dateForLastIndexRun != null) { %>
 <p>Indexes created <%= cfg.getEnv().getDateForLastIndexRun() %></p>
```
tulinkry: You retrieve the variable twice. Use the value you've retrieved earlier.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1504,7 +1512,13 @@ public void saveStatistics(OutputStream out) throws IOException {
...
      */
     public void loadStatistics() throws IOException, ParseException {
-        loadStatistics(new File(getConfiguration().getStatisticsFilePath()));
+        String statisticsFilePath = getConfiguration().getStatisticsFilePath();
+        if(statisticsFilePath != null && !statisticsFilePath.isEmpty()){
+            File statisticsFile = new File(statisticsFilePath);
+            if(statisticsFile != null && statisticsFile.exists()){
+                loadStatistics(statisticsFile);
+            }
+        }
```
tulinkry: I don't see the gain in this - if there is a problem with the file it raises an exception as expected - this is silently ignoring the problem and not logging anything. Change the test as I suggested instead of this check.

tulinkry: Or throw an exception if the `getStatisticsFilePath()` is `null` that's another option how to convert the NPE into IOException and logs the problem in the higher layer. And you wouldn'ť need to do anything to the original test I guess in this case.

test/org/opensolaris/opengrok/configuration/messages/StatsMessageTest.java
```diff
@@ -159,8 +159,8 @@ public void testInvalidReload() {
```
tulinkry: Actually I meant that you provide a non-existing file, like "/no/file/exists/g65g65d4g65v16516" and then the test will raise an exception. [see this comment](https://github.com/OpenGrok/OpenGrok/pull/1440/files/45776d90dfb381e32043317d9a853d06b3cac0b4#r102430071) which might solve the situation while not changing anything in the test

web/index.jsp
tulinkry: Yes this is an overkill. I'd patch it with the `canRead` method even though it's 100% valid on all systems. Also this question is for java6 and we have java8 now.

web/index.jsp
```diff
@@ -30,9 +30,13 @@ include file="projects.jspf"
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
     String sourceRootPath = cfg.getSourceRootPath();
-    if(sourceRootPath == null || sourceRootPath.isEmpty()){
+    if(sourceRootPath == null && sourceRootPath.isEmpty()){
         throw new java.io.FileNotFoundException("Configuration File Not Found");
     }
+    File sourceRootPathFile = new File(sourceRootPath);
```
tulinkry: this exists as a method of `cfg`. Something like `cfg.getSourceRootFile()`

web/index.jsp
```diff
@@ -30,9 +30,13 @@ include file="projects.jspf"
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
     String sourceRootPath = cfg.getSourceRootPath();
-    if(sourceRootPath == null || sourceRootPath.isEmpty()){
+    if(sourceRootPath == null && sourceRootPath.isEmpty()){
```
tulinkry: Eh? This leads to NPE if the path is null. I'm not sure why you changed that.

web/index.jsp
ozankiraz: @tulinkry , 
replaced with : File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();

web/index.jsp
ozankiraz: Changed back

ozankiraz:
@tulinkry Commits squashed into one and changes you mentioned made.

PR #1450, opened 2017-02-27
log exit value from ctags

### LINKED ISSUE #1449: report ctags exit value in doCtags()
doCtags() handles exit of ctags process like this:

```
241          if (ctags != null) {
242              try {
243                  ctags.exitValue();
244                  ctagsRunning = false;
245                  // ctags is dead! we must restart!!!
246              } catch (IllegalThreadStateException exp) {
247                  ctagsRunning = true;
248                  // ctags is still running :)
249              }
250          }
```

It would be useful to know the exit value on line 243 (even though ctags is effectively killed in close() by calling `ctags.destroy()` so that should probably be preceded with "Killing ctags" message or such so that it is possible to distinguish between abnormal exit and cleanup).

vladak:
fixes #1449

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/Ctags.java
+++ b/src/org/opensolaris/opengrok/analysis/Ctags.java
@@ -18,7 +18,7 @@
  */
 
  /*
- * Copyright (c) 2005, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
  */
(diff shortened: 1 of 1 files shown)


tarzanek:
thank you

vladak:
Tested by `kill` of `ctags` process:

```
$ grep -i exited /var/opengrok/log/opengrok.log.0
14:47:24 WARNING: Ctags process exited with exit value 15
$ kill -l TERM
15
```

Issue #1413, opened 2017-02-14
0.13 webapp won't start if config dir / file doesn't exist - regression


tarzanek:
0.12 used to start even without config.xml readable,
0.13 however does:
Feb-2017 17:07:51.941 WARNING [localhost-startStop-1] org.opensolaris.opengrok.web.WebappListener.contextInitialized OpenGrok Configuration error. Failed to read config file:
 java.io.FileNotFoundException: /var/opengrok/etc/configuration.xml (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:195)
        at java.io.FileInputStream.<init>(FileInputStream.java:138)
        at org.opensolaris.opengrok.configuration.Configuration.read(Configuration.java:788)
        at org.opensolaris.opengrok.configuration.RuntimeEnvironment.readConfiguration(RuntimeEnvironment.java:905)
        at org.opensolaris.opengrok.web.WebappListener.contextInitialized(WebappListener.java:58)
        at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4727)
        at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5189)
        at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
        at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:752)
        at org.apache.catalina.core.ContainerBase.addChild(ContainerBase.java:728)
        at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:734)
        at org.apache.catalina.startup.HostConfig.deployWAR(HostConfig.java:952)
        at org.apache.catalina.startup.HostConfig$DeployWar.run(HostConfig.java:1823)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)

and the webapp fails to start (repro on mac os x with howto from https://blog.simplypatrick.com/posts/2015/01-16-opengrok-on-mac-osx/ )
@nerakhon reported this ...

tarzanek:
[...]

PR #1453, opened 2017-02-28
Web app start fix when config file is missing #1413

ozankiraz:
Created New PR since I messed up previous one

### DIFF
--- a/src/org/opensolaris/opengrok/web/PageConfig.java
+++ b/src/org/opensolaris/opengrok/web/PageConfig.java
@@ -1402,4 +1402,20 @@ public final class PageConfig {
 
         return Util.htmlize(title);
     }
+    
+    public void checkSourceRootExistence() throws IOException {        
+        if (sourceRootPath == null || sourceRootPath.isEmpty()) {
+            throw new java.io.FileNotFoundException("Configuration File Not Found, Can not determine source root path");
+        }
+        
+        File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();
+        if (sourceRootPathFile.exists() && !sourceRootPathFile.canRead()) {
+            throw new java.io.IOException("Can not read configuration file");
+        }
+        
+        if(!sourceRootPathFile.exists()){
+            throw new java.io.FileNotFoundException("Source root path does not exist : " + sourceRootPathFile.getAbsolutePath());
+        }
+    }
+    
 }
--- a/web/index.jsp
+++ b/web/index.jsp
@@ -29,14 +29,7 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
-    String sourceRootPath = cfg.getSourceRootPath();
-    if(sourceRootPath == null || sourceRootPath.isEmpty()){
-    throw new java.io.FileNotFoundException("Configuration File Not Found");
-    }
-    File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();
-    if(sourceRootPathFile.exists() && !sourceRootPathFile.canRead()){
-    throw new java.io.IOException("Can not read configuration file");
-    }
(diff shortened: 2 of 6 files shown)


web/index.jsp
```diff
@@ -29,6 +29,14 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     cfg.setTitle("Search");
+    String sourceRootPath = cfg.getSourceRootPath();
+    if(sourceRootPath == null || sourceRootPath.isEmpty()){
+    throw new java.io.FileNotFoundException("Configuration File Not Found");
+    }
+    File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();
+    if(sourceRootPathFile.exists() && !sourceRootPathFile.canRead()){
```
tulinkry: This is not correct
```java
if(sourceRootPathFile.exists() && !sourceRootPathFile.canRead()){
```

Perhaps should have been:
```java
if(!sourceRootPathFile.exists() || !sourceRootPathFile.canRead()){
```

ozankiraz: @tulinkry I intended to do this actually, There are two exceptions, first one is FileNotFoundException, thrown when source path is null or empty, covers the scenario when source root path is not set,
Second exception is IOException, covers the scenario when file is there but cannot read

tulinkry: But when the file does not actually exist then there is no exception which is not correct, is it?

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1402,4 +1402,20 @@ public String getPathTitle() {
 
         return Util.htmlize(title);
     }
+    
+    public void checkSourceRootExistence() throws IOException {
```
ozankiraz: Moved logic here

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1402,4 +1402,20 @@ public String getPathTitle() {
...
+        if (sourceRootPath == null || sourceRootPath.isEmpty()) {
+            throw new java.io.FileNotFoundException("Configuration File Not Found, Can not determine source root path");
+        }
+        
+        File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();
+        if (sourceRootPathFile.exists() && !sourceRootPathFile.canRead()) {
+            throw new java.io.IOException("Can not read configuration file");
+        }
+        
+        if(!sourceRootPathFile.exists()){
```
ozankiraz: Added this check as required

web/index.jsp
```diff
@@ -29,14 +29,7 @@ include file="projects.jspf"
...
     cfg.setTitle("Search");
-    String sourceRootPath = cfg.getSourceRootPath();
-    if(sourceRootPath == null || sourceRootPath.isEmpty()){
-    throw new java.io.FileNotFoundException("Configuration File Not Found");
-    }
-    File sourceRootPathFile = RuntimeEnvironment.getInstance().getSourceRootFile();
-    if(sourceRootPathFile.exists() && !sourceRootPathFile.canRead()){
-    throw new java.io.IOException("Can not read configuration file");
-    }
+    cfg.checkSourceRootExistence();
```
ozankiraz: Shorter and more readable

ozankiraz:
@tulinkry Yes, you're right, I was confused about SRC_ROOT in web.xml and sourceRoot in configuration.xml
Moved logic to PageConfig class and updated jsp file. Thanks for pointing it out.

tulinkry:
Ok to me now

vladak:
@ozankiraz any news w.r.t. your OCA ?

ozankiraz:
Hi @vladak 
I got an email about OCA is processed and I can see my name in OCA list, but in email they didn't sent me OCA number.

vladak:
Okay, I see your name there, merging.

PR #1457, opened 2017-03-07
Authorization framework enhancements

tulinkry:
Adding a little bit of control to how the authorization is performed in opengrok.

Introducing three new plugin roles:
 - required
 - requisite
 - sufficient

I'll include more detailed description of these in the wiki page once this is merged. These roles should copy the characeristics of the same roles in pam list.

Also the order of plugin execution is determined by their order in the configuration and these three keywords control how to deal with the returned value of the authorization check.

**NOTE: This change is backward compatible with the previous version of auth framework as it adds all discovered plugins as `required`.** The system still remains plug-and-play when using just one plugin as the only action is to put it in the plugin directory.

The configuration entry (in read-only configuration) can look like this:
```xml
<void property="pluginConfiguration">
	<void method="add">
		<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
			<void property="role">
				<string>requisite</string>
			</void>
			<void property="classname">
				<string>tulinkry.my.RequisitePlugin</string>
			</void>
		</object>
	</void>	
	<void method="add">
		<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
			<void property="role">
				<string>sufficient</string>
			</void>
			<void property="classname">
				<string>Plugin</string>
			</void>
		</object>
	</void>
	<void method="add">
		<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
			<void property="role">
				<string>required</string>
			</void>
			<void property="classname">
				<string>ExamplePlugin</string>
			</void>
		</object>
	</void>		
</void>
```

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
@@ -1,0 +1,155 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+ /*
+ * Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opensolaris.opengrok.authorization;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ *
+ * @author Krystof Tulinger
+ */
+public class AuthorizationCheck implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Enum for avaliable authorization roles.
+     */
+    public enum AuthorizationRole {
+        /**
+         * Failure of such a plugin will ultimately lead to the authorization
+         * framework returning failure but only after the remaining plugins have
+         * been invoked.
+         *
+         */
+        REQUIRED("required"),
+        /**
+         * Like required, however, in the case that such a plugin returns a
+         * failure, control is directly returned to the application. The return
+         * value is that associated with the first required or requisite plugin
+         * to fail.
+         *
+         */
+        REQUISITE("requisite"),
+        /**
+         * If such a plugin succeeds and no prior required plugin has failed the
+         * authorization framework returns success to the application
+         * immediately without calling any further plugins in the stack. A
+         * failure of a sufficient plugin is ignored and processing of the
+         * plugin list continues unaffected.
+         */
+        SUFFICIENT("sufficient");
+
+        private final String role;
+
+        private AuthorizationRole(String role) {
+            this.role = role;
+        }
+
+        @Override
+        public String toString() {
+            return this.role;
+        }
+
+        public static AuthorizationRole get(String role) {
+            try {
+                return AuthorizationRole.valueOf(role.toUpperCase());
+            } catch (IllegalArgumentException ex) {
+                // role does not exist -> add some more info about which roles do exist
+                throw new IllegalArgumentException(
+                        String.format("No authorization role \"%s\", available roles are [%s]. %s",
+                                role,
+                                Arrays.asList(AuthorizationRole.values())
+                                        .stream()
+                                        .map(AuthorizationRole::toString)
+                                        .collect(Collectors.joining(", ")),
+                                ex.getLocalizedMessage()), ex);
+            }
+        }
+    };
+
+    /**
+     * One of "required", "requisite", "sufficient".
+     */
+    private AuthorizationRole role;
+
+    /**
+     * Canonical name of a java class.
+     */
+    private String classname;
+
+    public AuthorizationCheck() {
+    }
+
+    public AuthorizationCheck(AuthorizationRole role, String classname) {
+        this.role = role;
+        this.classname = classname;
+    }
+
+    /**
+     * Get the value of classname
+     *
+     * @return the value of classname
+     */
+    public String getClassname() {
+        return classname;
+    }
+
+    /**
+     * Set the value of classname
+     *
+     * @param classname new value of classname
+     */
+    public void setClassname(String classname) {
+        this.classname = classname;
+    }
+
+    /**
+     * Get the value of role
+     *
+     * @return the value of role
+     */
+    public AuthorizationRole getRole() {
+        return role;
+    }
+
+    /**
+     * Set the value of role
+     *
+     * @param role new value of role
+     */
+    public void setRole(AuthorizationRole role) {
+        this.role = role;
+    }
+
+    /**
+     * Set the value of role
+     *
+     * @param role new value of role
+     */
+    public void setRole(String role) {
+        this.role = AuthorizationRole.get(role);
+    }
+}
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
@@ -37,7 +37,7 @@ public class AuthorizationCheck implements Serializable {
     /**
      * Enum for avaliable authorization roles.
      */
-    public enum AuthorizationRole {
+    public enum AuthControlFlag {
         /**
          * Failure of such a plugin will ultimately lead to the authorization
          * framework returning failure but only after the remaining plugins have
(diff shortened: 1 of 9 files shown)


tulinkry:
Another points for discussion:
 - if there is a configuration entry for some plugin and the class is not found - what should happen? should it be considered as failed plugin or completely removed. Now it is removed.

vladak:
- I think the whole app should bail if any of the authorization plugins fail to load/initialize. This might seem too strict however in the presence of authorization plugins it is better not to allow anyone than do wrong decision.
  - +1 for moving the config
  - +1 for less singletons

tulinkry:
This first point is solved in the manner that if the plugin fails when loading then it's marked as failed (all plugin's operation are mocked as failed); therefore:
1. failed required plugin always fails the authorization
2. failed sufficient plugin never fails the authorization (as per design of the keyword sufficent)
3. failed requisite plugin always fails the authorization

The plugins, for which there is no configuration entry (sufficient, required, requisite), i. e. plugins which are just dropped into the plugin directory without any configuration - these are loaded as required and then the previous strategy takes place.

For the plugins which fail when loading their class this is not solved yet (removed from the list). I need to do a change for this.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1577,6 +1586,8 @@ public void applyConfig(Message m, boolean reindex) {
             // Force timestamp to update itself upon new config arrival.
             config.refreshDateForLastIndexRun();
         }
+
+        AuthorizationFramework.getInstance().reload();
```
tulinkry: I'll merge this in a proper manner with the #1465 once it's merged

tulinkry:
I implemented the proposed changes when the plugin fails when loading/testing.

If one want to try it. There is minimal configuration using three plugins: [plugins.zip](https://github.com/OpenGrok/OpenGrok/files/839377/plugins.zip). **You may want to change the authorization policy inside the java classes for your project/group setup.** I used this command for compiling, but you can use any other as long as there is `opengrok.jar` in the classpath.
```bash
alias plugin='/usr/local/java/latest/bin/javac -classpath `find ~/workspace/OpenGrok/lib -name "*.jar" -printf "%h/%f:"`~/workspace/OpenGrok/dist/opengrok.jar:. -d .'
cd /var/opengrok/plugins
# extract here the zip file so perhaps
unzip plugins.zip
plugin SamplePlugin
plugin Plugin
plugin tulinkry/my/RequisitePlugin
```


And an entry for the read only configuration:
```java
	<void property="pluginConfiguration">
		<void method="add">
			<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
				<void property="role">
					<string>requisite</string>
				</void>
				<void property="classname">
					<string>tulinkry.my.RequisitePlugin</string>
				</void>
			</object>
		</void>	
		<void method="add">
			<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
				<void property="role">
					<string>sufficient</string>
				</void>
				<void property="classname">
					<string>Plugin</string>
				</void>
			</object>
		</void>
		<void method="add">
			<object class="org.opensolaris.opengrok.authorization.AuthorizationCheck">
				<void property="role">
					<string>required</string>
				</void>
				<void property="classname">
					<string>SamplePlugin</string>
				</void>
			</object>
		</void>		
	</void>
```

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -275,107 +422,164 @@ private String getClassName(JarEntry f) {
...
-                LOGGER.log(Level.SEVERE, "Plugin \"" + plugin.getClass().getName() + "\" has failed with exception:", ex);
-                removePlugin(plugin);
+                LOGGER.log(Level.WARNING,
+                        String.format("Plugin \"%s\" has failed the testing of \"%s\" with an exception.",
+                                plugin.getClassname(),
+                                entity.getName()),
+                        ex);
+
+                // set the return value to false for this faulty plugin
+                if (!plugin.isSufficient()) {
```
vladak: if SUFFICIENT plugin throws exception, it should return failure too I think. In overall I wonder if any exception here should cause immediate termination of traversal and return false.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -275,107 +422,164 @@ private String getClassName(JarEntry f) {
...
+    private boolean performCheck(Nameable entity, Predicate<AuthorizationPluginWrapper> predicate) {
+        boolean overallDecision = true;
+        for (AuthorizationPluginWrapper plugin : getPlugins()) {
+            // run the plugin's test method
             try {
-                if (!predicate.test(plugin)) {
-                    m.put(name, false);
-                    request.setAttribute(cache, m);
-                    return false;
+                boolean pluginDecision = predicate.test(plugin);
```
vladak: it would be good to have some debug info (`FINE` log level) that would print some basic info about the plugin before the call to `test()` and the result after + the overall result - similarly to what PAM debug does.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: I don't think so, as I read in the man:

>  A failure of a sufficient module is ignored and processing of the PAM module stack continues unaffected.

Sufficient plugin should only decide if the stack can be left immediately if it returns true. In all other cases (false, failure) it should not affect the stack processing

For the other types it could be that the stack will be left immediately when the plugin has a failure (or even returns false). The only argument to leave as it is is a timing attack but I don't think that it's that important here.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: I added a log output. I haven't done it before because I was little worried about the performance - on the main page this causes `plugins * (projects + groups)` number of outputs which may be a slowdown.

tulinkry: > That way we would also get the timings (it would also be good to have them available in the Statistics framework).

I'd add this in different PR so we dont complicate things in one PR.

tulinkry: changed the level to `FINEST`

src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
```diff
@@ -0,0 +1,155 @@
...
+ * @author Krystof Tulinger
+ */
+public class AuthorizationCheck implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Enum for avaliable authorization roles.
+     */
+    public enum AuthorizationRole {
```
vladak: `AuthControlFlag` would be better name to align with PAM terminology

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -118,102 +152,200 @@ private AuthorizationFramework() {
...
-     * @return list of file with suffix
+     * Load all plugins in the current array. If the plugin has not been loaded
+     * yet it is marked as failed.
      */
-    private List<File> listFiles(String suffix) {
-        File[] files = directory.listFiles(new FilenameFilter() {
-            @Override
-            public boolean accept(File dir, String name) {
-                return name.endsWith(suffix);
+    public void loadAllPlugins() {
```
vladak: it would be good to log the list of plugins (including their order) at the end of the (re)loading so that we know how the stack looks like.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -236,22 +368,51 @@ private void loadClass(String classname) throws ClassNotFoundException,
...
                 loaded = true;
             }
         }
         if (!loaded) {
-            LOGGER.log(Level.INFO, "Plugin class \"{0}\" does not implement IAuthorizationPlugin interface.", classname);
+            LOGGER.log(Level.FINEST, "Plugin class \"{0}\" does not implement IAuthorizationPlugin interface.", classname);
+        }
+    }
+
+    private void loadClasses(List<File> classfiles, List<File> jarfiles) {
```
vladak: add a comment about what this method does exactly

src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: added a comment

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: added a log call when traversing the plugins

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
vladak: I guess it makes sense to treat exceptions and errors as one condition - after all exception can be used to signal error condition. Good point about the timing attack.

vladak:
pls resolve conflicts and merge

PR #1465, opened 2017-03-10
Authorization framework configuration in configuration.xml

tulinkry:
... instead of web.xml

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -53,7 +53,7 @@ import org.opensolaris.opengrok.logger.LoggerFactory;
 public final class AuthorizationFramework {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizationFramework.class);
-    private File directory;
+    private File pluginDirectory;
     private AuthorizationPluginClassLoader loader;
 
     private volatile static AuthorizationFramework instance = new AuthorizationFramework();
@@ -73,17 +73,17 @@ public final class AuthorizationFramework {
     /**
      * Get the plugin directory.
      */
-    public synchronized File getDirectory() {
-        return directory;
+    public synchronized File getPluginDirectory() {
+        return pluginDirectory;
     }
 
     /**
      * Set the plugin directory.
      *
-     * @param directory the directory
+     * @param pluginDirectory the directory
      */
-    public synchronized void setDirectory(File directory) {
-        this.directory = directory;
+    public synchronized void setPluginDirectory(File pluginDirectory) {
+        this.pluginDirectory = pluginDirectory;
     }
 
     /**
@@ -91,8 +91,8 @@ public final class AuthorizationFramework {
      *
      * @param directory the directory path
      */
-    public void setDirectory(String directory) {
-        setDirectory(directory != null ? new File(directory) : null);
+    public void setPluginDirectory(String directory) {
+        AuthorizationFramework.this.setPluginDirectory(directory != null ? new File(directory) : null);
     }
 
     /**
@@ -139,7 +139,7 @@ public final class AuthorizationFramework {
         String path = RuntimeEnvironment.getInstance()
                 .getPluginDirectory();
         plugins = new ArrayList<>();
-        this.directory = path == null ? null : new File(path);
+        setPluginDirectory(path);
         reload();
     }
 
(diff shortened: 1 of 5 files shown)


src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -83,7 +83,17 @@
...
+     * Directory with authorization plugins. Default value is
+     * dataRoot/../plugins (can be /var/opengrok/plugins if dataRoot is
+     * /var/opengrok/data).
+     */
     private String pluginDirectory;
+    /**
+     * Enable watching the plugin directory for changes in real time. Suitable
+     * for development.
+     */
+    private boolean authorizationWatchdog;
```
vladak: the name of the var should signify the type, what about `authorizationWatchdogEnabled` or `authorizationWatchdogOn` or similar ?

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -275,12 +297,14 @@ private String getClassName(JarEntry f) {
...
      *
-     * @see IAuthorizationPlugin#load() 
-     * @see IAuthorizationPlugin#unload() 
+     * @see IAuthorizationPlugin#load()
+     * @see IAuthorizationPlugin#unload()
      */
     @SuppressWarnings("unchecked")
     public synchronized void reload() {
         if (directory == null || !directory.isDirectory() || !directory.canRead()) {
+            LOGGER.log(Level.INFO, "plugin directory not found or not readable: {0}. "
```
vladak: ERROR level would be better I think - if someone configures the directory in config, he probably wants to use it so treating as error is in order.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -53,7 +53,7 @@
 public final class AuthorizationFramework {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizationFramework.class);
-    private final File directory;
+    private File directory;
```
vladak: `pluginDirectory` ?

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -113,22 +138,19 @@ public boolean test(IAuthorizationPlugin plugin) {
...
         String path = RuntimeEnvironment.getInstance()
                 .getPluginDirectory();
-        File directory = path == null ? null : new File(path);
-        if (path == null || directory == null || !directory.isDirectory() || !directory.canRead()) {
-            LOGGER.log(Level.INFO, "plugin directory not found or not readable: {0}. "
-                    + "The AuthorizationFramework will just pass requests through.", path);
-        }
         plugins = new ArrayList<>();
-        this.directory = directory;
+        this.directory = path == null ? null : new File(path);
```
vladak: use `setDirectory()` ?

src/org/opensolaris/opengrok/configuration/Configuration.java
vladak: Also, I'd expect some mention (ideally control of) the refresh/watch period.

tulinkry: > given this will be present in configuration.xml, the name of the var should signify the type, what about authorizationWatchdogEnabled or authorizationWatchdogOn or similar ?

changed

> Also, I'd expect some mention (ideally control of) the refresh/watch period.

There is no way to control this. It is done automatically with the WatchService and that is implementation dependent how the service watch for changes.

In case you meant this:

```java
Thread.sleep(THREAD_SLEEP_TIME); // experimental wait if file is being written right now
AuthorizationFramework.getInstance().reload();
```

It means something different than the refresh period. It is set to 2s and I think that it is not necessary to control this and also that is is enough.

I added a comment to this method (which has already been there for the `authorizationWatchdogEnabled` variable in configuration).

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: ERROR level does not exist. I changed that to WARNING

PR #1467, opened 2017-03-13
Json servlet updates

pmv:
- add ability to restrict searches to specific projects
- use application/json content type

### DIFF
--- a/src/org/opensolaris/opengrok/web/JSONSearchServlet.java
+++ b/src/org/opensolaris/opengrok/web/JSONSearchServlet.java
@@ -47,6 +47,7 @@ public class JSONSearchServlet extends HttpServlet {
     private static final String PARAM_PATH = "path";
     private static final String PARAM_HIST = "hist";
     private static final String PARAM_MAXRESULTS = "maxresults";
+    private static final String PARAM_PROJECT = "project";
     private static final String ATTRIBUTE_DIRECTORY = "directory";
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/web/JSONSearchServlet.java
```diff
@@ -47,6 +47,7 @@
     private static final String PARAM_PATH = "path";
```
tulinkry: wrong year in copyright

tulinkry:
Except the copyright this is ok to me

tarzanek:
lgtm, if OCA is sorted feel free to merge
thank you!

vladak:
@pmv if you don't have OCA number already, please start the process and state that you agree with the OCA terms conditions so that this PR can be merged.

tarzanek:
@pmv OCA : http://www.oracle.com/technetwork/community/oca-486395.html
if you say you agree with OCA here, we can merge immediatelly (and you can follow up with OCA in parallel), tia
L

tarzanek:
@pmv we plan on releasing very soon, any chance you can get OCA sorted?
I think this would be good to have this PR in

pmv:
Sorry for the delay.  I just received notice that the OCA has been approved from my side, so we should be good to go.

vladak:
Great, thanks !

PR #1471, opened 2017-03-16
more robust handling of merging history entries for files

vladak:
Sometimes in our deployment I saw some files having history entries twice - the history is formed by 2 identical history entry sequences. It's not clear to me how this can happen, it has to be caused by invalid merging of old and new history. It seems this happens for renamed files. This PR makes the area more robust by avoiding the merge for renamed files.

### DIFF
--- a/src/org/opensolaris/opengrok/history/FileHistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/FileHistoryCache.java
@@ -267,11 +267,12 @@ class FileHistoryCache implements HistoryCache {
      * @param cacheFile file to where the history object will be stored
      * @param histNew history object with new history entries
      * @param repo repository to where pre pre-image of the cacheFile belong
-     * @return merged history
+     * @return merged history (can be null if merge failed for some reason)
      * @throws HistoryException
      */
-    private History mergeOldAndNewHistory(File cacheFile, History histNew, Repository repo) throws HistoryException {
-        // Incremental update of the history for this file.
+    private History mergeOldAndNewHistory(File cacheFile, History histNew, Repository repo)
+            throws HistoryException {
+
         History histOld;
         History history = null;
 
@@ -314,7 +315,8 @@ class FileHistoryCache implements HistoryCache {
      * @param history history object to store
      * @param file file to store the history object into
      * @param repo repository for the file
-     * @param mergeHistory whether to merge the history with existing or store the histNew as is
+     * @param mergeHistory whether to merge the history with existing or
+     *                     store the histNew as is
      * @throws HistoryException
      */
     private void storeFile(History histNew, File file, Repository repo,
@@ -333,7 +335,13 @@ class FileHistoryCache implements HistoryCache {
             history = mergeOldAndNewHistory(cacheFile, histNew, repo);
         }
 
-        writeHistoryToFile(dir, history, cacheFile);
+        // If the merge failed, null history will be returned.
+        // In such case store at least new history as a best effort.
+        if (history != null) {
+            writeHistoryToFile(dir, history, cacheFile);
+        } else {
+            writeHistoryToFile(dir, histNew, cacheFile);
+        }
     }
 
     private void finishStore(Repository repository, String latestRev) {
--- a/src/org/opensolaris/opengrok/history/FileHistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/FileHistoryCache.java
@@ -129,7 +129,7 @@ class FileHistoryCache implements HistoryCache {
 
         File file = new File(root, filename);
         if (!file.isDirectory()) {
-            storeFile(hist, file, repository);
+            storeFile(hist, file, repository, !renamed);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -309,9 +302,35 @@ private void storeFile(History histNew, File file, Repository repo) throws Histo
...
+        if (!dir.isDirectory() && !dir.mkdirs()) {
+            throw new HistoryException(
+                    "Unable to create cache directory '" + dir + "'.");
+        }
+
+        if (mergeHistory && cacheFile.exists()) {
+            history = mergeOldAndNewHistory(cacheFile, histNew, repo);
         }
 
         writeHistoryToFile(dir, history, cacheFile);
```
Hlipa: What happens here when mergeOldAndNewHistory() returns null. Looks like it will fail in writeHistoryToFile() later.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -309,12 +303,45 @@ private void storeFile(History histNew, File file, Repository repo) throws Histo
...
+     * @param history history object to store
+     * @param file file to store the history object into
+     * @param repo repository for the file
+     * @param mergeHistory whether to merge the history with existing or
+     *                     store the histNew as is
+     * @throws HistoryException
+     */
+    private void storeFile(History histNew, File file, Repository repo,
+            boolean mergeHistory) throws HistoryException {
+
```
tulinkry: It would be nice if there was a overloaded version without the `mergeHistory` parameter.
```java
private void storeFile(History histNew, File file, Repository repo)
```

just an idea

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -309,12 +303,45 @@ private void storeFile(History histNew, File file, Repository repo) throws Histo
                 }
             }
         } catch (IOException ex) {
-            // Ideally we would want to catch the case when incremental update
-            // is done but the cached file is not there however we do not have
-            // the data to do it here.
```
tulinkry: This comment is no longer "valid"?

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -309,12 +303,45 @@ private void storeFile(History histNew, File file, Repository repo) throws Histo
...
+            LOGGER.log(Level.SEVERE, "Cannot open history cache file " + cacheFile.getPath());
```
tulinkry: use logger templates instead of string concatenation

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -309,12 +303,45 @@ private void storeFile(History histNew, File file, Repository repo) throws Histo
...
+     * @param repo repository for the file
+     * @param mergeHistory whether to merge the history with existing or
+     *                     store the histNew as is
+     * @throws HistoryException
+     */
+    private void storeFile(History histNew, File file, Repository repo,
+            boolean mergeHistory) throws HistoryException {
+
+        File cacheFile = getCachedFile(file);
+        History history = histNew;
```
tulinkry: is it necessary to assign the histNew value when it gets overwritten?

src/org/opensolaris/opengrok/history/FileHistoryCache.java
tulinkry: or append the exception and use String.format

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: I think this sort of info belongs to Issues, not to code. Moreover, the feasibility of this change is low so such comment does not have much value.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: fixed

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: it only gets overwritten if `mergeHistory` is true, that's the trick here.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: that's interesting idea. That said, I quite like this construct in `doFileHistory()`:

```
        if (!file.isDirectory()) {
            storeFile(hist, file, repository, !renamed);
        }
```

Also, it would like to code duplication which I'd like to avoid so keeping as is.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
tulinkry: Ok. Still it works the same when the `history = null` is used but this is ok

src/org/opensolaris/opengrok/history/FileHistoryCache.java
tulinkry: I meant that if you do this:
```java
private void storeFile(History histNew, File file, Repository repo) {
    storeFile(histNew, file, repo, false);
}

private void storeFile(History histNew, File file, Repository repo,
            boolean mergeHistory) {
    // the code as you have it
}
```

Then you keep your constructs and allow the callee not to worry about renamed parameter.
This is still valid
```java
        if (!file.isDirectory()) {
            storeFile(hist, file, repository, !renamed);
        }
```
As well as this:

```java
// retrieving the history takes too long, cache it!
storeFile(history, file, repository);
```

This doesn't really bring a code duplication. However all of these are just cosmetics.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
tulinkry: ok

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: will address this in separate PR

PR #1473, opened 2017-03-20
defaulting the cvs repository to automatically decide whether it is i…

tulinkry:
…nstalled or not

I don't know how it was supposed to work but with the previous settings (introduced in d8a4ee9) it always skips the tests for this repository even if the binary is installed on the system.

### DIFF
--- a/src/org/opensolaris/opengrok/history/CVSRepository.java
+++ b/src/org/opensolaris/opengrok/history/CVSRepository.java
@@ -67,6 +67,17 @@ public class CVSRepository extends RCSRepository {
             = Pattern.compile("([\\.\\d]+)\\W+\\((\\w+)");
 
     public CVSRepository() {
+        /**
+         * This variable is set in the anchestor to TRUE which has a side effect
+         * that this repository is always marked as working even though it does
+         * not have the binary available on the system.
+         *
+         * Setting this to null does restores the default behavior (as java
+         * default for reference is null) for this repository - detecting the
+         * binary and act upon that.
+         *
+         * @see #isWorking
+         */
         working = null;
         setType("CVS");
         datePatterns = new String[]{
--- a/src/org/opensolaris/opengrok/history/CVSRepository.java
+++ b/src/org/opensolaris/opengrok/history/CVSRepository.java
@@ -67,7 +67,7 @@ public class CVSRepository extends RCSRepository {
             = Pattern.compile("([\\.\\d]+)\\W+\\((\\w+)");
 
     public CVSRepository() {
-        working = Boolean.FALSE;
+        working = null;
         setType("CVS");
         datePatterns = new String[]{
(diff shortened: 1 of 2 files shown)


vladak:
pls fix the typo (`csv`) in the synopsis

tulinkry:
Done

vladak:
Could you provide some explanation on how is the detection supposed to work ?

Also, it would be good to check other repos too.

vladak:
while you're there could you fix the typo in RepositoryInfo.java:

```
85       * systems that use external binaries, the binary must be availabe etc)
```

src/org/opensolaris/opengrok/history/CVSRepository.java
```diff
@@ -67,7 +67,7 @@
             = Pattern.compile("([\\.\\d]+)\\W+\\((\\w+)");
 
     public CVSRepository() {
-        working = Boolean.FALSE;
+        working = null;
```
vladak: is this line even necessary ?

tulinkry:
> Could you provide some explanation on how is the detection supposed to work ?

Each repository has a `isWorking()` method which decide if the repository is available. For most of the repos this consists of something like:

```python
isWorking:
    if working != null:
        working = run_repo_command
    return working
```

This means that once you set the working variable this check will always return true without even checking if the repository binary is available. 

> is this line (the new line added) even necessary ?

The problem above exactly happens here as the CVSRepo inherits from RCSRepo where the variable is set to `true`. Apparently the easiest fix was to set the variable to `false` in CSVRepo however that bypasses the check - so setting it to null (as it usually is by java default) is good way here to ensure the check again.

> Also, it would be good to check other repos too.

No other repo contains the same problem. Several repos set the `working` variable to `true` by default. I don't have experience with these.

 - RazorRepository - is working by default
 - RepoRepository - is working by default
 - RCSRepository - is working by default

src/org/opensolaris/opengrok/history/CVSRepository.java
tulinkry: Yes because its anchestor sets it to TRUE. See [my comment below](https://github.com/OpenGrok/OpenGrok/pull/1473#issuecomment-287709907)

tulinkry:
Fixed the typo.

vladak:
I wonder if the RCS-CVS inheritance is a healthy one.

RCS and Razor repositories are parsed in Java so they don't need external binaries it seems. The 'repo' repository is special case of git AFAIK so it probably has some special handling.

tulinkry:
Comment added. I decided to leave the inheritance because of
 - we're about to do a new release
 - lack of experience with this kind of repository

vladak:
Fair enough, after all CVS was born from RCS so this makes sense at least conceptually.

PR #1476, opened 2017-03-22
javafy storeFile()

vladak:
as discussed in #1471


src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -560,6 +565,7 @@ public History get(File file, Repository repository, boolean withFiles)
                 storeFile(history, file, repository, false);
             }
```
tulinkry: Maybe use it directly here?

vladak: forgotten commit, fixed

tulinkry: ok

PR #1480, opened 2017-03-24
move project picker to the query table

### LINKED ISSUE #1370: move project selection list and buttons above the other search-text fields
Please, consider moving the project list above the other search-text fields. I hope, this should provide a better navigation on narrower screens (tables, smart phones). Of course, the "In project(s)" and the buttons could be differently aligned.. Also, the chosen project buttons/icons could pop-up under the project selection-list (between the project selection-list and the "Full Search" text field).

move_project_select

vladak:
fixes #1370

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -283,7 +283,7 @@ label {
 }
 
 /* *** full search navbar *** */
-#qtbl, #ptbl {
+#qtbl {
     display: inline-block;
     vertical-align: top;
 }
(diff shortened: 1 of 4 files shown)


vladak:
looks like this with default css:

scr

ChristopheBordieu:
Personaly, I would make button same size , text inside sams font/size and I would align buttons of Project  bar to the right like the ones of Search.

tulinkry:
> I would align buttons of Project bar to the right like the ones of Search

This stayed on the right side in order to not to be confused with the form controls - these are the project picker controls. The same reason applies for the size. But it may look better, I haven't tested it.

PR #1486, opened 2017-03-27
checking source root existence in the top level jsps

tulinkry:
After changes in #1453 not all pages are affected with the source root check.

The error page is not set in every case when the configuration is missing. This PR should set it everywhere.

Actual output when the configuration.xml does not exist (status 200 OK):
screenshot from 2017-03-27 15-33-09
screenshot from 2017-03-27 15-33-18

Expected output:
screenshot from 2017-03-27 15-35-46
screenshot from 2017-03-27 15-35-59
screenshot from 2017-03-27 15-35-38

### DIFF
--- a/web/enoent.jsp
+++ b/web/enoent.jsp
@@ -54,7 +54,7 @@ include file="menu.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     String configError = "";
-    if (cfg.getSourceRootPath().isEmpty()) {
+    if (cfg.getSourceRootPath() == null || cfg.getSourceRootPath().isEmpty()) {
         configError = "CONFIGURATION parameter has not been configured in "
             + "web.xml! Please configure your webapp.";
     } else if (!cfg.getEnv().getSourceRootFile().isDirectory()) {
--- a/web/enoent.jsp
+++ b/web/enoent.jsp
@@ -19,13 +19,14 @@ CDDL HEADER END
 Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
 Portions Copyright 2011 Jens Elkner.
 
---%><%@page session="false" isErrorPage="true" import="
+--%><%@page session="false" errorPage="error.jsp" isErrorPage="true" import="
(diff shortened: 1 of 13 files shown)


web/enoent.jsp
```diff
@@ -54,7 +54,7 @@ include file="menu.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
     String configError = "";
-    if (cfg.getSourceRootPath().isEmpty()) {
+    if (cfg.getSourceRootPath() == null || cfg.getSourceRootPath().isEmpty()) {
```
vladak: why is `checkSourceRootExistence()` not used here ?

tulinkry: I wasn't quite sure about the `errorPage="error.jsp" isErrorPage="true"` configuration as it's not really common use case however I checked that it works. Actually it makes more sense to return 500 instead of 404 with the same error message.

vladak: Still having troubles understanding how it works/should work. If the call to checkSourceRootExistence() ` in `enoent.jsp` results in exception, this will be handled by `error.jsp` ?

In other words, is the `configError` handling and display of exception needed in `enoent.jsp` ?

tulinkry: This `errorPage="error.jsp"` means that when there is an exception it will be passed to the error.jsp. This `isErrorPage="true"` means that this page can access the `exception` object. Using this in `enoent.jsp` means that any exception in `enoent.jsp` will be handled by `error.jsp` meanwhile `enoent.jsp` stays an error page for 404 errors.

The stuff in `enoent.jsp` is needed for the usual 404 (file not found) errors. Actually it should not be necessary for the "no configuration" error which is handled via the exception and `error.jsp`.

vladak: That's what I meant - the no config handling can be removed from `enoent.jsp`.

tarzanek:
merging, thank you

web/enoent.jsp
tulinkry: covered in #1496

PR #1492, opened 2017-03-28
READ_XML_CONFIGURATION -> OPENGROK_READ_XML_CONFIGURATION

### LINKED ISSUE #1487: Add OPENGROK_ prefix to the READ_XML_CONFIGURATION
Environment variable `READ_XML_CONFIGURATION` should have the prefix `OPENGROK_` like the other variables as it is opengrok related.

**The same naming scheme is used in `Groups` and `Messages`.**

vladak:
fixes #1487


vladak:
will let @tarzanek know to add it to the blog

Issue #1488, opened 2017-03-28
scrollbar shouldn't depend on .js


tarzanek:
When loading big file firefox won't print its scrollbar
due to .js errors

content shouldn't depend on .js to show

.js needs to be loaded at the end of the page

ilovemaui:
how long to fix this? and is this a complete rebuild? or is there a quick way to fix?

vladak:
According to @tulinkry there is fair amount of work associated with this change. Given the relatively low impact and impending release I am removing the stopper keyword.

stevenschlansker:
The original issue description isn't very precise, so I'm not sure if this is the same problem -- but we are noticing that our javascript files on the source viewer have an incorrect relative path.

e.g. the page `http://opengrok/xref/myproj/src/main.c` has this block at the bottom of the page:
```
	<script type="text/javascript" src="js/jquery-3.2.0.min.js" data-priority="10"></script>
	<script type="text/javascript" src="js/jquery-ui-1.12.0-custom.min.js" data-priority="11"></script>
	<script type="text/javascript" src="js/jquery-tablesorter-2.26.6.min.js" data-priority="12"></script>
	<script type="text/javascript" src="js/tablesorter-parsers-0.0.1.js" data-priority="13"></script>
	<script type="text/javascript" src="js/searchable-option-list-2.0.3.min.js" data-priority="14"></script>
	<script type="text/javascript" src="js/utils-0.0.9.js" data-priority="15"></script>
```
which is relative to the *project directory of the source being viewed* instead of the context root, which the code in [Scripts.java](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/web/Scripts.java#L116-L131) assumes.

Is this the issue reported above, or should I open a new issue report?  This causes any source file longer than a page to be inaccessible so I'd consider it probably a blocker.  Confirmed that the behavior is browser independent and affects OpenGrok 0.13 (is that the same as 1.0?)

vladak:
yep, 0.13 was renamed to 1.0 just before the release.

stevenschlansker:
Any comment on whether this is the same issue or I should open a new one?

vladak:
@stevenschlansker im the recent `master` branch (1.1-rc5+) I see this in the web page source when requesting content under `/source/xref/`:

```
<script type="text/javascript" src="/source/js/jquery-3.2.0.min.js" data-priority="10"></script>
[...]

PR #1493, opened 2017-03-28
moving javascripts to the page footer

### LINKED ISSUE #1455: move scripts to the bottom of the page
https://github.com/OpenGrok/OpenGrok/blob/master/web/httpheader.jspf#L78

scripts should be sourced in as last on page
(performance best practice)

### LINKED ISSUE #1490: Navigate window expects scopes window to be ready
This function in utils.js expects scopes window to be inizialized. The check should contain also that condition when the scopes window isnt available at all.

```js
updatePosition: function ($w) {
    var a = {}
    a.top = $.scopesWindow.is(':visible') ? $.scopesWindow.offset().top + $.scopesWindow.outerHeight() + 20 : this.getTopOffset() + 10;
    a.height = Math.min(parseFloat($w.css('max-height')) || 480, $(browserWindow).outerHeight() - a.top - ($w.outerHeight(true) - $w.height()) - 20);

    if (a.height == $w.height() && a.top == this.getTopOffset())
        return $w;

    if (this.$content.children().length === 0)
        // the window is empty
        delete a.height

    return $w.stop().animate(a)
},
```

tulinkry:
fixes #1455
fixes #1490
related to #1488

I double checked all jsps; this configuration appears to work.

Summary:
1. for the `document` variables I used functions so that the inner block does not need to be defined in advance
2. moved the js to footer
3. fixed observed problem described in #1490 
4. merged `$(document).ready()` calls to single call

### DIFF
--- a/web/repos.jspf
+++ b/web/repos.jspf
@@ -46,6 +46,14 @@ Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
 <%@page import="org.opensolaris.opengrok.web.PageConfig"%>
 <%
 {
+    /**
+     * This file depends on the javascript file repos.js which must be
+     * included after all other scripts. Currently it is included in
+     * index.jsp.
+     *
+     * Without this file the group and project listing will not work.
+     */
+
     PageConfig cfg = PageConfig.get(request);
     JSONArray messages;
 
--- a/web/repos.jspf
+++ b/web/repos.jspf
@@ -44,7 +44,6 @@ Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
 <%@page import="org.opensolaris.opengrok.configuration.Group"%>
 <%@page import="java.util.List"%>
 <%@page import="org.opensolaris.opengrok.web.PageConfig"%>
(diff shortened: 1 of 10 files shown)


tarzanek:
;) sleeps in tests made travis cry?

tarzanek:
I have restarted the build

tulinkry:
Aargh. Need to update them after the release.

tarzanek:
no prob, we've all been there and done that
sleeps are just evil ...

web/repos.jspf
```diff
@@ -44,7 +44,6 @@ Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
 <%@page import="org.opensolaris.opengrok.configuration.Group"%>
 <%@page import="java.util.List"%>
 <%@page import="org.opensolaris.opengrok.web.PageConfig"%>
-<script src="<%=request.getContextPath()%>/js/repos-0.0.1.js" type="text/javascript"></script>
```
vladak: pls add a comment to whoever is including `repos.jspf` to add the `script` tag at the end of the file

tulinkry: added a comment to the top

tarzanek:
thank you @tulinkry ! merging

Issue #1479, opened 2017-03-24
Message for merging configuration


tulinkry:
After changes in #1433 (user can send new configuration to the web app through the Messages tools) it would be quite useful if user could control if the new configuration will merge with the current or will overwrite it (current behaviour).

tulinkry:
Actually it could be useful to even have a possibility to set some primitive properities directly with something like
```bash
./Messages -n config -t set "messageLimit = 1000"
./Messages -n config -t set "userPage = http://example.com?user="
```

tulinkry:
The second part is implemented in 662c4211d55a90a4b0555965358f6508ca9cedeb

tulinkry:
Closing as the first part is sort of implemented by #1590 where user can merge several configuration into a single one and then send it as a single file.

PR #1499, opened 2017-03-29
Messages: adding an option 'set' for config message

tulinkry:
Mentioned in #1479 this implements the possibility to set a webapp option at runtime. It's not really necessary but appears to be quite handy with some options to test them out, like
 - `authorizationWatchdogEnabled` (#1465) 
 - `hitsPerPage`
 - `chattyStatusPage`
 - `revisionMessageCollapseThreshold`
 - `groupsCollapseThreshold`
 - `messageLimit`
and a couple of others. Not every option in `Configuration.java` has a meaning for the webapplication.

**The change is not persistent when server was restarted or webapp was redeployed or when the reindex was run.** For persistent configuration when reindexing use the `OPENGROK_READ_XML_CONFIGURATION` environment variable pointing to a xml file with changed options.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -137,7 +137,7 @@ Currently supported message types:
 4) ConfigMessage (config)
     This message retrieves or sends a configuration to the webapp,
     depending on tag. "setconf" tag sends config to webapp and requires
-    file as argument.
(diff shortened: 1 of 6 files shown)


tarzanek:
lgtm, can we merge?

Issue #1501, opened 2017-03-29
Navigate and Scopes windows superimpose


ChristopheBordieu:
These 2 windows superimpose when both activated at same time.
![image](https://cloud.githubusercontent.com/assets/26546513/24453965/2df9088a-148a-11e7-8c1c-d71f466c8309.png)
When both are empty, only Navigate window is visibe. 
When at least Navigate is non empty, Scopes is always in front.
I have no idea what happens when Navigate is empty but not Scopes ?

Maybe, these 2 windows when activated at same time should be side by side ?

tulinkry:
I currently updated this in #1502 and it should by fine when it gets merged.

The idea is (and was partially supported) that when both are open then the navigate moves down a bit.

tulinkry:
Closing as this is implemented in #1502. Feel free to reopen if the problem persists

PR #1502, opened 2017-03-29
speeding up the scopes window

tulinkry:
As suggested by @mrate I don't update the scopes window when it's not visible.

The whole change here is about this at the very bottom in `scopes_on_scroll()`:
```javascript
if($.scopesWindow && $.scopesWindow.initialized && !$.scopesWindow.is(':visible')) {
     return;
}
```

The rest is just to fully support this behaviour without any visible delay. I tested the results in a big file https://github.com/svn2github/valgrind-vex/blob/master/priv/guest_arm_toIR.c.

### DIFF
--- a/web/js/utils-0.0.8.js
+++ b/web/js/utils-0.0.8.js
@@ -592,12 +592,26 @@
                 };
 
                 /**
+                 * Display or hide the window.
+                 *
+                 * We override this method from jquery to manually
+                 * trigger the hide() and show() methods
+                 * which may be used in the descendants.
+                 *
+                 * @returns self
+                 */
+                $window.toggle = function() {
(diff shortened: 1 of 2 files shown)


web/js/utils-0.0.8.js
```diff
@@ -1634,6 +1674,7 @@ function pageReadyList() {
     }
     $('#navigate').click(function () {
         $.navigateWindow.toggle()
+        return false;
```
tulinkry: read as "result is that it wasn't clicked on this element" - this doesn't add the empty # in the url

mrate:
Nice! Thank you

tarzanek:
thank you guys, merging, let's get the auth stuff in and release!!! :)

tulinkry:
Should fix #1501

PR #1504, opened 2017-03-30
add option to disable handling renamed files history

### LINKED ISSUE #1500: Provide a way how to simplify turning off the renamed files
In #1491 and some others we discussed that the process of turning off the renamed files handling is somewhat cumbersome (adding a property to a xml file and using `OPENGROK_READ_XML_CONGURATION` variable). It would be nice if there was a quicker way through say `OPENGROK_RENAMED_FILES`. 

I know we're almost out of the alphabet in the command line but this appeared as really important for several users.

vladak:
fixes #1500


tarzanek:
lgtm, feel free to merge

PR #1506, opened 2017-03-31
bugs and reviews are replaced in both short and long summary

tulinkry:
Reported in internal environment. When there is a bug id (or a review number) which matches the `bugRegex` the commit message is replaced with the link to it. However this happened only for the "long" (original) summary not for the "short" summary. When user clicked "show more" the links disappeared.

Now the code supports both ways.

### DIFF
--- a/web/history.jsp
+++ b/web/history.jsp
@@ -324,12 +324,10 @@ document.domReady.push(function() {domReadyHistory();});
                 String cout = Util.htmlize(entry.getMessage());
 
                 if (bugPage != null && bugPage.length() > 0) {
-                    cout = bugPattern.matcher(cout).replaceAll("<a href=\""
-                        + bugPage + "$1\">$1</a>");
+                    cout = Util.linkifyPattern(cout, bugPattern, "$1", bugPage + "$1");
                 }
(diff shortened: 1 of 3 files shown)


web/history.jsp
```diff
@@ -338,6 +336,12 @@ document.domReady.push(function() {domReadyHistory();});
                     showSummary = true;
                     coutSummary = coutSummary.substring(0, summaryLength - 1);
                     coutSummary = Util.htmlize(coutSummary);
+                    if (bugPage != null && bugPage.length() > 0) {
+                        coutSummary = Util.linkifyPattern(coutSummary, bugPattern, "$1", bugPage + "$1");
+                    }
+                    if (reviewPage != null && reviewPage.length() > 0) {
+                        coutSummary = Util.linkifyPattern(coutSummary, reviewPattern, "$1", reviewPage + "$1");
```
vladak: the use of the `"$1"` literal assumes that the bug pattern is valid, pls file new issue to validate the pattern when Configuration is being set.

tulinkry: #1508

PR #1456, opened 2017-03-06
BitKeeper History Parsing

jas2701:
I have recently been setting up an OpenGrok server and I wrote this to handle some BitKeeper repositories.  I'm not an expert in BitKeeper (or Java for that matter) but I figured this was probably good enough to be useful to someone.

PR #1507, opened 2017-03-31
rename Bitkeeper .bk directory

vladak:
post PR #1456 fixup


tulinkry:
Can you also fix all the copyright commented in #1456 ?

vladak:
Not sure if the Oracle copyright is necessary as the code has not originated from our side.

PR #1522, opened 2017-04-07
Support for nested Git repositories (#1512)

### LINKED ISSUE #1512: support for nested git repositories
I'm using the latest OpenGrok 1.0 release, and I have an issue with the repository detection for the history creation.

Our project is using `repo` for the file versioning, allowing to have a Git repository inside another one. The folder structure looks like:
```
baseDirectory/.git
baseDirectory/anotherDirectory
baseDirectory/anotherDirectory/.git
```

The first Git repository `baseDirectory` is correctly detected and the history is retrieved, but the second one `anotherDirectory` is never seen. I have tried to use the `-z` command-line option, but it doesn't solve my issue.

Looking at [the code](https://github.com/OpenGrok/OpenGrok/blob/b9a9ac8ac8c99dbddc850331b5f8229ed0ac3509/src/org/opensolaris/opengrok/history/HistoryGuru.java#L339) of the `addRepositories` function, I'm under the impression that the case of a repository inside another repository is not supported. Am I right?

I know that I could create the history for all sub-repositories using the `-h` option, but this would be extremely painful as all absolute paths should be added manually. This would also not be easy to maintain if a repository is added or removed.

dohseven:
Fixes #1512, adds `supportsSubRepositories()` to Git.

And I accept the Oracle Contributor Agreement. I haven't signed it and sent it back, but apparently it is enough to specify it in the pull request comment.

### DIFF
--- a/src/org/opensolaris/opengrok/history/GitRepository.java
+++ b/src/org/opensolaris/opengrok/history/GitRepository.java
@@ -488,6 +488,11 @@ public class GitRepository extends Repository {
     }
 
     @Override
+    boolean supportsSubRepositories() {
+        return true;
+    }
+
+    @Override
     public boolean isWorking() {
         if (working == null) {
             ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);


vladak:
I wonder what is the interaction with `RepoRepository`.

As for OCA, it is expected that if you agree then you initiate the process so that we have our paperwork in order :-)

dohseven:
Well, `RepoRepository` does not support history (`fileHasHistory()` and `hasHistoryForDirectories()` both return false), so the `repo` repository is detected but no history is retrieved.
And the `git` repositories located in the subfolders are then correctly detected.

By the way I sent the signed OCA today, so everything should be in order.

PR #1525, opened 2017-04-10
using authorization stacks instead of a list

tulinkry:
1. Plugin stack

    This change is to introduce the ability to a plugin stack to be a substack of another plugin stack with similar attributes as the plugins have (1):
     - required
     - requisite
     - sufficient

    The substack can contain an arbitrary number of
     - other stacks
     - plugin classes

    The meaning of the three keywords in (1) is not changed and is extended for the entire stacks as well.

    There is a default **required** top level stack for the application with a name "default stack"

2. Plugin parameters

    Another friendliness is that the plugin can now accept an arbitrary parameter included in java `Map` object. This parameters can be configured in configuration. The parameters can be also used for the entire stack and then it is merged with the plugin specific parameters (the plugin overwrites the conflicts).

    **However note that this is BC break as the `IAuthorizationInterface` has changed slightly.**

A configuration example can be this configuration with:
- two standalone plugins (SamplePlugin, RequisitePlugin)
- a required substack
   - containing two other plugins

```xml
  <void property="pluginStack">
   <void method="add">
     <void property="name">
      <string>SamplePlugin</string>
     </void>
     <void property="role">
      <string>SUFFICIENT</string>
     </void>
    </object>
   </void>
   <void method="add">
     <void property="name">
      <string>tulinkry.my.RequisitePlugin</string>
     </void>
     <void property="role">
      <string>REQUIRED</string>
     </void>
     <void property="setup">
		<void method="put">
			<string>config</string>
			<string>/home/ktulinge/my.xml</string>
		</void>
		<void method="put">
			<string>group</string>
			<object idref="group"/>
		</void>
     </void>
    </object>
   </void>
   <void method="add">
    <object class="org.opensolaris.opengrok.authorization.AuthorizationStack">
     <void property="name">
      <string>stack for group 1</string>
     </void>
     <void property="role">
      <string>REQUIRED</string>
     </void>
     <void method="add">
		 <void property="name">
		  <string>org.opensolaris.opengrok.authorization.plugins.GroupPlugin</string>
		 </void>
		 <void property="role">
		  <string>SUFFICIENT</string>
		 </void>
		 <void property="setup">
		 	<void method="put">
		 	  <string>groups</string>
              <array class="java.lang.String" length="2">
               <void index="0">
                 <string>1</string>
               </void>
               <void index="1">
                 <string>2</string>
               </void>
              </array>
		 	</void>
         </void>
		</object>
     </void>
     <void method="add">
		 <void property="name">
		  <string>Plugin</string>
		 </void>
		 <void property="role">
		  <string>REQUIRED</string>
		 </void>
		</object>
     </void>
    </object>
   </void>
  </void>
```

Also note this parts which is how you can set the parameters to the plugin class - **property setup** which is a `Map`.

```xml
  <void property="groups">
   <void method="add">
    <object class="org.opensolaris.opengrok.configuration.Group"  id="group">
     <void property="name">
      <string>3</string>
     </void>
     <void property="pattern">
      <string>.*\([456]th copy\)</string>
     </void>
    </object>
   </void>
  </void>

  <!-- group with id="group" declared above and the reference can be used later in this file -->

   <void method="add">
     <void property="name">
      <string>tulinkry.my.RequisitePlugin</string>
     </void>
     <void property="role">
      <string>REQUIRED</string>
     </void>
     <void property="setup">
		<void method="put">
			<string>config</string> <!-- PARAMETER KEY -->
			<string>/home/ktulinge/my.xml</string> <!-- PARAMETER VALUE -->
		</void>
		<void method="put">
			<string>group</string> <!-- PARAMETER KEY -->
			<object idref="group"/> <!-- PARAMETER VALUE => here a group declared in this configuration before this statement -->
		</void>
     </void>
    </object>
   </void>
```


src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
```diff
@@ -0,0 +1,95 @@
+/*
```
tulinkry: This file is refactored from the `AuthorizationCheck.java` class

src/org/opensolaris/opengrok/authorization/AuthorizationCheck.java
```diff
@@ -1,155 +0,0 @@
-/*
```
tulinkry: Refactored to:
 - `AuthorizationEntity.java`
 - `AuthControlFlag.java`

test/org/opensolaris/opengrok/authorization/AuthorizationFrameworkTest.java
```diff
@@ -22,326 +22,655 @@
...
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.web.DummyHttpServletRequest;
 
-import static org.junit.Assert.assertEquals;
-
+@RunWith(Parameterized.class)
 public class AuthorizationFrameworkTest {
 
     private static String pluginDirectory;
-
```
tulinkry: This file is heavily refactored and changed to easily form the test cases for the authorization and to support a test cases for substacks. All important test functions in the previous version have been converted to the array of parameters in the new version.

test/org/opensolaris/opengrok/web/ProjectHelperTestBase.java
```diff
@@ -252,23 +256,12 @@ public static void tearDownClass() {
...
 
     protected void invokeRemoveAll() {
-        try {
-            Method method = AuthorizationFramework.class.getDeclaredMethod("removeAll");
-            method.setAccessible(true);
-            method.invoke(AuthorizationFramework.getInstance());
-        } catch (Exception ex) {
-            Assert.fail("invokeRemoveAll should not throw an exception");
-        }
+        env.setPluginStack(new AuthorizationStack(AuthControlFlag.REQUIRED, "default-stack"));
```
tulinkry: This is necessary because the application changes the stack at runtime directly into `env` and does not have its local copy in the auth. framework.

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
```diff
@@ -0,0 +1,95 @@
...
+     * authorization framework returns success to the application immediately
+     * without calling any further plugins in the stack. A failure of a
+     * sufficient plugin is ignored and processing of the plugin list continues
+     * unaffected.
+     */
+    SUFFICIENT("sufficient");
+
+    private final String role;
+
+    private AuthControlFlag(String role) {
```
vladak: there is still a bit of schizophrenia in the naming - role vs. control flag.

tulinkry: so all should be flag as in pam list?

tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationEntity.java
```diff
@@ -0,0 +1,214 @@
...
+package org.opensolaris.opengrok.authorization;
+
+import java.io.Serializable;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.function.Predicate;
+import org.opensolaris.opengrok.configuration.Nameable;
+
+/**
+ *
```
vladak: should have some description of the purpose

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param s the check to remove
+     */
+    public void remove(AuthorizationEntity s) {
+        s.unload();
+        this.stack.remove(s);
+    }
+
+    /**
+     * Load all authorization checks in this stack.
```
vladak: "checks" should probably be replaced with different term.

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public boolean isRequisite() {
+        return REQUISITE.equals(this);
+    }
+
+    public boolean isSufficient() {
+        return SUFFICIENT.equals(this);
+    }
+
+    public static AuthControlFlag get(String flag) {
```
vladak: javadoc

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            return AuthControlFlag.valueOf(flag.toUpperCase());
+        } catch (IllegalArgumentException ex) {
+            // role does not exist -> add some more info about which roles do exist
```
vladak: role

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
```diff
@@ -0,0 +1,95 @@
...
+            throw new IllegalArgumentException(
+                    String.format("No authorization flag \"%s\", available flags are [%s]. %s",
```
vladak: control flag

src/org/opensolaris/opengrok/authorization/AuthorizationEntity.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @see IAuthorizationPlugin#unload()
+     */
+    abstract public void unload();
+
+    /**
+     * Test the given entity if it should be allowed with this authorization
+     * check.
+     *
+     * @param entity the given entity
```
vladak: project/group - the terminology clashes a bit here

src/org/opensolaris/opengrok/authorization/AuthorizationPlugin.java
```diff
@@ -0,0 +1,219 @@
...
+                || !getName().equals(plugin.getClass().getName())) {
+            return false;
+        }
+        if (hasPlugin()) {
+            unload();
+        }
+        try {
+            /**
+             * The exception should not happen here as we already have an
+             * instance of IAuthoriazationPlugin. But is is required by the
```
vladak: typo

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -0,0 +1,236 @@
...
+    public void remove(AuthorizationEntity s) {
+        s.unload();
+        this.stack.remove(s);
+    }
+
+    /**
+     * Load all authorization entities in this stack.
+     *
+     * <p>
+     * If the method is unable to load all the entities contained in this stack
```
vladak: does not match the code below + explain the rationale (why this is done)

src/org/opensolaris/opengrok/authorization/AuthorizationPlugin.java
tulinkry: fixed

src/org/opensolaris/opengrok/authorization/AuthorizationEntity.java
tulinkry: added a more detailed comment

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthControlFlag.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -0,0 +1,236 @@
...
+    public void add(AuthorizationEntity s) {
+        this.stack.add(s);
+    }
+
+    /**
+     * Remove the given authorization entity from this stack.
+     *
+     * @param s the entity to remove
+     */
+    public void remove(AuthorizationEntity s) {
```
vladak: how does the unload/remove work in general w.r.t. incoming requests ? If the `Configuration` is being reloaded and the authorization stack deconstructed what mechanism is used to ensure that the evaluation of the request waits until the stack is fully loaded ?

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        LOGGER.log(Level.FINEST, "Authorization for \"{0}\" => {1}",
+                new Object[]{entity.getName(), overallDecision ? "true" : "false"});
+        return overallDecision;
+    }
+
+    /**
+     * Set the plugin to all classes in this stack which requires this class in
+     * the configuration. This creates a new instance of the plugin for each
+     * class which needs it.
```
vladak: explain why this is necessary

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
tulinkry: changed both code and the comment + added the purpose

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
tulinkry: explained

vladak:
Other than the partial comments I had this looks good, feel free to merge.

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
tulinkry: I added a synchronized block over the stack replacement and changed couple of stuff:
 - the stack is copied from the configuration (not just the reference)

The reload is made like:
1. construct a new copy of the configuration stack
2. load plugins into this new copy
3. replace the old stack with the new stack in the synchronized block (this should be fast as it's just a reference swap)
4. free the old stack

tulinkry:
Check the new addition, it should be fine by now

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -455,22 +463,29 @@ public Object run() {
...
-        loadClasses(
+        loadClasses(newStack,
                 IOUtils.listFilesRec(pluginDirectory, ".class"),
                 IOUtils.listFiles(pluginDirectory, ".jar"));
 
         // fire load events
-        loadAllPlugins();
+        loadAllPlugins(newStack);
+
+        // replace the stack in a synchronized block
```
vladak: explain why synchonized is used

vladak:
Ok, merge then.

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: explained, waiting for the travis to finish

PR #1526, opened 2017-04-10
EXUBERANT_CTAGS -> OPENGROK_CTAGS

### LINKED ISSUE #1514: rename EXUBERANT_CTAGS env var
The `OpenGrok` shell script uses `EXUBERANT_CTAGS` which has 2 problems:
   - does not have the `OPENGROK_` suffix
  - OpenGrok supports Universal ctags as well so the name could be confusing

So, it should be renamed to something more fitting, e.g. `OPENGROK_CTAGS`

vladak:
fixes #1514


tarzanek:
lgtm, please merge

PR #1527, opened 2017-04-10
Easily discover groups in authorization plugins

tulinkry:
**This expects the #1525 to be merged as it already implements the new interface.**

This change moves the responsibility of a group discovery
 - which repositories are under this group (even indirectly)
 - which projects are under this group (even indirectly)
 - descendant groups
 - parent groups

to the opengrok code and the plugin can just call the new methods if it wants to add all subgroups/repositories/.. to the allowed set.

### DIFF
--- a/plugins/src/main/java/HttpBasicAuthorizationPlugin.java
+++ b/plugins/src/main/java/HttpBasicAuthorizationPlugin.java
@@ -81,6 +81,20 @@ public class HttpBasicAuthorizationPlugin implements IAuthorizationPlugin {
         }
     }
 
+    /**
+     * Add this group, all parent groups, all subgroups, all projects in this
+     * group, all repositories in this group, all projects in the subgroups and
+     * all repositories in the subgroups among the allowed entities for the
+     * authorization.
+     *
+     * <p>
+     * The purpose of this is when user allows a particular group then the
+     * expectation is to allow all included groups/projects/repositories.
+     * </p>
+     *
+     * @param group string name of the group to be discovered
+     * @param request the requests containing the user information
+     */
     private void discoverGroup(String group, HttpServletRequest request) {
         Group g;
         if ((g = Group.getByName(group)) != null) {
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/configuration/Group.java
```diff
@@ -134,6 +134,42 @@ public void addGroup(Group g) {
...
+
+    /**
+     * Collect all group's projects and repositories included in this group and
+     * in any subgroup.
+     *
+     * @return all collected projects and repositories
+     */
+    public Set<Project> getAllProjects() {
+        Set<Project> projectsTmp = new TreeSet<>();
+        for (Project p : getRepositories()) {
```
vladak: the `getRepositories()` name is confusing to me as it clashes with the `Repository` object.

tulinkry: I agree however I couldn't think of a better name for the projet-repository type (listed on the main page under the search form). Do you have better idea?

vladak: If I understand it correctly, the `repositories` contains projects with at least one `Repository` (according to `populateGroups()` so maybe it shall be named `projectsWithRepo` or such ?

plugins/src/main/java/HttpBasicAuthorizationPlugin.java
```diff
@@ -72,42 +72,25 @@ private void init(HttpServletRequest request) {
...
 
-            discoverGroup(group, request, descendants);
+            discoverGroup(group, request);
         }
-
-        userGroups.get(request.getUserPrincipal().getName()).addAll(descendants);
     }
 
-    private void discoverGroup(String group, HttpServletRequest request, Set<String> descendants) {
+    private void discoverGroup(String group, HttpServletRequest request) {
```
vladak: javadoc would be nice

src/org/opensolaris/opengrok/configuration/Group.java
tulinkry: filed #1553

plugins/src/main/java/HttpBasicAuthorizationPlugin.java
tulinkry: added

PR #1528, opened 2017-04-10
Two plugin classes implemented in OpenGrok - groups and projects restrictions

tulinkry:
**This expects the #1525 to be merged as it already implements the new interface.**

This change provides two simple plugins to statically enforce some authorization for groups or projects.

 - [GroupPlugin.java](https://github.com/tulinkry/OpenGrok/blob/0f16fa25ccb45e1412b1c1790bf3a96774334f65/src/org/opensolaris/opengrok/authorization/plugins/GroupPlugin.java)
 - [ProjectPlugin.java](https://github.com/tulinkry/OpenGrok/blob/0f16fa25ccb45e1412b1c1790bf3a96774334f65/src/org/opensolaris/opengrok/authorization/plugins/ProjectPlugin.java)

Both performing authorization for groups (or projects) based on statically initialized set of groups (projects) which are allowed. The initialization can be one of
 - a single string instance
 - a string array (the example bellow)
 - a list of strings

Example:
(note the `<string>1</string>` is real group name in my setup)

```xml
 <void property="name">
  <string>org.opensolaris.opengrok.authorization.plugins.GroupPlugin</string>
 </void>
 <void property="role">
  <string>SUFFICIENT</string>
 </void>
 <void property="setup">
    <void method="put">
      <string>groups</string>
      <array class="java.lang.String" length="2">
       <void index="0">
         <string>1</string>
       </void>
       <void index="1">
         <string>2</string>
       </void>
      </array>
    </void>
 </void>
</object
```

**There are also the changes for #1525 which I'll merge manually after the PR gets merged**

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -38,6 +38,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.servlet.http.HttpServletRequest;
 import org.opensolaris.opengrok.authorization.AuthorizationCheck.AuthControlFlag;
+import org.opensolaris.opengrok.authorization.plugins.GroupPlugin;
+import org.opensolaris.opengrok.authorization.plugins.ProjectPlugin;
 import org.opensolaris.opengrok.configuration.Configuration;
 import org.opensolaris.opengrok.configuration.Group;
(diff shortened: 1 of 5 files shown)


tulinkry:
A good idea could be to use also a blacklist, not only the whitelist

tulinkry:
With the new addition in #1560 this PR is quite meaningless

PR #1531, opened 2017-04-11
Configuration checks

tulinkry:
Approaches #1508 but it's not complete for the urls. However as stated in #1517 it would be nice if the urls did not include the hostname so the checks for it are not possible

For the
 - bugPattern it validates the pattern and if it contains at least one capture group
 - reviewPattern it validates the pattern and if it contains at least one capture group

PR #1532, opened 2017-04-11
logging exceptions in linkify + support urls without the hostname

### LINKED ISSUE #1508: Validate configuration for regexps/urls
When setting `userPage` or when setting some regexps in the configuration; the app should try to validate if the patterns/urls are valid and that they will not throw any unchecked exception. This is similar to the #1443 but it should be expanded to all problematic properities.

Also mentioned in [a comment here](https://github.com/OpenGrok/OpenGrok/pull/1506#discussion_r109160439).

### LINKED ISSUE #1515: log exceptions in linkifyPattern()
```
public static String linkifyPattern(String text, Pattern pattern, String name, String url) {
        try {
            String buildLink = buildLink(name, url, true);
            return pattern.matcher(text).replaceAll(buildLink);
        } catch (URISyntaxException | MalformedURLException ex) {
            return text;
        }
    }
```

the exception should be logged.

### LINKED ISSUE #1517: linkifyPattern() should accept not fully formed URIs
After the changes in cset 6611b858 such config stopped working:

```
  <void property="bugPage">
   <string>/cgi-bin/buglookup.pl?bug=</string>
  </void>
```

since `buildLink()` throws `MalformedURLException`.

To workaround it I used

```
  <void property="bugPage">
   <string>http://foo/cgi-bin/buglookup.pl?bug=</string>
  </void>
```

however I don't like it too much - having to include hostname makes the config less generic and cannot be moved between machines easily.

tulinkry:
fixes #1515
fixes #1517
also together with #1531 (**needs to be merged**) fixes #1508

### DIFF
--- a/src/org/opensolaris/opengrok/web/Util.java
+++ b/src/org/opensolaris/opengrok/web/Util.java
@@ -1513,7 +1513,48 @@ public final class Util {
             String buildLink = buildLink(name, url, true);
             return pattern.matcher(text).replaceAll(buildLink);
         } catch (URISyntaxException | MalformedURLException ex) {
+            LOGGER.log(Level.INFO, "The given url '{0}' is not valid", url);
             return text;
+        }
+    }
+
+    /**
+     * Try to complete the given url part into full url with server name, port,
+     * scheme, ...
+     * <dl>
+     * <dt>for request http://localhost:8080/source/xref/xxx and part
+     * /cgi-bin/user=</dt>
+     * <dd>http://localhost:8080/cgi-bin/user=</dd>
+     * <dt>for request http://localhost:8080/source/xref/xxx and part
+     * cgi-bin/user=</dt>
+     * <dd>http://localhost:8080/source/xref/xxx/cgi-bin/user=</dd>
+     * <dt>for request http://localhost:8080/source/xref/xxx and part
+     * http://users.com/user=</dt>
+     * <dd>http://users.com/user=</dd>
+     * </dl>
+     *
+     * @param url the given url part, may be already full url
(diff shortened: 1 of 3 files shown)


src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -1513,7 +1513,48 @@ public static String linkifyPattern(String text, Pattern pattern, String name, S
...
+                StringBuffer prepUrl = req.getRequestURL();
+                if (!url.isEmpty()) {
+                    prepUrl.append('/').append(url);
+                }
+                return new URI(prepUrl.toString()).toString();
+            }
+            return url;
+        } catch (URISyntaxException ex) {
+            LOGGER.log(Level.INFO,
+                    String.format("Unable to convert given url part '%s' to complete url", url),
```
vladak: url -> URL

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -1513,7 +1513,48 @@ public static String linkifyPattern(String text, Pattern pattern, String name, S
             String buildLink = buildLink(name, url, true);
             return pattern.matcher(text).replaceAll(buildLink);
         } catch (URISyntaxException | MalformedURLException ex) {
+            LOGGER.log(Level.INFO, "The given url '{0}' is not valid", url);
```
tulinkry: changed

src/org/opensolaris/opengrok/web/Util.java
tulinkry: changed

Issue #1519, opened 2017-04-05
convert project picker to pure jquery


vladak:
When the project picker is being loaded, the classical `select`  is displayed first, to be rewritten by the jquery version shortly afterwards. This creates a visual blip which is a bit disturbing. Given most of the pages require JS anyway, maybe converting it to pure jquery would be in order.

tulinkry:
After #1535 is merged consider using pure javascript for this picker as well. However there shouldn't be the ugly jump as the number of values in this picker is rather small (and does not change in any deployment)

PR #1535, opened 2017-04-18
adding a select picker to the file type

### LINKED ISSUE #1510: Type selector should use jquery
The Type selector should switch to jquery to match project picker and allow to search by typing.

tulinkry:
fixes #1510

screenshot from 2017-04-18 15-08-52

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -345,6 +345,17 @@ label {
     height: 100%;
 }
 
+/** file type selection container (next to the file type select box) */
+
+#type-select-container .sol-selected-display-item,
+#type-select-container .sol-results-count,
+#type-select-container .sol-selected-display-item-text{
+    margin-bottom: 0;
+    margin-right: 0;
+    padding-top: 0;
+    padding-bottom: 0;
+}
+
 #form-controls {
     white-space: nowrap;
     /* 11ex = width(#qtbl td:first-child) + margin-left(body) */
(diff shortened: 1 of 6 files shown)


vladak:
I wonder how this could work with multiple selected file projects.

tulinkry:
I'm curious as well. I guess that the backend is not implemented

vladak:
Can you file a new enhancement to track this ?

vladak:
how does it look like when no file type is selected ?

vladak:
Maybe consider an alternative: instead of displaying the selected type next to the selector, the selected type(s) can be displayed in the selector.

tulinkry:
> how does it look like when no file type is selected ?

screenshot from 2017-04-18 16-42-18

> Can you file a new enhancement to track this ?

1537

tulinkry:
> Maybe consider an alternative: instead of displaying the selected type next to the selector, the selected type(s) can be displayed in the selector.

I tried to play with this but it's not fairly simple with the select-picker we use. That could be the next step after this.

vladak:
Alright. Will this exhibit same behavior as described in #1519 ?

tulinkry:
Yes this is just a replacement for the old select which is hidden afterwards. However there are less values in the project picker so there should not be that ugly jump.

But with #1519 this should be possibly taken into account as well.

vladak:
Okay, merging.

Issue #1509, opened 2017-04-03
main page not valid HTML


tarzanek:
http://src.couchbase.org/source/

https://validator.w3.org/check?uri=http%3A%2F%2Fsrc.couchbase.org%2Fsource%2F&charset=%28detect+automatically%29&doctype=Inline&group=0&No200=1&user-agent=W3C_Validator%2F1.3+http%3A%2F%2Fvalidator.w3.org%2Fservices

the div and js are not well formed

tulinkry:
and for the second part it seems to me that the extra `<div>` element was added in cset dc2627429268b6ece694374ed41014269dd8a976 by mistake as a part of pr #1342. @vladak can you please review this again? it seems to me that

```html
        <div id="results">
            <%= PageConfig.get(request).getEnv().getConfiguration().getBodyIncludeFileContent() %>
        </div> <!-- THIS ONE IS EXTRA -->
```

this `<div>` can be safely removed to make the main page valid again

vladak:
Good catch, the terminating div indeed seems superfluosly added.

PR #1536, opened 2017-04-18
adding a container for javascripts and write them out in the footer

tulinkry:
approaches #1509 (main page is still invalid because of a traling `<div>`)


web/httpheader.jspf
```diff
@@ -66,6 +67,13 @@ org.opensolaris.opengrok.web.PageConfig"
     <link rel="stylesheet" type="text/css" href="<%=styleDir%>/jquery.combo.css%>" />
 --%>
 <%
+    cfg.addScript(ctxPath + "/js/jquery-3.2.0.min.js", 0);
```
vladak: If I understand this correctly, the intention is to assemble a list of scripts and write them to the page in order according to priority. It seems that the ordering is known beforehand so instead of using the hand-tuned priority values, it might be better to have the list (sans versions) hardcoded in the correct order in the `Scripts` class and just record what scripts are necessary for given JSP (with `addScript()`) and them write them out in `getScripts()`)

tulinkry: Could be better solution as we dont need anything extra generic for our low number of js files. Will try.

tulinkry:
Reworked to contain the predefined list and using shortcuts from jsps.

src/org/opensolaris/opengrok/web/Scripts.java
```diff
@@ -0,0 +1,238 @@
...
+ * @author Krystof Tulinger
+ */
+public class Scripts implements Iterable<Scripts.Script> {
+
+    /**
+     * A script wrapper.
+     */
+    static abstract public class Script {
+
+        protected String script;
```
vladak: scriptName ?

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1185,6 +1189,45 @@ public String getUriEncodedPath() {
...
+     * Add a new file script to the page.
+     *
+     * @param path web path to the file
+     * @param priority script priority (scripts are sorted based on the
+     * priority)
+     * @return this
+     *
+     * @see Scripts.FileScript#FileScript(java.lang.String, int)
+     */
+    public PageConfig addScript(String path, int priority) {
```
vladak: is it necessary to expose the priority ?

test/org/opensolaris/opengrok/web/ScriptsTest.java
```diff
@@ -0,0 +1,138 @@
...
+                        + " data-priority=\"0\"></script>"));
+        assertTrue(scripts.toHtml()
+                .contains("<script type=\"text/javascript\""
+                        + " src=\"http://example.com/main3.js\""
+                        + " data-priority=\"0\"></script>"));
+    }
+
+    @Test
+    public void testLookup() {
+        scripts.addScript("", "utils");
```
vladak: would the test benefit from having non-empty context path ?

src/org/opensolaris/opengrok/web/Scripts.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Scripts which will be written to the page.
+     */
+    private final List<Script> outputScripts = new ArrayList<>();
+
+    /**
+     * Sort the page scripts according to the priority.
+     */
+    public void sort() {
+        outputScripts.sort((a, b) -> a.getPriority() - b.getPriority());
```
vladak: I wonder if there is a way how to make the Scripts self-sorting - i.e. without having to call `sort()` in `toHtml()` and elsewhere.

src/org/opensolaris/opengrok/web/Scripts.java
tulinkry: for the inline script this is not really correct, changed to `scriptData`

src/org/opensolaris/opengrok/web/PageConfig.java
tulinkry: no it's not yet

test/org/opensolaris/opengrok/web/ScriptsTest.java
tulinkry: added a test case for non-empty context path

tulinkry: The difference is in the expectations:
```java
scripts.toHtml()
        .contains("<script type=\"text/javascript\""
                + " src=\"" + contextPath + '/' + s.getValue().getScriptData() + "\""
                //            ^^^^^^^^^^^^^^^^^
                + " data-priority=\"" + s.getValue().getPriority() + "\"></script>"));
 ```

File: src/org/opensolaris/opengrok/web/Scripts.java (hunk 5 above)
Comment (tulinkry) [inline src/org/opensolaris/opengrok/web/Scripts.java]: changed

Scripts are added into the array in a sorted manner using `binarySearch` to find the insertion point

This is not the most efficient way (`O(n)`) however most readable (than using `Map<int, List<Script>>`) and for small number of scripts (which is expected) this does not really matter

File: src/org/opensolaris/opengrok/web/Scripts.java [hunk 6]
```diff
@@ -40,18 +42,18 @@
      */
     static abstract public class Script {
 
-        protected String script;
+        protected String scriptData;
```
vladak: pls add a comment about what kind of data this member can hold

src/org/opensolaris/opengrok/web/Scripts.java
```diff
@@ -204,24 +199,45 @@ public boolean isEmpty() {
...
+        int index = Collections.binarySearch(outputScripts, script, new Comparator<Script>() {
+            @Override
+            public int compare(Script a, Script b) {
+                return a.getPriority() - b.getPriority();
+            }
+        });
+        if (index < 0) {
+            /**
+             * Key is not found in the list the index equals to -(insertion
+             * index) - 1.
```
vladak: pls keep the formula on single line

vladak:
pls fix the remaining nit(s) and merge

src/org/opensolaris/opengrok/web/Scripts.java
tulinkry: added an artifitial line break

src/org/opensolaris/opengrok/web/Scripts.java
tulinkry: added

PR #1540, opened 2017-04-21
checking if the line element taken from the hash exists in the page

### LINKED ISSUE #1495: js error in console when configuration is missing and url has # in it
FF console when accessing http://localhost:8080/source/xref/OpenGrok/testdata/sources/c/myclass.cc#12 and the real source root directory does not exist.

```log
TypeError: $(...).offset(...) is undefined[Learn More]  utils-0.0.7.js:281:34
```

tulinkry:
fixes #1495

### DIFF
--- a/web/js/utils-0.0.9.js
+++ b/web/js/utils-0.0.9.js
@@ -275,14 +275,22 @@
                     return
    
                 var lines = inner.getLinesParts(window.location.hash);
-                if(lines.length > 0) {
-                   var line = lines[0] // first line
-                   $("#content").animate({
-                      scrollTop: $(inner.format(inner.options.linkSelectorTemplate, {
-                          parent: inner.options.parent,
-                          n: line
-                      })).offset().top - $(inner.options.parent).offset().top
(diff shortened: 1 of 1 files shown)


web/js/utils-0.0.9.js
```diff
@@ -275,14 +275,21 @@
...
-                      })).offset().top - $(inner.options.parent).offset().top
-                   }, inner.options.autoScrollDuration);
+                if (lines.length > 0) {
+                    var line = lines[0]; // first line
+                    var $line = $(inner.format(inner.options.linkSelectorTemplate, {
+                        parent: inner.options.parent,
+                        n: line
+                    }));
+                    // there is no such element in the page where we should scroll to
+                    if ($line.length > 0) {
```
tulinkry: This is the actual change

vladak: The comment reads strange - should it say 'if there is such element...' ?

tulinkry: changed

PR #1558, opened 2017-04-28
adding basic stats for authorization decisions

tulinkry:
this will provide some info measurements for the authorization performance

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -46,6 +46,7 @@ import org.opensolaris.opengrok.configuration.Project;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.util.IOUtils;
+import org.opensolaris.opengrok.web.Statistics;
 
 /**
  * Placeholder for performing authorization checks.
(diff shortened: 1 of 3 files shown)


src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -599,18 +600,28 @@ private boolean checkAll(HttpServletRequest request, String cache, Nameable enti
...
             // cache hit
+            stats.addRequest(request, "authorization_cache_hits");
             return val;
         }
 
+        stats.addRequest(request, "authorization_cache_misses");
+
+        long time = System.currentTimeMillis();
+
         boolean overallDecision = performCheck(entity, predicate);
```
vladak: I'd add this to the stats too. Might be good to detect brute-forcing attacks.

tulinkry: added

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -599,18 +600,28 @@ private boolean checkAll(HttpServletRequest request, String cache, Nameable enti
...
 
+        stats.addRequestTime(request, "authorization", System.currentTimeMillis() - time);
+        stats.addRequestTime(request, "authorization_of_" + entity.getName(), System.currentTimeMillis() - time);
```
vladak: these 2 lines should probably use the same time value

tulinkry: changed that already

PR #1560, opened 2017-04-28
targeting stacks for particular groups or projects

tulinkry:
Allow user to specify target groups/projects for particular stack (or even a plugin).

When the authorization traverses the stack of plugins/stacks it checks if the plugin/stack is targeted only for some groups/projects and does not execute if it that is not the case.

The syntax is pretty straight-forward (note the section forGroups):

```xml
<void method="add">
        <void property="forGroups"> <!-- or forProjects -->
             <string>Group 1</string>
        </void>
        <!-- or for more groups -->
        <void property="forGroups"> <!-- or forProjects -->
              <void method="add">
                   <string>Group 1</string>
              </void>
              <void method="add">
                   <string>Group 2</string>
              </void>
        </void>
        <void property="name">
            <string>opengrok.auth.plugin.SomePlugin</string>
        </void>
        <void property="flag">
            <string>REQUISITE</string>
        </void>
                <void property="setup">
                        <void method="put">
                                <string>config</string>
                                <string>/home/my.xml</string>
                        </void>
                 </void>
    </object>
</void>
```

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -35,7 +35,6 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.function.Predicate;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.logging.Level;
@@ -145,10 +144,25 @@ public final class AuthorizationFramework {
                 request,
                 "plugin_framework_project_cache",
                 project,
-                new Predicate<IAuthorizationPlugin>() {
+                new AuthorizationEntity.PluginDecisionPredicate() {
             @Override
-            public boolean test(IAuthorizationPlugin plugin) {
+            public boolean decision(IAuthorizationPlugin plugin) {
                 return plugin.isAllowed(request, project);
+            }
+        }, new AuthorizationEntity.PluginSkippingPredicate() {
+            @Override
+            public boolean shouldSkip(AuthorizationEntity entity) {
+                // shouldn't skip if there is no setup
+                if (entity.forProjects().isEmpty() && entity.forGroups().isEmpty()) {
+                    return false;
+                }
+
+                // shouldn't skip if the project is contained in the setup
+                if (entity.forProjects().contains(project.getName())) {
+                    return false;
+                }
+
+                return true;
             }
         });
     }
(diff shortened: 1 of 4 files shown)


src/org/opensolaris/opengrok/authorization/AuthorizationPlugin.java
```diff
@@ -158,11 +165,15 @@ public synchronized void unload() {
      * @see IAuthorizationPlugin#isAllowed(HttpServletRequest, Group)
      */
     @Override
-    public boolean isAllowed(Nameable entity, Predicate<IAuthorizationPlugin> predicate) {
+    public boolean isAllowed(Nameable entity,
+            AuthorizationEntity.PluginDecisionPredicate pluginPredicate,
+            AuthorizationEntity.PluginSkippingPredicate skippingPredicate) {
```
vladak: should check the skippingPredicate here ?

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -145,11 +144,26 @@ public boolean isAllowed(HttpServletRequest request, Project project) {
...
-                new Predicate<IAuthorizationPlugin>() {
+                new AuthorizationEntity.PluginDecisionPredicate() {
             @Override
-            public boolean test(IAuthorizationPlugin plugin) {
+            public boolean decision(IAuthorizationPlugin plugin) {
                 return plugin.isAllowed(request, project);
             }
+        }, new AuthorizationEntity.PluginSkippingPredicate() {
+            @Override
+            public boolean shouldSkip(AuthorizationEntity entity) {
```
vladak: would be more readable if `entity` was called `authEntity` or such. similarly for the other `isAllowed()` method

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -579,7 +604,7 @@ public void setPluginVersion(int pluginVersion) {
      * <p>
      * The order of plugin invokation is given by the configuration
      * {@link RuntimeEnvironment#getPluginStack()} and appropriate actions are
-     * taken when traversing the list with set of keywords, such as:</p>
+     * taken when traversing the stack with set of keywords, such as:</p>
      *
      * <h4>required</h4>
```
vladak: is the explanation of the keywords duplicate from the enum code ?

tulinkry: yes

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: changed

src/org/opensolaris/opengrok/authorization/AuthorizationPlugin.java
tulinkry: actually no, I added a comment

```java
        /**
         * We don't check the skippingPredicate here as this instance is
         * <b>always</b> a part of some stack (may be the default stack) and the
         * stack checks the skipping predicate before invoking this method.
         *
         * @see AuthorizationStack#processStack
         */
```

PR #1563, opened 2017-05-02
using statistics with synchronized methods

tulinkry:
We may loose some information without the synchronization when there are more parallel requests.

### DIFF
--- a/src/org/opensolaris/opengrok/web/Statistics.java
+++ b/src/org/opensolaris/opengrok/web/Statistics.java
@@ -67,7 +67,7 @@ public class Statistics {
     /**
      * Adds a single request into all requests.
      */
-    public void addRequest(HttpServletRequest req) {
+    synchronized public void addRequest(HttpServletRequest req) {
         maybeRefresh();
 
         requestsPerMinute++;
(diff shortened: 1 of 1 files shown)


vladak:
As a stop gap measure this should work however I wonder if `AtomicInteger` would do the job too ?

tulinkry:
More like `AtomicLong`. It would do the job too I'm just a little worried about a map of (quite a lots of) `AtomicLong` values. I've never experimented with it

vladak:
I'd not worry about it. After all, the atomic types just use the underlying CPU operations making them a bit more expensive than regular types w.r.t. arithmetics.

tulinkry:
Ok, I changed that.

src/org/opensolaris/opengrok/web/Statistics.java
```diff
@@ -103,11 +126,11 @@ protected void maybeRefresh() {
      * @param category category
      */
     public void addRequest(HttpServletRequest req, String category) {
-        Long val = requestCategories.get(category);
+        AtomicLong val = requestCategories.get(category);
         if (val == null) {
-            val = new Long(0);
+            val = new AtomicLong(0);
```
vladak: Not sure this would work. If 2 threads are racing for this code while `val` is still `null`, both can read `val` as `null` so both will create new instance and 1 count will be lost.

tulinkry: The design with the synchrinized blocks/methods was simpler in my opinion than this

vladak: yep, certainly more robust. For the simple counters, the `atomic` types would work, for the maps and complex statements (like `maybeRefresh()`) probably not so much.

vladak:
Let's merge this with `synchronized`, can always worry about performance later on.

PR #1565, opened 2017-05-02
more efficient locking for shared resource with more readers

tulinkry:
The scenario before needlessly slowed the authorization process as it exclusively allowed only one thread processing the stack of plugins.

The problem with the synchronization here is that we don't want to concurrently share the stack when there is a reload of plugins while there is some authorization in progress. ReadWrite lock is more accurate for this situation.

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -33,6 +33,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Predicate;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -74,6 +76,11 @@
     AuthorizationStack stack;
 
     /**
+     * Lock for safe reloads.
+     */
+    private ReadWriteLock lock = new ReentrantReadWriteLock();
```
vladak: explain why it needs to be reentrant

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -646,6 +657,11 @@ private boolean checkAll(HttpServletRequest request, String cache, Nameable enti
      * @return true if entity is allowed; false otherwise
      */
     private synchronized boolean performCheck(Nameable entity, Predicate<IAuthorizationPlugin> predicate) {
```
vladak: the `synchronized` should still be present ?

tulinkry: no it should not, that is my merge problem

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: because that is the only class implementing the `ReadWriteLock` interface

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: fixed

PR #1574, opened 2017-05-04
obfuscating sensitive svn login credentials when running subcommands

### LINKED ISSUE #1561: User/password printed during indexing
During indexing the user and the password are printed to the console.

505       9205 11803 15 07:53 ?        00:00:00 /usr/bin/svn log --non-interactive --username XXXXX --password XXXX --trust-server-cer

This information is available to any user connected to the machine by the command pf -ef. Is there any way to avoid this?

tulinkry:
fixes #1561


tarzanek:
Krystof, I understood this is not only about our logs, but also about ps -ef process tree
so basically we should avoid the command this way and rather find a way to run the command with password/login in a file ... (or whatever safer way svn provides ... )

tulinkry:
Ok, that is my misunderstanding. There was a typo in the command.

PR #1591, opened 2017-05-12
display default 403 forbidden page when there is no customization

### LINKED ISSUE #1585: Display default eforbidden page when there is no customization
In connection to #1575 when the user does not specify his own error page, some default should be displayed. Currently empty page is displayed.

tulinkry:
fixes #1585

### DIFF
--- a/web/WEB-INF/web.xml
+++ b/web/WEB-INF/web.xml
@@ -200,10 +200,6 @@
     <error-code>500</error-code>
     <location>/error</location>
   </error-page>
-  <error-page>
-      <error-code>403</error-code>
-      <location>/eforbidden</location>
-  </error-page>
   <jsp-config>
       <jsp-property-group>
       <url-pattern>*.jsp</url-pattern>
--- a/web/eforbidden.jsp
+++ b/web/eforbidden.jsp
@@ -17,14 +17,15 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
---%><%@page  session="false" isErrorPage="true" import="
+--%><%@page  session="false" import="
              org.opensolaris.opengrok.web.PageConfig"
 %><%
-    /* ---------------------- eforbidden.jspf start --------------------- */
-    {
+/* ---------------------- eforbidden.jspf start --------------------- */
+{
+    response.setStatus(HttpServletResponse.SC_FORBIDDEN);
 %>
-<%= PageConfig.get(request).getEnv().getConfiguration().getForbiddenIncludeFileContent()%>
+<%= PageConfig.get(request).getEnv().getConfiguration().getForbiddenIncludeFileContent() %>
 <%
-    }
-    /* ---------------------- eforbidden.jspf end --------------------- */
+}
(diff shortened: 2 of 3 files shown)


web/WEB-INF/web.xml
```diff
@@ -200,10 +200,6 @@
     <error-code>500</error-code>
     <location>/error</location>
   </error-page>
-  <error-page>
-      <error-code>403</error-code>
-      <location>/eforbidden</location>
-  </error-page>
```
tulinkry: this is necessary otherwise it would **always** trigger the customized page

web/WEB-INF/web.xml
```diff
@@ -200,10 +200,6 @@
...
   <jsp-config>
       <jsp-property-group>
       <url-pattern>*.jsp</url-pattern>
```
vladak: why is this needed ?

vladak:
what would be actually displayed with no custom config ?

web/WEB-INF/web.xml
tulinkry: https://github.com/OpenGrok/OpenGrok/pull/1591#discussion_r116225626

web/eforbidden.jsp
```diff
@@ -17,14 +17,15 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
---%><%@page  session="false" isErrorPage="true" import="
```
vladak: why `isErrorPage` is no longer needed ?

web/eforbidden.jsp
```diff
@@ -17,14 +17,15 @@ information: Portions Copyright [yyyy] [name of copyright owner]
...
+--%><%@page  session="false" import="
              org.opensolaris.opengrok.web.PageConfig"
 %><%
-    /* ---------------------- eforbidden.jspf start --------------------- */
-    {
+/* ---------------------- eforbidden.jspf start --------------------- */
+{
+    response.setStatus(HttpServletResponse.SC_FORBIDDEN);
 %>
 <%= PageConfig.get(request).getEnv().getConfiguration().getForbiddenIncludeFileContent()%>
```
vladak: nit: missing space at the end of function call

web/WEB-INF/web.xml
vladak: Oh, I meant the `*.jsp` part however that is pretty standard stuff.

tulinkry:
> what would be actually displayed with no custom config ?

screenshot from 2017-05-29 15-01-47

And the custom page if there is one.

web/eforbidden.jsp
tulinkry: fixed

web/WEB-INF/web.xml
tulinkry: that's just the way how to add the new servlet page into the request lifecycle

web/eforbidden.jsp
tulinkry: I removed the definition that the eforbidden is a handler for 403 errors; in order not to confuse anybody I removed the declaration from this place as well.

This parameter in jsp means that there will be an `exception` object available for this jsp which caused the error which led to this jsp.

Issue #1569, opened 2017-05-03
add new message to trigger authorization plugin reload


vladak:
Currently the only way how to reload plugin configuration is to set new configuration. This can take long time due to `invalidateRepositories()` being called. New message directed just to reload auth plugin config would be nice to have, maybe with a possibility to specify concrete plugin type to be reloaded.

tulinkry:
started thinking about this

> maybe with a possibility to specify concrete plugin type to be reloaded.

and I'm not sure if this is really a gain because our plugins use session to store some metadata and reloading the plugin will not necessarily invalidate the session; so this would mean to increase the plugin generation for all of the plugins - which eventually leads to that all plugins revalidate their session however this is not a single plugin reload

vladak:
Okay, ditch the per-plugin idea.

vladak:
Given the authorization framework refresh Message was implemented in c36f8e6106d16c5440e24746977aa2beec3c8f8b I think this one can be closed.

PR #1592, opened 2017-05-12
adding a tag to the config message to reload all authorization plugins

tulinkry:
approaches #1569


vladak:
looks good, pls resolve and merge

PR #1600, opened 2017-05-22
'media_query_support'

### LINKED ISSUE #1370: move project selection list and buttons above the other search-text fields

### LINKED ISSUE #1521: Project selector should be placed based on screen resolution
As part of fix for issue 1370, project selector was moved above the query table. The change might seem fine for tablet and smart phone users but it does not look nice on big screens. Seems left side of the screen is cluttered where as there are lots of space available on the right. Previous layout of showing project selector on right was good for big screens. 

It would be nice to to change the layout dynamically based on screen resolution. If the screen resolution is greater than certain size, show the project selector on right. If the screen resolution is small, then show the project selector on top of query table.

yrdivya:
This fixes #1521. I have reverted the changes done as part of #1370. Added media query support for the same. Until the width of the screen is 770px, the UI will look the same way it did that is before #1370. If the width is lesser than 770px, the UI will behave the same way as it does after the fix #1370. 

Fix: I am reordering the qtbl, ptbl divs based on the resolution to achieve this.

### DIFF
--- a/web/menu.jspf
+++ b/web/menu.jspf
@@ -49,6 +49,7 @@ org.opensolaris.opengrok.web.Util"
 /* ---------------------- menu.jspf start --------------------- */
 {
     PageConfig cfg = PageConfig.get(request);
+    boolean isResponsive = cfg.isResponsiveUI();
     ProjectHelper ph = ProjectHelper.getInstance(cfg);
     JSONArray messages;
     Set<Project> projects = ph.getAllProjects();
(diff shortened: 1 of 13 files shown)


web/menu.jspf
```diff
@@ -71,24 +71,153 @@ org.opensolaris.opengrok.web.Util"
             pMap.put(name, esc);
         }
 %>
+<script type="text/javascript" src="<%= request.getContextPath() %>/js/jquery-3.2.0.min.js"></script>
```
tarzanek: loading of  scripts happens at the end in footer, so this is not needed

tarzanek:
Are your changes in menu.jspf w3c  valid?
https://validator.w3.org/
thnx
L

web/menu.jspf
```diff
@@ -71,24 +71,153 @@ org.opensolaris.opengrok.web.Util"
...
-    document.domReady.push(function() { domReadyMenu(); });
+    document.xrefPath = '<%= request.getContextPath() + Prefix.XREF_P
+    %>';document.domReady.push(function() { domReadyMenu(); });
 /* ]]> */
+var initialWidth = null;
+var isFirst = true;
+var checkWidth = function(){
+    var width = (window.innerWidth > 0) ? window.innerWidth : screen.width;
+	console.log(width);	
+	if(width <= 770) {
```
vladak: it would be really nice if the default value could be configurable

web/menu.jspf
```diff
@@ -71,24 +71,153 @@ org.opensolaris.opengrok.web.Util"
...
+		//$(detach).insertAfter($(this).find(".teams"));
+		//})
+		$("#sbox").each(function() {
+		var detach = $(this).find("#qtbl").detach();
+		$(detach).insertAfter($(this).find("#ptbl"));
+		var detach1 = $(this).find("#ltbl").detach();
+		$(detach1).insertBefore($(this).find("#qtbl"));
+		});
+		console.log('width changed');
+	}else{
```
vladak: please keep the coding style aligned with the rest of the file, in particular the white space next to keywords.

web/menu.jspf
vladak: the default value should be configurable. In our environment we'd really like to preserve the behavior as of #1370

web/menu.jspf
```diff
@@ -71,24 +71,153 @@ org.opensolaris.opengrok.web.Util"
...
+		//revert back
+		$("#sbox").each(function() {
+		var detach = $(this).find("#qtbl").detach();
+		$(detach).insertBefore($(this).find("#ptbl"));
+		});	
+	}	
+	initialWidth = width;
+};
+$(document).ready(function(){
```
tulinkry: Also I'm affraid that this wont work when jquery is loaded at the bottom of the page

web/menu.jspf
```diff
@@ -71,24 +71,153 @@ org.opensolaris.opengrok.web.Util"
             pMap.put(name, esc);
         }
 %>
+<script type="text/javascript" src="<%= request.getContextPath() %>/js/jquery-3.2.0.min.js"></script>
 <script type="text/javascript">/* <![CDATA[ */
-    document.domReady.push(function() { domReadyMenu(); });
+    document.xrefPath = '<%= request.getContextPath() + Prefix.XREF_P
+    %>';document.domReady.push(function() { domReadyMenu(); });
 /* ]]> */
+var initialWidth = null;
```
tulinkry: I think this isn't really necessary to achieve the result. Combination of HTML markup duplication and hiding and showing based on css media queries looks cleaner (and perhaps faster than this).

web/menu.jspf
yrdivya: I am thinking of moving the methods added in menu.jspf to utils-0.0.9.js. In that case, since the util file is loaded after jquery, it have no issues. Would that be ok?

web/menu.jspf
tulinkry: You can duplicate the qtbl ptbl and ltbl in the markup (and use something different than "id" cause it wont be unique anymore) and then hide the ones you dont want (using display: none) and make visible the ones you want (display: block) and the other way round inside some media query

so its not a real "reordering" it is rather showing/hiding the parts you want/dont want to see

yrdivya:
I have made the responsiveness as configurable. As a default, this is turned off. When turned off, it will display the UI as per #1370. I also tried to make the width configurable. But this messes up with UI in lower resolutions.

mrashmiranjan:
Any chance of the new change getting reviewed and merged to the master ?

vladak:
@tulinkry can you review this ?

tulinkry:
Yes, I need to play with this but I'm currently fixing #1488

tulinkry:
I played with this a bit; it appears that the most effective (and also simple to code) way is to use the javascript in the suggested way. However I have some comments to this.

Main problems:
 - changing lines which arent related to this change (imports, relative imports)
 - uniform code style with the rest of our project
 - i dont prefer the "responsiveUI" over the "3columnslayout"
 - cleanup the javascript part

When using -Z off the page (should look as it used to look till now) and it looks like this:
screenshot from 2017-07-10 17-06-59
which is far away from the current state:
screenshot from 2017-07-10 17-08-38

web/menu.jspf
tulinkry: yes

web/menu.jspf
tulinkry: the width is not really important to be configurable, rather the value if using the "old" and "new" design is important

vladak:
No response for almost 2 years, closing.

PR #1604, opened 2017-05-22
removing singleton pattern from authorization framework

### LINKED ISSUE #1595: get rid of the singleton on AuthorizationFramework

tulinkry:
fixes #1595

the tests should be ok when #1603 is merged.


vladak:
what's the advantage of doing this ?

tulinkry:
1. reduce number of singletons in the application (we need just the RE)
2. easier testing and instantiating the authorization framework with no secret dependency (this is not quite true since there is still a dependency on RE in `reload()`)

vladak:
I see, especially the call to ` AuthorizationFramework.getInstance()` in WebappListener to get the framework to load is inconspicious/unclear.

PR #1613, opened 2017-05-29
improve logging of 403 HTTP errors

### LINKED ISSUE #1611: improve doFilter() logging
I see entries like this in the Tomcat logs:

```
23-May-2017 23:09:42.081 INFO [http-nio-8888-exec-96] org.opensolaris.opengrok.web.AuthorizationFilter.doFilter access denied for uri: /source/xref/foo/
```

Ideally it should log the user identity.

Also, `uri` should be spelled in capital lettes.

tulinkry:
fixes #1611

### DIFF
--- a/src/org/opensolaris/opengrok/web/AuthorizationFilter.java
+++ b/src/org/opensolaris/opengrok/web/AuthorizationFilter.java
@@ -54,18 +54,23 @@ public class AuthorizationFilter implements Filter {
 
         Project p = config.getProject();
         if (p != null && !config.isAllowed(p)) {
-            LOGGER.log(Level.INFO, "access denied for uri: {0}", httpReq.getRequestURI());
+            if (httpReq.getRemoteUser() != null) {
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/web/AuthorizationFilter.java
```diff
@@ -54,13 +54,19 @@ public void doFilter(ServletRequest sr, ServletResponse sr1, FilterChain fc) thr
 
         Project p = config.getProject();
         if (p != null && !config.isAllowed(p)) {
-            LOGGER.log(Level.INFO, "access denied for uri: {0}", httpReq.getRequestURI());
+            if (httpReq.getRemoteUser() != null) {
+                LOGGER.log(Level.INFO, "Access denied for user {0} for URI: {1}",
```
vladak: possibly enclosing the user identity in quotes would help if it has some long/strange value

tulinkry: fixed

tulinkry:
fixed and conficts resolved

PR #1616, opened 2017-05-29
perform any necessary action when the configuration is read from a file

### LINKED ISSUE #1609: pluginDirectory null even if specified in 1.1-rc4
Was running 1.1-rc3 with non-empty authorization config. Then upgraded to 1.1-rc4, deployed the webapp and Tomcat logs contained this:

```
23-May-2017 22:22:52.390 INFO [localhost-startStop-5] org.apache.catalina.startup.HostConfig.deployWAR Deploying web application archive /var/tomcat8/webapps/source.war
23-May-2017 22:22:53.062 WARNING [localhost-startStop-5] org.opensolaris.opengrok.authorization.AuthorizationFramework.reload Plugin directory not found or not readable: null. All 
requests allowed.
```

although `pluginDirectory` is defined and valid in `configuration.xml`.

tulinkry:
fixes #1609

Be aware that this still issues the first warning that the plugin directory is null (when env is instantiated) and after that it sets the directory and reloads the plugins.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -132,9 +132,9 @@ public final class RuntimeEnvironment {
     private Statistics statistics = new Statistics();
 
     /**
-     * Instance of authorization framework. Allowing every request by default.
+     * Instance of authorization framework.
      */
-    private AuthorizationFramework authFramework = new AuthorizationFramework(null);
+    private AuthorizationFramework authFramework;
 
     /* Get thread pool used for top-level repository history generation. */
     public static synchronized ExecutorService getHistoryExecutor() {
@@ -1552,7 +1552,7 @@ public final class RuntimeEnvironment {
      * @see AuthorizationFramework#getPluginVersion()
      */
     public int getPluginVersion() {
-        return authFramework.getPluginVersion();
+        return getAuthorizationFramework().getPluginVersion();
     }
 
     /**
@@ -1560,7 +1560,10 @@ public final class RuntimeEnvironment {
      *
      * @return the framework
      */
-    public AuthorizationFramework getAuthorizationFramework() {
+    synchronized public AuthorizationFramework getAuthorizationFramework() {
+        if (authFramework == null) {
+            authFramework = new AuthorizationFramework(threadConfig.get().getPluginDirectory());
+        }
         return authFramework;
     }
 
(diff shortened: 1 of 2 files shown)


vladak:
The double messaging is what I was afraid of. Is there a way how to make it better or get rid of it completely ?

vladak:
Also, does this mean that there will be a window where requests will be allowed before the configuration is set ? This has to be avoided.

tulinkry:
> The double messaging is what I was afraid of. Is there a way how to make it better or get rid of it completely ?

I'm sure there is.

> Also, does this mean that there will be a window where requests will be allowed before the configuration is set ? This has to be avoided.

Not really because this code is invoked **before** any request and the container waits sequentially after this method is finished.

tulinkry:
I change the code so that the framework is instantiated after the configuration value is known.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1561,6 +1561,9 @@ public int getPluginVersion() {
      * @return the framework
      */
     public AuthorizationFramework getAuthorizationFramework() {
```
vladak: I wonder if this method should be synchronized ?

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -74,6 +75,13 @@ public void contextInitialized(final ServletContextEvent servletContextEvent) {
             }
         }
 
+        /**
+         * Create a new instance of authorization framework. If the code above
+         * (reading the configuration failed) then the plugin directory is
```
vladak: the 'failed' should be outside of the parens, I believe.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: further, what about getAuthorizationFramework()/setAuthorizationFramework() races ?

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: fixed

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
tulinkry: I think that this won't occur in our setup however yes that could cause a problem. I marked them as synchronized.

PR #1617, opened 2017-05-29
removing the size for a directory in directory listing

### LINKED ISSUE #1552: consider suppressing length for directories in directory listing
`printDateSize()` in `DirectoryListing.java` contains this curious piece of commented out code:

```
84          // if (isDir) {
85              out.write(Util.readableSize(child.length()));
86          // }
```

Thing is, the `length()` value for directories is unspecified. On some systems it is a reference count, on others some different value. Also, I think in the case of simplified directory listing wrong directory might be taken for `length()`. So we should consider not writing the length for directories.

tulinkry:
fixes #1552

Replaced with '-'.

screenshot from 2017-05-29 15-56-30
screenshot from 2017-05-29 15-56-36

Tablesorters return 0 for all directories (unparseable size):
```
        if (parts === null || parts.length < 3) {
            return 0;
        }
```

### DIFF
--- a/test/org/opensolaris/opengrok/web/DirectoryListingTest.java
+++ b/test/org/opensolaris/opengrok/web/DirectoryListingTest.java
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
-/*
+ /*
  * Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.web;
@@ -28,6 +28,7 @@ import java.io.FileOutputStream;
 import java.io.StringWriter;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.List;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import org.junit.After;
@@ -50,31 +51,46 @@ import static org.junit.Assert.*;
  */
 public class DirectoryListingTest {
 
+    private static int DIRECTORY_INTERNAL_SIZE = -2;
+
     private File directory;
     private FileEntry[] entries;
     private SimpleDateFormat dateFormatter;
 
     class FileEntry implements Comparable {
+
         String name;
         String href;
         long lastModified;
-        String size;
+        int size;
+        List<FileEntry> subdirs;
 
         FileEntry() {
             dateFormatter = new SimpleDateFormat("dd-MMM-yyyy");
         }
 
-        FileEntry(String name, String href, long lastModified, String size) {
+        FileEntry(String name, String href, long lastModified, int size, List<FileEntry> subdirs) {
             this.name = name;
             this.href = href;
             this.lastModified = lastModified;
             this.size = size;
+            this.subdirs = subdirs;
         }
 
         private void create() throws Exception {
             File file = new File(directory, name);
-            if (!file.exists()) {
-                assertTrue("Failed to create file", file.createNewFile());
+
+            if (subdirs != null && subdirs.size() > 0) {
+                // this is a directory
+                assertTrue("Failed to create a directory", file.mkdirs());
+                for (FileEntry entry : subdirs) {
+                    entry.name = name + File.separator + entry.name;
+                    entry.create();
+                }
+            } else {
+                if (!file.exists()) {
+                    assertTrue("Failed to create file", file.createNewFile());
+                }
             }
 
             long val = lastModified;
@@ -83,13 +99,12 @@ public class DirectoryListingTest {
             }
 
             assertTrue("Failed to set modification time",
-                       file.setLastModified(val));
+                    file.setLastModified(val));
 
-            if (!"-".equals(size)) {
-                int s = Integer.parseInt(size);
-                if (s > 0) {
+            if (subdirs == null) {
+                if (size > 0) {
                     FileOutputStream out = new FileOutputStream(file);
-                    byte[] buffer = new byte[s];
+                    byte[] buffer = new byte[size];
                     out.write(buffer);
                     out.close();
                 }
@@ -104,13 +119,16 @@ public class DirectoryListingTest {
 
                 // @todo verify all attributes!
                 if (name.compareTo(fe.name) == 0
-                        && href.compareTo(fe.href) == 0
-                        && size.compareTo(fe.size) == 0) {
-                    ret = 0;
+                        && href.compareTo(fe.href) == 0) {
+                    if ((subdirs == null && size == fe.size)
+                            || (subdirs != null && size == DIRECTORY_INTERNAL_SIZE)) {
+                        ret = 0;
+                    }
                 }
             }
             return ret;
         }
+
     }
 
     public DirectoryListingTest() {
@@ -129,27 +147,27 @@ public class DirectoryListingTest {
         directory = FileUtilities.createTemporaryDirectory("directory");
 
         entries = new FileEntry[3];
-        entries[0] = new FileEntry("foo.c", "foo.c", 0, "1");
-        entries[1] = new FileEntry("bar.h", "bar.h", Long.MAX_VALUE, "0");
-        entries[2] = null;
+        entries[0] = new FileEntry("foo.c", "foo.c", 0, 112, null);
+        entries[1] = new FileEntry("bar.h", "bar.h", Long.MAX_VALUE, 0, null);
+        // Will test getSimplifiedPath() behavior for ignored directories.
+        // Use DIRECTORY_INTERNAL_SIZE value for length so it is checked as the directory
+        // should contain "-" (DIRECTORY_SIZE_PLACEHOLDER) string.
+        entries[2] = new FileEntry("subdir", "subdir/", 0, DIRECTORY_INTERNAL_SIZE, Arrays.asList(
+                new FileEntry[]{
+                    new FileEntry("SCCS", "SCCS", 0, DIRECTORY_INTERNAL_SIZE, Arrays.asList(
+                            new FileEntry[]{
+                                new FileEntry("version", "version", 0, 312, null)
+                            })
+                    )}
+        ));
 
         for (FileEntry entry : entries) {
-            if (entry != null) {
-                entry.create();
-            }
+            entry.create();
         }
-        // Create the entry that will be ignored separately.
-        FileEntry hgtags = new FileEntry(".hgtags", ".hgtags", 0, "1");
-        hgtags.create();
 
-        // Will test getSimplifiedPath() behavior for ignored directories.
-        // Use negative value for length so it is not checked as the return value
-        // of length() is unspecified for directories.
-        entries[2] = new FileEntry("subdir", "subdir/", 0, DirectoryListing.DIRECTORY_SIZE_PLACEHOLDER);
-        File subdir = new File(directory, "subdir");
-        subdir.mkdir();
-        File SCCSdir = new File(subdir, "SCCS");
-        SCCSdir.mkdir();
+        // Create the entry that will be ignored separately.
+        FileEntry hgtags = new FileEntry(".hgtags", ".hgtags", 0, 1, null);
+        hgtags.create();
 
         // Need to populate list of ignored entries for all repository types.
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
@@ -160,6 +178,7 @@ public class DirectoryListingTest {
     public void tearDown() throws Exception {
         if (directory != null && directory.exists()) {
             removeDirectory(directory);
+            directory.delete();
         }
     }
 
@@ -167,7 +186,6 @@ public class DirectoryListingTest {
         File[] childs = dir.listFiles();
         if (childs != null) {
             for (File f : childs) {
-
                 if (f.isDirectory()) {
                     removeDirectory(f);
                 }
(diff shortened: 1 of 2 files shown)


vladak:
what about `DirectoryListingTest.java` ?

tulinkry:
I updated the test case to cover this directory size set to "-"

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -84,11 +85,15 @@ private void create() throws Exception {
...
-                out.close();
+            try {
+                int s = Integer.parseInt(size);
+                if (s > 0) {
+                    FileOutputStream out = new FileOutputStream(file);
+                    byte[] buffer = new byte[s];
+                    out.write(buffer);
+                    out.close();
+                }
+            } catch (NumberFormatException e) {
```
vladak: empty catch ?

tulinkry: this is on purpose; if the size isn't numeric then nothing should be written to the file and it is a test so I suppose any logging isn't really necessary here.

vladak: I think the better way would be to explicitly check for `-`. If the string cannot be parsed as integer, then the test should fail. Also, it is generally considered bad practice to include empty catch statements (at least some explanatory test should be added).

tulinkry: changed that to "if" statement

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -84,11 +85,14 @@ private void create() throws Exception {
             assertTrue("Failed to set modification time",
                        file.setLastModified(val));
 
-            if (size > 0) {
-                FileOutputStream out = new FileOutputStream(file);
-                byte[] buffer = new byte[size];
-                out.write(buffer);
-                out.close();
+            if (!"-".equals(size)) {
```
vladak: use `DIRECTORY_SIZE_PLACEHOLDER` ?

vladak: Also, while you're at it, change create() to use mkdir() for the directories.and fix setUp()

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -276,8 +275,8 @@ private void validateEntry(Element element) throws Exception {
         entry.lastModified = getLastModified(nl.item(3));
         try {
             entry.size = getSize(nl.item(4));
-        } catch (Exception e) {
-            entry.size = -1;
+        } catch (NumberFormatException e) {
```
vladak: can this actually happen given that getSize() does not parse the integer anymore ?

tulinkry:
I reworked the test case a little bit
 - creating all the file entries is simplified
 - size is integer again, -2 is to signal that "-" has been found in the output

tulinkry:
@vladak can you review again?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -238,25 +257,34 @@ private long getLastModified(Node item) throws Exception {
...
-    private int getSize(Node item) throws Exception {
+    private int getSize(Node item) throws NumberFormatException {
         Node val = item.getFirstChild();
         assertNotNull(val);
         assertEquals(Node.TEXT_NODE, val.getNodeType());
-        return Integer.parseInt(val.getNodeValue().trim());
+        try {
+            return Integer.parseInt(val.getNodeValue().trim());
+        } catch (NumberFormatException ex) {
+            if (DirectoryListing.DIRECTORY_SIZE_PLACEHOLDER.equals(val.getNodeValue().trim())) {
```
vladak: why not check the `-` equality first and then try to convert to integer ? Checking the character only after exception looks a bit convoluted.

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -28,52 +28,69 @@
...
     private FileEntry[] entries;
     private SimpleDateFormat dateFormatter;
 
     class FileEntry implements Comparable {
+
         String name;
         String href;
         long lastModified;
-        int size; // If negative, don't check it.
+        int size;
```
vladak: restore/update the previously present comment ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -28,52 +28,69 @@
...
 import static org.junit.Assert.*;
-import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
-import org.opensolaris.opengrok.history.RepositoryFactory;
 
 /**
  * JUnit test to test that the DirectoryListing produce the expected result
  */
 public class DirectoryListingTest {
 
+    private static int DIRECTORY_INTERNAL_SIZE = -2;
```
vladak: add a comment explaining how this is used and why the value is -2 ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -82,13 +99,15 @@ private void create() throws Exception {
...
-
-            if (size > 0) {
-                FileOutputStream out = new FileOutputStream(file);
-                byte[] buffer = new byte[size];
-                out.write(buffer);
-                out.close();
+                    file.setLastModified(val));
+
+            if (subdirs == null) {
+                if (size > 0) {
```
vladak: join this if clause with the one on the precedent line ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -238,25 +257,34 @@ private long getLastModified(Node item) throws Exception {
...
-            : dateFormatter.parse(value).getTime();
+                ? Long.MAX_VALUE
+                : dateFormatter.parse(value).getTime();
     }
 
     /**
      * Get the size from the: &lt;td&gt;&lt;tt&gt;size&lt;/tt&gt;&lt;/td&gt;
+     *
      * @param item the node representing &lt;td&gt;
      * @return The size
```
vladak: adjust the comment a bit reflecting negative values ?

vladak:
added a bunch more comments

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -28,52 +28,69 @@
...
-        int size; // If negative, don't check it.
+        int size;
+        List<FileEntry> subdirs;
 
         FileEntry() {
             dateFormatter = new SimpleDateFormat("dd-MMM-yyyy");
         }
 
-        FileEntry(String name, String href, long lastModified, int size) {
+        FileEntry(String name, String href, long lastModified, int size, List<FileEntry> subdirs) {
```
vladak: pls keep the old method signature and make it a wrapper so that old calls do not have to be adjusted with the `null` argument.

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -28,52 +28,69 @@
...
+
+            if (subdirs != null && subdirs.size() > 0) {
+                // this is a directory
+                assertTrue("Failed to create a directory", file.mkdirs());
+                for (FileEntry entry : subdirs) {
+                    entry.name = name + File.separator + entry.name;
+                    entry.create();
+                }
+            } else {
+                if (!file.exists()) {
```
vladak: is this check necessary ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -28,52 +28,69 @@
...
-                assertTrue("Failed to create file", file.createNewFile());
+
+            if (subdirs != null && subdirs.size() > 0) {
+                // this is a directory
+                assertTrue("Failed to create a directory", file.mkdirs());
+                for (FileEntry entry : subdirs) {
+                    entry.name = name + File.separator + entry.name;
+                    entry.create();
+                }
+            } else {
```
vladak: add Assert for the case when size > 0 and subdirs != null && subdirs.size() > 0 (badly specified arguments) or use different API (see above).

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
vladak: or use different signature for the subdirs case without the `size` argument

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -129,28 +147,28 @@ public void setUp() throws Exception {
...
-        entries[1] = new FileEntry("bar.h", "bar.h", Long.MAX_VALUE, 0);
-        entries[2] = null;
+        entries[0] = new FileEntry("foo.c", "foo.c", 0, 112, null);
+        entries[1] = new FileEntry("bar.h", "bar.h", Long.MAX_VALUE, 0, null);
+        // Will test getSimplifiedPath() behavior for ignored directories.
+        // Use DIRECTORY_INTERNAL_SIZE value for length so it is checked as the directory
+        // should contain "-" (DIRECTORY_SIZE_PLACEHOLDER) string.
+        entries[2] = new FileEntry("subdir", "subdir/", 0, DIRECTORY_INTERNAL_SIZE, Arrays.asList(
+                new FileEntry[]{
+                    new FileEntry("SCCS", "SCCS", 0, DIRECTORY_INTERNAL_SIZE, Arrays.asList(
```
vladak: should this have the `href` argument with trailing slash given this is a directory ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: fixed

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: ok, changed

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: comment added

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: joined

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: changed

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: added two different constructors and made the original private

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: no it's not, that was just for my local testing. good catch

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: changed the api

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: yes it should, changed

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -51,7 +51,15 @@
...
-    private static int DIRECTORY_INTERNAL_SIZE = -2;
+    /**
+     * Indication of that the file was a directory and so that the size given by
+     * the FS is platform dependent.
+     */
+    private static final int DIRECTORY_INTERNAL_SIZE = -2;
+    /**
+     * Indication of unparseable file size.
+     */
+    private static final int DIRECTORY_INVALID_SIZE = -1;
```
vladak: just name this INVALID_SIZE ?

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
```diff
@@ -101,16 +141,15 @@ private void create() throws Exception {
...
                     file.setLastModified(val));
 
-            if (subdirs == null) {
-                if (size > 0) {
-                    FileOutputStream out = new FileOutputStream(file);
+            if (subdirs == null && size > 0) {
+                try (FileOutputStream out = new FileOutputStream(file)) {
                     byte[] buffer = new byte[size];
                     out.write(buffer);
-                    out.close();
```
vladak: why is the close() not needed anymore ?

tulinkry: cause netbeans converted the block to the try-with-resource - it will be closed automatically

test/org/opensolaris/opengrok/web/DirectoryListingTest.java
tulinkry: changed

PR #1621, opened 2017-05-31
SearchAllProjects functionality added

dhireng:
Here's a summary of the changes:
PageConfig.java
1. added getAllRequestedProjects function to return all projects. This is used when seachall projects button is clicked
2. prepareSearch now has a boolean input. this is set to true when all projects is selected and is used to determine if the message "You must select a project!" is to be displayed.
utils-0.0.8.js
added function SearchAll. this is called when selectAllprojects button is clicked. It deselects a) all projects b) sets the hdnsearchall to true and then submits the form
menu.jspf
1. if selectall projects is true, it clears pRequested list to ensure none of the projects are listed
2. SearchAllProjects button is added
3. hidden variable hiddenSearchAll is created to support the SearchAllProjects
projects.jspf
1. resetcookie function to remove cookie PageConfig.OPEN_GROK_PROJECT when searchall projects is clicked
2. if condition added to either call remove cookie or continue with current logic
search.jsp
1. added flag to prepareSearch call. This is used to decide if error message "You must select a project!" is required or not
2. call to search updated with allprojects if that is selected --> cfg.getAllRequestedProjects()

Note: OCA will be signed


src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -752,6 +752,20 @@ public String getRequestedProjectsAsString() {
         return requestedProjects;
     }
     
+	public SortedSet<String> getAllRequestedProjects() {
+		    TreeSet<String> set = new TreeSet<>();
+            Set<Project> allProjects = getProjectHelper().getAllProjects();
```
tulinkry: so basically this whole method is useless, call to `getProjectHelper().getAllProjects()` is enough anywhere when you would call this method

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1277,12 +1290,12 @@ public File getDataRoot() {
      *
      * @return a search helper.
      */
-    public SearchHelper prepareSearch() {
+    public SearchHelper prepareSearch(boolean isSearchAll) {
```
tulinkry: I would rather get this boolean value inside this function (request is a member variable in PageConfig) and keep this method without the parameter

web/search.jsp
```diff
@@ -65,11 +65,18 @@ include file="projects.jspf"
     PageConfig cfg = PageConfig.get(request);
 
     long starttime = System.currentTimeMillis();
-
-    SearchHelper searchHelper = cfg.prepareSearch();
+	SearchHelper searchHelper = null;
+    if (request.getParameter("hdnSearchAll") != null && request.getParameter("hdnSearchAll").equals("true"))
```
tulinkry: `request.getParameter("hdnSearchAll") != null && request.getParameter("hdnSearchAll").equals("true")`

I would suggest to make a method of PageConfig to check this, something like `isSearchAll` or similar.

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -752,6 +752,20 @@ public String getRequestedProjectsAsString() {
         return requestedProjects;
     }
     
+	public SortedSet<String> getAllRequestedProjects() {
```
tulinkry: Also isn't better idea to move the check of `is this a search all?` to the `getRequestedProjects` and act upon there? It would make the change more transparent I think

web/menu.jspf
```diff
@@ -202,10 +204,12 @@ org.opensolaris.opengrok.web.Util"
     </tbody>
 </table>
 <div id="form-controls">
+    <input tabindex="9" class="submit btn" type="button" value="Search All Projects" onclick="javascript: SearchAll();"/>
```
tulinkry: Firstly this should be after the `Search` button to avoid confusion with already existing button.

Secondly I would use just `Search All` text or `Search all`

web/menu.jspf
```diff
@@ -202,10 +204,12 @@ org.opensolaris.opengrok.web.Util"
...
     <input tabindex="9" class="submit btn" type="submit" value="Search"/>
     <input tabindex="10" class="submit btn" onclick="javascript: clearSearchFrom();"
            type="button" value="Clear"/>
     <a href="help.jsp" target="_blank" class="help btn" title="Help" tabindex="11" />Help</a>
+	<input type="hidden" id="hdnSearchAll" name="hdnSearchAll" value="false">
```
tulinkry: The name is actually a little clumsy here because it is a part of an URL which you can send to any user. 

I'd prefer `all` or `search_all` or `searchAll`.

web/menu.jspf
```diff
@@ -92,7 +92,9 @@ org.opensolaris.opengrok.web.Util"
     <select tabindex="8" class="q" id="project"
         name="project" multiple="multiple" size="<%=
         Math.min(15, projectsSize) %>"><%
-        SortedSet<String> pRequested = cfg.getRequestedProjects();
+		SortedSet<String> pRequested = cfg.getRequestedProjects();
+		if (request.getParameter("hdnSearchAll") != null && request.getParameter("hdnSearchAll").equals("true"))
+			pRequested.clear();
```
tulinkry: explain in a code comment why this is necessary so that someone in the future does not remove it impetuously

web/projects.jspf
```diff
@@ -30,22 +30,45 @@ java.util.TreeSet,
...
-    PageConfig cfg = PageConfig.get(request);
-    cfg.getEnv().setUrlPrefix(request.getContextPath() + Prefix.SEARCH_R + "?");
-
-    String projects = cfg.getRequestedProjectsAsString();
-    if (projects.length() != 0) {
-        Cookie cookie = new Cookie(PageConfig.OPEN_GROK_PROJECT, URLEncoder.encode(projects, "utf-8"));
-        response.addCookie(cookie);
-    }
+	if (request.getParameter("hdnSearchAll") !=null && request.getParameter("hdnSearchAll").equals("true")) {
+		resetCookie(PageConfig.OPEN_GROK_PROJECT, request, response);
```
tulinkry: If you follow my comment

> Also isn't better idea to move the check of is this a search all? to the getRequestedProjects and act upon there? It would make the change more transparent I think

then this part would not be necessary I think you will store all projects in the cookie or is there any problem with it?

tulinkry:
And also a big problem with this PR is that you include the merge commit 

```
 Merge pull request #1 from OpenGrok/master
dhireng committed on GitHub 2 hours ago

latest updates
```

which basically adds all of the commits in opengrok twice

you should include only the commit with your changes (the other one)

tulinkry:
Which you can do by:

```bash
$ git log --format=%B -n 1 8424b72bea3e021e659bcf66ab115a5a9372e4c7 > ../commit.tmp
$ git reset fc55df9580d76bd4eeefc16086c5fd7bcab6f708
$ git add .
$ git commit -m "$(cat ../commit.tmp)"
$ rm ../commit.tmp
```

src/org/opensolaris/opengrok/web/PageConfig.java
dhireng: I thought there is a need to use authFramework.isAllowed to check if a search is allowed?

src/org/opensolaris/opengrok/web/PageConfig.java
dhireng: okay.. will look cleaner

vladak:
Let's not add any redundant elements to the index page. It looks like the button basically performs what can be now done by clicking on 'Select all' and 'Search'.

dhireng:
Select All and Search makes a long URL. Web Application Firewalls don't like it.  load balancers have scripts that fail. too difficult to share. too long to type.

tarzanek:
@dhireng thank you for OCA, we just got it

dhireng:
is the latest version good? any further areas to improve?

vladak:
I don't like the addition of the new button. In the past release, we were striving to make the UI minimal and I'd like to continue in the trend. Also, usually, if any sort of UI change is done, screenshot should be included.

dhireng:
searchallinopengrok

davecahill:
We implemented a similar feature due to our many repos.
We avoided the need for a "search all" button by:
* Adding a "fake" project called "!all!" (so that it sorts to the top)
* The "Select all" button now deselects everything and selects the "!all!" project

Not sure if that helps here, but thought I'd mention.

vladak:
That's a neat idea with the special project. It just needs to be correctly implemented w.r.t. authorization.

src/org/opensolaris/opengrok/web/PageConfig.java
tulinkry: the call to the framework is ensured by `getProjectHelper().getAllProjects()`

vladak:
Closing for lack of response. Feel free to reopen a new PR with fresh code.

Issue #1622, opened 2017-06-06
on demand xref generation does not produce Scopes


vladak:
It seems that if the indexer is run with -e (generate xrefs on demand), the Scopes are not generated. Not sure if this is a inherent limitation or could be fixed. @mrate ?

mrate:
Yep, scopes were formely generated separately but due to some optimizations we merged them to xref generation. So if xrefs are disabled then scopes will not be generated... This could be fixed that if scopes are enabled and xrefs are not then run xref but ommit output and just generate the scopes.

tarzanek:
I think with -e xrefs are generated on demand, it could very well be, that webapp container is not able to properly run them on top of source (e.g. webapp is missing ctags in path, etc. )

PR #1631, opened 2017-06-07
on demand xref generation does not produce Scopes (#1622)

mrate:
Simple fix for #1622

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
+++ b/src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
@@ -105,6 +105,11 @@ public class PlainAnalyzer extends TextAnalyzer {
         }
         
         if (scopesEnabled && xrefOut == null) {
+            /*
+             * Scopes are generated during xref generation. If xrefs are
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/util/NullWriter.java
```diff
@@ -0,0 +1,46 @@
...
+ /*
+ * Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opensolaris.opengrok.util;
+
+import java.io.IOException;
+import java.io.Writer;
+
+/**
+ *
```
vladak: some comment here explaining for what is this class used would be handy

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
```diff
@@ -102,6 +103,10 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut) throws IOExc
                 }
             }
         }
+
```
vladak: pls add comment explaining why this is being done

PR #1635, opened 2017-06-14
Improve UTF-8/16 BOM handling.

tetsurom:
Currently OpenGrok recognizes UTF-8/16 BOM only in `/xref` without parameter such as revision or annotate.
This patch enable to handle it in `/xref` (with any parameter),  `/search` and `/diff` in same way.

I have already signed OCA.


tarzanek:
Thank you Matsumura san
merging

PR #1642, opened 2017-06-23
use lambda function for sorting

vladak:


tarzanek:
let the lambdas begin!
functional programming ... give it to us baby!

PR #1643, opened 2017-06-26
disable history completely if Indexer is run without -H

vladak:


tulinkry:
What happens when I use the search url with hist parameter? `/source/search?hist=abcd` ?

vladak:
The result will be 'Your search hist:abcd did not match any files.'

vladak:
Requess for `http://grok/source/history/foo/bar` will end up with:

> Error: File not found!
> The requested resource is not available.
> Resource lacks history info. 
> ...

which seems correct to me.

Issue #817, opened 2014-04-17
Solaris IPS package should deliver to /var/share


vladak:
Comment from Jens about IPS Solaris package:

> /var/opengrok is a bad choice. Should be /var/share/opengrok/src [or as
> long as lazy installer people haven't fixed their stuff and you wanna
> make it 100% "bad workaround" compliant /var/.migrate/opengrok/src =8-( ].
> Otherwise you may loose data when switching the BE.

Hence, we should switch from `/var/opengrok` to `/var/share/opengrok`

jelmd:
NOTE: currently IPS does not support this directly, i.e. having a startup/setup script, which populates datadir, etc. below /var/share/ is ok, but the package itself is not allowed to deliver anything below /var/share/ or /var/.migrate/! See also SUNBT7081759 .

vladak:
According to the IPS documentation [*] it is necessary to perform the migration of data in script run by a temporary SMF service. Such a script exist (`/lib/svc/share/migrate_shared_files.py`, run by `fs-minimal` service if `is_self_assembly_boot()` i.e. if `/.SELF-ASSEMBLY-REQUIRED` file exists which seems does not unless special circumstances are met) and although it is not a public interface, we can use it quite safely until better way is possibly delivered in Solaris 11.x updates.

[*] https://docs.oracle.com/cd/E53394_01/html/E54820/gnxft.html#PKDEVglyrx

PR #1644, opened 2017-06-27
IPS packaging changes

vladak:
- added more checks
- cleanup
- fix for #817 to share data across boot environments

### DIFF
--- a/platform/solaris/ips/create.sh
+++ b/platform/solaris/ips/create.sh
@@ -146,13 +146,21 @@ then
    exit 1
 fi
 
+cleanup()
+{
+	rm -rf "$PKG_REPO_NAME"
+	if [[ -n $manifest ]]; then
+		rm -f "$manifest"
+	fi
+}
+
 PKG()
 {
    "$@"
    if [ $? != 0 ]
    then
       echo Command failed: "$@"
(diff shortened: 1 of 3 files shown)


platform/solaris/ips/create.sh
```diff
@@ -185,33 +194,68 @@ do
    PKG pkgsend add dir mode=0755 owner=root group=bin path=${dir}
 done
 
-for dir in /var/opengrok /var/opengrok/data /var/opengrok/etc \
-           /var/opengrok/log /var/opengrok/src
+PKG pkgsend add dir mode=0755 owner=webservd group=webservd path=/var/opengrok
```
tarzanek: shouldn't there be also /var/share/opengrok
too ?

vladak: This will be created automatically by the migration script. In fact I think `pkglint` would complain if the package delivered content under `/var/share`.

tarzanek: ok, please merge

vladak: Also, the reason why packages must not deliver under `/var/share` is simple - if a BE A installs a package bits to /var/share and then BE B uninstalls the package and BE A is activated, its world will be corrupted.

PR #1653, opened 2017-07-03
adding per stack statistics for authorization

### LINKED ISSUE #1568: per authorization stack stats
would like to have timing info for authorization sub-stacks

tulinkry:
fixes #1568

### DIFF
--- a/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
+++ b/src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
@@ -156,10 +156,10 @@ public final class AuthorizationFramework {
                 "plugin_framework_project_cache",
                 project,
                 new AuthorizationEntity.PluginDecisionPredicate() {
-            @Override
-            public boolean decision(IAuthorizationPlugin plugin) {
-                return plugin.isAllowed(request, project);
-            }
+                    @Override
+                    public boolean decision(IAuthorizationPlugin plugin) {
+                        return plugin.isAllowed(request, project);
+                    }
         }, new AuthorizationEntity.PluginSkippingPredicate() {
             @Override
             public boolean shouldSkip(AuthorizationEntity authEntity) {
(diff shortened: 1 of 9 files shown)


src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -160,6 +160,12 @@ public boolean isAllowed(HttpServletRequest request, Project project) {
             public boolean decision(IAuthorizationPlugin plugin) {
                 return plugin.isAllowed(request, project);
             }
+
+            @Override
```
vladak: The indentation is a bit confusing here, it'd be nice if the overrides were shifter to the right past the first column of the `new AuthorizationEntity` line

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -269,6 +275,18 @@ protected boolean processStack(Nameable entity,
                 }
             }
         }
+        time = System.currentTimeMillis() - time;
+
+        stats.addRequestTime(pluginPredicate.getRequest(),
+                String.format("authorization_in_stack_%s_%s", getName(), overallDecision ? "positive" : "negative"),
```
vladak: introduce `decisionToString()` ?

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
```diff
@@ -269,6 +275,18 @@ protected boolean processStack(Nameable entity,
...
+                time);
+        stats.addRequestTime(pluginPredicate.getRequest(),
```
vladak: I am baffled by the `request` argument of `addRequestTime()` which calls `addRequest()` but that is not using the `req` argument either. So why have it at all ?

tulinkry: actually a very good point :-D

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
tulinkry: I changed the indentation however this was the default NB format feature so maybe in the future it will again shift to the left

src/org/opensolaris/opengrok/authorization/AuthorizationStack.java
tulinkry: I don't think this is necessary
 - we dont do this for plugins (only stacks)
 - there is no duplication in the format strings

PR #1659, opened 2017-07-07
fixed header instead of javascript content height resize action

### LINKED ISSUE #1288: [Wish-list] Can't cooperative with many famous chrome plugins for taking sticky notes on code pages
I want to use some chrome plugins like: [Note Anywhere](https://chrome.google.com/webstore/detail/note-anywhere/bohahkiiknkelflnjjlipnaeapefmjbh) and 
[Web page sticky notes](https://chrome.google.com/webstore/detail/web-page-sticky-notes/alpjieidnmmkljnceakgpeajlngabnee) to recording my thoughts when I am reading codes in opengrok:

But the sticky notes can not synchronize with the code pages (they should be) when I am scrolling code window down.

I have tried more than 10+ different note-on-web-pages plugins in chrome. The notes can truely stick to the origin position on one web page, when I scroll around.

They works very well on any other web sites, except the code pages generated by OpenGrok.

Can OpenGrok generate some sticky note friendly pages in the feature version ??  I want not only read the code, but also take some notes right in the code. Previously, I was using vim with cscope, I could write notes by directly adding comments to the current file in vim. Now I switch to the great project because I want to read code on my iPad. But I found I can just read in OpenGrok and can not record my thoughts right on the code, when I am reading.

After installing some famous chrome note plugins, I sadly found OpenGrok just can't cooperative with them (synchronizing issues):

stickynote

Could you please tell me a way I can take notes on the code with OpenGrok ?

### LINKED ISSUE #1488: scrollbar shouldn't depend on .js

tulinkry:
fixes #1488
might fix #1288

I removed the silly content resize javascript and introduced a fixed header for all code related pages. The rest of the changes in the js code are just changes to make it work with the new fixed header.

The result of this change is that viewing a file is still available even if any of our javascript beauty fails when executing.

Anyway I had to introduce a new color for the header because it looks quite ugly when it's white and there is a space between the line number (the header must have a fixed height).

screenshot from 2017-07-07 14-18-37
screenshot from 2017-07-07 14-18-51

### DIFF
--- a/src/org/opensolaris/opengrok/web/Scripts.java
+++ b/src/org/opensolaris/opengrok/web/Scripts.java
@@ -125,9 +125,9 @@ public class Scripts implements Iterable<Scripts.Script> {
         SCRIPTS.put("jquery-tablesorter", new FileScript("js/jquery-tablesorter-2.26.6.min.js", 12));
         SCRIPTS.put("tablesorter-parsers", new FileScript("js/tablesorter-parsers-0.0.1.js", 13));
         SCRIPTS.put("searchable-option-list", new FileScript("js/searchable-option-list-2.0.3.min.js", 14));
-        SCRIPTS.put("utils", new FileScript("js/utils-0.0.11.js", 15));
+        SCRIPTS.put("utils", new FileScript("js/utils-0.0.12.js", 15));
         SCRIPTS.put("repos", new FileScript("js/repos-0.0.1.js", 20));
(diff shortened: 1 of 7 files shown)


tarzanek:
lgtm, merging

PR #1661, opened 2017-07-07
removing sleeps as the tests arbitrarily fail

### LINKED ISSUE #1470: get rid of sleeps in testExpirationSingle if possible
for (int i = 0; i < 5; i++) {
            Assert.assertEquals(1, env.getMessagesInTheSystem());
            Assert.assertNotNull(env.getMessages());
            Assert.assertEquals(new TreeSet<Message>(Arrays.asList(makeArray(m1))), env.getMessages());
            sleep(100);
        }
        sleep(30);
        Assert.assertEquals(0, env.getMessagesInTheSystem());

junit/org.opensolaris.opengrok.configuration.messages/ExpirationNormalMessageTest/testExpirationSingle/

started to fail randomly, that means we have to find a better way than sleeps to check if the messages arrived ...

tulinkry:
fixes #1470
Removing sleeps - the timer thread is now tested just for the concurrent performance.

### DIFF
--- a/test/org/opensolaris/opengrok/configuration/messages/ExpirationNormalMessageTest.java
+++ b/test/org/opensolaris/opengrok/configuration/messages/ExpirationNormalMessageTest.java
@@ -22,13 +22,10 @@
  */
 package org.opensolaris.opengrok.configuration.messages;
 
-import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


tarzanek:
lgtm, tests pass, merging

PR #1662, opened 2017-07-10
adding more colors for symbol highlighting

### LINKED ISSUE #1634: More highlighting colors in xref!
Highlighting symbols in xref is a very useful feature.

As an enhancement I'd like to request addition of more colors, and e.g. binding them to keys 3, 4, 5, 6. Current binding for "3", which currently means "deselect all", is kind of a waste anyway, and can be sacrificed in favor of the new colors.

tulinkry:
fixes #1634
screenshot from 2017-07-10 18-08-39

### DIFF
--- a/web/js/utils-0.0.10.js
+++ b/web/js/utils-0.0.10.js
@@ -807,10 +807,20 @@
                                 break;
                             case 50: // 2
                                 if (that.symbol) {
-                                    that.unhighlight(that.symbol).length === 0 && that.highlight(that.symbol);
+                                    that.unhighlight(that.symbol).length === 0 && that.highlight(that.symbol, 'yellow');
                                 }
                                 break;
                             case 51: // 3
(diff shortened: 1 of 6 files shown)


web/js/utils-0.0.10.js
```diff
@@ -807,10 +807,20 @@
                                 break;
                             case 50: // 2
                                 if (that.symbol) {
-                                    that.unhighlight(that.symbol).length === 0 && that.highlight(that.symbol);
+                                    that.unhighlight(that.symbol).length === 0 && that.highlight(that.symbol, 'yellow');
```
vladak: naming the styles in color-agnostic way would be better I think (second/third/... comes to mind)

tulinkry:
I changed the names to `highlight-color-<n>` where n is 1, 2 or 3. And I updated the polished and offwhite styles.

Issue #1390, opened 2017-02-07
Project centric indexing and management


vladak:
In our deployment syncing source code before indexing takes a lot time for some projects, so a scheme is used where syncing is done in parallel and then `indexpart` is used for individual projects and once all these partial indexes are done, one global index run is done to discover newly added projects and remove data of deleted projects. The global index takes non-trivial time to run and can be avoided by having `addproject` and `delproject` sub-commands of the `OpenGrok` shell script.

`addproject` would scan project's repos, add the project to the configuration. The `indexpart` could then index it and the config can be sent to the webapp (e.g. via the changes for #1355 or more fine grained Message).

`delproject` would remove the project from configuration (and update the webapp) and remove project's index/xref/history data. This should take care of #1307.

To make this fully functional, some other changes are needed, like #1187.

TotoXe:
Very interested by this great feature! 
We scan with OpenGrok more than 800 projects for more than 9500 repos. At the moment, I am running daily 1 indexing job but I would like using indexpart plus other features to add easily new project and delete useless ones.

vladak:
Also, something needs to be done with the `Indexes created ...` timestamp in the page footer. Once a setup is fully project driven, this will cease to have significance.

vladak:
Here's a proposal of how I think it can work:
  - `addproject` will be a Message sent to the webapp. The webapp will add the project to the list and do the scan of project's repositories and add them to the config as well.
  - `delproject` will also be a Message. The webapp will remove the project and its repositories from config, perform the cleanup of indexSearchers/searcherManagers. It can also try remove project's data from data root (but not the actual source code).
  - to sync the config from the webapp to `configuration.xml`,  the configuration can be downloaded from the webapp via new Message (#1426) so that `indexpart` can use it.
  - It should be possible to send multiple `addproject`/`delproject` messages to the webapp in parallel. If needed, create new thread that will be handling the work of scanning repositories etc.

tulinkry:
Maybe better only one message `project` with text "add" or "delete" and appropriate tags.

vladak:
Yep, new `project` message will be introduced that has 3 flavors:
  - add
  - delete
  - indexed (internal; to mark the project as searchable, sent automatically once indexing is done for the first time for given project)

vladak:
Documented on the wiki: https://github.com/OpenGrok/OpenGrok/wiki/Per-project-management

Issue #767, opened 2014-02-27
remove history cache automatically


vladak:
When working on #729 it occurred to me that similarly like xref's are deleted for files which no longer exist in the repo, the same could be done for history indexes. I.e. if the repository is no longer around it's history cache should be cleared automatically.

This would make maintenance of large OpenGrok instance easier.

PR #1663, opened 2017-07-11
Project addition/deletion

vladak:
The overall goal is to avoid indexing of all projects (i.e. `OpenGrok index /var/opengrok/src`) in order to discover new and remove deleted projects from the configuration. Thus one can start with no index and no projects and then incrementally add them and index them.

It basically works like this:

0. bootstrap initial configuration:
  `./OpenGrok bootstrap`
   - this will create `/var/opengrok/etc/configuration.xml` with basic set of properties. If more is needed `./tools/Messages -n config -t set "set foo = bar"` can be used
1.  add a new project `foo`:
  `./tools/Messages -t foo -n project add`
2. the project `foo` is now visible in the configuration however is not yet searchable. Store the config in a file so that indexer can see the project and its repositories.
  `./tools/Messages -n config -t getconf > /var/opengrok/etc/configuration.xml`
3. index the project. It will become searchable after that.
  `OPENGROK_READ_XML_CONFIGURATION=/var/opengrok/etc/configuration.xml ./OpenGrok indexpart /var/opengrok/src /foo`
4. make the project "indexed" status of the project persistent so that if webapp is redeployed the project will be still visible:
  `./tools/Messages -n config -t getconf > /var/opengrok/etc/configuration.xml`

If an `add` message is sent that matches existing project, the list of repositories for that project will be refreshed. This is handy when repositories are added/deleted.


tarzanek:
nice!
I will have a look and test this on Friday
if you want, please merge

vladak:
I'd like someone to take a look at the code first.

vladak:
Part of the changes that introduce some noise are:
  - `runHgCommand()` changed to use elipsis
  -

vladak:
The index timestamp will be updated also for each partial reindex. This means the timestamp will no longer reference all projects' indexes, rather date of any last reindex.

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -64,10 +64,11 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(FileHistoryCache.class);
 
     private final Object lock = new Object();
-    private String historyCacheDirName = "historycache";
-    private String latestRevFileName = "OpenGroklatestRev";
+    private final String historyCacheDirName = "historycache";
```
tulinkry: may be static?

src/org/opensolaris/opengrok/history/FileHistoryCache.java
```diff
@@ -64,10 +64,11 @@
...
+    private final String latestRevFileName = "OpenGroklatestRev";
     private boolean historyIndexDone = false;
-    
+    public final static String historyDirName = "historycache";
```
tulinkry: what is the point of having these two?
`private final String historyCacheDirName = "historycache";`
`public final static String historyDirName = "historycache";`

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -635,6 +641,12 @@ public void removeCache(Collection<String> repositories) throws HistoryException
                         new Object[]{r.getDirectoryName(), e.getLocalizedMessage()});
             }
         }
+
+        return repos;
+    }
+
+    public void removeCache(Collection<String> repositories) throws HistoryException {
```
tulinkry: a brief javadoc would be nice

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -764,7 +776,7 @@ public void invalidateRepositories(Collection<? extends RepositoryInfo> repos) {
             boolean verbose = RuntimeEnvironment.getInstance().isVerbose();
 
             if (verbose) {
-                LOGGER.log(Level.FINE, "invalidating repositories");
+                LOGGER.log(Level.FINE, "invalidating " + repos.size() + " repositories");
```
tulinkry: using log template is preferred

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1024,6 +1045,14 @@ public void setScopesEnabled(boolean scopesEnabled) {
         threadConfig.get().setScopesEnabled(scopesEnabled);
     }
 
+    public boolean isProjectsEnabled() {
+        return threadConfig.get().isProjectsEnabled();
+    }
```
tulinkry: - shouldn't we use this in Util.dumpConfiguration?
- shouldn't we use this in env.hasProjects?

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -0,0 +1,275 @@
...
+import org.opensolaris.opengrok.index.Indexer;
+import static org.opensolaris.opengrok.util.IOUtils.removeRecursive;
+import org.opensolaris.opengrok.util.TestRepository;
+
+/**
+ *
+ * @author vkotal
+ */
+@ConditionalRun(condition = RepositoryInstalled.MercurialInstalled.class)
+// XXX need svn + git
```
tulinkry: ?

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -0,0 +1,275 @@
...
+        m.apply(env);
+        Assert.assertEquals(3, env.getProjects().size());
+        Assert.assertEquals(3, env.getRepositories().size());
+        Assert.assertEquals(3, env.getProjectRepositoriesMap().size());
+
+        // Run the indexer (ala 'indexpart') so that data directory is populated.
+        ArrayList<String> subFiles = new ArrayList<>();
+        subFiles.add("/git");
+        subFiles.add("/mercurial");
+        subFiles.add("/git");
```
tulinkry: git twice?

tools/Messages
```diff
@@ -62,6 +62,14 @@ Usage()
...
     echo "                      Returns text describing the action."
     echo "                      The change is NOT persistent when reindex or redeploy!"
+    echo "    project:"
+    echo "     - project names to be added are specified as <tags>"
+    echo "     - command is specified in message <text>:"
+    echo "       - \"add\" - adds project(s) and its repositories to the configuration."
+    echo "                   If the project already exists, refresh list of its repositories."
+    echo "       - \"delete\" - removes project(s) and its repositores from the configuration"
+    echo "                      Also deletes its data under data root (but not the source code)."
+    echo "       - \"indexed\" - mark the project(s) as indexed so it becomes visible in the UI"
```
tulinkry: + info in README.txt

web/menu.jspf
```diff
@@ -93,11 +93,15 @@ org.opensolaris.opengrok.web.Util"
...
         Math.min(15, projectsSize) %>"><%
         SortedSet<String> pRequested = cfg.getRequestedProjects();
-        for ( Group group : ph.getGroups() ) {
+        for (Group group : ph.getGroups()) {
             Set<Project> groupProjects = ph.getAllGrouped(group);
-            if ( groupProjects.size() > 0) {
+            if (groupProjects.size() > 0) {
                 %><optgroup label="<%= group.getName() %>"><%
                 for (Project p : groupProjects) {
+                    if (!p.isIndexed()) {
```
tulinkry: There can be a situation when `groupProjects.size() > 0` this is true but with this `if` statement no project will be printed. Is that ok?

web/menu.jspf
```diff
@@ -117,7 +121,11 @@ org.opensolaris.opengrok.web.Util"
         if (ph.getGroups().size() > 0 && ph.getAllUngrouped().size() > 0) {
             %><optgroup label="Other"><%
         }
-        for ( Project p : ph.getAllUngrouped()) {
+        for (Project p : ph.getAllUngrouped()) {
```
tulinkry: same comment as above

tulinkry:
> add a new project foo:
./tools/Messages -t foo -n project add

what if I search the project manually now after the first step (for example I have a wrong url containing the project name)?

what if I access the xref manually now? especially the root xref; is the project visible there now?

is there any problem with autorization of an existing yet not searchable project?

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1970,7 +2028,7 @@ public SuperIndexSearcher getIndexSearcher(String proj) throws IOException {
      * change so we go through the SearcherManager objects and close those where
      * the corresponding project is no longer present.
      */
-    private void refreshSearcherManagerMap() {
+    synchronized public void refreshSearcherManagerMap() {
```
tulinkry: is it necessary for this to be synchronized when the map is concurrent?

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -0,0 +1,248 @@
...
+
+                // In case this project has just been incremetally indexed,
+                // its IndexSearcher needs a poke.
+                env.maybeRefreshIndexSearchers(getTags());
+
+                env.refreshDateForLastIndexRun();
+                break;
+        }
+
+        return null;
```
tulinkry: no "useful" output for the user? like "project abcd added" etc

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -620,6 +620,15 @@ public void createCache(Collection<String> repositories) {
         createCacheReal(getReposFromString(repositories));
     }
 
+    private HistoryCache getHistoryCache() {
+        HistoryCache cache = historyCache;
+        if (cache == null) {
+            cache = new FileHistoryCache();
```
tulinkry: shouldn't this update also the `historyCache` variable?

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -0,0 +1,240 @@
...
+import org.opensolaris.opengrok.logger.LoggerFactory;
+import org.opensolaris.opengrok.util.IOUtils;
+
+
+/**
+ * @author Vladimir Kotal
+ */
+public class ProjectMessage extends Message {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RuntimeEnvironment.class);
```
vladak: fixed

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: yep

src/org/opensolaris/opengrok/history/FileHistoryCache.java
vladak: I wanted to use the dirname either in the tests or elsewhere (which is bad idea really as everything from within `IndexDatabase` et al. should go through `HistoryGuru` and not directly to `FileHistoryCache`) so made a public duplicate. Removed the public object.

vladak: I guess the `FileHistoryCacheTest` can live with hardcoded `"historycache"`. If the name is changed the test will fail anyway.

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: added

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: yes and yes :-)

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
vladak: could not figure out how to add multiple conditions - the test file needs RepositoryInstalled.MercurialInstalled.class, RepositoryInstalled.SubversionInstalled.class, RepositoryInstalled.CvsInstalled.class

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
vladak: fixed

tools/Messages
vladak: added

web/menu.jspf
vladak: I think so. Initially the menu listing can be empty and only group names will be displayed. I think this is better than suppressing empty groups. Alternatively, some count (indexed/not indexed) could be displayed.

web/menu.jspf
vladak: I wondered whether changing `ProjectHelper#cacheProjects()` to suppress the projects would be good idea (to make the counts on the index page next to the groups reflecting only indexed projects) however changing the caching might have undesired effects.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: should not be necessary (in the worst case the iterator will not reflect current state), removed.

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
vladak: fixed

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: removed this piece of code, will submit separate PR for #767

vladak:
If the project is searched after it's added and before indexed the search will result in the `Failed to initialize search. Check the index.` error.

directory listings and xrefs will be displayed normally as they are generated on demand by default.

As for the authorization I think the indexed property of a project should be checked only after authorization which it currently is, at least in menu.jspf.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -130,6 +133,8 @@
     private int messagesInTheSystem = 0;
 
     private Statistics statistics = new Statistics();
+    
+    private static IndexTimestamp indexTime = new IndexTimestamp();
```
tarzanek: the truth is, that every project has its own index, so this should be a per project variable, but this is a good step forward, so let's start with this

tarzanek:
lgtm, please merge, let's start testing this in real life

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
tulinkry: Simple way is to somehow trick it in `setUp` using Assume.assumeTrue which causes the test to be ignored when the condition is not met

The nice but more complicated is to implement the ConditionalRun annotation to be repeatable
https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html may be a starter

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
tulinkry: this is to be approached with #1390

vladak: The question is how these per-project timestamps should be reflected in the UI.

tulinkry:
> If the project is searched after it's added and before indexed the search will result in the Failed to initialize search. Check the index. error.

There should be something more intelligent - skip the project. Create an issue for this?

> Directory listings and xrefs will be displayed normally as they are generated on demand by default.

I wouldn't allow this as the not indexed project means "not ready for use". Could you create an issue to disable this for not indexed projects?

src/org/opensolaris/opengrok/history/FileHistoryCache.java
tulinkry: Yes the test can live without it. Or if you make the property protected it should be accessible from the test as it is the same package.

Doesn't NB hint you with naming conventions for constants?

web/menu.jspf
tulinkry: Yes this is exactly the place where the change should be made if necessary. But as you said if empty groups look fine to you just keep it.

For me it would be quite confusing why some groups arent displayed (unauthorized content - empty groups; no projects in the group) and some are (empty group but contains not indexed projects).

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
vladak: hacked this around by using Assume and filed #1669 for the repeatable annotation.

web/menu.jspf
vladak: The latest change I did was to modify `cacheProjects()` to filter not-indexed projects so the changes in the cycle are probably not needed anymore.

vladak:
>> If the project is searched after it's added and before indexed the search will result in the Failed to initialize search. Check the index. error.

>There should be something more intelligent - skip the project (current behaviour when the project does not exist). Create an issue for this?

Added a proper message to `searchHelper#prepareExec()`. The method receives the list already filtered by authorization framework.

>> Directory listings and xrefs will be displayed normally as they are generated on demand by default.

>I wouldn't allow this as the not indexed project means "not ready for use". Could you create an issue to disable this for not indexed projects?

I think that's fine as long as this functionality works (aka best effort) and gives proper error message when trying search.

tulinkry:
> >> If the project is searched after it's added and before indexed the search will result in the Failed to initialize search. Check the index. error.
> 
> >There should be something more intelligent - skip the project (current behaviour when the project does not exist). Create an issue for this?
> 
> Added a proper message to `searchHelper#prepareExec()`. The method receives the list already filtered by authorization framework.

Ok

> >> Directory listings and xrefs will be displayed normally as they are generated on demand by default.
> 
> >I wouldn't allow this as the not indexed project means "not ready for use". Could you create an issue to disable this for not indexed projects?
> 
> I think that's fine as long as this functionality works (aka best effort) and gives proper error message when trying search.

But what for the root xref? The project will be available for browsing but not for searching that seems inconsistent. But still why not.

vladak:
I realized (seeing that test case `PageConfigTest.java#testGetResourceFileList()` failed) that the `/xref` filtration is done via the changes done `cacheProjects()` so this should address the last concern.

Resynced with upstream master, all tests are passing now.

PR #1668, opened 2017-07-17
reflecting the scopes and folding in xref generation

### LINKED ISSUE #1627: Scopes lost in Annotate view
Once Annotate view is triggered on a page where Scopes correctly displays function definitions and then disabled, Scopes loses information about the definitions.

Also, one would expect that Scopes window should display the definitions in Annotate view.

tulinkry:
fixes #1627


tarzanek:
I think you just miss a fix for unit tests ... otherwise it looks OK

tulinkry:
I adapted the tests by turning off scopes and I added one tests where scopes are enabled

vladak:
From maintenance perspective I don't like the addition of verbatim HTML code too much. Ideally it should be parsed to be get the scopes.

tulinkry:
For test purpose this seems ok to me now, taken into account that I only followed the other test case in this package. Of course a kind of less strict test would be welcomed.

src/org/opensolaris/opengrok/analysis/plain/AbstractSourceCodeAnalyzer.java
```diff
@@ -76,6 +77,8 @@ static protected void writeXref(JFlexXref lxref, Reader in, Writer out, Definiti
             lxref.reInit(in);
             lxref.annotation = annotation;
             lxref.project = project;
+            lxref.setScopesEnabled(RuntimeEnvironment.getInstance().isScopesEnabled());
```
vladak: As far as I understand this and `AnalyzerGuru.java#writeXref()` are used only for testing. Otherwise during normal indexing the xref is created via this path:

```
IndexDatabase#addFile()
  FileAnalyzer fa = AnalyzerGuru.getAnalyzer() // fa can be e.g. CAnalyzer
  AnalyzerGuru#populateDocument()
    fa.analyze()
    AbstractSourceCodeAnalyzer#analyze()
      super.analyze()
        PlainAnalyzer#analyze()
          writeXref()
            xref = newXref() // xref can be e.g. CXref
            JFlexXref.write()

```

where the scopes/folding properties are in effect thanks to the inheritance and these lines in `IndexDatabase#addFile()`:

```
662          fa.setScopesEnabled(RuntimeEnvironment.getInstance().isScopesEnabled());
663          fa.setFoldingEnabled(RuntimeEnvironment.getInstance().isFoldingEnabled());
```

So having these properties set into the `writeXref()` seems like short-circuiting however there is no other way given the method is static.

The only nit I have is that the call to `reInit()` in `JFlexXref` above contains:

```
        scopes = new Scopes();
```

which is not necessary when scopes are disabled. Granted, this method does not do anything however it's a bit of wasted memory.

src/org/opensolaris/opengrok/analysis/JFlexXref.java
```diff
@@ -236,7 +236,7 @@ protected String getProjectPostfix(boolean encoded) {
     }
 
     protected void startScope() {
-        if (scopesEnabled && scope == null) {
+        if (scopesEnabled && scope == null && defs != null) {
```
vladak: is this fixing any misbehavior ?

Issue #1666, opened 2017-07-14
Travis failed with NPE in IndexerRepoTest


tulinkry:
Yesterday our build failed (after integrating #1643 @vladak), not sure what happened but we shouldn't forget about it.

https://travis-ci.org/OpenGrok/OpenGrok/builds/253203610

```log
[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 6.552 s <<< FAILURE! - in org.opensolaris.opengrok.index.IndexerRepoTest
[ERROR] testMainWithoutH(org.opensolaris.opengrok.index.IndexerRepoTest)  Time elapsed: 2.751 s  <<< ERROR!
java.lang.NullPointerException
	at org.opensolaris.opengrok.index.IndexerRepoTest.checkNumberOfThreads(IndexerRepoTest.java:82)
	at org.opensolaris.opengrok.index.IndexerRepoTest.testMainWithoutH(IndexerRepoTest.java:113)
```

vladak:
The method looks like this:

```
71      private void checkNumberOfThreads() {
72          /*
73           * There should not be any threads in the renamed pool now.
74           * We need to check it like this since the test framework tears
75           * down the threads at the end of the test case run so any
76           * hangs due to executors not being shut down would not be visible.
77           */
78          ThreadGroup mainGroup = Thread.currentThread().getThreadGroup();
79          Thread[] threads = new Thread[mainGroup.activeCount()];
80          mainGroup.enumerate(threads);
81          for (int i = 0; i < threads.length; i++) {
82              if (threads[i].getName() == null) {
83                  continue;
84              }
85              assertEquals(false, threads[i].getName().contains("renamed-handling"));
86          }
87      }
```

So it seems `threads[i]` for a `i` was `null`.

PR #1673, opened 2017-07-19
use ctags binary from RuntimeEnvironment in IndexerRepoTest.java

vladak:
When investigating #1666 I found out that ctags binary executed by tests in  IndexerRepoTest.java is not necessarily the one set by `setCtags()` in these tests. This is because the indexer is run via `Indexer.main()` with distinct set of args that are supplied directly so settings in RuntimeEnvironment such as ctags binary do not get into effect unless specified as these args.

Also, given ctags are checked in individual tests, the `@BeforeClass` check is unnecessary.


tarzanek:
lgtm, merging
I hope helius will still work too ;)

PR #1675, opened 2017-07-21
remove historycache entries automatically

### LINKED ISSUE #767: remove history cache automatically

vladak:
fixes #767

Also, done some semi-related code cleanup.

### DIFF
--- a/src/org/opensolaris/opengrok/history/HistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/HistoryCache.java
@@ -127,7 +127,7 @@ interface HistoryCache {
 
     /**
      * Clear entry for single file from history cache.
-     * @param file path to the file
+     * @param file path to the file relative to the source root
      */
     void clearFile(String file);
 
--- a/src/org/opensolaris/opengrok/history/HistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/HistoryCache.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2006, 2014, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2006, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.history;
 
--- a/src/org/opensolaris/opengrok/history/HistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/HistoryCache.java
@@ -126,6 +126,12 @@ interface HistoryCache {
     void clear(Repository repository) throws HistoryException;
 
     /**
+     * Clear entry for single file from history cache.
+     * @param file path to the file
+     */
+    void clearFile(String file);
+
+    /**
      * Get a string with information about the history cache.
      *
      * @return a free form text string describing the history cache instance
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -81,6 +81,7 @@ public final class HistoryGuru {
         HistoryCache cache = null;
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         scanningDepth = env.getScanningDepth();
+
         if (env.useHistoryCache()) {
             cache = new FileHistoryCache();
 
@@ -627,12 +628,17 @@ public final class HistoryGuru {
      * @throws HistoryException 
      */
     public List<Repository> clearCache(Collection<String> repositories) throws HistoryException {
-        List<Repository> repos = getReposFromString(repositories);
+        List<Repository> repos = new ArrayList<>();
         HistoryCache cache = historyCache;
 
-        for (Repository r : repos) {
+        if (!useCache()) {
+            return repos;
+        }
+
+        for (Repository r : getReposFromString(repositories)) {
             try {
                 cache.clear(r);
+                repos.add(r);
                 LOGGER.log(Level.INFO,
                         "History cache for {0} cleared.", r.getDirectoryName());
             } catch (HistoryException e) {
@@ -646,6 +652,10 @@ public final class HistoryGuru {
     }
 
     public void clearCacheFile(String path) {
+        if (!useCache()) {
+            return;
+        }
+
         historyCache.clearFile(path);
     }
 
@@ -656,6 +666,10 @@ public final class HistoryGuru {
      * @throws HistoryException 
      */
     public void removeCache(Collection<String> repositories) throws HistoryException {
+        if (!useCache()) {
+            return;
+        }
+
         List<Repository> repos = clearCache(repositories);
         invalidateRepositories(repos);
     }
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -620,6 +620,15 @@ public final class HistoryGuru {
         createCacheReal(getReposFromString(repositories));
     }
 
+    private HistoryCache getHistoryCache() {
+        HistoryCache cache = historyCache;
+        if (cache == null) {
+            cache = new FileHistoryCache();
+        }
+
+        return cache;
+    }
+
     /**
      * Remove history data for a list of repositories
      * @param repositories list of repository paths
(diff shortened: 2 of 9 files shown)


src/org/opensolaris/opengrok/history/HistoryCache.java
```diff
@@ -126,6 +126,12 @@ String getLatestCachedRevision(Repository repository)
     void clear(Repository repository) throws HistoryException;
 
     /**
+     * Clear entry for single file from history cache.
+     * @param file path to the file
+     */
+    void clearFile(String file);
```
tulinkry: why isn't this a `File` instead of a path?

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
```diff
@@ -37,6 +38,9 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import org.opensolaris.opengrok.configuration.Project;
+import org.opensolaris.opengrok.history.HistoryGuru;
+import org.opensolaris.opengrok.history.RepositoryFactory;
```
tulinkry: import ordering is wrong in this case I think

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
```diff
@@ -45,22 +49,31 @@
 
     private static TestRepository repository;
 
+    private final static String ctagsProperty = "org.opensolaris.opengrok.analysis.Ctags";
```
tulinkry: isn't this declared somewhere else? Ctags or elsewhere?

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -629,9 +629,7 @@ public void createCache(Collection<String> repositories) {
     public List<Repository> clearCache(Collection<String> repositories) throws HistoryException {
         List<Repository> repos = getReposFromString(repositories);
         HistoryCache cache = historyCache;
-        if (cache == null) {
```
tulinkry: why this is not needed anymore?

src/org/opensolaris/opengrok/history/HistoryCache.java
tulinkry: any kind of return value which could be useful for the caller in the future?

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
```diff
@@ -77,27 +90,85 @@ public static void tearDownClass() throws Exception {
...
-        File f2 = new File(repository.getSourceRoot() + "/c/foobar.d");
+        File f2 = new File(repository.getSourceRoot() + "/git/foobar.d");
         Definitions defs2 = IndexDatabase.getDefinitions(f2);
         assertNull(defs2);
     }
+
+    private void checkDataExistence(String fileName, boolean shouldExist) {
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+        for (String dirName : new String[]{"historycache", IndexDatabase.XREF_DIR}) {
```
tulinkry: `FileHistoryCache.HISTORY_DIR`?

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
```diff
@@ -77,27 +90,85 @@ public static void tearDownClass() throws Exception {
...
+        Assert.assertTrue(new File(gitRoot, fileName).exists());
+
+        // Check that the file was indexed successfully in terms of generated data.
+        checkDataExistence(projectName + File.separator + fileName, true);
+        origNumFiles = idb.getNumFiles();
+        Assert.assertEquals(6, origNumFiles);
+
+        // Remove the file and reindex using IndexDatabase directly.
+        File file = new File(repository.getSourceRoot(), projectName + File.separator + fileName);
+        file.delete();
```
tulinkry: assertTrue(file.delete(), "file not deleted");

src/org/opensolaris/opengrok/history/HistoryCache.java
vladak: the path is relative to source root. It cannot be a `File` object because that would bypass HistoryGuru as a generic interface to concrete history cache backend (remember the days when OpenGrok supported JavaDB history cache next to File history cache).

vladak: It mimicks the `clear()` interface. It just logs the error and goes on. The same thing is done for removal of xref files. It can be changed to `boolean` or `int` and just ignore the values in `IndexDatabase#removeFile()` however this seems unnecessary to me.

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
vladak: Netbeans does not complain however they should be better grouped.

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
vladak: yep, it's defined in number of places throughout `test/` directory. filed #1694

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
vladak: I'd rather keep it not exposed. Changing the field to `protected` does not help.

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
vladak: I don't like having actions done in asserts, makes the code less readable I think.

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: I think this can't happen - the `historyCache` is created in `HistoryGuru` constructor.

tarzanek:
feel free to merge, lgtm
(I hope this won't be too memory hungry)

src/org/opensolaris/opengrok/history/HistoryCache.java
tulinkry: ok

tulinkry: okey

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
tulinkry: weird

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
tulinkry: perfect

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
tulinkry: ok

test/org/opensolaris/opengrok/index/IndexDatabaseTest.java
tulinkry: ok

src/org/opensolaris/opengrok/history/HistoryGuru.java
tulinkry: in the constructor it is in an statement `if (env.useHistoryCache()) {`, is it ensured that this `clearCache` is not called when this is not in effect?

vladak:
As for "memory hungy" I bet you're referring to `IndexDatabase#getFiles()` however that is dead code and it is not called.

src/org/opensolaris/opengrok/history/HistoryCache.java
vladak: comment added

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: good catch; added bunch of `useCache()` calls.

tulinkry:
ok to me now

PR #1681, opened 2017-07-24
fixed header for popup windows with scrollable content inside

### LINKED ISSUE #1350: Navigate window does not have fixed header
It's probably all of the windows. When the content is longer then the max-height of the window the header does not stay at the top and scrolls away as the user scrolls down.

screenshot from 2017-01-27 12-21-36
screenshot from 2017-01-27 12-21-52

tulinkry:
fixes #1350

Navigate window now has fixed header and scrollable content. Other windows don't need it.

screenshot from 2017-07-24 16-05-50

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -820,16 +820,6 @@ a.scope            { /* scope              */ color: steelblue; font-weight: bol
     margin-bottom: 1em;
 }
 
-/* "Navigate Style" */
-.diff_navigation_style {
-    border: solid 1px #c0c0c0;
-    border-radius: 5px;
-    box-shadow: 10px 10px 5px #888888;
-    background-color: rgb(255,255,204);
-    position: absolute;
-    padding: 1em;
-}
-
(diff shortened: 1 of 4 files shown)


web/default/style.css
```diff
@@ -819,10 +809,34 @@ a.scope            { /* scope              */ color: steelblue; font-weight: bol
     position: fixed;
     font-size: 12px;
     font-family: monospace;
-    overflow: auto;
+    overflow: hidden;
```
vladak: if the content is clipped, what mechanism will add the scrollbar now ?

tulinkry: The content is in the body and that is set to overflow auto

tulinkry:
conflicts resolved

tarzanek:
@tulinkry one more please, will merge after you resolve it

tarzanek:
@tulinkry :) one more conflict resolve, please (and ping me on IM so I will merge asap)

PR #1682, opened 2017-07-24
send Message before indexing with "set projectsEnabled = true" if run…

### LINKED ISSUE #1678: Projects have to be enabled - incovenient migration
Using OpenGrok without projects (or now just indexing after #1663) indexer emits several times an exception (one per project I guess):

```log
Jul 24, 2017 11:50:25 AM org.opensolaris.opengrok.index.IndexDatabase$2 run
SEVERE: Problem updating lucene index database: 
java.io.IOException: Message was not accepted by the remote server - Projects have to be enabled in order to use this message (error 0x4).
	at org.opensolaris.opengrok.configuration.messages.Message.throwIfError(Message.java:385)
	at org.opensolaris.opengrok.configuration.messages.Message.write(Message.java:307)
	at org.opensolaris.opengrok.index.IndexDatabase.update(IndexDatabase.java:423)
	at org.opensolaris.opengrok.index.IndexDatabase$2.run(IndexDatabase.java:186)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

As a result no project is displayed on the main page and one has to run the indexer twice or restart the webapplication (the configuration.xml already contains the projectsEnabled = true after the first run).

Problems:
1. It should not be a problem to turn on projects when application is already running
2. It should not be a problem to migrate to PR #1663

vladak:
…ning with -P

fixes #1678

### DIFF
--- a/src/org/opensolaris/opengrok/index/Indexer.java
+++ b/src/org/opensolaris/opengrok/index/Indexer.java
@@ -48,6 +48,7 @@ import org.opensolaris.opengrok.analysis.AnalyzerGuru;
 import org.opensolaris.opengrok.configuration.Configuration;
 import org.opensolaris.opengrok.configuration.Project;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
+import org.opensolaris.opengrok.configuration.messages.Message;
 import org.opensolaris.opengrok.history.HistoryException;
 import org.opensolaris.opengrok.history.HistoryGuru;
 import org.opensolaris.opengrok.history.Repository;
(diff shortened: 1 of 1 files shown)


src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -582,6 +583,18 @@ public static void main(String argv[]) {
...
+                // If the webapp is running with a config that does not contain
+                // 'projectsEnabled' property (case of upgrade or transition
+                // from project-less config to one with projects), set the property
+                // using a message so that the 'project/indexed' messages
+                // emitted during indexing do not cause validation error.
+                if (addProjects && host != null && port > 0) {
+                    Message m = Message.createMessage("config");
+                    m.addTag("set");
+                    m.setText("projectsEnabled = true");
+                    m.write(host, port);
```
tulinkry: If this fails sending will the index process continue?

vladak: If e.g. `write()` throws exception, It will stop.

vladak: ..which is good I think as it will avoid surprise at the end of the indexing (if the webapp is permanently down).

tulinkry: I think it is not because that completely prevents you to prepare the configuration offline.

vladak: If the host/port is not set the message will not be sent.

tulinkry: Ok, my bad. Let's merge.

PR #1692, opened 2017-07-25
making the conditional rules in tests repeatable

### LINKED ISSUE #1669: @ConditionalRun needs to be repeatable
`ProjectMessageTest` needs multiple conditions (check whether Git, Mercurial and Subversion repositories are working). In order to achieve this in elegant way, `@ConditionalRun` annotation needs to be repeatable

tulinkry:
fixes #1669

### DIFF
--- a/test/org/opensolaris/opengrok/condition/ConditionalRun.java
+++ b/test/org/opensolaris/opengrok/condition/ConditionalRun.java
@@ -18,19 +18,25 @@
  */
 
  /*
- * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.condition;
 
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Repeatable annotation for automatic test skipping.
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(value = {ElementType.METHOD, ElementType.TYPE})
+@Repeatable(ConditionalRunRepeatable.class)
 @Documented
 public @interface ConditionalRun {
+
     Class<? extends RunCondition> condition();
 }
--- a/test/org/opensolaris/opengrok/condition/ConditionalRun.java
+++ b/test/org/opensolaris/opengrok/condition/ConditionalRun.java
@@ -24,6 +24,7 @@ package org.opensolaris.opengrok.condition;
 
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
 import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
@@ -33,8 +34,9 @@ import java.lang.annotation.Target;
  * Repeatable annotation for automatic test skipping.
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(value = {ElementType.METHOD, ElementType.TYPE})
+@Target(value = {ElementType.METHOD, ElementType.TYPE, ElementType.PACKAGE})
 @Repeatable(ConditionalRunRepeatable.class)
+@Inherited
 @Documented
 public @interface ConditionalRun {
 
--- a/test/org/opensolaris/opengrok/condition/ConditionalRunRepeatable.java
+++ b/test/org/opensolaris/opengrok/condition/ConditionalRunRepeatable.java
@@ -18,11 +18,26 @@
  */
 
  /*
- * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opensolaris.opengrok.condition;
 
-public interface RunCondition {
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 
-    public boolean isSatisfied();
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Annotation allowing the {@link ConditionalRun} to be repeatable annotation.
+ *
+ * @author Krystof Tulinger
+ */
+@Retention(value = RUNTIME)
+@Documented
+@Target(value = {ElementType.METHOD, ElementType.TYPE})
+public @interface ConditionalRunRepeatable {
+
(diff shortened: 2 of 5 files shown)


tulinkry:
One downside of this is that there will always be 2 skipped tests in junit (which indicates correct behaviour).

vladak:
While you are at it could you use the repeatable condition in `ProjectMessageTest` ?

test/org/opensolaris/opengrok/condition/ConditionalRunRule.java
```diff
@@ -64,34 +68,70 @@ public Statement apply(Statement aStatement, Description aDescription) {
...
+            throw new RuntimeException(ex);
+        }
     }
 
     private static RunCondition getIgnoreConditionOnMethod(Description aDescription) {
-        ConditionalRun annotation = aDescription.getAnnotation(ConditionalRun.class);
-        return new IgnoreConditionCreator(aDescription.getTestClass(), annotation).create();
+        try {
+            // this is possible because test methods must not have any argument
+            ConditionalRun[] annotation = aDescription.getTestClass().getMethod(aDescription.getMethodName()).getAnnotationsByType(ConditionalRun.class);
```
vladak: annotations ?

tulinkry: fixed

tulinkry:
> While you are at it could you use the repeatable condition in ProjectMessageTest ?

I wanted to do this in a different PR.

tulinkry:
@vladak ?

test/org/opensolaris/opengrok/condition/RunningRepeatableConditionTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @ConditionalRun(condition = TrueRunCondition.class)
+    @ConditionalRun(condition = FalseRunCondition.class)
+    @ConditionalRun(condition = TrueRunCondition.class)
+    @Test
+    public void testSkippedTest() throws NoSuchMethodException {
+        Assert.assertTrue("This test must be skipped", false);
+    }
+
+    public static class TrueRunCondition implements RunCondition {
```
vladak: somehow it seems strange to me having the `{True,False}RunCondition` being shared like this. Why not have them in separate file ?

vladak:
Separate PR for the actual use in tests is fine, as well as separate commit in this PR.

test/org/opensolaris/opengrok/condition/ConditionalRunRule.java
```diff
@@ -64,34 +68,70 @@ public Statement apply(Statement aStatement, Description aDescription) {
...
+            }
+            return true;
+        }
+    }
+
+    protected static class IgnoreConditionCreator {
 
         private final Class<?> mTestClass;
-        private final Class<? extends RunCondition> conditionType;
+        private final List<Class<? extends RunCondition>> conditionType;
```
vladak: rename to signify it is now holding more conditions ?

test/org/opensolaris/opengrok/condition/ConditionalRunRule.java
```diff
@@ -64,34 +68,70 @@ public Statement apply(Statement aStatement, Description aDescription) {
...
 
-        IgnoreConditionCreator(Class<?> aTestClass, ConditionalRun annotation) {
+        public IgnoreConditionCreator(Class<?> aTestClass, ConditionalRun[] annotation) {
             this.mTestClass = aTestClass;
-            this.conditionType = annotation.condition();
+            this.conditionType = new ArrayList<>(annotation.length);
+            for (int i = 0; i < annotation.length; i++) {
```
vladak: would `addAll()` work here ?

test/org/opensolaris/opengrok/condition/ConditionalRun.java
```diff
@@ -18,19 +18,25 @@
...
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Repeatable annotation for automatic test skipping.
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(value = {ElementType.METHOD, ElementType.TYPE})
```
vladak: just wondering if expanding this with `PACKAGE` would make sense

test/org/opensolaris/opengrok/condition/ConditionalRun.java
```diff
@@ -18,19 +18,25 @@
...
+@Repeatable(ConditionalRunRepeatable.class)
 @Documented
 public @interface ConditionalRun {
```
vladak: use [`Inherited`](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Inherited.html) ?

test/org/opensolaris/opengrok/condition/ConditionalRunRepeatable.java
```diff
@@ -0,0 +1,43 @@
...
+ * Annotation allowing the {@link ConditionalRun} to be repeatable annotation.
+ *
+ * @author Krystof Tulinger
+ */
+@Retention(value = RUNTIME)
+@Documented
+@Target(value = {ElementType.METHOD, ElementType.TYPE})
+public @interface ConditionalRunRepeatable {
+
+    ConditionalRun[] value();
```
vladak: I am slightly confused by the `value` used in this file. Does it have any special significance to name the array so ?

vladak: Oh, this seems to be the rule for containing annotation type as per https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html

tulinkry: yes it must be like this

test/org/opensolaris/opengrok/condition/ConditionalRunRule.java
tulinkry: fixed

test/org/opensolaris/opengrok/condition/ConditionalRunRule.java
tulinkry: no cause you still need to call annotation[i].condoition()

test/org/opensolaris/opengrok/condition/ConditionalRun.java
tulinkry: i have no idea how this should behave but we can experiment with it

test/org/opensolaris/opengrok/condition/ConditionalRun.java
tulinkry: added

test/org/opensolaris/opengrok/condition/RunningRepeatableConditionTest.java
tulinkry: I made them protected. There's no need to use them anywhere else.

tarzanek:
let's merge

PR #1693, opened 2017-07-25
project customizations in read-only configuration

tulinkry:
User can specify project customizations in read only configuration by opening the `projects` property and filling the map entry according to the project name and desired customizations.

```xml
  <void property="projects">
   <void method="put">
    <string>OpenGrok</string>
     <void property="navigateWindowEnabled">
      <boolean>true</boolean>
     </void>
    </object>
   </void>
  </void>
```

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Project.java
+++ b/src/org/opensolaris/opengrok/configuration/Project.java
@@ -63,9 +63,9 @@ public class Project implements Comparable<Project>, Nameable, Serializable {
     private boolean navigateWindowEnabled = false;
 
     /**
-     * This marks the project as (not)ready before initial index is done.
-     * this is to avoid all/multi-project searches referencing this project
-     * from failing.
+     * This marks the project as (not)ready before initial index is done. this
+     * is to avoid all/multi-project searches referencing this project from
+     * failing.
      */
     private boolean indexed = false;
 
@@ -78,13 +78,38 @@ public class Project implements Comparable<Project>, Nameable, Serializable {
     public Project() {
     }
 
+    /**
+     * Create a project with given name.
+     *
+     * @param name the name of the project
+     */
     public Project(String name) {
         this.name = name;
     }
 
+    /**
+     * Create a project with given name and path
+     *
+     * @param name the name of the project
+     * @param path the path of the project relative to the source root
+     */
     public Project(String name, String path) {
         this.name = name;
         this.path = path;
+    }
+
+    /**
+     * Create a project with given name and path and default configuration
+     * values.
+     *
+     * @param name the name of the project
+     * @param path the path of the project relative to the source root
+     * @param cfg configuration containing the default values for project
+     * properties
+     */
+    public Project(String name, String path, Configuration cfg) {
+        this(name, path);
+        completeWithDefaults(cfg);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -757,7 +766,11 @@ public void prepareIndexer(RuntimeEnvironment env,
                     // This is an existing object. Reuse the old project,
                     // possibly with customizations, instead of creating a
                     // new with default values.
-                    projects.put(name, oldProjects.get(name));
+                    Project p = oldProjects.get(name);
+                    // Set the path which may be missing in the project if
+                    // the project was customized.
+                    p.setPath(path);
```
vladak: what about `tabSize` ? Perhaps more generic approach with filling in all fields (copy constructor ?) would be in order.

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -535,6 +534,16 @@ public static void main(String argv[]) {
                     }
                 }
 
+                // If an user used customizations for projects he perhaps just
+                // used the key value for project without a name but the code
```
vladak: do you have an example on how the key value config looks like ?

tulinkry: In the PR description:

```xml
  <void property="projects">
   <void method="put">
    <string>OpenGrok</string>
     <void property="navigateWindowEnabled">
      <boolean>true</boolean>
     </void>
    </object>
   </void>
  </void>
```

src/org/opensolaris/opengrok/index/Indexer.java
tulinkry: I worked with it a litle and I added a method for such case

src/org/opensolaris/opengrok/index/Indexer.java
vladak: I see

test/org/opensolaris/opengrok/configuration/ProjectTest.java
```diff
@@ -134,4 +132,52 @@ public void testGetProjectDescriptions() {
...
         assertEquals(2, descs.size());
     }
+
+    /**
+     * Insert the value from configuration.
+     */
+    @Test
+    public void testMergeProjects1() {
+        Configuration cfg = new Configuration();
+        cfg.setTabSize(8);
```
vladak: using some crazy number as a value might be more robust (in case default value gets set to 8)

src/org/opensolaris/opengrok/configuration/Project.java
```diff
@@ -224,6 +249,29 @@ public void addGroup(Group group) {
...
+     * @param cfg configuration with default values if applicable
+     */
+    final public void completeWithDefaults(Configuration cfg) {
+        Configuration defaultCfg = new Configuration();
+        /**
+         * Choosing strategy for properties (tabSize here):
+         * <pre>
+         * this       cfg        defaultCfg
+         * ===========================================
+         *  |5|        4             0
```
vladak: Not sure the table helps to understand; at least for me it makes things confusing. Maybe some extended description would help ? Or add another column with resulting value ?

test/org/opensolaris/opengrok/configuration/ProjectTest.java
vladak: or use something like:
```
final Configuration defaultConfig = new Configuration();
Configuration cfg = new Configuration();
cfg.setTabSize(defaultConfig.getTabSize() + 3333);
```
to make sure the value differs from default config.

test/org/opensolaris/opengrok/configuration/ProjectTest.java
```diff
@@ -134,4 +132,52 @@ public void testGetProjectDescriptions() {
...
+    /**
+     * Do not overwrite customized project property.
+     */
+    @Test
+    public void testMergeProjects2() {
+        Configuration cfg = new Configuration();
+        cfg.setTabSize(4);
+
+        Project p1 = new Project();
+        p1.setTabSize(5);
```
vladak: ditto

vladak:
apart from the nits it looks good, pls fix them and merge.

test/org/opensolaris/opengrok/configuration/ProjectTest.java
tulinkry: fixed

test/org/opensolaris/opengrok/configuration/ProjectTest.java
tulinkry: fixed

src/org/opensolaris/opengrok/configuration/Project.java
tulinkry: actually the table was somewhat wrong

I corrected the table and added a brief description of what's happening. The point was to fill the project properities only when it is set to default.

tulinkry:
@vladak ?

tarzanek:
merging, looks like this got lost in emails

PR #1698, opened 2017-07-26
fixing serialization bug with order of class properties

tulinkry:
The serialization bug we found is based on unexpected order of the field serialization in xml (and then deserializing them).

This has three possible solutions:
 - marking the properties methods with `@Transient` annotation
 - removing the setter method for these properties
 - marking the properties as transient like in this PR

Note that just using `transient` keyword is not enough because the XML serialization is not a java serialization mechanism and it ignores this modifier. The serialization algorithm is based on presence of "get" and "set" methods to the corresponding fields.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Group.java
+++ b/src/org/opensolaris/opengrok/configuration/Group.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import org.opensolaris.opengrok.util.ClassUtil;
 
 /**
  * Placeholder for the information about subgroups of projects and repositories.
@@ -37,6 +38,10 @@ import java.util.regex.PatternSyntaxException;
  * @version $Revision$
  */
 public class Group implements Comparable<Group>, Nameable {
+
+    static {
+        ClassUtil.remarkTransientFields(Group.class);
+    }
 
     private String name;
     /**
@@ -60,13 +65,13 @@ public class Group implements Comparable<Group>, Nameable {
      */
     private Pattern compiledPattern = Pattern.compile("()");
     private Group parent;
-    private int flag;
-
     private Set<Group> subgroups = new TreeSet<>();
-    private Set<Group> descendants = new TreeSet<>();
-    private Set<Project> projects = new TreeSet<>();
-    private Set<Project> repositories = new TreeSet<>();
-    private Set<Group> parents;
+
+    private transient int flag;
+    private transient Set<Group> descendants = new TreeSet<>();
+    private transient Set<Project> projects = new TreeSet<>();
+    private transient Set<Project> repositories = new TreeSet<>();
+    private transient Set<Group> parents;
 
     /**
      * No-arg constructor is needed for deserialization.
--- a/src/org/opensolaris/opengrok/util/ClassUtil.java
+++ b/src/org/opensolaris/opengrok/util/ClassUtil.java
@@ -28,13 +28,18 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.opensolaris.opengrok.configuration.Project;
+import org.opensolaris.opengrok.logger.LoggerFactory;
 
 /**
  *
  * @author Krystof Tulinger
  */
 public class ClassUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClassUtil.class);
 
     /**
      * Mark all transient fields in {@code targetClass} as @Transient for the
@@ -61,6 +66,7 @@ public class ClassUtil {
                 }
             }
         } catch (IntrospectionException ex) {
+            LOGGER.log(Level.WARNING, "An exception ocurred during remarking transient fields:", ex);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/org/opensolaris/opengrok/configuration/Group.java
```diff
@@ -38,6 +39,10 @@
  */
 public class Group implements Comparable<Group>, Nameable {
 
+    static {
+        ClassUtil.remarkTransientFields(Group.class);
```
vladak: is there any way this could be done via inheritance ?

src/org/opensolaris/opengrok/util/ClassUtil.java
```diff
@@ -0,0 +1,66 @@
...
+            for (Field f : Project.class.getDeclaredFields()) {
+                if (Modifier.isTransient(f.getModifiers())) {
+                    for (int i = 0; i < propertyDescriptors.length; ++i) {
+                        if (propertyDescriptors[i].getName().equals(f.getName())) {
+                            propertyDescriptors[i].setValue("transient", Boolean.TRUE);
+                        }
+                    }
+                }
+            }
+        } catch (IntrospectionException ex) {
```
vladak: log this somewhere ?

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.IOException;
+import java.util.LinkedList;
+import junit.framework.AssertionFailedError;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ *
```
vladak: add comment about what bug this reproduces and how exactly

vladak: actually, this could be generic set of tests for "serialization"

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * @author Krystof Tulinger
+ */
+public class ConfigurationTest {
+
+    @Test
```
vladak: the per-test specific comment should be here

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        final LinkedList<Exception> exceptions = new LinkedList<>();
+
+        try (ByteArrayInputStream in = new ByteArrayInputStream(out.toByteArray());
+                XMLDecoder dec = new XMLDecoder(in, null, (Exception e) -> {
+                    exceptions.addLast(e);
+                })) {
+
+            cfg = (Configuration) dec.readObject();
+            assertNotNull(cfg);
+            dec.close();
```
vladak: is this necessary given the try-with-resources above ?

vladak:
I see - https://stackoverflow.com/questions/8813631/how-do-you-mark-a-property-as-transient-in-a-javabean-in-order-to-avoid-serializ

This is a bit of a hack with nice surface :-)

src/org/opensolaris/opengrok/util/ClassUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    public static void remarkTransientFields(Class targetClass) {
+        try {
+            BeanInfo info;
+            info = Introspector.getBeanInfo(targetClass);
+            PropertyDescriptor[] propertyDescriptors = info.getPropertyDescriptors();
+            for (Field f : Project.class.getDeclaredFields()) {
+                if (Modifier.isTransient(f.getModifiers())) {
+                    for (int i = 0; i < propertyDescriptors.length; ++i) {
+                        if (propertyDescriptors[i].getName().equals(f.getName())) {
+                            propertyDescriptors[i].setValue("transient", Boolean.TRUE);
```
vladak: is the "transient" somewhere defined as constant ?

src/org/opensolaris/opengrok/util/ClassUtil.java
tulinkry: done

src/org/opensolaris/opengrok/configuration/Group.java
tulinkry: unfortunately not as efficient as this is

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
tulinkry: comment added to the test method

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
tulinkry: added

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
tulinkry: no, removed

src/org/opensolaris/opengrok/util/ClassUtil.java
tulinkry: digging in the codes of java this string has private access; so no

test/org/opensolaris/opengrok/configuration/ConfigurationTest.java
tulinkry: for the tests there is an issue #1699

PR #1702, opened 2017-07-28
removeFile() needs to be smarter to allow incremental history indexing

vladak:
The important piece is this line in `indexDown()`:

```
removeFile(!termPath.equals(path));
```


tarzanek:
thank you sir!
lgtm, merging

PR #1706, opened 2017-07-31
CSS cleanup

vladak:
Obviously, the `#whole_header` height and `#content`'s `margin-top` need to match. Sadly, there is no way how to compute one from the other in CSS (https://stackoverflow.com/questions/23076550/css-use-calc-to-keep-widths-of-elements-the-same)

Also, to avoid fixing the non-default styles, I opted for their removal :-)


tarzanek:
hmm, there are sites/people who use other styles
do we really want to do this?

vladak:
The non default styles are getting broken as it seems no one is using them. The styling infrastructure is left in place so anyone can still add the styles. Also, The issue has been opened for some time and there were no objections.

tarzanek:
then let's merge this

PR #1715, opened 2017-08-02
Added support for paginating the JSONSearchServlet

jpitz:
Right now, this search servlet doesn't support paginating results. This can be problematic for large code bases that have more then the hard configured limit of MAXRESULTS. Paginating will keep the endpoint performant and allow these larger companies to get at data easily.

### DIFF
--- a/src/org/opensolaris/opengrok/web/JSONSearchServlet.java
+++ b/src/org/opensolaris/opengrok/web/JSONSearchServlet.java
@@ -124,15 +124,12 @@ public class JSONSearchServlet extends HttpServlet {
 
             int pageStart = getIntParameter(req, PARAM_START, 0);
 
-            int maxResults = MAX_RESULTS;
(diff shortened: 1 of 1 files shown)


jpitz:
Submitted Signed OCA

vladak:
while you're at it, can use getIntParameter() for other parameters. Otherwise this looks good.

jpitz:
Can do. I wanted to keep this change as minimal as possible.

vladak:
As long as the other changes are present as separate changesets that's fine, will merge them without squashing.

jpitz:
Sounds good. I only see 1 other param in this servlet (maxResults).

The real value would be to apply the same method call across the project. That seems a bit out of scope for the course of this change. Want me to update the one other case in this class or do a separate ticket for the rest of the params?

jpitz:
Applied to the one other param in the current servlet. I can go through the rest of the project in another PR

PR #1717, opened 2017-08-03
cache history for top-level directory in repository

vladak:
This change caches history of the top-level directory of repository. This is easy to do since the history is available anyway to generate per-file history cache. New history will be incrementally merged with existing cache upon reindex.


tarzanek:
lgtm, merge

Issue #1478, opened 2017-03-22
better repository table formatting in presence of big column widths


ChristopheBordieu:
Hello
With 0.13-rc10, I manage many projects with, for each, many repos inside, with some repo paths very long.
Result is that in homepage, Mirror column is very large leading to colum Current version splitted on several lines.  Thus, each line for project or repo is very high.

![capture](https://cloud.githubusercontent.com/assets/26546513/24215096/f2a6519a-0f37-11e7-8887-f975b70b8515.PNG)

The root cause is that, by default, all repos into a project are displayed.
Is it possible to fold all projects by default ? And to unfold only clicked project ?

tulinkry:
It should be possible to limit the scan to only top level directories with this for the wrapper:
```txt
#   - OPENGROK_SCAN_DEPTH         how deep should scanning for repos go
#                                 (by default 3 directories from SRC_ROOT)
```
or this with java
```bash
# options.add(new Option('z', NUMBER, "depth of scanning for repositories in directory structure relative to source root. Default is "+Configuration.defaultScanningDepth+" ."));
java -jar opengrok.jar -z 1
```

You loose the subrepository listing with this; it's not folded/hidden or anything like that.

ChristopheBordieu:
But I want to scan my project dirs to get repos inside. My OPENGROK_SCAN_DEPTH is set to 2 for that purpose: all repos are level 2 below SRC_ROOT.
Not sure I understand your comment...
I would like the UI offers the possibility to fold or not the project directory, not showing or showing the repos inside. Default would be not showing the repos. Then clicking for example to a project, repos will be unfolded and showed.
Sorry, maybe me explanation (in French English) is not clear.

tulinkry:
Yes I understood the problem. I just tried to offer you a temporary workaround before this gets fixed in master. However you use that intentionally so it's not a solution for you.

ChristopheBordieu:
Hello,
Is there any news regarding this UI bug ?

ChristopheBordieu:
Hello,
I have played with groups thinking it could help me to get the folding feature I was requesting when entering this issue.
Actually, there is indeed a bug. I have not noticed it because I was focused on display...
So, I  explain again. M?aybe it will help you to reproduce the issue.
The structure of my src directory is:
SRC/
  projA/
      repoA1
      repoA2
      repoA3
  ProjB/
       repoB1
  projC/
       repoC1
       repoC2
   ...

[...]

PR #1718, opened 2017-08-04
tunable for displaying listing of repositories on index page

vladak:
to provide workaround for the problems raised in #1478


vladak:
Tested by:
  1. deploy webapp, display index - contains repo listing
  2. disable via `./tools/Messages -n config -t set "displayRepositories = false"` , refresh - no repos displayed
  3. enable again via `./tools/Messages -n config -t set "displayRepositories = true"`and refresh  - repos displayed again

tarzanek:
lgtm, thank you sir!

PR #1721, opened 2017-08-04
Have .opengrok_skip_history seen by AccuRev

shaehn:
Changed HistoryGuru::addRepository so that it will recognize .opengrok_skip_history files in folders prior to scanning the folder/directory.

### DIFF
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -72,8 +72,6 @@ public final class HistoryGuru {
     private Map<String, Repository> repositories = new ConcurrentHashMap<>();
 
     private final int scanningDepth;
-
-    private String ignoreRepositoryName = ".opengrok_skip_history";
     
     /**
      * Creates a new instance of HistoryGuru, and try to set the default source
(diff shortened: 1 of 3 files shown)


vladak:
The latter 2 changesets look okay, will need to look at the first one in detail.

Do you have a OCA ?

vladak:
I noticed that there are no AccuRev tests under the `test` directory - care to add at least some basic tests ?

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -73,6 +73,8 @@
 
     private final int scanningDepth;
 
+    private String ignoreRepositoryName = ".opengrok_skip_history";
```
vladak: can be `final`

vladak: and `static`

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -341,69 +343,95 @@ public void addRepositories(File[] files, Collection<RepositoryInfo> repos,
      */
```
vladak: the javadoc should probably say something about the special skip-history file.

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -341,69 +343,95 @@ public void addRepositories(File[] files, Collection<RepositoryInfo> repos,
...
-                                file.getAbsolutePath());
-                    } else if (depth <= scanningDepth) {
-                        addRepositories(subFiles, repos, ignoredNames, depth + 1);
-                    }
+                // This check for ignoring a repository is for the
+                // kind of repository, such as AccuRev, where the
+                // entire project folder layout (workspace) is
+                // considered as the repository. Here we need to
+                // look inside the folder for the magic file.
+                if( file.isDirectory() ) {
```
vladak: pls use uniform coding style - there should be no spaces around the condition

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -341,69 +343,95 @@ public void addRepositories(File[] files, Collection<RepositoryInfo> repos,
...
+                    ignoreRepository = skipRepository.exists();
                 }
+            } catch( IOException exp ) {
+                LOGGER.log(Level.WARNING,
+                    "Failed to get canonical path for {0}: {1}",
+                    new Object[]{file.getAbsolutePath(), exp.getMessage()});
+            }
+            
+            if( ignoreRepository ) {
+                LOGGER.log(Level.INFO,
```
vladak: rather than introducing another level of indentation, `continue` can be used here

src/org/opensolaris/opengrok/history/HistoryGuru.java
```diff
@@ -341,69 +343,95 @@ public void addRepositories(File[] files, Collection<RepositoryInfo> repos,
      */
     private void addRepositories(File[] files, Collection<RepositoryInfo> repos,
             IgnoredNames ignoredNames, boolean recursiveSearch, int depth) {
+
+        // This check for ignoring a repository only works for embedded
```
vladak: I think the previous algorithm is actually wrong. If the traversal descends into the directory that happens to be a repository, it already added the repository to the list, so the check for the ".opengrok_skip_history" file inside will do nothing. See pull request #1727. Maybe you can change the fix accordingly.

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: I can certainly add that, but your later comments about not using .opengrok_skip_history at all warrants a complete rethink of the whole mechanism.

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: When you say javadoc, do you mean adding info to the comment above the addRepositories method?

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: Sorry about that. This is a style I use in all the code I write normally.

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: ok

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: No, it just needs a minor change, see the pull req mentioned below.

src/org/opensolaris/opengrok/history/HistoryGuru.java
vladak: yep, see http://www.oracle.com/technetwork/articles/java/index-137868.html for details

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: Vlad, I am at a total loss right now. I am looking and answering this message via github web site and i don't see what pull request you are mentioning. (Or even how to go about looking at it)

shaehn: OK, I found the pull request #1727. Is this the one you are speaking about?

vladak: yes, that's the one.

src/org/opensolaris/opengrok/history/HistoryGuru.java
shaehn: OK, let me test my understanding. You have a pull request with the suggested changes. Do you want me to copy the code out of that and apply it to my forked repository along with adding the Unit Test you created, then promote it back into my repository so that it gets associated with my original pull request?

You sent me some other mail talking about the secret file not being a good mechanism and it should be taken care of via Configuration mechanism. Is that a longer term situation and these changes just a band-aid for now?

vladak: Yes, exactly. It will hopefully fix it for your use case. The long-term solution is something which needs more thought/testing/discussion.

shaehn:
Vlad are there any changes I am missing still here? The state seems to still be set as Changes Requested. When does an actual pull request get merged?

tarzanek:
I think @vladak just needs some time to review and start the merge,
this looks ok to me, not sure if the path resolving is needed in the IndexDatabase, if it will break something later on we can always remove it

vladak:
Looks good now, merging.

PR #1725, opened 2017-08-07
Replace long-defunct example of OpenGrok user

fitojb:
I release this trivial patch to the public domain.


tarzanek:
thank you Adolfo , merging

PR #1727, opened 2017-08-09
Historyguru ignore test

vladak:
This is mainly to supply the test case, the fix will be done in pull request #1721 I believe.


vladak:
The changes will be delivered via #1721, hence closing.

PR #1739, opened 2017-08-21
refresh current version of project's repositories upon "indexed" message

vladak:
I contemplated a bit whether to add another message to trigger the `RepositoryInfo` refresh however was concerned that there might be too many messages in the system when reindexing many projects in parallel.

Also, there is a problem when webapp is running for a while and configuration is not stored persistently after each reindex cycle that the webapp will have more fresh configuration - with refreshed RepositoryInfo objects. This is left to be solved by whatever wrapper script is used to run the indexing (not the `OpenGrok` shell script but something which is wrapping around it).


tarzanek:
lgtm , please merge

PR #1741, opened 2017-08-21
send the "indexed" message even if the project is marked as indexed

vladak:
Forgot to change the most important piece for the fix to actually work.


tarzanek:
please push, looks ok

PR #1743, opened 2017-08-22
always provide revision number for xrefs

vladak:


vladak:
I found out this does not work yet - clicking on line number erases the revision part of the URL. Needs a bit more work.

src/org/opensolaris/opengrok/web/PageConfig.java
```diff
@@ -1190,6 +1193,50 @@ public File findDataFile() {
...
+        return rev.equals(getLatestRevision());
+    }
+
+    /**
+     * Get the location of cross reference for given file containing the current
+     * revision.
+     * @return location to redirect to or null if failed
+     */
+    public String getLatestRevisionLocation() {
+        return new String(req.getContextPath() + Prefix.XREF_P +
```
tarzanek: use StringBuilder and append please

tarzanek:
please fix the commend and merge

tarzanek:
(and thank youuuuu sir! :) )

src/org/opensolaris/opengrok/web/PageConfig.java
vladak: what's the advantage of `StringBuilder` ? Seems less readable to me.

vladak:
Still need to figure out the hash URL rewrite. I suspect there is some JS meddling with the URL.

PR #1749, opened 2017-08-23
project message fixes and enhancements

vladak:
also finally gets rid of project descriptions


vladak:
The `list-indexed` can be used in a setup where solely partial indexing is used to avoid certain projects from being indexed/displayed by OpenGrok, e.g.:

```
Messages -n project "list-indexed" | while read project; do
    sync $project # performs 'hg pull -u' etc.
    OpenGrok indexpart /var/opengrok/src $project
done
```

vladak:
The changes in `ProjectMessageTest` will make running the tests a bit slower as I had to get rid of the sharing `repositories.zip` since the tests started to influence each other too much.

tarzanek:
how do the test influence each other?
do we have some misuse of junit ?
or order of tests is a problem? (tests shouldn't depend on each other and be isolated)

vladak:
The tests change the repository which was shared before this change (setup in `@BeforeClass`). Even if each test tried to make sure to revert its changes to the repository, I think it could still happen that other tests will see the repository changed if the test ends with exception. So, not sharding the repository makes sense.

tarzanek:
ok, then please merge, full tests are run by robots anyways, so if it takes few more seconds there is no prob, especially with the ssd disks

PR #1752, opened 2017-08-24
add HTML encoding of links generated by history search

vladak:
This should make sure that all links passed to the `diff` JSP are URIencoded and HTMLized properly (e.g. the `&` character in `href` needs to be `&amp;`).


tarzanek:
thank you, merging

PR #1755, opened 2017-08-28
Subversion + current version fixes

vladak:
- fills missing features of Subversion repository
- unifies output format for current version

### DIFF
--- a/src/org/opensolaris/opengrok/history/SubversionRepository.java
+++ b/src/org/opensolaris/opengrok/history/SubversionRepository.java
@@ -494,8 +494,9 @@ public class SubversionRepository extends Repository {
         try {
             History hist = getHistory(new File(getDirectoryName()), null, 1);
             HistoryEntry he = hist.getHistoryEntries().get(0);
-            curVersion = he.getDate() + ": " + he.getRevision() +
-                    " " + he.getAuthor() + " " + he.getMessage();
+            curVersion = outputDateFormat.format(he.getDate()) + " " +
+                    he.getRevision() + " " + he.getAuthor() + " " +
+                    he.getMessage();
         } catch (HistoryException ex) {
             LOGGER.log(Level.WARNING, "cannot get current version info for {0}",
                     getDirectoryName());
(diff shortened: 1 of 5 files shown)


src/org/opensolaris/opengrok/history/SubversionRepository.java
```diff
@@ -441,23 +451,55 @@ String determineParent() throws IOException {
...
+        return getInfoPart("URL:");
     }
 
     @Override
-    String determineBranch() {
-        return null;
+    String determineBranch() throws IOException {
+        String branch = null;
+
+        String url = getInfoPart("URL:");
```
tarzanek: you could have made "URL:" a constant

src/org/opensolaris/opengrok/history/SubversionRepository.java
```diff
@@ -441,23 +451,55 @@ String determineParent() throws IOException {
...
+    @Override
+    public String determineCurrentVersion() throws IOException {
+        String curVersion = null;
+
+        try {
+            History hist = getHistory(new File(getDirectoryName()), null, 1);
+            HistoryEntry he = hist.getHistoryEntries().get(0);
+            curVersion = outputDateFormat.format(he.getDate()) + " " +
+                    he.getRevision() + " " + he.getAuthor() + " " +
+                    he.getMessage();
```
tarzanek: stringbuilder ?
I am actually not sure if some newer jdks won't already optimize like this so it might be unnecessary ...

tarzanek:
other than above comments lgtm, feel free to merge

src/org/opensolaris/opengrok/history/SubversionRepository.java
vladak: yep

src/org/opensolaris/opengrok/history/SubversionRepository.java
vladak: according to https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java it looks like that the compiler will substitute it itself so leaving as is.

PR #1757, opened 2017-08-29
getLatestRevision() fix

vladak:
getLatestRevision() was not careful enough with checking `HistoryEntry` list and its called did not fullfil its promise to return `null` when `getLatestRevision()` fails.


tarzanek:
lgtm, merging

PR #1758, opened 2017-08-29
directory name of RepositoryInfo should be stored as relative path

vladak:
This is a bit of a hack to store only relative paths in `RepositoryInfo` object encoding.


tarzanek:
merging
I suppose all this transient stuff is just to have proper config.xml ?

tarzanek:
ah, ok, cannot merge
can you please solve conflicts and merge @vladak ?

thnx
L

vladak:
Even though I merged with upstream/master according to https://help.github.com/articles/syncing-a-fork/, Github seems to think that this has conflict even though they are not listed.

vladak:
Anyhow, I found that there is some deduplication to do in the `SubversionRepository` code.

PR #1762, opened 2017-08-30
Rewrite README.txt to use markdown syntax

Orviss:
You can check out the result here: https://github.com/Orviss/OpenGrok/tree/readme-txt-to-markdown


tarzanek:
lgtm, merging, thank you Adam!

PR #1771, opened 2017-09-14
refresh group lists in project message handling

### LINKED ISSUE #1769: add/remove project message handling fails to refresh group list
After project add (via messagge) and per-project indexing the project is not displayed in the listing. I suspect this is because the project is part of a group (more precisely a sub-group) so it seems there is a list that needs to be refreshed.

vladak:
fixes #1769


tarzanek:
lgtm, thank you, merging

PR #1774, opened 2017-09-15
disable generating full history for renamed files by default

### LINKED ISSUE #1773: disabled renamed files handling by default
the renamed files handling is useful however for repos with large history this makes indexing too slow.

vladak:
fixes #1773


tarzanek:
merging, thank you <3 ! :-)

PR #1779, opened 2017-09-19
Folding icon selection fix

shaehn:
This change fixes the folding icon selection problem (jump to top of page).


vladak:
I assume this was tested on all major browsers.

shaehn:
Yes, it was tested on Internet Explorer, Firefox, Chrome (Hmm... we had this conversation earlier, but I don't see it listed here)

vladak:
That's why I ask - to have the testing aspect confirmed in the pull request. Anyway, merging now.

PR #1780, opened 2017-09-19
Dash B option enhancement

shaehn:
This allows a user to enter -B "none" to disable the information provider link on the annotation page.


shaehn:
I am open to suggestions. My original thought process is to minimize change of original intent because I certainly don't know the original reasons why things are set up the way they are.

So are you saying that you want the -B default value to be **null** instead of the example bogus string that it is originally set up?  Also, I am not sure what you mean by the "user page". You will have to be a little more specific so I may look at the correct item to which you are referring.

vladak:
Yep, make the default value `null`.

Actually, there are 3 similar tunables:
  - bug page (links to bug database)
  - user page (links to user database)
  - architecture page (links to architecture review database)

The default values are set in set in `Configuration` constructor:

```
setBugPage("http://bugs.myserver.org/bugdatabase/view_bug.do?bug_id=");
setUserPage("http://www.myserver.org/viewProfile.jspa?username=");
setReviewPage("http://arc.myserver.org/caselog/PSARC/");
```

shaehn:
Ok, I see those two other settings, but looking further, I don't see them being used anywhere, So does it really matter at all? (I would just comment out the setBugPage and the setReviewPage calls.)

vladak:
Pretty sure they are used, we use the links generated via them in our deployment every day.

vladak:
See `history.jsp`

vladak:
The above are setters so if these are used, their getters have to be called somewhere - voila, searching for (this is OpenGrok after all, so why to use it to search its own source code; alternatively just use Netbeans which can perform caller/callee/usage search) getUserPage() and getReviewPage() reveals `history.jsp`.

shaehn:
OK,  I think I was looking for the setters and not the getters.  If I set the review page and bug page elements to null, how will they be set to anything else? I looked at the command line switches and was not finding anything there that did that. That would mean that once they are set to null, they will stay that way.

On another topic, I was finding a radical problem with my current build in this branch. The resulting OpenGrok page changed in unexpected ways and lots of things (buttons, links) stopped working.

image

The Projects element now seems to be a scrolling window, whereas before it was a pull down menu. The buttons below, select all, invert selection, and clear do nothing at all.

Then when I went to an indexed file's content, the Annotate, Line#, Scopes#, and Navigate no longer worked. They all have entries like **localhost:8080/source/xref/file.ps1#** which do nothing more than pop the view back to the top of the page. Any ideas as to what is going on?

vladak:
The configuration can be changed:
  - using the readonly option of indexer (`-R`)
  - using the `setconf` option of `Messages` tool (`config` message type)
  - setting individual properties, using the `set` option with the `Messages` tool (`config` message type)

In our deployment we use all three options, the last one only infrequently though mainly for quick fixups.

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
```diff
@@ -298,19 +299,21 @@ public void testIgnoredNames() {
     @Test
     public void testUserPage() {
         RuntimeEnvironment instance = RuntimeEnvironment.getInstance();
-        String page = "http://www.myserver.org/viewProfile.jspa?username=";
-        assertEquals(page, instance.getUserPage());
-        instance.setUserPage(page.substring(5));
-        assertEquals(page.substring(5), instance.getUserPage());
```
vladak: this should work. why is this change necessary ?

vladak: I mean - the test should call setBugPage() at least once and verify that the setter works with getBugPage(). It **may** verify the default value as well. The more coverage we get the better.

shaehn: If you are referring to lines 142, this fails on a Windows box. The canonical path returned is C:\opengrok_automatic_test\foo\bar. The problem I was encountering has to do with the getCanonicalPath method in RuntemeEnvironment.java. It tests for existence, and when that fails, it returns the original string. So I was forced to make the original string be the canonical path to prevent the test failing further down.

So do you want to see something like this for the testBugPage:

        assertNull(instance.getBugPage());
        instance.setBugPage(instance.getBugPage());
        assertNull(instance.getBugPage());

vladak: No, I am referring to the testUserPage(), testBugPage(), testReviewPage(). These tests should do something like this:

1. use getter to verify the default value is `null`
2. use setter to set a value
3. use getter to verify that the value was set

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
```diff
@@ -299,21 +299,19 @@ public void testIgnoredNames() {
...
     public void testUserPage() {
         RuntimeEnvironment instance = RuntimeEnvironment.getInstance();
-        //String page = "http://www.myserver.org/viewProfile.jspa?username=";
-        String page = null;
-        assertNull(instance.getUserPage());
-        //instance.setUserPage(page.substring(5));
-        //assertEquals(page.substring(5), instance.getUserPage());
+        String page = "http://www.myserver.org/viewProfile.jspa?username=";
+        assertNull(instance.getUserPage());   // default value is null
+        instance.setUserPage(page.substring(5));
```
vladak: I wonder what was the purpose of these `substring` operations. Perhaps they can be removed ?

shaehn: I can do that. They were just there from before.

PR #1782, opened 2017-09-20
get rid of GNU sed'ism so that file substitution works on all platforms

vladak:
tested with:
```
rm /var/tomcat8/webapps/source.war
OPENGROK_INSTANCE_BASE=/foo ./OpenGrok deploy
cd /tmp
jar xf /var/tomcat8/webapps/source.war
```

and check contents of `/tmp/WEB-INF/web.xml` - should contain this:

```
  <context-param>
    <description>Full path to the configuration file where OpenGrok can read its configuration</description>
    <param-name>CONFIGURATION</param-name>
    <param-value>/foo/etc/configuration.xml</param-value>
  </context-param>
```

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -915,12 +915,10 @@ DeployWar()
         FatalError "Web Application Installation FAILED"
     fi
 
-    # Fix for opengrok issue https://github.com/OpenGrok/OpenGrok/issues/863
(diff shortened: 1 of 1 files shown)


tarzanek:
posix ftw!
thank you, merging
L

PR #1785, opened 2017-09-23
Autorefresh after project removal from the chosen set

### LINKED ISSUE #1369: autorefresh after project removal from the chosen set
The main search page could auto-refresh (automatically resend the search request) when one
closes one from the chosen projects icons. Say, in the example below, when one closes the "openssl-1.0.0" button, the search should automatically refresh. The current behavior is that one has to click the "Search" button after the chosen project button is closed -- which seem rather superfluous (subjective).

autorefresh

Orviss:
fixes #1369 

After clicking on the 'x' icon from the project set it will deselect the project and refresh the search. 

**Important**: this can also lead to some weird behaviour – if there is only one project then clicking on 'x' will not remove this project but only refresh the search (this happens also when you deselect the last project from the project list and click on the search). In my opinion in this case it should redirect to the main page. I did not want to mix two things together so I did not add fix for this in this change. (Just tell me if you think otherwise.)


tarzanek:
@tulinkry can you please review ?

web/js/searchable-option-list-2.0.4.js
```diff
@@ -1037,10 +1037,11 @@
                 if ((this.config.multiple || this.config.allowNullSelection) && !$changedItem.prop('disabled')) {
                     $('<span class="sol-quick-delete"/>')
                         .html(this.config.texts.quickDelete)
-                        .click(function () {
+                        .click(function () { // deselect the project and refresh the search
```
tulinkry: Add there this note
```javascript
/*
 * Modified 2017
 */
```

Orviss: 1. The page is refreshed – would you like me to add there a check if we are on the main page and then ignore the form submit (or do you have something different in mind)?
2. Project picker is not affected (since I thought that reloading page while selecting/deselecting in project picker would not be very wise)
3. --||--

tulinkry: > add there a check if we are on the main page and then ignore the form submit

I think this is preferable. Consider a scenario when the user fills the form with something and then cares about the projects and a removal takes him to search results immediately. That doesn't seem intuitive.

I think that I added a class to body to reflect the current page name so you can maybe use that.

tulinkry: 2. and 3. Ok

vladak:
The single project deselection works in 1.1-rc10 so that would be a regression.

PR #1786, opened 2017-09-24
Java 9 compilation support

Orviss:
Compilation error: "Unsupported Java version: 9. Make sure that the Java version is 1.8 or greater."

This fix adds Java 9 to supported versions.


vladak:
Okay. I was about to say update README etc. however this can wait until we get more experience with OpenGrok and Java 9.

tarzanek:
there are a bit more changes needed for java 9, let's not rush

PR #1790, opened 2017-09-27
Opengrok src root

vladak:
also removes last bits of JavaDB support.


tarzanek:
there was also a way to override source root by 
./Opengrok index /path/to/source/root/
merging , thank you

PR #1796, opened 2017-09-30
searchable-option-list workaround for less jumpy page loading

emergie:
Minor UI fix.

searchable-option-list plugin adds functionality: autocomplete + multi selection handling.
Unfortunately it also causes UI to be jumpy, because it has to be loaded client-side after the initial page load.

Proposed solution is not pretty, but it does the job.

before:
peek sol before

after:
peek sol after

I accept OCA for this patch

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -1242,6 +1242,22 @@ h6.message-group-caption {
     *zoom: 1;
 }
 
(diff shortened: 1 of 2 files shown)


vladak:
This has been bothering me for a long time so accepting the workaround. The proper fix will be delivered via #1519 I guess.

PR #1797, opened 2017-09-30
search form focus restoring after submit triggered by return key

emergie:
Minor UI enhancement.

If search form submission is triggered by pressing the return key in one of `input[type=text]` fields it would be nice to have focus back in the field after the search completes.

Proposed implementation of this idea is based on adding `input[name=si]`  to the form on _submit_  event. Additional input contains _id_ of active field, so the focus can be restored on the next page load.

For example:
Pressing enter in `Full Search` input field adds `si=q` to the query url.
After page load javascript function looks for a field `input[type=text][id="q"]`, if found then it gets focus and cursor is moved at the end of text.

In action:
peek 2017-09-30 19-24

I accept OCA for this patch.


vladak:
same thing - pls bump version and rename.

vladak:
Also pls resolve conflicts.

web/js/utils-0.0.17.js
```diff
@@ -1941,3 +1942,43 @@ function scope_on_scroll() {
...
+    }
+}
+
+/**
+ * Restores focus on page load
+ *
+ * @see #searchSubmit
+ */
+function restoreFocusAfterSearchSubmit() {
+    var siParam = $.grep(window.location.href.split('&'), function (queryParam) {
```
tulinkry: maybe use `function getParameter(p)`

emergie: fixed

PR #1802, opened 2017-10-03
Message get repos

vladak:
This will be handy for yet-to-be synchronization library.


tulinkry:
There should be also a note in readme

tarzanek:
good stuff, you're slowly developing a reasonable project API, cool, let's merge! :)

vladak:
Well, it's a bit heavy through Messages. One day it could be HTTP API.

PR #1804, opened 2017-10-03
remove extraneous message emitted to stderr about extra ctags config …

vladak:
This has been bothering me for quite some time. This message is extra naughty because:
  - it claims it is being sent with `INFO` prirority
  - is produced during every indexer run (with the `-o` option)
  - somehow it gets propagated through layers of processes to the session leader process so it ends up in cron emails etc.


tarzanek:
I am ok with this, but you might wanted to change it from .err to .out so your filtering later would work (or ideally send it to logger)

vladak:
I think in this case the logger might not be initialized yet.  Anyway the message is redundant - if the set succeeds it should be silent.

Issue #1503, opened 2017-03-29
use long options for Indexer.java


vladak:
As mentioned in #1500, the getopt parsing in Indexer.java has reached the point where we are almost out of alphabet. Also the letter choice for some of the options do not make sense. The time has come to switch purely to longopts - without the hybrid mode.

PR #1812, opened 2017-10-06
Option parser #1503

shaehn:
Introducing OptionParser to OpenGrok. Allows use of long name options, and much more.


shaehn:
So, does this code need any addendum or changes? I would like to start working on Indexer.java, but not unless this code passes muster first.

idodeclare:
Hi, these new files read "Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved." May I ask what is their provenance?

shaehn:
I am not sure what you mean be provenance.  I wrote these files from scratch. The header is a standard header required in this project.

idodeclare:
The copyright indicating "Oracle" is not a standard part of the header. The CDDL HEADER START/END is standard, but then contributors include or update their copyright notice as per CDDL 3.3. Required Notices: 

> 
> You must include a notice in each of Your Modifications that
>     identifies You as the Contributor of the Modification.  You may
>     not remove or alter any copyright, patent or trademark notices
>     contained within the Covered Software, or any notices of licensing
>     or any descriptive text giving attribution to any Contributor or
>     the Initial Developer.
> 

(See LICENSE.TXT.)

As these are your own files from scratch, and because CDDL is file-based, you are their "Initial Developer" per the CDDL. These new files should only have a copyright notice for your legal entity name.

shaehn:
Thanks for the clarification. I was just using the Netbeans IDE which I think put this header when I initially created the file.

idodeclare:
@shaehn, you're welcome.

One additional recommendation: please consider `git pull --rebase` instead of just `git pull` so that your contributed branches do not have excessive merges (I count six above). In fact, this way a pull request would not start with any merges at all — yet it could still be kept on a fresh tip of master for submission.

(After raising a Github pull request, you should no longer rebase. If necessary after that point, you would use `git pull` to merge a fresh master.)

shaehn:
@idodeclare, currently I am struggling with git and its concepts of merging/syncing/rebase etc. Syncing is just driving me nuts. I try to follow the "create-a-new-branch-for-each-issue" mechanism off of my OpenGrok fork. 

After that, the nightmare/pain/uncertainty begins.

OpenGrok
   myOpenGrok
       myIssue1
       myIssue2
       
So for example, I am creating these new files in myIssue1 branch.
Now I want to use those new files in the myIssue2 branch, but I don't see them yet.
Do I need to merge them to myOpenGrok fork so that myIssue2 can see them?
How does that affect the pull request I have made to the OpenGrok repository?
Will there be some problems in myOpenGrok when the pull request is actually merged into OpenGrok?

For most of my pushing up to my remote repository, I have been using the Git Desktop application.
It just makes my interaction to git a little more pleasant.
(I normally use AccuRev which is GUI based and don't have these type of problems.)

idodeclare:
@shaehn,

Reiterating the rule I recommended above: prior to submitting a pull-request, keep rebasing (`git pull --rebase`) your feature branches (and avoid merge). Once submitted as a pull-request, however, GitHub does not like rebased branches, so stop using it — and only merge from master to your submitted feature only if absolutely necessary to get master's latest commits.

For your example, while myIssue1 and myIssue2 are independent, they should just be rebased on master (e.g., `git checkout myIssue1 && git pull --rebase`).

You've described that you sometimes need to make myIssue2 dependent on myIssue2. Then for myIssue2: `git checkout myIssue2 && git rebase myIssue2`). Now this brings up a third rule: only create pull requests for branches off of master. As myIssue2 is now rebased on myIssue1, it should not be submitted until later (e.g., after myIssue1 is merged) after the following is run: `git checkout myIssue2 && git rebase master`.

Continuing this example, once myIssue2 is dependent on myIssue1, then after any new commits are made to myIssue1 (i.e., `git checkout myIssue1; ...; git commit`), then myIssue2 must once again be rebased on (the HEAD of) myIssue1: `git checkout myIssue2 && git rebase myIssue1`.

Hope this helps!

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+        if (optind == -1) {
+            remainingArgs = args;
+        } else if (optind < args.length) {
+            remainingArgs = Arrays.copyOfRange(args, optind, args.length);
+        }
+        
+        return remainingArgs;
+    }
+    
+    private String program() {
```
vladak: I don't like this - surely there is better way how to get program name other than traversing call stack ?

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+    }
+    
+    /**
+     * Obtain option summary
+     * @return option summary
+     */
+    public String getUsage() {
+        
+        StringWriter wrt = new StringWriter();
+        String indent = "  ";
```
vladak: would be nice if the indent was configurable

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+        try (PrintWriter out = new PrintWriter(wrt)) {
```
vladak: would be nice to be able to tell whether the output should go to `stdout` or `stderr`

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+            out.flush();
+        }
+        return wrt.toString();
+    }
+    
+    /**
+     * Print out option summary.
+     */
+    public void help() {
+        System.out.println(getUsage());
```
vladak: same here w.r.t. where the output should be printed

vladak:
general nit comment: pls insert space after command in method parameters

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+        List<String> allowedValues;  // list of restricted values
+        Boolean mandatory;           // true/false when argument present
+        StringBuilder description;   // option description for summary
+        Consumer<Object> action;     // code to execute when option encountered
+        
+        public Option() {
+            names = new ArrayList<String>();
+        }
+                
+        public void set(String option, String arg) throws IllegalArgumentException {
```
vladak: use more descriptive method name ?

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+    public String prologue;  // text emitted before option summary
+    public String epilogue;  // text emitted after options summary
+    
+    class DuplicateOptionName extends Exception {
+        public DuplicateOptionName(String message) {
+            super(message);
+        }
+    }
+    
+    public class Option {
```
vladak: general note for this class - all public methods should have a javadoc (maybe besides getters/setters)

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+            }
+            description.append(descrip);
+            description.append("\n");
+        }
+        
+        /**
+         * Store user defined action with option object
+         * @param action 
+         */
+        public void Do(Consumer<Object> action) {
```
vladak: `setAction` ?

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+    
+    private String getOption(String arg) {
+        String option = null;
+        
+        if (arg.startsWith("-")) {
+            if (arg.startsWith("--")) {
+                option = arg;                 // long name option (--longOption)
+            } else if (arg.length() > 2) {
+                option = arg.substring(0,2);  // short name option (-xValue)
+            } else {
```
vladak: more robust checking ? the else branch should check that the arg indeed starts with `-` and the else would then throw error if it starts with anything else.

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+        // to be the option's value. Note that if the argument
+        // is actually another option (starts with '-') then
+        // there is no value available. If the option is required
+        // to have a value, null is returned. If the option
+        // does not require a value, an empty string is returned.
+        String value = arg;
+        Boolean isOption = value.startsWith("-");
+        
+        if (mandatory) {
+            if(isOption ) {
```
vladak: style

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+
+            if (!option.startsWith("-")) {
+                break;
+            } else {
+                if (option.equals("--")) {
+                    optind = ii + 1;
+                    break;
+                }
+
+                if (unknown(option, ii)) {
```
vladak: If `unknown()` already did the job and found partial match, it would be wasteful to repeat this work. Also, I don't see how partial matching could actually work - if `unknown()` returns `false` then the partial option would get matched in `options.get()` which cannot succeed.

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+        
+        if (!found) {
+            // Now check to see if initial substring was entered.
+            for (String key: options.keySet()) {
+                if (key.startsWith(option)) {
+                    candidates.add(key);
+                }
+            }
+            if (candidates.size() == 1 ) {
+                found = true;
```
vladak: `false` ?

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+                option = arg.substring(0,2);  // short name option (-xValue)
+            } else {
+                option = arg;                 // short name option (-x)
+            }
+        }
+        return option;
+    }
+
+    // Allows user to enter initial substring of a long option name.
+    private boolean unknown(String option, int index) throws ParseException {
```
vladak: this should be covered by a test

tulinkry:
I have just one question, why do you prefer this solution over any kind of stable library already solving this problem?

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: I was hoping for something already available as an interface in java, but web searching wasn't popping up anything easily relevant. Some of the solutions were downright scary looking. So I chose the least scary.  If you have pointers for a better way... let me know.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Hmm... I am not sure what you are asking here, more documentation?  I pulled this out of CommandLineOptions.java

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Again, here do you want me to state in the documentation that the output is going to stdout? (Would you prefer that it goes to stderr?  It doesn't matter to me one way or the other.)

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: yes I can do that. Originally, this was not public, but I had to make it so because when I started working in Indexer.java, it couldn't see its methods.

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: I'd like to have the flexibility to choose where it will go.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: This one is seen/used by the caller and I want it to appear to be similar in name to the same named method in OptionParser. For example,

OptionParser openGrok = OptionParser.Do(parser -> {

    parser.on("-?", "-h", "--help", "Display this usage.").Do( v -> {
	parser.help();
    });

So when one reads the code they can think, 'on' seeing this option string, the code will 'do' the following...

With your suggestion, the code would look like this, which doesn't read the same.

OptionParser openGrok = OptionParser.Do(parser -> {

    parser.on("-?", "-h", "--help", "Display this usage.").setAction( v -> {
	parser.help();
    });

tulinkry: I prefer `do` but I don't get why it's with a capital letter

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: how does one put a private method in a test?

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: by style do you mean I missed a space between 'if' and '(' ?

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: maybe just specify the progname while constructing the parser..

src/org/opensolaris/opengrok/util/OptionParser.java
tulinkry: Use protected

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Hmm.... you can only get to the else statement when the 'arg' starts with a '-' and only that character  and that the size of the 'arg' is of length 2, or less....... Ahhh.... the only other way to get to the else is that the argument is simply '-'. So I could put in a test that explicitly asks for the length to be of size '2', or just test for arg.equals("-") and holler at that.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: I would have chosen 'do', but the compiler didn't like that, it is a keyword to the java language.

shaehn:
@tulinkry   I did look at other solutions, but they all seemed rather cumbersome/complicated. I also wasn't sure if such libraries could be easily incorporated into OpenGrok, otherwise I would have thought that would have been done a long time ago. I saw that OpenGrok already rolled its own Getopt, so that seemed the simplest path to take.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: but Vlad, you do have the flexibility....
From the calling program say, System.err.println(parser.getUsage());
I just added the 'help' method as a convenience mechanism.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: I see your point. I thought my original test covered this, but it doesn't. So I need to write another test. My cure will be to change 'unknown' into 'candidate' which will return the actual option name found, or null when not found.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Note that this is only a fallback mechanism (default) when the programmer doesn't specify a prologue. So in a way, the program name can already be specified there.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Would another method getUsage(indent) satisfy this itch?

shaehn:
Vlad, as to your

> general nit comment: pls insert space after command in method parameters

would you please show me where I am not doing what you are expecting, because I am not seeing it.

vladak:
ad style: e.g. this `arg.substring(0,2);` should be `arg.substring(0, 2);`

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: yep + missing space between `if` and left paren

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: I think so.

shaehn:
Vlad,   I was rummaging through CommandLineOptions.java just now to see if I covered everything that needed covering so that the file could be removed and I noticed a method in there for producing a manual page. Is that used at all? Does OptionParser need to mimic it?

src/org/opensolaris/opengrok/util/OptionParser.java
tulinkry: ah :-D

tarzanek:
last time I checked the manual page used the options to generate man page automagically ...
so yes, it'd be cool if we wouldn't loose this convenience if possible
(we try to make most of stuff to be automated, we don't have a full team to take care of the code base, so the more is automated, the better)

shaehn:
Hmmm...   Looking at the man page generation code, I am not sure I understand what it is producing.  How do I activate it from the command line to see what is generated? It appears that it wants opengrok.xml and I see 4 different ones in the baseline.

vladak:
As for where the man page gets generated with `ant`, here's the relevant part of `build.xml`:

```
    <target name="-post-jar" depends="-copy-lib-without-netbeans, sample-plugins">
        <war destfile="${dist.war}" webxml="web/WEB-INF/web.xml">
            <fileset dir="web" excludes="META-INF/context.xml"/>
            <lib dir="${dist.dir}" includes="opengrok.jar"/>
            <lib dir="${dist.dir}/lib" includes="*.jar" excludes="${servlet-api.jar}"/>
        </war>
        <echo message="Generating man page.."/>
        <java classname="org.opensolaris.opengrok.index.CommandLineOptions"
              output="${dist.dir}/opengrok.1" failonerror="true" fork="true">
            <classpath>
                <pathelement location="dist/opengrok.jar"/>
                <pathelement path="${java.class.path}"/>
            </classpath>
        </java>
    </target>
```

so something like this gets to run:

`java -classpath dist/opengrok.jar org.opensolaris.opengrok.index.CommandLineOptions`

vladak:
As for the validity of the XML produced, this seems to be a separate bug. After I removed the `<?Pub Inc>` line and fixed the `DOCTYPE` line, `xmllint` is happy.

shaehn:
Vlad, when you said that you 'fixed' the DOCTYPE line, how did you alter it?
As to the invocation to generate this man page, I will have to do something completely different. The Indexer is going to have to have a --man option put in place and thus the generated command is going to have to look like:

`java -classpath dist/opengrok.jar --man `

So first I will have to see how 'ant' declares arguments on the command line.
The problem that I currently have when I manually run from the command line is that the code croaks looking for opengrok.xml.

try (BufferedReader reader = new BufferedReader(new InputStreamReader(
                     getClass().getResourceAsStream("opengrok.xml"), "US-ASCII"))) {

and I am not sure how to specify where it should look for the file. (It is not included in the opengrok.jar file)

vladak:
I just found some valid document and copied this:

```
<!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook XML V4.1.2//EN"
                "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd">
```

vladak:
The `opengrok.xml` is part of the repository - it lives in [src/org/opensolaris/opengrok/index/opengrok.xml](https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/index/opengrok.xml) . When you reference files using `getResourceAsStream()` it will look in the same Java package.

shaehn:
The implementation of the manual generation did not turn out to be too difficult after enough hints were given, but wanting to add a test to make sure empty option descriptions were caught turned out to be brutal. I fought with the XMLEncoder stuff most of the day. I finally gave up trying to encode an Option because XMLEncoder just could not handle it, so instead I encoded a simple string array with the offending option names.  This required me to add testing hooks into OptionParser.

If there are no further comments on OptionParser, I will then upload the new Indexer next.

shaehn:
I know that there will be lots of interest/questions/changes wanted now that I have pushed up the Indexer.java code. I tried to add long names for all the options that are currently present, but there were some I just didn't have a clue what would be wanted. You may want a different naming scheme altogether, so I am sure this is just a first stab at the content. So people, review, review, review.

src/org/opensolaris/opengrok/util/OptionParser.java
tulinkry: maybe just add an overloaded method help(stream) and use the help() to call it with System.out. I think that should be enogh

src/org/opensolaris/opengrok/util/OptionParser.java
tulinkry: This is fine the output goes to string

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,701 @@
...
+    // List of options in order of declaration
+    final private List<Option> optionList;
+    
+    // Keeps track of separator elements placed in option summary
+    final private List<Object> usageSummary;
+    
+    private boolean scanning = false;
+    
+    public String prologue;  // text emitted before option summary
+    public String epilogue;  // text emitted after options summary
```
tulinkry: don't expose thi directly please, use getter and setters

src/org/opensolaris/opengrok/util/OptionParser.java
tulinkry: I would suggest the main class to be a parameter of a constructor as this class is not very useful fro any other use than with a main class. How about that?

tulinkry: Or better part of this call opts.parse(); where the args is quite tight to the main class

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: So do you want this stream to be a  PrintStream, or something more general?

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: When someone wants to name the program, this is how its done currently:
    
    String program = "opengrok.jar";
    openGrok = OptionParser.Do(parser -> {
        parser.setPrologue(
	     String.format("\nUsage: %s [options] [subDir1 [...]]\n", program));

I don't see a need for anything else. I reiterate that this 'program'method is just an internal mechanism to generate a placeholder name when someone doesn't specify a proglogue. I am pretty sure most people will be setting the prologue value to their particular needs.

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                    "      will disable the c-analyzer for for all files ending with .c").
+                Do( analyzerSpec -> {
+                    String[] arg = ((String)analyzerSpec).split(":");
+                    String fileSpec = arg[0];
+                    String analyzer = arg[1];
+                    configureFileAnalyzer(fileSpec, analyzer);
+                }
+            );
+
+            parser.on("-a", "--leadingWildCards", "=on/off", onOff, Boolean.class,
```
vladak: no short

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                    configureFileAnalyzer(fileSpec, analyzer);
+                }
+            );
+
+            parser.on("-a", "--leadingWildCards", "=on/off", onOff, Boolean.class, 
+                "Allow or disallow leading wildcards in a search.").Do( v -> {
+                cfg.setAllowLeadingWildcard((Boolean)v);
+            });
+
+            parser.on("-B", "--infoProvider", "=URL",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                "Base URL of the user Information provider.",
+                "Example: \"http://www.myserver.org/viewProfile.jspa?username=\".",
+                "Use \"none\" to disable link.").Do( v -> {
+                cfg.setUserPage((String)v);
+            });
+
+            parser.on("-C", "--progress",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-                        System.exit(1);
+            parser.on("-c", "--ctags","=/path/to/ctags", 
+                "Path to Exuberant Ctags from http://ctags.sf.net",
+                "By default takes the Exuberant Ctags in PATH.").
+                Do( ctagsPath -> {
+                    cfg.setCtags((String)ctagsPath);
+                }
+            );
+
+            parser.on("-D", "--renamedHistory", "=on/off", onOff, Boolean.class,
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+            });
+
+            parser.on("-i", "--ignore", "=pattern", 
+                "Ignore the named files (prefixed with 'f:')",
+                "or directories (prefixed with 'd:').",
+                "Supports wildcards (example: -i *.so -i *.dll)").Do( pattern -> {
+                cfg.getIgnoredNames().add((String)pattern);
+            });
+
+            parser.on("-K", "--listRepos", "List all repository paths and exit.").Do( v -> {
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                listRepos = true;
+            });
+
+            parser.on("-k", "--deleteHistory", "=/path/to/repository",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                listRepos = true;
+            });
+
+            parser.on("-k", "--deleteHistory", "=/path/to/repository", 
+                "Delete the history cache for the given repository and exit.",
+                "Use '*' to delete the cache for all repositories.").Do( repo -> {
+                zapCache.add((String)repo);
+            });
+
+            parser.on("-L", "--style", "=path",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-                getInstance().prepareIndexer(env, searchRepositories, addProjects,
-                        defaultProjects,
-                        listFiles, createDict, subFiles, repositories,
-                        zapCache, listRepos);
-                if (listRepos || !zapCache.isEmpty()) {
-                    return;
+                System.exit(status);
+            });
+
+            parser.on("-N", "--symlink", "=/path/to/symlink",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-            } catch (IndexerException ex) {
-                LOGGER.log(Level.SEVERE, "Exception running indexer", ex);
-                System.err.println(cmdOptions.getUsage());
-                System.exit(1);
-            } catch (Throwable e) {
-                System.err.println("Exception: " + e.getLocalizedMessage());
-                LOGGER.log(Level.SEVERE, "Unexpected Exception", e);
+            );
+
+            parser.on("-X", "=url:suffix",
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-                System.err.println("Exception: " + e.getLocalizedMessage());
-                LOGGER.log(Level.SEVERE, "Unexpected Exception", e);
+            );
+
+            parser.on("-X", "=url:suffix", 
+                "URL Suffix for the user Information provider. Default: \"\"").Do( suffix -> {
+                cfg.setUserPageSuffix((String)suffix);
+            });
+
+            parser.on("-y", "Populate the webapp with bare configuration and exit.").Do( v -> {
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
+                noindex = true;
+            });
+
+            parser.on("-z", "--depth", "=number", Integer.class,
```
vladak: ditto

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-                 * directory and not per project data root directory).
-                 * For the check we need to have 'env' already set.
-                 */
-                for (String path : subFilesList) {
-                    String srcPath = env.getSourceRootPath();
-                    if (srcPath == null) {
-                        System.err.println("Error getting source root from environment. Exiting.");
-                        System.exit(1);
+            parser.on("-c", "--ctags","=/path/to/ctags", 
+                "Path to Exuberant Ctags from http://ctags.sf.net",
```
vladak: - say Exuberant or Universal ctags
- remove the http link

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -101,542 +109,658 @@ private static void A_usage() {
...
-                            progress);
+            );
+
+            parser.on("-o", "--ctagOpts", "=path", 
+                "File with extra command line options for ctags.").
+                Do( path -> {
+                    String CTagsExtraOptionsFile = (String)path;
+                    File CTagsFile = new File(CTagsExtraOptionsFile);
+                    if (!(CTagsFile.isFile() && CTagsFile.canRead())) {
+                        die("File '"+ CTagsExtraOptionsFile+ "' not found for the -o option");
```
vladak: style: surround binary operators with spaces

vladak: or use fomatter

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: better to have no fallback than hacky fallback I think

src/org/opensolaris/opengrok/util/OptionParser.java
```diff
@@ -0,0 +1,843 @@
...
+    public void separator(String text) {
+        usageSummary.add(text);
+    }
+    
+    /**
+     * Obtain option summary
+     * @param indent a string to be used as the option summary initial indent.
+     * @return 
+     */
+    public String getUsage(String indent) {
```
vladak: it would be really nice if this function could wrap the lines to terminal width (80 columns by default), it's one of the things that I find irritating with the current option parser.

src/org/opensolaris/opengrok/index/Indexer.java
shaehn: So are you asking to remove the -a option mentioned here and just use --leadingWildCards?

src/org/opensolaris/opengrok/index/Indexer.java
shaehn: Do you have a suggestion for the long version name for this one? I couldn't think of one at the time.

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: I wasn't expecting to get into reformatting here. That is one reason why the descriptions have multiple entries. The intent is that the programmer can format the option information to look exactly as it would on the command line in the program.  For example,

image

produces:

image

You would not be able to guide the formatting as well (and as simply) if we start playing around moving text around programmatically.

src/org/opensolaris/opengrok/index/Indexer.java
shaehn: done

src/org/opensolaris/opengrok/index/Indexer.java
vladak: `--user-page-suffix`

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: I agree this is a bit orthogonal to this change so file a new Issue for that pls.

src/org/opensolaris/opengrok/index/Indexer.java
vladak: yep

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: Vlad, why do you think this code is too hacky (as in a bad sense)?  After I obtained this from the net, I did a step through the code during a debugging session to see what it was doing. It certainly makes sense that the bottom of the stack is going to be the entry point of the program. Look at the enclosed snapshot I took when I activated the Indexer program.

image

So you would rather just have the default for prologue be,    Usage: MyProgram [options]
than the name of the actual entry point?

src/org/opensolaris/opengrok/index/Indexer.java
shaehn: OK, I noticed that you are using hyphens to separate the names as opposed to CamelCase.  Do you want me to go back and do the same for the other options too?

src/org/opensolaris/opengrok/util/OptionParser.java
shaehn: OK, I added issue #1824

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: yes.

src/org/opensolaris/opengrok/index/Indexer.java
vladak: That's a good question. I am not strongly bound for either, feel free to stick with CamelCase.

src/org/opensolaris/opengrok/index/Indexer.java
shaehn: You do realize that removal of these single character options will be breaking what exists in the field?

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: As for the hackiness, I think traversing the stack which is behind-the-covers data structure to obtain program name is too much hassle. Given the `argv[0]` in Java is not the program name, it better be supplied by hand.

src/org/opensolaris/opengrok/util/OptionParser.java
vladak: `PrintStream` is fine with me

shaehn:
This is what the usage summary looks like at this point: (so are we done yet?)

image
image
image

vladak:
Seems fine to me. Does the `OpenGrok` script need any modifications ?

shaehn:
I have no idea, I have not looked at that file at all. Let me check.

vladak:
merged, thanks !

shaehn:
Ackkkkk.   You merged too soon. I just got back from appointment and found build failure from a dumb typo (what happens when in a hurry). We also still have one housekeeping item to take care of, removal of CommandLineOptions.java

vladak:
No need to open new issue, just submit another PR.

PR #1814, opened 2017-10-09
Feature/perlwork

idodeclare:
Hello,

Please consider this patch to PerlXref.lex with the principal change:

* Handle Perl quote-like operators generally plus Here-documents    

I made some minor additional PerlXref revisions too:
* do not link from inside non-interpolating quote-like operators
* allow multi-line strings -- i.e., do not terminate "s"-state prematurely
* add a '#' to {URIChar}
* update main.pl test file to include quote-like- and Here- examples

(I e-mailed my OCA to oracle-ca_us today.)


idodeclare:
Sorry, I didn't realize github doesn't work for force-pushed revisions. (Closing and reopening.)

Issue #1809, opened 2017-10-05
project add Message does not refresh repository list in HistoryGuru


vladak:
project add Message handling calls into `HistoryGuru` to get a list of repositories using:

```
private List<RepositoryInfo> getRepositoriesInDir(RuntimeEnvironment env, File projDir) {
91          List<RepositoryInfo> repos = new ArrayList<>();
92          HistoryGuru hg = HistoryGuru.getInstance();
93  
94          // There is no need to perform the work of invalidateRepositories(),
95          // since addRepositories() calls getRepository() for each of
96          // the repos.
97          hg.addRepositories(new File[]{projDir}, repos,
98              env.getIgnoredNames());
99  
100          return repos;
101      }
```

however the only addition `addRepositories()` performs is to the list in the argument of the method. As a result, the list of repositories in `HistoryGuru` is not populated and as a result history for the added project is not shown in the UI.

This is a follow-up to #1390.

PR #1816, opened 2017-10-10
populate HistoryGuru repositories list from add project Message

vladak:
In the process of addressing #1809 I tried to fix the HistoryGuru API and its usage, namely:
  - addRepositories() not adding repositories to the HistoryGuru map (this gist of the fix for #1809) and needlessly calling invalidateRepositories()
    - renamed addRepositories(dir) to setRepositories(dir) to better reflect what it is doing and moved it to RuntimeEnvironment since it also modifies the configuration
  - removeCache() using invalidateRepositories() needlessly and actually in a wrong way (invalidateRepositories() needs to be called with full set of repositories since it sets the map from scratch)
  - setConfiguration() not setting the repository list in configuration after invalidateRepositories() (the latter can remove some repositories)


tarzanek:
merging

PR #1817, opened 2017-10-10
Improve AnalyzerGuru magic identification

idodeclare:
- Fix so magics (TreeMap) lookup can work by inspecting it with full
  two- (e.g. "/usr/bin/env perl") or one-word (e.g. "/usr/bin/perl") magics
  extracted from file contents.
- Condense spaced, extracted hashbangs for lookup (e.g.,
  " #! /usr/bin/env  perl" to "#!/usr/bin/env perl").
- Add Comparator so longer magic strings are checked first during the
  imprecise phase of matching (so ShAnalyzerFactory's short "#!" magic is not
  accidentally preferred).
- Add property, isPreciseMagic, to FileAnalyzerFactory.Matcher so byte-precise
  magic detection can be done early. (Only ZipAnalyzerFactory now has this set
  to true.)
- Use a UTF BOM as a precise indication of plain text in the otherwise
  now-imprecise PlainAnalyzerFactory.Matcher.isMagic().

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -19,17 +19,21 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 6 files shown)


tarzanek:
How big is the impact on speed of file detection?
(old Vs new approach)

idodeclare:
@tarzanek,

This would only impact detection for files which did not get a chosen analyzer based on their file extension. It must be a small minority of files, because I've perceived the failure to match magics for at least five years (user since 2005) — but I don't know that anyone else has noticed it.

For these affected files, the amount of work done at worst is roughly the same — i.e., comparing a signature from the start of the file to every defined magic string — although the current implementation does it twice for each string (see the existing `private static FileAnalyzerFactory find(byte[] signature)`) while this patch only compares each magic string at most once with `startsWith()` if it arrives at this imprecise stage.

(To be fair, it could be that the existing implementation is lightning fast during the imprecise stage since it's oddly comparing `sig.startsWith(entry.getKey())` — i.e., "short string starts with (generally) long string". The shortest, magic string — `"#!"` for `ShAnalyzerFactory` — is possibly the only magic that passes this odd comparison.)

For any of these affected files with a hashbang known to OpenGrok, however, the new patch will be much faster. The guru will be able to match a magic string in the `magic` hash table before even bothering to loop to compare `startsWith()` for every magic. (In the current implementation, the hash table lookup is not fruitful given that it searches a strictly-capped-to-eight-bytes string.)

idodeclare:
*Correction: I believe I am overestimating the current cost, because `"#!"` probably is the first magic in the current implementation's SortedTreeMap. Only a single comparison or two would be needed in the current implementation to decide on `ShAnalyzer` for files that start with `"#"`.

The new patch does make good use of the `magics` map for known hashbangs to avoid the exhaustive comparison — but I suppose in the worst case, the new patch likely does compare *every* magic string because of the new reverse-length sort. Files that did not even start with `"#"`, however, would quickly proceed to the fallbacks.

PR #1819, opened 2017-10-11
Improve Perl // heuristic to allow arbitrary whitespace between tokens

idodeclare:
Hello,

Please consider for integration the attached patch to improve the Perl // heuristic to allow arbitrary whitespace between tokens instead of just a single LF.

Also:
- remove no-longer-needed splitEol() and the fields: value1, value2

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
+++ b/src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
@@ -176,25 +176,56 @@ import org.opensolaris.opengrok.web.Util;
   }
 
   // Begins a quote-like state for a heuristic match of the shorthand // of m//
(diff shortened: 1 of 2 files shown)


tarzanek:
merging, thank you
what about some better junit tests? have a look at php or others, this way you will set safeguards for  future ...

idodeclare:
@tarzanek , yes, I'll look at writing some Perl tests in the near future.

It will be after I complete a feature branch I'm working on sporadically now to get OpenGrok tests fully working on macOS. There are a handful of breakages e.g. because of Mac's /var -> /private/var sym-linking, because of a test that doesn't acknowledge that org.opensolaris.opengrok.analysis.Ctags might be defined to specify an explicit Exuberant ctags, etc.

PR #1820, opened 2017-10-11
Feature/macos tests

idodeclare:
Hello,

Please consider for integration this pull request with some fixes for testing on macOS:

- Update test to allow for having set org.opensolaris.opengrok.analysis.Ctags
- Use getCanonicalPath() to allow for macOS's /var to /private/var symlink
- Decorate several methods as requiring "MercurialInstalled"
- Better to throw a HistoryException than NullReferenceException
- Do not throw NullReferenceException from a tearDown
- Store return value in local var before assertion for easier debugging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 nbproject/*~
+/nbproject/configs/*.properties
 opengrok-web-nbproject/nbproject/*~
 opengrok-web-nbproject/nbproject/build-impl.xml~
 tags
--- a/src/org/opensolaris/opengrok/analysis/IteratorReader.java
+++ b/src/org/opensolaris/opengrok/analysis/IteratorReader.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis;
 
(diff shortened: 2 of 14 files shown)


tarzanek:
travis build is a prereq, so we need to figure out why it failed (it might be it will need improving)
travis builds run on ubuntu, so something you changed either doesn't play well there, or uncovered a badly written test ... could you please investigate?

idodeclare:
Hmm, nothing I touched should have caused "IndexerTest.testRemoveFileOnFileChange:370 expected:<1> but was:<0>".

@tarzanek, do you know what is the filesystem type on that Travis ubuntu server for File.createTempFile()? opengrok is at the mercy of the filesystem precision for mtime — so if it's only up to e.g. ext3, then it is only a one second resolution (as with HFS+ on my MacBook). A short, one second sleep in testRemoveFileOnFileChange() would work around that.

What do you think?

tarzanek:
I don't like sleeps in code, I kinda can tolerate sleeps in test code, but if we can make stuff deterministic, there is no need for sleeps ... can we fix the test to avoid this ?
(and it seems it is default fs for ubuntu so ext4)

tarzanek:
the only thing I don't like is the sleep ... but ... I am going to merge now
thank you for all the fixes!

Is there a way to avoid the sleep?

tarzanek:
(and fwiw, I checked and compared to my mac failures and it seems your patch fixes all of them)

idodeclare:
> Is there a way to avoid the sleep?

The evidence indicates to me that the transient failure is due to limitation of mtime resolution on /tmp on the Travis CI server. If only a one second resolution,  then without the `sleep`, the checkout and subsequent update during test can happen close enough to be indistinguishable from their mtimes.

If that's the case, I can't think of any way to work around it except by delay.

PR #1821, opened 2017-10-12
Feature/perl xref tests

idodeclare:
Hello,

Please consider for integration this patch for the Perl xref test you suggested plus a minor tweak to PhpXrefTest:

- Add PerlXrefTest with a hand-curated addition of main.pl's xref
- Fix PhpXrefTest: transposition of expected v actual and unnecessary trim()

Thank you.


test/org/opensolaris/opengrok/analysis/perl/PerlXrefTest.java
```diff
@@ -0,0 +1,104 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2012, 2016, Oracle and/or its affiliates. All rights reserved.
```
tarzanek: you create new files, so this should be just 2017  (not 2012, 2016)
same for all other new files, license basically traces since when does the copyright start

tarzanek:
other than the copyrights this looks very good, once copyright years are fixes I'll hapilly merge, thank you for this Chris!

test/org/opensolaris/opengrok/analysis/perl/PerlXrefTest.java
vladak: Actually, I don't think the year needs bumping. Rather, @idodeclare should add his own copyright.

idodeclare: This was not my all my original work. I file-copied PhpXrefTest.java — hence the inherited copyright — and then reworked it for PerlXref — hence my own contributor notice.

idodeclare:
I believe the copyrights are correct. To reiterate, I file-copied PhpXrefTest.java and reworked it — hence the inherited copyright notice plus my own.

tarzanek:
hmm, but the file didn't exist before ... and I think the copyright tracks when the file was first created (and not really if it was copied from some other file, after all you changed most of the stuff)
but then I am no lawyer

idodeclare:
@tarzanek ,

"I file-copied PhpXrefTest.java and reworked it." That means that clause 1.9.B of the CDDL is in effect:
> Any new file that contains any part of the Original Software or previous Modifications."

The observation that I "changed most of the stuff" is true, but clause 1.9.A of the CDDL encompasses that anyway:
> Any file that results from an addition to, **deletion from** or
           modification of the contents of a file containing Original
           Software or previous Modifications;

"...hence the inherited copyright notice" follows from paragraph 3.3 of the CDDL:

> You may
    not remove or alter any copyright, patent or trademark notices
    contained within the Covered Software, or any notices of licensing
    or any descriptive text giving attribution to any Contributor or
    the Initial Developer.

These clauses certainly contradict that CDDL copyright tracks "not really if it was copied from some other file."

tarzanek:
hehe, then let's merge ;)

tarzanek:
thank you very much @idodeclare !

PR #1822, opened 2017-10-12
(NOT FOR MERGING) Add tests of magics to show broken behavior

idodeclare:
Hello,

This is just a push to show the broken magics matching on the CI server. I have pushed the same tests to pull request #1817 to show the fixed behavior.

Regards.


tarzanek:
OK, so actually after #1817 is merged, we can merge this, too, cann't we ?

idodeclare:
Well, it's exactly identical to commit ff310623 on #1817, but there it has a message "Add tests of fixed matching of magics" and here a message "Add tests of magics to show broken behavior."

PR #1827, opened 2017-10-18
project sync tools

vladak:
Here's set of simple scripts we've been using lately internally for setup with projects. They make it possible to handle projects in parallel, speeding the overall indexing. The handling can be a sequence of:
  - send a Message that project is synced+indexed
  - sync the project (e.g. `git pull`)
  - index the project
  - remove the previous Message

The `filelock.py` comes from https://github.com/benediktschmitt/py-filelock/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ egrok/bin
 pmd.bat
 *.iml
 .idea
+*.pyc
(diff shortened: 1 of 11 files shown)


tarzanek:
fire at will, it seems Travis is broken, but please merge, this doesn't need travis build

vladak:
Okay, will send another pull req once I rewrite the syncing script from shell to Python.

tarzanek:
you can also send request for the shell script, I don't mind in what language is it written as long as it works ;)

PR #1832, opened 2017-10-19
Feature/magics

idodeclare:
(This is a recreation of #1817 which is rebased since AnalyzerGuruTest had conflicts from recent merges and with an additional commit to "move BOM-related methods to IOUtils where other BOM-methods live".)

- Fix so magics (TreeMap) lookup can work by inspecting it with full
  two- (e.g. "/usr/bin/env perl") or one-word (e.g. "/usr/bin/perl") magics
  extracted from file contents.
- Condense spaced, extracted hashbangs for lookup (e.g.,
  " #! /usr/bin/env  perl" to "#!/usr/bin/env perl").
- Add Comparator so longer magic strings are checked first during the
  imprecise phase of matching (so ShAnalyzerFactory's short "#!" magic is not
  accidentally preferred).
- Add property, isPreciseMagic, to FileAnalyzerFactory.Matcher so byte-precise
  magic detection can be done early. (Only ZipAnalyzerFactory now has this set
  to true.)
- Use a UTF BOM as a precise indication of plain text in the otherwise
  imprecise PlainAnalyzerFactory.Matcher.isMagic().
- Add tests of fixed matching of magics
- Fix historically broken Java class magic
- Add test of ELF magic
- Move BOM-related methods to IOUtils where other BOM-methods live

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -101,6 +101,7 @@ import org.opensolaris.opengrok.history.HistoryGuru;
 import org.opensolaris.opengrok.history.HistoryReader;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.search.QueryBuilder;
+import org.opensolaris.opengrok.util.IOUtils;
 import org.opensolaris.opengrok.web.Util;
 
 /**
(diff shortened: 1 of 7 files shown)


tarzanek:
thank you for this, I think based on the talk in #1817 we could just merge this

tarzanek:
thank you for this, I think based on the talk in #1817 we could just merge this

PR #1833, opened 2017-10-19
Feature/perl symbols

idodeclare:
Hello,

Please consider for merging this patch with another set of Perl improvements:

* improvements to parsing of quote-like operators, including a fix to include symbols in quoting characters, a fix to recognize spaces around braced identifiers, etc.
* extract Perl symbols using the equivalent parsing as done for xref, and add a test of extracted Perl symbols.
* add Perl pragmas as keywords
* recognize Perl special identifiers, and format as keywords
* add PLX as a Perl suffix, to align with ctags perl.c
* improve POD recognition
* clean up to use existing yypush(), yypop(); and add yyjump()

I added test lines for all the improvements and fixes. Now that feature/perl_xref_tests is merged, I can raise this PR.

Thank you.


idodeclare:
Oh, I should point out that to rationalize PerlXref.lex and PerlSymbolTokenizer.lex where the latter was case-sensitive and the former was not and also in order to improve POD handling, I revised PerlXref.lex to be case-sensitive in 77053af1d7. This required modifying just the POD-related patterns and the {FileExt} pattern.

tarzanek:
I don't know why the build suddenly failed with javadoc errors, I hoped travis builds are ignoring the javadoc errors, seems like we need to update pom.xml to get a fix for the plugin for javadoc

test/org/opensolaris/opengrok/analysis/perl/sample.pl
```diff
@@ -1,280 +0,0 @@
-#!/usr/bin/perl
```
tarzanek: symlink ... this will make development problematic on windows ...

tarzanek:
lgtm, let's merge ...
@vladak @tulinkry  any comments ?

idodeclare:
I've been extending this pull request commit-by-commit. Each commit has stood on its own as a working version (along with its predecessors) and could have been merged at any time.

I finished rationalizing the differences between PerlXref and PerlSymbolTokenizer and pushed up a last modification for them to %include the same PerlProductions.lex.

tarzanek:
great, includes is something I've been longing to get to jflex files for a long time!
let me review

tarzanek:
fwiw the problem with includes was that they didn't really accept relative paths, not sure if this improved in last jflex or not

tarzanek:
but if it works, then I guess we can merge ...

idodeclare:
@tarzanek , in debugging the %include syntax, where the 1.6.1 manual reads to use quotation marks but where that doesn't actually work, I did try at one point an absolute path with and without quotation marks without success. Possibly only local %include works.

tarzanek:
yeah, local works, I tried to abstract stuff from all lex files to top dir and that didn't work

tarzanek:
ok, let's merge this

idodeclare:
@tarzanek , perhaps it would do to revise the &lt;run-jflex&gt; and &lt;run-jflex-and-disable-buffer-expansion&gt; targets to copy(if-updated) the source .lex files to an intermediate directory, including shared includes?

tarzanek:
hmm , that could be a good idea and we could sacrifice space/time for easier maintainance of lex files ... and especially keeping fixes and improvements in sync

PR #1834, opened 2017-10-20
Double click on project "bubble" should lead to project directory

### LINKED ISSUE #1807: double click on project "bubble" should lead to project directory browsing
Just like double clicking on project in the project picker leads to browsing the project, the same should happen when double clicking on project "bubble".

Orviss:
fixes #1807 

Double clicking on project "bubble" now invokes double click on the item in search list.

One question: I do not like 'I' cursor on the bubble. I think it would be wise to change it to default cursor (arrow). Solution is quite easy (adding `cursor:default;` to css). What do you think? 

Thanks :)


tulinkry:
> What do you think?

yes

PR #1837, opened 2017-10-20
"Help" button should have the same style as "Search" and "Clear"

### LINKED ISSUE #1362: "Help" button font size differs
The "Help" button font size is different to the other two buttons on the main page.

help-font

Orviss:
fixes #1362

I wanted to preserve the link (`<a>`) but I was not sure if I styled the link in the same way as `<input>` that it would look the same on different browsers.

Thanks :)

### DIFF
--- a/web/menu.jspf
+++ b/web/menu.jspf
@@ -219,7 +219,8 @@ org.opensolaris.opengrok.web.Util"
     <input tabindex="9" class="submit btn" type="submit" value="Search"/>
     <input tabindex="10" class="submit btn" onclick="javascript: clearSearchFrom();"
            type="button" value="Clear"/>
-    <a href="help.jsp" target="_blank" class="help btn" title="Help" tabindex="11" />Help</a>
+    <input tabindex="11" class="submit btn" onclick="window.open('help.jsp', '_blank');"
+           type="button" value="Help"/>
 </div>
 </div>
 <div id="ltbl">


web/menu.jspf
```diff
@@ -219,7 +219,8 @@ org.opensolaris.opengrok.web.Util"
     <input tabindex="9" class="submit btn" type="submit" value="Search"/>
     <input tabindex="10" class="submit btn" onclick="javascript: clearSearchFrom();"
            type="button" value="Clear"/>
-    <a href="help.jsp" target="_blank" class="help btn" title="Help" tabindex="11" />Help</a>
+    <input tabindex="11" class="submit btn" onclick="window.open('help.jsp', '_blank');"
```
tarzanek: shouldn't it be enough to just change the class ?

Orviss: No, the styles that would need to be added to achieve the same effect:
```css
font-weight: bold;
padding-top: 2px;
padding-bottom: 3px;
color: black; /* or something like that */
```

screen shot 2017-10-20 at 21 01 50

As you can see on the screenshot, the padding was taken form `input` and for specific input type. I do not know if this style will be consistent across all browsers or if it won't change in the future. That's why I chose this implementation. (I might be wrong, I am not very skilled in frontend stuff)

tulinkry: @tarzanek no that's not enought. I fought with this several times already

vladak: looks like the consensus was reached, merging.

PR #1840, opened 2017-10-20
Replace Java 9 deprecated calls with appropriate substitutes

Orviss:
When compiling with Java 9 deprecation warnings were shown so I repaired them.

Thanks for review :)

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -31,6 +31,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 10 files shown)


tarzanek:
I suppose these changes won't compile with 1.8 anymore, will they ?

Orviss:
Why wouldn't they? You can also see that the build passes and I believe that travis has still Java 8.

vladak:
Could you maybe summarize the types of problems being fixed in this PR ? Otherwise I have no other comments.

Orviss:
Mainly I changed:
<https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--> to `getDeclaredConstructor().newInstance()` which propagated some new exceptions
and
<https://docs.oracle.com/javase/9/docs/api/java/lang/Long.html#Long-long-> for which I used autoboxing/unboxing

vladak:
thanks, merging.

PR #1842, opened 2017-10-21
Feature/cache control

idodeclare:
Hello,

Please consider for integration this patch to set HTTP cache-controls for most OpenGrok content. Rather than use a container-specific filter (e.g.. org.catalina...), which I think would not be accepted for OpenGrok, I imported a simple filter implementation that is licensed freely from the book support site of "Servlets and JavaServer Pages; the J2EE Web Tier" by Jayson Falkner.

Because the filter is simple, it applies only by url-pattern (and not by e.g., content-type as Tomcat's might do). I set a one-half hour expiration for `/download/*`, `/history/*`, `/xref/*`, and `/raw/*` and one day for `/default/*` and `/js/*`. Any other content (e.g. `/Help.jsp`) will not be affected by this patch.

I split it into two commits. The first is just a re-save of web.xml from the NetBeans editor for its formatting revisions, and the second is the addition of and configuration of the filters.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
+++ b/src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
@@ -1,0 +1,46 @@
+/*
+ * The contents of this file are Copyright © 2003 Jayson Falkner made available
+ * under free license: "All of the code from Servlets and JSP the J2EE Web Tier
+ * is freely available. Additonally all of the code the book's code relies on
+ * is also freely available. With the exception of Sun's Java Development Kit,
+ * everything the book relies on is also open-source. You are encouraged to
+ * look at, learn from, and use everything!"
+ */
+
+package org.opensolaris.opengrok.web;
+
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import java.util.*;
+
+public class ResponseHeaderFilter implements Filter {
+    FilterConfig fc;
+
+    @Override
+    public void doFilter(ServletRequest req, ServletResponse res,
+        FilterChain chain)
+        throws IOException, ServletException {
+
+        HttpServletResponse response = (HttpServletResponse) res;
+
+        // set the provided HTTP response parameters
+        for (Enumeration e = fc.getInitParameterNames(); e.hasMoreElements();) {
+            String headerName = (String)e.nextElement();
+            response.addHeader(headerName, fc.getInitParameter(headerName));
+        }
+
+        // pass the request/response on
+        chain.doFilter(req, response);
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) {
+        this.fc = filterConfig;
+    }
+
+    @Override
+    public void destroy() {
+        this.fc = null;
+    }
+}
--- a/src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
+++ b/src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
@@ -5,6 +5,7 @@
  * is also freely available. With the exception of Sun's Java Development Kit,
  * everything the book relies on is also open-source. You are encouraged to
  * look at, learn from, and use everything!"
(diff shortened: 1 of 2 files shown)


vladak:
ad the re-save of web.xml: it lost the comments referencing the constants in the code (`<!-- DOWNLOAD_P -->`)

src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
```diff
@@ -0,0 +1,46 @@
...
+ * under free license: "All of the code from Servlets and JSP the J2EE Web Tier
+ * is freely available. Additonally all of the code the book's code relies on
+ * is also freely available. With the exception of Sun's Java Development Kit,
+ * everything the book relies on is also open-source. You are encouraged to
+ * look at, learn from, and use everything!"
+ */
+
+package org.opensolaris.opengrok.web;
+
+import java.io.*;
```
vladak: pls use specific imports, in OpenGrok code base we try to avoid wildcard imports.

vladak:
That's pretty neat. In my setup, I use Apache's  `Header set Cache-Control` directive to manipulate the expires. Having this centrally in OpenGrok setup is better IMO. Fix the imports and this can be merged I think.
I wonder if there is way how to test this in unit tests.

idodeclare:
@vladak , I'm not sure of a way to test this in unit tests. It might be nice to have higher-level integration tests at some point that can test real web endpoints. That would allow testing this patch as well as e.g. catching runtime problems with the OpenGrok script.

idodeclare:
@vladak, I pushed a small revision to the imported ResponseHeaderFilter to check that `response.containsHeader()` is false to allow designating ExpiresHalfHourFilter as the default (by its order) without overriding ExpiresOneDayFilter.

idodeclare:
Oops, let me push a revision to get rid of those now-superfluous url-patterns.

src/org/opensolaris/opengrok/web/ResponseHeaderFilter.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void doFilter(ServletRequest req, ServletResponse res,
+        FilterChain chain)
+        throws IOException, ServletException {
+
+        HttpServletResponse response = (HttpServletResponse) res;
+
+        // set the provided HTTP response parameters
+        for (Enumeration e = fc.getInitParameterNames(); e.hasMoreElements();) {
+            String headerName = (String)e.nextElement();
```
vladak: get this be null ?

idodeclare: I do not believe so — but if it were I think it would still be OK in order to reveal the configuration error. As a related example, if you have an invalid url-pattern, the servlet will not start up.

vladak:
merging, thanks !

tarzanek:
I actually hope this will solve the problem with js caching
BUT then I fear this will increase the traffic to opengrok servers, since we were caching the stuff for longer by default

tarzanek:
@idodeclare the defaults you used here ... how do they compare to e.g. previous defaults on tomcat 8 ?

idodeclare:
@tarzanek , in my testing, Cache-Control was undefined, so only ETag comparison was done.

For e.g. `js/*` and `default/*`, ETag checking seems reliable. With this feature/cache-control patch, ETag checking will be done more infrequently (no more often than 24 hours).

For e.g. `xref/*`, the ETag checking did not seem reliable, so usually even if content should have changed (e.g., by git-update) or formatting should have changed (e.g., a new OpenGrok deployment of Xref-logic), I would still have to force-refresh my browser.

idodeclare:
@tarzanek , having written "ETag checking did not seem reliable" for `xref/*`", I can say that the Cache-Control patch will not have addressed that — except that the ETags will be checked less frequently with non-zero Max-Age.

Let me investigate some more — but in any case I argue that non-zero Max-Age will not result in more server load, because the ETags would have been checked anyway.

PR #1844, opened 2017-10-22
Feature/fix ioexception

idodeclare:
Hello,

Please consider for integration this patch to fix an IOException that is occurring for code files larger than 8192 bytes in size. I created a sample, a.csproj; added a (breaking) test in the first commit; and then fixed it in the second.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -125,7 +125,14 @@ public class AnalyzerGuru {
      * Define very close to OPENING_MAX_CHARS so the input stream stays easily
      * in the <code>mark</code> limit
      */
-    private static final int OPENING_BUF_SIZE = 102;
+    private static final int OPENING_BUF_SIZE = 128;
+
+    /**
+     * Set to 16K -- though debugging shows it would do with only 8K+3
+     * (standard buffer for Java BufferedInputStream plus 3 bytes for largest
+     * UTF BOM) 
+     */
+    private static final int MARK_READ_LIMIT = 1024 * 16;
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AnalyzerGuru.class);
 
@@ -850,7 +857,7 @@ public class AnalyzerGuru {
     private static String readOpening(InputStream in, byte[] sig)
         throws IOException {
 
-        in.mark(512);
+        in.mark(MARK_READ_LIMIT);
 
         String encoding = IOUtils.findBOMEncoding(sig);
         if (encoding == null) {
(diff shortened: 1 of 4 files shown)


src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
```diff
@@ -125,7 +125,14 @@
      * Define very close to OPENING_MAX_CHARS so the input stream stays easily
      * in the <code>mark</code> limit
      */
-    private static final int OPENING_BUF_SIZE = 102;
+    private static final int OPENING_BUF_SIZE = 128;
```
vladak: why is this being bumped to 128 ?

test/org/opensolaris/opengrok/analysis/AnalyzerGuruTest.java
```diff
@@ -244,6 +242,15 @@ public void getAnalyzerFactoryClass() {
     }
 
     @Test
+    public void shouldNotThrowGettingCsprojOpening() throws IOException {
+        InputStream res = getClass().getClassLoader().getResourceAsStream(
+            "org/opensolaris/opengrok/analysis/a.csproj");
```
vladak: I wonder if we should stick the test files under subdirectory or elsewhere. There does not seem to be a convention for tests yet.

vladak:
This is kind of sad bug (given the conditions it wouldn't be fair to call the fix likewise). `addFile()` constructs the stream like this:

```
        try (InputStream in = new BufferedInputStream(new FileInputStream(file))) {
            fa = AnalyzerGuru.getAnalyzer(in, path);
        }
```

and this gets eventually passed to `readOpening()`. So the chain of streams is FileInputStream -> BufferedInputStream -> InputStream and the chain of readers is InputStreamReader -> BufferedReader.

Now, BufferedReader has its own buffer as well as BufferedInputStream. `readOpening()` reads the data one character at a time, however the underlying BufferedReader reads by its own buffer size and in turn BufferedInputStream uses its own buffer size that advances the position in the output stream. So even though max 100 characters (`OPENING_MAX_CHARS`) can be read by the cycle in reality it could be 8K characters.

The last constant seems to be unreachable from  `readOpening()` (https://stackoverflow.com/a/42572182)

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
```diff
@@ -125,7 +125,14 @@
      * Define very close to OPENING_MAX_CHARS so the input stream stays easily
      * in the <code>mark</code> limit
```
vladak: given this bug this comment is not very relevant. What about unifying OPENING_MAX_CHARS with OPENING_BUF_SIZE ?

idodeclare: You're right. It's not useful as a distinct value. I pushed a removal.

test/org/opensolaris/opengrok/analysis/AnalyzerGuruTest.java
idodeclare: I thought about that — but a lower-cased file name does at least show it at the bottom of the test/org folder in NetBeans.

vladak: Yeah, no big deal. We should probably rethink and change this for all tests.

PR #1845, opened 2017-10-22
Drop static writeXref()s; use existing thread-cached analyzers instead

idodeclare:
Hello,

Please consider for integration this patch that simplifies the FileAnalyzer and FileAnalyzerFactory subclasses by dropping static writeXref() methods and dropping FileAnalyzerFactory.writeXref() and all its overrides — and by just relying on the existing classes' newAnalyzer() and newXref() methods enlisted from AnalyzerGuru and TextAnalyzer.

Also:
- Move "&lt;pre&gt;"-manipulation and div#man writing into TroffXref so TroffAnalyzer doesn't have to know HTML formatting details.
- Do not force scopesEnabled,foldingEnabled to true for PowershellAnalyzer.
- Refactor (non-static) writeXref() to use new WriteXrefArgs class.

Thank you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 *~
-nbproject/*~
+*.swp
 /nbproject/configs/*.properties
-opengrok-web-nbproject/nbproject/*~
-opengrok-web-nbproject/nbproject/build-impl.xml~
 tags
 TAGS
 build
@@ -39,8 +37,6 @@ webrev
 opengrok-web-nbproject/nbproject/private/
 opengrok-web-nbproject/nbproject/genfiles.properties
 run_my.bat
-*.xml~
-opengrok-web-nbproject/nbproject/build-impl.xml~
 build.bat
 jacoco
 jacoco.exec
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -460,7 +460,14 @@ public class AnalyzerGuru {
             // spaces to match the project's tab settings.
             input = ExpandTabsReader.wrap(in, project);
         }
-        factory.writeXref(input, out, defs, annotation, project);
+
+        WriteXrefArgs args = new WriteXrefArgs(input, out);
+        args.setDefs(defs);
+        args.setAnnotation(annotation);
+        args.setProject(project);
+
+        FileAnalyzer analyzer = factory.getAnalyzer();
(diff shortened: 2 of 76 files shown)


shaehn:
So if you removed the scopesEnabled and foldingEnabled for Powershell, how do we get those items to be activated? At the time I created the parser, that was the only way I could get those elements to appear in the web UI.

tarzanek:
interesting cleanup, won't too many arg instances be created ?

idodeclare:
@shaehn , the values are meant to come from Configuration and defaulted to `true` in its constructor. This will accomplish what you would like — without the hard-coding.

idodeclare:
@tarzanek , I don't believe the args instances would cause a problem. They are tiny wrapper objects (so they won't be the cause of memory pressure that initiates GC compared to the lexers themselves and to the string manipulation involved therein), and an arg instance only survives the duration of the writeXref() call (so it will be young-generation-eden only and quickly handled).

shaehn:
I don't know if this would be considered a little wonky, but where you have introduced throwing IllegalArgumentException, the leading character surrounding the item being declared null is not an apostrophe character ('), but rather an accent grave (`).
```
In TextAnalyzer.java:
    IllegalArgumentException("`args' is null");
In WriteXrefArgs.java: 
    IllegalArgumentException("`in' is null");    
    IllegalArgumentException("`out' is null");
```

idodeclare:
@shaehn , that's just a quoting style. I use it when the quoted part is a language symbol.

tarzanek:
from where I stand this looks ok, it simplifies the xref calling
@vladak @tulinkry any comments ?
otherwise I guess we can merge

tulinkry:
This is perfect. I'm so glad you did this.

.gitignore
```diff
@@ -39,8 +37,6 @@ webrev
 opengrok-web-nbproject/nbproject/private/
 opengrok-web-nbproject/nbproject/genfiles.properties
 run_my.bat
-*.xml~
-opengrok-web-nbproject/nbproject/build-impl.xml~
 build.bat
```
tulinkry: Why are some changes in this file needed?

idodeclare: I added an entry to ignore vim's .swp files and also noticed the redundancy with entries for files ending with "~" already excluded by line 1.

tulinkry: ok

tarzanek:
ok, so let's have fun, will merge now ;)

PR #1846, opened 2017-10-25
fix relative paths for symlinked repositories

### LINKED ISSUE #1831: Searching for repository stopped working for projects using symbolic links.
Previously, (say about 3 months ago) I was able to use symbolic links for Accurev workspaces. Today this no longer works and I think it is due to the change from #1738. The problem lies in the RepositoryFactory.java at line 104, 

repo.setDirectoryName(file.getCanonicalPath());

Using getCanonicalPath is what is causing the problem.  It turns the symbolic link path into the real path and then inside setDirectoryName when it attempts to compare the path to the source root so that it can strip it off, it fails. The full canonical path becomes the directoryNameRelative instead and evrything goes down hill from there.

When I change the line to:

repo.setDirectoryName(file.getPath());

I get the previous behavior that I expected, because setDirectoryName is able to strip off the source root directory.

So why is getCanonicalPath being used?

vladak:
fixes #1831

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -392,17 +392,18 @@ public final class RuntimeEnvironment {
         String canonicalPath = file.getCanonicalPath();
         String sourceRoot = getSourceRootPath();
         
-        if(sourceRoot == null){
+        if (sourceRoot == null) {
             throw new FileNotFoundException("Source Root Not Found");
         }
         
         if (canonicalPath.startsWith(sourceRoot)) {
             return canonicalPath.substring(sourceRoot.length() + stripCount);
         }
+        
         for (String allowedSymlink : getAllowedSymlinks()) {
             String allowedTarget = new File(allowedSymlink).getCanonicalPath();
             if (canonicalPath.startsWith(allowedTarget)) {
-                return canonicalPath.substring(allowedTarget.length()
+                return file.getAbsolutePath().substring(sourceRoot.length()
                         + stripCount);
             }
         }
--- a/src/org/opensolaris/opengrok/history/FileHistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/FileHistoryCache.java
@@ -614,23 +614,27 @@ class FileHistoryCache implements HistoryCache {
     public boolean hasCacheForDirectory(File directory, Repository repository)
             throws HistoryException {
         assert directory.isDirectory();
-        Repository repos = HistoryGuru.getInstance().getRepository(directory);
-        if (repos == null) {
+        Repository repo = HistoryGuru.getInstance().getRepository(directory);
+        if (repo == null) {
             return true;
         }
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         File dir = env.getDataRootFile();
-        dir = new File(dir, this.historyCacheDirName);
+        dir = new File(dir, FileHistoryCache.historyCacheDirName);
         try {
-            dir = new File(dir, env.getPathRelativeToSourceRoot(
-                new File(repos.getDirectoryName())));
+            dir = new File(dir, env.getPathRelativeToSourceRoot(new File(repo.getDirectoryName())));
         } catch (IOException e) {
             throw new HistoryException("Could not resolve " +
-                    repos.getDirectoryName()+" relative to source root", e);
+                    repo.getDirectoryName()+" relative to source root", e);
         }
         return dir.exists();
     }
 
+    @Override
+    public boolean hasCacheForFile(File file) throws HistoryException {
+        return getCachedFile(file).exists();
+    }
+    
     public String getRepositoryHistDataDirname(Repository repository) {
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         String repoDirBasename;
@@ -645,7 +649,7 @@ class FileHistoryCache implements HistoryCache {
         }
 
         return env.getDataRootPath() + File.separatorChar
-            + this.historyCacheDirName
+            + FileHistoryCache.historyCacheDirName
             + repoDirBasename;
     }
 
--- a/src/org/opensolaris/opengrok/history/HistoryCache.java
+++ b/src/org/opensolaris/opengrok/history/HistoryCache.java
@@ -93,6 +93,14 @@ interface HistoryCache {
      */
     boolean hasCacheForDirectory(File directory, Repository repository)
             throws HistoryException;
+    
+    /**
+     * Check if the specified file is present in the cache.
+     * @param file the file to check
+     * @return {@code true} if the file is in the cache, {@code false}
+     * otherwise
+     */
+    boolean hasCacheForFile(File file) throws HistoryException;
 
     /**
      * Get the revision identifier for the latest cached revision in a
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -810,14 +810,7 @@ public final class HistoryGuru {
     }
 
     protected Repository getRepository(File path) {
-        File file;
-
-        try {
-            file = path.getCanonicalFile();
-        } catch (IOException e) {
-            LOGGER.log(Level.WARNING, "Failed to get canonical path for " + path, e);
-            return null;
-        }
+        File file = path;
 
         while (file != null) {
             Repository r = repositories.get(file.getAbsolutePath());
--- a/src/org/opensolaris/opengrok/history/HistoryGuru.java
+++ b/src/org/opensolaris/opengrok/history/HistoryGuru.java
@@ -391,7 +391,6 @@ public final class HistoryGuru {
                         }
                     }
                 } else {
-                    repository.setDirectoryName(path);
                     if (RuntimeEnvironment.getInstance().isVerbose()) {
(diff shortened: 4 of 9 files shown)


shaehn:
I looked at the test again a little closer and it really didn't matter that what was set up was a symlink or not. With the given version of setDirectoryName, its going to lop off source root, or not, end of story.

vladak:
Reorganized the PR into set of commits, where the commit with the fix is on the top. If I revert that commit and run the test case I get this exception:

```
java.io.IOException: Cannot run program "hg" (in directory "/tmp/src3853862132069608017opengrok/tmp/real3845045261582926117opengrok"): error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at org.opensolaris.opengrok.util.Executor.exec(Executor.java:173)
	at org.opensolaris.opengrok.history.MercurialHistoryParser.parse(MercurialHistoryParser.java:79)
	at org.opensolaris.opengrok.history.MercurialRepository.getHistory(MercurialRepository.java:596)
	at org.opensolaris.opengrok.history.Repository.createCache(Repository.java:330)
	at org.opensolaris.opengrok.history.HistoryGuru.createCache(HistoryGuru.java:560)
	at org.opensolaris.opengrok.history.HistoryGuru.access$000(HistoryGuru.java:59)
	at org.opensolaris.opengrok.history.HistoryGuru$1.run(HistoryGuru.java:611)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 12 more
```

which detects the problem.

vladak:
@shaehn Sorry, I do not understand your comment. (might be because of the late hour in this part of the planet) If I modify the test to clone into source root directly, it will pass.

shaehn:
@vladak  .... sorry your up so late.  

Here is what I was thinking. If setDirectoryName does nothing more than strip off the source root (when it finds it) and all other inputs are just considered to be the relative path, calling the setDirectoryName with '/source/root/realDirectory' will set directoryRelative to '/realDirectory'. Calling setDirectoryName with '/source/root/symlinkedDirectory' will set directoryRelative to '/symlinkedDirectory', whether it really is a symlink or not. So basically RepositoryInfo is agnostic to symlinks (which is what you wanted, I think). So we never want to use canonicalPath, or any other interface that obtains a real path prior to calling setDirectoryName or we get the problem that I encountered.

I am guessing that most repositories don't care one way or the other about symlinks. That is not the case for AccuRev. It only knows how to get to its data through real paths. I have attempted to handle that in the AccuRev repository code itself.  I originally did not thing the change that you made was going to fix the problem because in my early testing it just wasn't working. After all the flurry of programming activity I went through and your comments indicating that my pull request wasn't correct, I had to let the dust settle a bit and try to think about the problem again. I picked up your change to RepositoryFactory (removal of canonical naming), rebuilt and ran the test again. This time it worked (finally...)

If I appear a little stressed, sorry. I have been working on this all summer and don't know how much longer I will get the chance to do more.  I so want to get a stable enough version that I can use for my work team so I can deploy it on our vast repository.  This newer version of OpenGrok is so much better than the .11 we currently use.

I will still have to call out another issue, but its totally AccuRev related (uncovered during all my testing with this directory naming stuff).

shaehn:
@vladak 

OK, I see what the problem is and why I solved it the way I did in my pull request. What originally worked before the changes you made to setDirectoryName/getDirectoryName was that the call to getCanonicalName caused the real directory data to be put into the RepositoryInfo. Accurev was happy. It only likes the real thing.

When you made the change to those functions with your previous change and the fix in this pull request, those functions no longer store the real directory when symbolic links are involved. Accurev repository is busted. (History and Annotate links are not available in web browser). 

Here is where the problem comes into play. There is a call to PageConfig.hasHistory in minisearch.jspf and PageConfig.hasAnnotations that both call the HistoryGuru's hasHistory method.  Now when the HistoryGuru's repositories map has Accurev associated with a symbolic name key and it searches for the repository (see HistoryGuru.getRepository(file)), that method uses canonical and absolute names. It will never find the repository. Thus, no links on web page.  Here is a sample of my debug messages.
(Note: 24 is symbolically linked to /export/home/Browser/workspaces/CP.24.0_win)

```
PageConfig resourceFile: /export/home/Browser/snapshots/active/src/24/print_server/Modules-pre.mk
locate repository for: /export/home/Browser/snapshots/active/src/24/print_server/Modules-pre.mk
looking: /export/home/Browser/workspaces/CP.24.0_win/print_server/Modules-pre.mk
looking: /export/home/Browser/workspaces/CP.24.0_win/print_server
looking: /export/home/Browser/workspaces/CP.24.0_win
looking: /export/home/Browser/workspaces
looking: /export/home/Browser
looking: /export/home
looking: /export
looking: /
CANT find dang repo
```
So to get this to work, I think the HistoryGuru.getRepository is going to have to get agnostic about symlink paths too.  I am going to spend tomorrow testing that premise.

shaehn:
@vladak 
The following change to HistoryGuru.getRepository fixes the problem of not seeing History and Annotate links on web page.

```
    protected Repository getRepository(File path) {
        File file=path;

        while (file != null) {
            Repository r = repositories.get(file.getPath());
            if (r != null) {
                return r;
            }
            file = file.getParentFile();
        }

        return null;
    }

```
The results of my test produced the following output: (Remember that 24 is the symbolically linked directory)

```
PageConfig resourceFile: /export/home/Browser/snapshots/active/src/24/print_server/Modules-pre.mk
locate repository for: /export/home/Browser/snapshots/active/src/24/print_server/Modules-pre.mk
looking: /export/home/Browser/snapshots/active/src/24/print_server/Modules-pre.mk
looking: /export/home/Browser/snapshots/active/src/24/print_server
looking: /export/home/Browser/snapshots/active/src/24
```

I basically removed the getCanonicalFile and getAbsolutePath calls which turned symbolically linked paths in to real paths and substituted a getPath call.

The usage of canonical and absolute names fixes paths that have '..' or '.' characters used in their names, it unfortunately brings out the real path.  This fix will works assuming that the path entered never has any of the '..' or '.' characters embedded in it.

shaehn:
@vladak 
There is one more "gotcha" in the area of historycache directory creation. I observered that there were no sub directories created for each of the separate projects.  The problem lies in RuntimeEnvironment.getPathRelativeToSourceRoot(). It is using a canonical path and attempting to see if the source root is part of it to strip it off (just like it is being done in RepositoryInfo.setDirectoryName()).

I made the following change in my local copy and it cured the problem.

image

shaehn:
@vladak 
Have you had a chance to go over the changes I have suggested?  I have been running with the changes locally here and things are working as expected/desired.  Is there something else that you are expecting me to do? It would be nice to get the changes pulled into the main branch so that I don't have to keep a custom version locally.

vladak:
Sorry for the delay, had to do a bit of firefighting with auth plugins to fix internal deployment. I will try to get on this one early next week.

vladak:
@shaehn process wise: if you agree that the changes in this PR can be done as a basis for further changes, let's merge this in and you can submit PRs that will provide refinements.

As for your suggested changes in `getPathRelativeToSourceRoot()` I think they need to be a bit more sophisticated. Like it is, I am worried that the whole idea of "allowed symlinks" (see `Indexer.java` lines 154 and below) would be short-circuited (we should really have a test for that).

vladak:
And yes, I can see the history cache not being created for symlinked repositories.

vladak:
Another problem is that symlinked repositories are not listed on the index page.

vladak:
The symlinked repository is not present in the list of repositories (it is present in the project picker though) on the index page because it is missing in the `RuntimeEnvironment.repository_map` which is a consequence of `getPathRelativeToSourceRoot()` (called from `generateProjectRepositoriesMap()` which populates `repository_map`) not working properly as mentioned above.

vladak:
Okay, I think most of the problems should be fixed now. If AccuRev (or any other SCM) needs canonical paths, the canonicalization should be done within its code.

vladak:
@shaehn you probably meant line 832/833 which is in HistoryGuru.java#getRepository():

```
protected Repository getRepository(File path) {
        File file = path;

        while (file != null) {
            Repository r = repositories.get(file.getAbsolutePath());
            if (r != null) {
                return r;
            }
            file = file.getParentFile();
        }

        return null;
    }
```

The `repositories` is defined like this:

```
/**
     * map of repositories, with {@code DirectoryName} as key
     */
    private Map<String, Repository> repositories = new ConcurrentHashMap<>();
```

and populated like this in `addRepositories()`:

```
                    repositories.put(repository.getDirectoryName(), repository);
```

so I think it needs absolute path because `getDirectoryName()` is absolute.

As for the changes in `RuntimeEnvironment`, I believe (it is always better to put comments in the PR if possible to avoid confusion) you meant `getPathRelativeToSourceRoot()`. This method still needs to contain the check whether the symlink is allowed or not and for that the canonical path is handy.

Also, you can clone my branch and run it in your environment to see if it works there.

vladak:
closing in favor of #1911

Issue #1754, opened 2017-08-26
RepositoryInfo directoryName values should be relative to source root


vladak:
The `directoryName` member of `RepositoryInfo` should be relative to source root to make it easier to use the same configuration for different environment (thinking near instant reindex).

PR #1848, opened 2017-10-26
Fixes relative path problem with symbolic linked repositories

shaehn:
This change pays attention to the incoming path name such that when the name does not contain the initial source root directory obtained from the command line, it uses an heuristic to determine if the name is a simple relative name (/directory), or a completely different path, assuming it came from a symbolic link (/alternate/real/path/obtained/by/enumerating/symbolic/link)


vladak:
This is a dup of #1846.

Also, as explained in the bug, adding non-relative paths to `RepositoryInfo` is no go.

shaehn:
I don't understand why my pull requests keep showing all the merges that were performed way back when. I have tried doing the git pull --rebase on my master branch and I thought that was supposed to take care of it, but clearly it has not. Suggestions on how to fix this?

shaehn:
@vladak   I pushed this stuff up because you did not respond to my previous ask. So now that I have your attention, I don't know what you mean by 

"adding non-relative paths to RepositoryInfo is no go"

Can you give me an example?  The code accepts /directory a relative path, or it accepts the full path provided by a call to the getCanonicalPath used in RepositoryFactory. So what exactly is the problem?

shaehn:
So what are the steps to take to back out this change from the master branch here?

vladak:
as for the pull requests containing already merged stuff: I think you just need to follow the advice given to you by @idodeclare in #1812

vladak:
The responses will come out eventually, I just need to digest all the info :-) (also keep in mind that currently all OpenGrok core contributors have OpenGrok as hobby project) The goal of the changes for #1754 was to have just single `sourceRoot` absolute path in the configuration, the rest is relative. This helps to construct a setup where (with the help of file-system snapshots/clones) one gets resync/reindex with nearly zero inconsistency window. Also, it keeps the config tidy.

idodeclare:
@shaehn , your master branch looks to have deviated. I would recommend this:

1) create a branch from your current, clean master working directory to save its state:
    $ git checkout master && git checkout -b probmaster

2) reset your master back to the upstream state (assuming you have named the remote 'upstream'):
    $ git checkout master && git reset --hard upstream/master

3) create a branch to hold a squashed version of your 'probmaster', named for the issue at hand:
    $ git checkout master && git checkout -b fix_symlink_path && git merge --squash probmaster

Hopefully it merges cleanly, but if any merge conflicts arise, they should be resolved, and the merge should be committed.

(N.b., the `git checkout master` at the start of each step can be redundant, but I just was reinforcing that branch starting point is important.)

shaehn:
@idodeclare  Thank you for the instructions.  Before I go down that path, how do I back out the changes that I put up in my master branch that is part of this pull request, or will your instructions fix that?

shaehn:
@vladak I want to close this pull request, so that your fixes (along with my suggested needed changes in #1846) can be used to fix the symlink problem.

PR #1850, opened 2017-10-27
Re-jflex if *.lexh is newer than the .lex

idodeclare:
Hello,

Please consider for integration this patch to touch a .lex file if any peer *.lexh is newer. This way, an affected .lex will be re-processed if its %include file(s) are newer.

The conditional logic is dependent on the if/unless XML namespaces introduced in Apache Ant 1.9.1 (May 21, 2013).

(The Perl %include file is renamed PerlProductions.lexh.)

Thank you.


tarzanek:
I thought that jflex detects a change in an include and will rebuild, won't it ?

idodeclare:
No, my testing shows that the &lt;jflex&gt; task just examines the modification time of the source .lex file w.r.t. the generated output without inspecting the source file contents.

tarzanek:
so this is a jflex bug then ... ok, let's merge this workaround, but I guess jflex should fix this

tarzanek:
(will you file a jflex bug please?)

idodeclare:
Yes I will plan to file a bug.

tarzanek:
feel free to link the bug here, github is good at tracking links between repos

idodeclare:
The issue is jflex-de/jflex#217.

PR #1851, opened 2017-10-28
Add MandocAnalyzer and OPENGROK_MANDOC setting

idodeclare:
Hello,

Please consider for integration this patch to allow designating a mandoc(1) binary to be used for formatting mdoc(5) and man(5) documents — or if no binary is specified, then to detect and use the PlainTextAnalyzer for mdoc(5) documents, since TroffXref does not properly format these files.

As part of this patch, I added a DocumentMatcher class to be used by both TroffAnalyzerFactory and the new MandocAnalyzerFactory. The formerly-specified "magics" for TroffAnalyzerFactory were too imprecise to be used for distinguishing mdoc(5) and man(5) — but because they were "MAGICS" they were treated as precisely matching.

To reiterate: if a mandoc(1) binary is designated, then man(5) too will be formatted using mandoc(1). I.e., the MandocAnalyzerFactory MATCHER will also match ".TH" if a mandoc(1) binary is specified.

A single file, mandoc.css, is imported from http://mdocml.bsd.lv with its one-clause license included under LICENSE-mandoc.txt.

Thank you.


tarzanek:
I will merge after you resolve the conflict @idodeclare

tarzanek:
I am actually very surprised about all your changes, very nicely done

idodeclare:
> very nicely done

@tarzanek , thank you!

idodeclare:
OK that resolved it.

tarzanek:
merging, let's get it rolling ...

PR #1853, opened 2017-10-28
Feature/improve here docs (Perl)

idodeclare:
Hello,

Please consider for integration this patch with an improvement for parsing of stacked HERE-documents.

Thank you.


idodeclare:
Oh I’ve realized I could parse without any caveats by treating a Here-document declaration as a deferred operator, which it is.

PR #1854, opened 2017-10-29
Perl improvements 

idodeclare:
Hello,

Please consider for integration this patch with an improvements to parsing of stacked HERE-documents as well as some quote-like operator fixes identified after inspecting the OpenGrok cross-referencing of a few large Perl projects (DBI, JSON, and Text-CSV).

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/perl/PerlLexHelper.java
+++ b/src/org/opensolaris/opengrok/analysis/perl/PerlLexHelper.java
@@ -22,6 +22,8 @@
  */
 package org.opensolaris.opengrok.analysis.perl;
 import java.io.IOException;
+import java.util.LinkedList;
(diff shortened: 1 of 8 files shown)


tarzanek:
can I merge ?

tarzanek:
from my side it looks good, so if you're OK with it, I will merge

tarzanek:
I also think the best thing would be to test your analyzer changes on bigger data sets (not sure what is the biggest/most complicated perl repo) and ev. compare the stats with 1.0 (luke can check for some stats)

idodeclare:
@tarzanek , well I have been testing it with open-source Perl repos of tens of thousands of lines  — that's how I find new edge cases. But I have to review by eye — and then upon finding an issue, I add to the unit test file and fix the issue.

PR #1857, opened 2017-11-01
Remove a bit of redundancy

idodeclare:
Hello,

Please consider for integration this patch with a minor cleanup of some redundancy in shared lex logic between Xrefs and SymbolTokenizers by:

* Adding JFlexXref yypush(string) overload to align with JFlexTokenizer's.
* Publicizing JFlexXref startNewLine() so a separate, public method mediator is not needed.

Also:
* Remove unneeded pushback() and switchState() mediators which I mistakenly thought were needed.

Thank you.


tarzanek:
;) thank you

Issue #958, opened 2015-07-01
add Ruby analyzer


potashj:
The shanalyzer appears capable of finding ruby files. Editing the html to submit "Ruby" in the "Type" field yields results consisting largely of .rb, .erb, and Gemfiles in addition to some other unexpected types, but mostly what you'd expect. It'd be great to have this as select-able option by default!

tarzanek:
Why not adding a ruby analyzer?

Lookup history of commits, find perl or python analyzer, or any recent commit for analyzers
exuberant ctags supports ruby out of box, so all should be easy as far as writing a jflex gramar for it (which should be easy too)

xing the fingers
L

tarzanek:
(I guess if you have or learn lexical analysis (it's enough to read other analyzers, find closest one and copy it for your purposes and fix its lex files, read jflex docs on what and how to do it) , then this task for ruby analyzer should be a matter of few hours, it's a nice java/OpenGrok/jflex learning exercise too! :) )

tarzanek:
@potashj care to add the analyzer?
look at golang pull req, or scala, etc.  
have a look at pull req  #998

vladak:
fixed in 414fc60

PR #1861, opened 2017-11-02
Add classes to support Ruby analysis

idodeclare:
Hello,

Please consider for integration this patch to support analysis for the Ruby programming language (issue #958).

There are some related changes as well:
* extract a CtagsReader class from Ctags for testability.
* make slight revisions to Ada and Perl analyzers with lessons learned from this Ruby work.

Thank you.


tarzanek:
@idodeclare you roll man! respect!

tarzanek:
wondering if we shouldn't create a license folder so we stop polluting the main dir with license files

idodeclare:
Thanks, @tarzanek !

I did think the license notices were supposed to be prominent, so I put them in the main dir — but an alternative might be to collapse them as addenda to the main LICENSE file.

src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
```diff
@@ -41,17 +43,27 @@ import org.opensolaris.opengrok.web.Util;
...
-    private final PerlLexHelper h;
+    /**
+     * Reinitialize the lexer with new reader.
+     *
+     * @param reader new reader for this lexer
+     */
+    @Override
+    public void reInit(Reader reader) {
+        super.reInit(reader);
+        h = getNewHelper();
```
tarzanek: ???
the design of analyzers should be that they are static and can statically process any new stream ...
if you create new object in every reinit it kinda beats this design ... are you sure you want to do this?
it also will break contract with lucene

src/org/opensolaris/opengrok/analysis/JFlexXref.java
```diff
@@ -182,7 +182,7 @@ public void reInit(char[] contents, int length) {
      *
      * @param reader new reader for this lexer
      */
-    public final void reInit(Reader reader) {
+    public void reInit(Reader reader) {
```
tarzanek: I have a bad feeling this breaks how lucene analyzers are working, but I will have a look and confirm ... see below

tarzanek:
can the helpers be statical, so you don't respawn them for every new stream please ?

tarzanek:
ok, I guess you keep state in the lexers helpers ... well wondering if we should do this, I mean, we can try, but isn't there a better way?
(and we can definitely improve in next iteration, I mean, with states we can do more semantic analysis per file)

src/org/opensolaris/opengrok/analysis/perl/PerlXref.lex
idodeclare: Oh I do not agree that they are static. There is a lot of state in a jflex class, and even more in the OpenGrok JFlexTokenizer and JFlexXref abstract classes.

idodeclare:
@tarzanek ,

There are multiple problems with sharing code between the two strains (tokenizer and xref) of lexers:

1) They are already in distinct class hierarchies: JFlexTokenizer extends Tokenizer, but JFlexXref only implicltly extends Object.
2) A JFlexTokenizer subclass must yylex() to boolean, while a JFlexXref subclass must yylex to int().
3) (most importantly) **It is a huge drag to code Java in the .lex files in NetBeans or any IDE**.

So though they might share logic entirely through included .lexh files, point 3) is IMHO the biggest obstacle.

My solution has been: 1) to use a common delegate (a language-specific LexHelper class) with state data absolutely permitted; and 2) to bridge the intractable differences of the distinct inheritance chains by implementing a small-footprint, language-specific Listener interface in the .lex files along with a tiny number of methods in the .lex files that are common in name but different in return value (e.g. getSymbolReturn() returning `int` in an Xref lexer but `boolean` in a SymbolTokenizer).

src/org/opensolaris/opengrok/analysis/JFlexXref.java
idodeclare: I believe the JFlexXref subclasses are not used with Lucene — only the JFlexTokenizer subclasses.

tarzanek:
for 3) I actually did hope to abstract all(most of) java to JFlex* abstract classes

for 2) if possible move xref to use boolean too, we try to model the tokenizers based on current lucene state

for 1) the original design was to have both inspired by lucene, but over time I only fixed the ones used by lucene (so sorry for my ignorance, that I didn't realize this and got confused)

idodeclare:
> I actually did hope to abstract all(most of) java to JFlex* abstract classes

I see, but the existing abstractions are what makes it cumbersome. Since the xref- and tokenizing-lexers are divergent (i.e., JFlexXref- and JFlexTokenizer-derived), a delegate pattern — as with PerlLexHelper, AdaLexHelper, RubyLexHelper — is one way to extend them with common functionality. (Another way of course as mentioned is through shared code-gen via .lexh files, though I'll continue to argue that "it's a huge drag to code Java in the .lex files in NetBeans or any IDE.")

I hope you can approve this strategy for this new Ruby analysis.

> if possible move xref to use boolean too, we try to model the tokenizers based on current lucene state

I am skeptical. The xrefers and tokenizers would still be in divergent class hierarchies. Moreover, the xrefers' design is to produce non-Lucene documents.

I would argue that it would be better to decouple OpenGrok from Lucene (while still keeping it as a principal technology of course). I.e., to refactor JFlexTokenizer as a non-abstract class that *uses* the language-specific, JFlex symbol matchers. This would allow symbol matchers at least to inherit from the same common ancestor (below Object) as the xrefers. I have a branch where I've done this already, since it was straight-forward.

I am looking into whether it's straight-forward to go a step further and make JFlexXref a non-abstract class that likewise *uses* the language-specific, JFlex xrefers. This way, the symbol matchers and xrefers could inherit from the same base class, and thus the language=specific delegates could be moved into language-specific base classes. (Though I will say the place I arrived at as described in the previous paragraph is quite satisfying too. If you accept this pull request, I could raise a new request to show you that work.)

tarzanek:
ok, I saw the other pull req, let's continue, this area needs some love (and I am happy you care)

Issue #1859, opened 2017-11-02
generate testdata/plugins/plugins.jar


vladak:
Assuming #1620 will ship with `TruePlugin.java` and `FalsePlugin.java` it would make sense to generate the `testdata/plugins/plugins.jar` file from them during the build.

vladak:
It seems this is not really possible without introducing additional complexity since Maven builds the plugins after the indexer and there is a hook in `opengrok-indexer/build.xml` where the test plugins jar would have to be generated however the hook runs in the Indexer module, i.e. before the Plugins module and Maven does not seem to have a phase where this could be moved to.

PR #1862, opened 2017-11-03
add set of authorization plugins

vladak:
This is a set of authorization plugins written by @tulinkry.


plugins/LdapPlugin/src/opengrok/auth/plugin/LdapFilter.java
```diff
@@ -0,0 +1,153 @@
...
+     * Special values are:
+     * <ul>
+     * <li>%uid% - to be replaced with LDAP uid value</li>
+     * <li>%mail% - to be replaced with LDAP mail value</li>
+     * <li>%username% - to be replaced with OSSO user value</li>
+     * <li>%guid% - to be replaced with OSSO guid value</li>
+     * </ul>
+     *
+     * Use \% for printing the '%̈́' character.
+     *
```
tulinkry: This doc actualy lacks info that it replaces any other ldap attribute which would not be ambiguous (lines 119-126)

vladak: you can add it in later PR

vladak:
Also, the NetBeans projects are probably not working out of the box, can fix them later.

plugins/LdapPlugin/src/opengrok/auth/plugin/LdapFilter.java
vladak: or just send me the diff and I will slap on the top

tulinkry:
Just to be sure, so it eventually builds the plugins into a jar file and runs the tests or?

tulinkry:
And why is junit needed? I thought we already have it since we run some tests

vladak:
The tests of the plugins are actually not run. I've added the junit csets as a basis for running the tests via `ant` - not directly related to the auth plugins yet.

vladak:
One more note w.r.t. junit - on Travis the tests are run via Maven which I guess knows where to get Junit.jar from. For `ant` though, you had to download the junit.jar yourself and place it in the `lib` directory. It's sort of surprising that I got to fix this only now since it was a nuisance for me from the very start.

tulinkry:
ah right, but doing that from netbeans downloads the dependencies automaticaly when you add the library in the test dependencies, right?

vladak:
yep, in NB this is easily fixable.

tulinkry:
I can't see that maybe I'm just blind but did you delete the file "plugins.jar" from testdata? It seems to me that it is not used anymore.

vladak:
It was an attempt to fix #1859 which failed, reverted and renamed test plugins to `testplugins.jar` so they are not confused with `plugins.jar` which actually gets shipped.

vladak:
@tulinkry mentioned on Slack that he has not further comments, merging.

PR #1864, opened 2017-11-05
Feature/common lexh

idodeclare:
Hello,

Please consider for integration this patch to define some "Common" lexh files and to update the build to copy lex and lexh files (language-specific and Common) to intermediate, language-specific areas for running jflex.

In this patch, I also demonstrate via the Ada test an insufficiency with the current URI matching rules that use {URIChar} (all languages); then define in Common.lexh a new {BrowseableURI} regex per RFC-3986; and then update the Ada analyzers to use the Common*.lexh files and the superior {BrowseableURI} regex.

Thank you.


tarzanek:
sounds good, so this way we can abstract most of common stuff to top level lexh
(and avoid so much duplication in all analyzers)

PR #1866, opened 2017-11-06
Add getYYEOF() to CommonXref.lexh to avoid reflection.

idodeclare:
Hello,

Please consider for integration this patch to include CommonXref.lexh in the language Xref lex files, with a definition of getYYEOF() and the removal of the reflection in the constructor of JFlexXref.

Thank you.


tarzanek:
thank you sir, merging

PR #1867, opened 2017-11-06
Include CommonTokenizer.lexh, and revise tokenizers as %int lexers

idodeclare:
Hello,

Please consider for integration this patch to include CommonTokenizer.lexh in the language Tokenizer lex files.

This patch also converts the tokenizers to `%type int` lexers to align them with the Xref lexers and to demonstrate the start of a decoupling of the Lucene Tokenizer logic (while still keeping it as a principal technology of course) from the JFlex-based lexers.

Also:
- Remove redundant `<<EOF>>` rules when `%eofval{` is also defined. The former would have won, so move those statements into the `%eofval{` block.
- Ensure all tokenizers set finalOffset in `%eofval{`. The value is not actually used, but retiring it is left for another patch.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/CommonTokenizer.lexh
+++ b/src/org/opensolaris/opengrok/analysis/CommonTokenizer.lexh
@@ -20,3 +20,11 @@
 /*
  * Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
  */
+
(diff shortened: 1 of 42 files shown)


tarzanek:
nice, this looks good, thank you again

PR #1868, opened 2017-11-07
change LdapFilter so that it actually works

### LINKED ISSUE #1863: LdapFilter plugin allows anyone with non-zero attributes
The `LdapFilter authorization plugin (as found in #1862) basically allows any user that has some attributes.

There are multiple problems there:
  - `LdapFilter#fillSession()` calls `lookupLdapContent(user, filter)` however the filter argument ends up being ignored in `AbstractLdapProvider.java`:
```
public Map<String, Set<String>> lookupLdapContent(User user, String filter) {
        return lookupLdapContent(user, null, null);
}
```
  - the actual lookup in `LdapFacade.java#lookupLdapContent()` is performed with  `user.getUsername()` as `dn` which cannot work - it has to be just the search base so the filter can be generic (e.g. `(&amp;(objectclass=posixGroup)(cn=foo_grp*)(memberUid=%uid%))`).
  - the check for the result done in `LdapFilter.java#fillSession()` considers strictly more than 0 attributes  to be success however it is more efficient to search just for a single matching object (e.g. "dn") which will still be positive however will not have any attributes. Without searching for concrete object a list of attributes will be returned which is sub-optimal (more CPU cycles burned on the LDAP server + extra network traffic; in the example above, a list of user IDs matching a sub-group will be returned).

vladak:
fixes #1863

### DIFF
--- a/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapFilter.java
+++ b/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapFilter.java
@@ -54,6 +54,7 @@ public class LdapFilter extends AbstractLdapPlugin {
         if ((ldapFilter = (String) parameters.get(FILTER_PARAM)) == null) {
             throw new NullPointerException("Missing param [" + FILTER_PARAM + "] in the setup");
         }
+        LOGGER.log(Level.FINE, "LdapFilter plugin loaded");
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


vladak:
Still haven't figured out how to configure `logging.properties` in Tomcat to actually display log messages from auth plugins with level higher than `INFO` however adding these `FINE*` messages anyway.

tarzanek:
lgtm, please merge

PR #1871, opened 2017-11-07
lucene 7.2.1 , bcel 6.2

tarzanek:
and first stab on jdk8/jdk9 support in jar/class analyser ...


tarzanek:
don't merge, there is still a prob with:
15:20:21 WARNING: Failed to add file /private/var/opengrok/src/OpenGrok/egrok/lib/json-simple-1.1.1.jar
java.lang.IllegalArgumentException: startOffset must be non-negative, and endOffset must be >= startOffset, and offsets must not go backwards startOffset=1,endOffset=10,lastStartOffset=333 for field 'defs'

src/org/opensolaris/opengrok/analysis/executables/JarAnalyzer.java
```diff
@@ -60,15 +60,15 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut) throws IOExc
                     xrefOut.append("</b>");
                 }
 
-                doc.add(new TextField("full", ename, Store.NO));
+//                doc.add(new TextField("full", ename, Store.NO)); we cannot do multiple adds from lucene 7, so later call will merge this
```
tarzanek: find a better way how to enable multiple .add calls
it might be a lucene bug, that you cannot call it twice ... (since before it worked)
alt. this might be some safeguard on proper stream implementation ?

tarzanek:
yes, reindexing from scratch will be needed + I will need to fix the 
" startOffset must be non-negative, and endOffset must be >= startOffset, and offsets must not go backwards  " first before I can merge ...

tarzanek:
this PR is also missing changes in webapp

tarzanek:
webapp should work now, so just the defs for jar are broken and it might be that my trick with injecting symbol tokenizer for defs no longer works as said in #1898 (which might be the root cause for broken defs? ... just speculating)

tarzanek:
ah, 7.2 not released yet ... anyways, the problems with the jars are not fixed yet .... so this is still WIP

tarzanek:
(will respin the build once 7.2 will be released, which will be soon)

platform/solaris/ips/create.sh
```diff
@@ -277,7 +277,7 @@ PKG pkgsend add file doc/ctags.config \
```
chocobrownie: hey

tarzanek: hello

idodeclare:
> anyways, the problems with the jars are not fixed yet

@tarzanek , FYI I have a branch, `feature/new_highlighter` (a728e4ab), that installs the modern Lucene [UnifiedHighlighter](https://www.slideshare.net/lucidworks/solr-highlighting-at-full-speed-presented-by-timothy-rodriguez-bloomberg-david-smiley-d-w-smiley-llc) to supplant `PlainFullTokenizer`-based context presentations in `search.jsp` and `more.jsp` (while preserving OpenGrok's look and feel).

To support `UnifiedHighlighter`, OpenGrok needs to index postings and offsets etc. — so I too had to fix out-of-order tokens and redundant fields for `JarAnalyzer`/`JavaClassAnalyzer` (e961f36ee4).

`feature/new_highlighter` is done, and I will be submitting a PR soon, after a number of intermediate PRs which I'll be raising in order.

tarzanek:
that sounds very nice, since this is the only thing that blocks this PR ...
I have also changes for JavaClassAnalyzer that will support java 8(9) classes properly (need to finish them)
so if your changes fix the lucene errors "startOffset must be non-negative, and endOffset must be >= startOffset, and offsets must not go backwards " when indexing with 7.2.1, then we will be good to go!
thank you @idodeclare !

tarzanek:
ok, so current trunk has already problems with java class / jar analyser
so I guess I can just merge this and we can try to fix the analyser as stopper
this change should get bcel to latest version, which should help with the problem

idodeclare:
OK, I will plan to raise a PR with cherry-picked 8630b28b to fix.

tarzanek:
fwiw we have a lucene compatibility tests which are disabled, so I will try to use them to avoid such situations

idodeclare:
Ahh sounds good

tarzanek:
after merging with latest trunk  and last fixes (thnx @idodeclare , your class/jar analyser fixes worked), I am OK to merge this, java class analyser needs improvement for properly supporting jdk9 classes, but this can be tracked by a new bug ...

tarzanek:
@idodeclare can you review pls?

src/org/opensolaris/opengrok/analysis/FileAnalyzer.java
```diff
@@ -211,34 +232,36 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut)
...
         throw new UnsupportedOperationException(
-            "Base FileAnalyzer cannot write xref");
+                "Base FileAnalyzer cannot write xref");
     }
-    
-    protected final JFlexTokenizer symbolTokenizer;
-    
+
+    // you analyzer HAS to override this to get proper symbols in results
+    protected JFlexTokenizer symbolTokenizer;
```
idodeclare: This may be an accidental regression from `final`.

src/org/opensolaris/opengrok/analysis/executables/JavaClassAnalyzer.java
```diff
@@ -88,11 +93,11 @@ protected JavaClassAnalyzer(FileAnalyzerFactory factory) {
     @Override
     public void analyze(Document doc, StreamSource src, Writer xrefOut) throws IOException {
         try (InputStream in = src.getStream()) {
-            analyze(doc, in, xrefOut);
+            analyze(doc, in, xrefOut,null);
         }
     }
 
-    void analyze(Document doc, InputStream in, Writer xrefOut) throws IOException {
+    void analyze(Document doc, InputStream in, Writer xrefOut, String extra) throws IOException {
```
idodeclare: `extra` is new but not used

src/org/opensolaris/opengrok/analysis/executables/JarAnalyzer.java
```diff
@@ -88,7 +88,7 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut) throws IOExc
                     }
                     JavaClassAnalyzer jca =
                             (JavaClassAnalyzer) fac.getAnalyzer();
-                    jca.analyze(doc, new BufferedInputStream(zis), xrefOut);
+                    jca.analyze(doc, new BufferedInputStream(zis), xrefOut,ename);
```
idodeclare: arg is new but not used

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -356,7 +349,12 @@ private void markProjectIndexed(Project project) throws IOException {
                 Message m = Message.createMessage("project");
                 m.addTag(project.getName());
                 m.setText("indexed");
-                m.write(env.getConfigHost(), env.getConfigPort());
+                try {
+                    m.write(env.getConfigHost(), env.getConfigPort());
+                } catch (ConnectException ce) {
```
idodeclare: I'm not quite clear why this exception is now being swallowed in a Lucene upgrade.

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -263,7 +264,12 @@ public static void main(String argv[]) {
                 Message m = Message.createMessage("config");
                 m.addTag("set");
                 m.setText("projectsEnabled = true");
+                try {
                 m.write(host, port);
+                } catch (ConnectException ce) {
```
idodeclare: I'm not quite clear why this exception is now being swallowed in a Lucene upgrade.

src/org/opensolaris/opengrok/analysis/executables/JavaClassAnalyzer.java
tarzanek: fixed

src/org/opensolaris/opengrok/analysis/executables/JarAnalyzer.java
tarzanek: fixed

src/org/opensolaris/opengrok/index/IndexDatabase.java
tarzanek: so this exception should be really a warning
since when implemented it forced a local webapp (or setup webapp) to be listening to update ... which really doesn't have to be a case, when you index from a different box and you don't want to update anyone, just create an index ... (e.g. parallel reindex from scratch)
I know that @vladak or @tulinkry might not like it, but I am just relaxing what they forced, since I don't think that an error of update should break or stop whole indexing ...

src/org/opensolaris/opengrok/index/Indexer.java
tarzanek: same as above, error connecting to webapp shouldn't block indexing

src/org/opensolaris/opengrok/analysis/FileAnalyzer.java
tarzanek: yes, sorry, reverting

tarzanek: (and good catch, thank you!)

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: OK I see. The PR looks good to me.

PR #1873, opened 2017-11-08
Include Common *.lexh, and use common macros

idodeclare:
Hello,

Please consider for integration this patch to include Common *.lexh files and to use shared definitions of paths, {WhiteSpace}, URIs, etc.

To minimize the build-time warnings about unused macros, I extracted from Common.lexh three files: CommonPath.lexh, CommonURI.lexh, and CommonLaxFPath.lexh.

The former versions of http,https,ftp rules (all identical and insufficient to match some URLs) in any Xref or Tokenizer *.lex are replaced with use of {BrowseableURI}.

Following are special notes about whitespace-related conversions:

* CSharpXref: do not include \r in a custom {WhiteSpace}; use Common.lexh version.

- ErlangXref: preserve existing definitions using renamed, specialized {ErlangWhspChar} and {ErlangWhiteSpace} that supplement Common.lexh's {WhspChar}.

- PhpSymbolTokenizer, PhpXref: do not exclude \f in a custom {WhiteSpace}; use Common.lexh version. For the Tokenizer, it's certain this will not affect results; for Xref, it should not degrade results.

- KotlinXref, SwiftXref: do not include \n in a custom {WhiteSpace}; use Common.lexh version. The former version would have had a bug in not calling startNewLine() when its version of {WhiteSpace} matched a LF.

- PowershellSymbolTokenizer, PowershellXref, ShXref, TroffXref, UuencodeXref, XMLXref: a private {WhiteSpace} was identical to Common.lexh's {WhspChar}, so translate to Common.lexh macros.

- PowershellXref, ShXref, XMLXref: do not output trailing tabs or spaces in lines.

Following are special notes about path handling:

- PlainXref, ShXref, TroffXref: use a lax version of {FPath} from CommonLaxFPath.lexh.

- PowershellXref: rename specialized macro as {AnyFPath} that only uses {FNameChar} to constrain path segments.

- XMLXref: rename specialized macro as {AlmostAnyFPath} that primarily uses {FNameChar} to constrain path segments, with a constraint that the last segment ending in a number or ASCII letter.

Thank you.


src/org/opensolaris/opengrok/analysis/CommonLaxFPath.lexh
```diff
@@ -0,0 +1,28 @@
...
+
+/*
+ * Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ */
+
+/*
+ * Relaxed from {FPath} in that only the last path segment is required
+ * to end in {ASCII_ALPHA} or {DIGIT}.
+ */
+LaxFPath = "/"? [a-zA-Z]{FNameChar}* ("/" [a-zA-Z]{FNameChar}*)+[a-zA-Z0-9]
```
tarzanek: LaxFPath is kinda confusing name, would it make sense to call it e.g.
LastSegmentFPath ?

src/org/opensolaris/opengrok/analysis/csharp/CSharpXref.lex
```diff
@@ -43,25 +43,17 @@ import org.opensolaris.opengrok.web.Util;
...
   @Override
   protected int getLineNumber() { return yyline; }
   @Override
   protected void setLineNumber(int x) { yyline = x; }
 %}
 
-/* Must match WHITE_SPACE constant */
-WhiteSpace     = [ \t\f\r]+
-EOL = \r|\n|\r\n|\u2028|\u2029|\u000B|\u000C|\u0085
+CsharpEOL = {EOL}|\u2028|\u2029|\u000B|\u000C|\u0085
```
tarzanek: we should finally fully support UTF, so I hope we will end up with common UTF eol detection regexp eventually

tarzanek:
respect!
so besides the self explanatory variable names I have no other comments & I'd be happy to merge, once you respond to my Q

src/org/opensolaris/opengrok/analysis/CommonLaxFPath.lexh
idodeclare: I'm not sure about "LastSegmentFPath". It's really the middle that has been relaxed from {FPath}; would "RelaxedMiddleFPath" do?

idodeclare:
I added tests C[xx]XrefTests with pre-BrowsebleURI output showing how previously a URL embedded at the end of a sentence would have excluded the sentence's '.', even though it is a valid URI character.

I restored this behavior with a special pushback in 2d10b85. I'm in contact with JFlex about the idea of support for negative lookbehind which could more gracefully effect this subtle behavior.

src/org/opensolaris/opengrok/analysis/CommonLaxFPath.lexh
tarzanek: yeah, better than Lax (which stands for Relaxed? )

tarzanek:
so if a sentence would end with url and "." it would be a bit of a problem to detect, no?

idodeclare:
@tarzanek , well the scanner can't know truly what is a sentence, but it can decide whether a trailing '.' belongs to a URL or to the rest of the document. I've restored the OpenGrok convention that the dot belongs to the rest of the document.

src/org/opensolaris/opengrok/analysis/CommonLaxFPath.lexh
idodeclare: Well "lax" is an adjective (Webster's:  "Not strict or stringent; not exact; loose; weak; vague; equivocal.") and "relax" is a verb ("to make lax") — but "relaxed" is an adj. too. I'll push that change.

tarzanek:
thank you sir!
merging

PR #1875, opened 2017-11-09
LDAP fixes

vladak:
This PR contains mainly these 2 changes:
  - `AuthorizationFramework` not to load plugins not contained in the configuration
    - the gist of this change is in `AuthorizationFramework#loadClasses()` where I basically removed the call to `addPlugin()`:
```
-            if ((pf = handleLoadClass(classname)) != null && !stack.setPlugin(pf)) {
-                // if there is not configured space -> append it to the stack
-                addPlugin(stack, pf);
+            // Load the class in memory and try to find a configured space for this class.
+            if ((pf = handleLoadClass(classname)) != null) {
+                stack.setPlugin(pf);
             }
```
  - `LdapUserPlugin` to use just the Common Name to perform the lookups
    - this avoids problems in internal deployment where Location in OSSO headers does not match the location in LDAP. Instead of using full DN as base for search, only CN is used with a filter. The filter now contains mandatory `objectclass` parameter so that only single object is returned from the search.

### DIFF
--- a/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
+++ b/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2016, 2017 Oracle and/or its affiliates. All rights reserved.
  */
 package opengrok.auth.plugin;
@@ -30,13 +30,18 @@ import opengrok.auth.plugin.entity.User;
 import org.opensolaris.opengrok.configuration.Group;
 import org.opensolaris.opengrok.configuration.Project;
 
+/**
+ * Authorization plug-in to extract user's LDAP attributes.
+ *
+ * @author Krystof Tulinger
+ */
 public class LdapUserPlugin extends AbstractLdapPlugin {
 
     public static final String SESSION_ATTR = "opengrok-ldap-plugin-user";
 
     /**
      * Check if the session exists and contains all necessary fields required by
-     * this plugin.
+     * this plug-in.
      *
      * @param req the HTTP request
      * @return true if it does; false otherwise
--- a/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
+++ b/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
@@ -24,6 +24,10 @@ package opengrok.auth.plugin;
 
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
 import opengrok.auth.entity.LdapUser;
 import opengrok.auth.plugin.entity.User;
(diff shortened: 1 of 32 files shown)


plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -17,26 +17,73 @@
...
+            char c = str.charAt(i);
+            if (!Character.isDigit(c) && !Character.isLetter(c)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+    
+    private Pattern usernameCnPattern = null;
```
tulinkry: can this be in the same block as other properities?

plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -17,26 +17,73 @@
...
+            throw new NullPointerException("Missing param [" + OBJECT_CLASS +
+                    "] in the setup");
+        }
 
+        if (!isAlphanumeric(objectClass)) {
+            throw new NullPointerException("object class '" + objectClass +
+                    "' contains non-alphanumeric characters");
+        }
+    
+        usernameCnPattern = Pattern.compile("(cn=[a-zA-Z_]+)");
```
tulinkry: why here and not in the property definition?

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -452,15 +452,15 @@ private IAuthorizationPlugin loadClass(String classname) throws ClassNotFoundExc
...
             if (classname.isEmpty()) {
                 continue;
             }
-            // load the class in memory and try to find a configured space for this class
-            if ((pf = handleLoadClass(classname)) != null && !stack.setPlugin(pf)) {
-                // if there is not configured space -> append it to the stack
-                addPlugin(stack, pf);
+            // Load the class in memory and try to find a configured space for this class.
+            if ((pf = handleLoadClass(classname)) != null) {
+                stack.setPlugin(pf);
```
tulinkry: maybe some log that this plugin is not configured in the stack?

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -473,10 +473,9 @@ private void loadClasses(AuthorizationStack stack, List<File> classfiles, List<F
                     if (!entry.getName().endsWith(".class") || classname.isEmpty()) {
                         continue;
                     }
-                    // load the class in memory and try to find a configured space for this class
-                    if ((pf = handleLoadClass(classname)) != null && !stack.setPlugin(pf)) {
-                        // if there is not configured space -> append it to the stack
-                        addPlugin(stack, pf);
+                    // Load the class in memory and try to find a configured space for this class.
+                    if ((pf = handleLoadClass(classname)) != null) {
+                        stack.setPlugin(pf);
```
tulinkry: maybe some log that this plugin is not configured in the stack?

plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
tulinkry: i mean that you instantly see what are the class properities

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
vladak: If I do this:

```
             // Load the class in memory and try to find a configured space for this class.
             if ((pf = handleLoadClass(classname)) != null) {
                 stack.setPlugin(pf);
+            } else {
+                LOGGER.log(Level.INFO, "plugin {0} is not configured in the stack",
+                        classname);
             }
         }
```

then this will also match classes like `opengrok.auth.plugin.ldap.LdapFacade$ContentAttributeMapper` etc.

Issue #1860, opened 2017-11-02
Ctags.java creates org.opensolaris.opengrok.util.Interner in loop


idodeclare:
Hello,

While refactoring Ctags.java to split out the line-parsing logic to a different class for testability, I noticed that the existing implementation is creating org.opensolaris.opengrok.util.Interner instances in a line-reading loop.

I believe creating in a loop renders the purpose of using Interner in Ctags moot since Interner is supposed to be used across a larger number of identical strings than just a line of data. The other user, SubversionHistoryParser, has the Interner as a class field.

Moreover, the documented purpose of org.opensolaris.opengrok.util.Interner is that it "works similar to java.lang.String#intern, but it stores the canonical objects on the heap instead of in the permgen space to address bug #15956." My research indicates, however, that String#intern does not behave that way anymore:

> In JDK 7, interned strings are no longer allocated in the permanent generation of the Java heap, but are instead allocated in the main part of the Java heap (known as the young and old generations), along with the other objects created by the application.

Might org.opensolaris.opengrok.util.Interner be retired to resolve the Ctags loop problem (as well as acknowleding its present redundancy)?

tarzanek:
Yeah, if it improved, then I am definitely for this ... there might be more such gotchas that are improved in JDK8/9 now ...

vladak:
fixed in 690f131

PR #1878, opened 2017-11-10
Retire Interner added in commits f4355bfc8, bfbf5b419

idodeclare:
Hello,

Please consider for integration this patch to retire the Interner class added in commits f4355bfc8, bfbf5b419.

Java 7 moved `intern()` out of permgen, so Interner is deprecated. Its use in Ctags[Reader] was buggy, since a new Interner instance was created per line, negating the value of the Interner.

This patch also revises all HistoryParser subclasses to `addFile()` with `intern()` paths — i.e. to use the same logic restored to SubversionHistoryParser in reverting f4355bfc8.

Thank you.


idodeclare:
I should remark that this resolves issue #1860.

tarzanek:
let's merge this, it also makes sense to compare memory usage for big repos ... @vladak make sure you note down how much did last index take before using this in new RC ... especially for (big) files with lots of symbols ...

PR #1880, opened 2017-11-13
Extract JFlexLexer, JFlexStateStacker, JFlexXrefSimple

idodeclare:
Hello,

Please consider for integration this patch with a bit of refactoring JFlexXref to extract some interfaces and to simplify it as a base class.

After analysis, only PhpXref was truly depending on the complicated, stacked "popStrings" API, so I moved that down into a PHP-specific class. The other languages were more straight-forward and needed only the intuitive case of disjoint (i.e., non-stacked) output-formatting spans (e.g., strings, numbers, comments, etc.) while still allowing their language lex states to be stacked.

E.g., Ruby xref only needs disjoint output-formatting spans (like every xref), but the language lex states themselves can be deeply stacked for interpolated strings.

- Refactor JFlexXref to have same, single-value stacked state data as JFlexTokenizer with language-specific, additional state downstream.
- Add JFlexXref `disjointSpan()` to manage non-stacked output spans (independent of and not clashing with Scopes).
- Extract PhpXrefSpanner for that language's particular "popStrings" method.
- Update C(xx)XrefTests data to accommodate escaping of &quot; and &apos;.

(Using `disjointSpan()` will also allow the CXref behavior of 056576c84 — "Make sure ... [to close] open tags if it reaches EOF in the middle of a string or a comment" — to apply to all languages that use the new method. Languages will have to be revised in future patches to use it. Right now I have this minor open-tag bug but also more importantly fixes for symbol-matching for Fortran, Erlang, Python, JSON, Clojure, Golang, CSharp, Ruby, Perl, and Ada fixes pending on this patch.)

Thank you.


tarzanek:
bear with me, I will try to review during this week

src/org/opensolaris/opengrok/web/HtmlConsts.java
```diff
@@ -33,4 +33,13 @@
     public static final String SPAN_N = "<span class=\"n\">";
     public static final String SPAN_S = "<span class=\"s\">";
```
tarzanek: will these be replaced by below ?

idodeclare: Yes, these public constants later could be mostly relocated and privatized.

(ZSPAN would survive here, and the others would continue to be useful — privately — in web/Util.)

tarzanek: ok, so I suppose this can happen next, I will merge this, let's move forward
thank you Chris!

idodeclare: My pleasure, @tarzanek !

I have a queue of pull requests for language tests and fixes that I'll start submitting now.

Issue #1876, opened 2017-11-10
project bubble(s) placed below search form


vladak:
Using the latest `master` I noticed this:

![bubble-bug](https://user-images.githubusercontent.com/2934284/32663755-54fd64ba-c62f-11e7-9d38-684252364320.png)

i.e. the "bubble" with the project is placed below the search form, along its side.

vladak:
@Orviss is there a chance this could be related to one of your commits since 1.1-rc16 ?

Orviss:
Possibly, I will take a look later today.

Orviss:
This is caused by this line 
```css
* {		clear: both }
```
in `mandoc.css` on line `38` which was added by @idodeclare

idodeclare:
Oops, that definition escaped my attention when I translated mandoc.css for import to OpenGrok. I'll push a fix.

vladak:
fixed in 2daafcd

PR #1882, opened 2017-11-15
Fix to escape '*' under div#mandoc descendant combinator

idodeclare:
Hello,

Please consider for integration this patch to fix the CSS problem identified in issue #1876. This patch also fixes to use only descendant combinators w.r.t. div#mandoc.

Thank you.


tarzanek:
thanks for the fix

PR #1885, opened 2017-11-20
Plugin fixes for internal deployment

vladak:
- Looks like the regexp used in `LdapUser` plugin is too strict so I relaxed it a bit.
- the fake OSSO header decoder could be initialized better + make the fake nature more prominent by using `fake-` prefix

### DIFF
--- a/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
+++ b/plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
@@ -47,7 +47,7 @@ public class LdapUserPlugin extends AbstractLdapPlugin {
     protected static final String OBJECT_CLASS = "objectclass";
     
     private String objectClass;
-    private Pattern usernameCnPattern = Pattern.compile("(cn=[a-zA-Z_]+)");
+    private final Pattern usernameCnPattern = Pattern.compile("(cn=[a-zA-Z0-9_-]+)");
 
     private boolean isAlphanumeric(String str) {
         for (int i = 0; i < str.length(); i++) {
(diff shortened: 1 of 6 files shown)


vladak:
Also, I bumped log level for error conditions in `LdapUser` plugin to `WARNING`. This concern is that this might result in message flurry however usually this plugin resides on the top of the authorization stack so should not have so big impact, @tulinkry ?

plugins/UserPlugin/src/opengrok/auth/plugin/decoders/FakeOSSOHeaderDecoder.java
```diff
@@ -23,20 +23,22 @@
...
  * This class should therefore only be used for debugging.
  *
  * @author Krystof Tulinger
  */
 public class FakeOSSOHeaderDecoder extends OSSOHeaderDecoder {
 
+    private final String PREFIX = "fake-";
+    
     public FakeOSSOHeaderDecoder() {
-        OSSO_COOKIE_TIMESTAMP_HEADER = "my-osso-cookie-timestamp";
```
tulinkry: This is not important now but maybe this should be called just SSO and not OSSO.

plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -47,7 +47,7 @@
     protected static final String OBJECT_CLASS = "objectclass";
     
     private String objectClass;
-    private Pattern usernameCnPattern = Pattern.compile("(cn=[a-zA-Z_]+)");
+    private final Pattern usernameCnPattern = Pattern.compile("(cn=[a-zA-Z0-9_-]+)");
 
     private boolean isAlphanumeric(String str) {
```
tulinkry: I know that this change is not about this function but it should really be in StringsUtils and I think we even have this class in opengrok already

tulinkry:
Also this is done when the session is created/renewed so this happens only once in a while for a particular user

tulinkry:
I would move the isAlpha... to the utils class but you can do it in a different PR.

I'm ok with the rest

plugins/UserPlugin/src/opengrok/auth/plugin/decoders/FakeOSSOHeaderDecoder.java
vladak: I think these are specific to Oracle SSO so I'd leave them as they are.

plugins/LdapPlugin/src/opengrok/auth/plugin/LdapUserPlugin.java
vladak: sure

PR #1883, opened 2017-11-18
Bugfix #1830 #1877

emergie:
Two commits to make clearer what changed:
1. revert of revert of #1830 bugfix
2. rework for #1877

Resolves #1830.
Resolves #1877.

I accept OCA for this fix.

PR #1887, opened 2017-11-21
bugfix: #1810 Keyboard shortcuts are active when typing into search b…

### LINKED ISSUE #1810: Keyboard shortcuts are active when typing into search box
When typing into the search text box on a file-view page, keyboard shortcuts are still active. So, for example, typing "b1" into the box also will move the page to the bottom ('b') and bring up the Intelligence window ('1').

emergie:
>bugfix: #1810 Keyboard shortcuts are active when typing into search box + type=hidden for input created in searchSubmit as it sometimes is visible in new Chromium before actual form submission.

Because of global (document level) keypress hooks defined in the _Diff jumper_ it is impossible to write `n` or `b` in the _Search_ input on any /diff/ page.
Both `prevHandler` and `nextHandler` call Event#preventDefault which disables keyboard event propagation.

A bit simillar problem occurs in _Intelligence window_ hooks for `1`-`5`, `n` and `b` keys.
There is no #preventDefault call, so it is possible to write those characters, but pressing keys while typing in the Search input triggers both _Intelligence window_ hook and text field modification. This is not a proper UI behaviour:
peek 2017-11-18 20-16

Intentionally I have not updated utils-0.0.18.js version number as it would introduce conflict with PR https://github.com/OpenGrok/OpenGrok/pull/1883.

Resolves #1810.

I accept OCA for this fix.

### DIFF
--- a/web/js/diff-0.0.2.js
+++ b/web/js/diff-0.0.2.js
@@ -17,6 +17,8 @@
  * CDDL HEADER END
  */
 
+/* global textInputHasFocus */
+
 /*
  * Copyright (c) 2016, 2017 Oracle and/or its affiliates. All rights reserved.
  */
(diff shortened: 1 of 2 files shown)


tarzanek:
thank you, lgtm, merging

PR #1889, opened 2017-11-22
override RepositoryInfo.equals() so that removeAll() works for Projec…

### LINKED ISSUE #1886: project "delete" message sometimes fails to delete repositories
Sometimes project "delete" message does not delete repositories, leading to orphaned `RepositoryInfo` objects to remain in the configuration.

vladak:
…t Message

fixes #1886

### DIFF
--- a/src/org/opensolaris/opengrok/history/RepositoryInfo.java
+++ b/src/org/opensolaris/opengrok/history/RepositoryInfo.java
@@ -24,6 +24,7 @@ package org.opensolaris.opengrok.history;
 
 import java.io.Serializable;
 import java.text.SimpleDateFormat;
+import java.util.Objects;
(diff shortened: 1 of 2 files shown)


tarzanek:
usually hashcode is overriden with equals too ...
would it make sense to override it too?

vladak:
Yes, this is generally recommended (https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java) and should be safe in this case since two `RepositoryInfo` objects that have equal `directoryNameRelative` should also have the rest of the private members equal.

Used Netbeans to generate the `hashCode()`.

tarzanek:
thank you, merging!

PR #1891, opened 2017-11-23
Run plugins test

vladak:
- maven only
- still not possible to execute the `test` (or `compile`) goals just for `plugins` project


tarzanek:
thank you,
lgtm, feel free to merge(if you don't want to add anything)

PR #1895, opened 2017-11-26
Feature/c cxx symbol tests

idodeclare:
Hello,

Please consider for integration this patch to add symbol tests for C/C++ and to fix the exposed bugs.

Thank you.


src/org/opensolaris/opengrok/analysis/c/CXref.lex
```diff
@@ -80,7 +78,12 @@ Number = (0[xX][0-9a-fA-F]+|[0-9]+\.[0-9]+|[1-9][0-9]*)(([eE][+-]?[0-9]+)?[ufdlU
 
 {Identifier} {
     String id = yytext();
-    writeSymbol(id, Consts.kwd, yyline);
+    // For historical reasons, CXref does not link Identifiers of length=1
```
tarzanek: the reason was that lucene wasn't that performant back then (and it made no sense to index 1 char identifiers), we can lift this now, since lucene should be able to handle 1 char identifiers now ...

src/org/opensolaris/opengrok/analysis/c/Cxx.lexh
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/*
+ * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ */
+
+Identifier = [a-zA-Z_] [a-zA-Z0-9_]*
+
+Number = (0[xX][0-9a-fA-F]+|[0-9]+\.[0-9]+|[1-9][0-9]*)(([eE][+-]?[0-9]+)?[ufdlUFDL]*)?
```
tarzanek: Isn't this a dup ? 
shouldn't we have just one include for both ?
btw. Identifier and Number definitions are most probably shared between lots of languages in exactly the same form as above ...

tarzanek: but I don't have a list of languages where this actually really is like that, so for some it might make sense to share identifier and number definition, for others they will need improvement

idodeclare: I would argue the standards evolve at their own pace. E.g., C++14 adds support for binary literals — 0b11111100 — but that's not yet handled by this {Number} regex and not necessarily supported either by C.

tarzanek:
Well this talk about lengths of identifiers got me to a bug (I didn't have time to root cause it)
C Symbol Tokenizer does this:
Identifier = [a-zA-Z_] [a-zA-Z0-9_]*
{Identifier} {String id = yytext();
                if(!Consts.kwd.contains(id)) {
                        setAttribs(id, yychar, yychar + yylength());
                        return yystate(); }
so basically per above we save 1 char identifiers, which is OK,
a test could be a file with e.g. "c" , so if you index OpenGrok repo
BUT you will see that c def https://github.com/OpenGrok/OpenGrok/blob/master/testdata/sources/c/main.c#L7
won't be found in definitions, although full search will mark it as definition
but let's deal with it in a new bug

tarzanek:
thank you, will merge

PR #1901, opened 2017-11-27
Feature/sh tests

idodeclare:
Hello,

Please consider for integration this patch to add SH tests and to fix the exposed bugs.

Thank you.


src/org/opensolaris/opengrok/analysis/sh/ShSymbolTokenizer.lex
```diff
@@ -38,48 +37,65 @@ super(in);
...
                 if(!Consts.shkwd.contains(id)){
                         setAttribs(id, yychar, yychar + yylength());
                         return yystate(); }
               }
+ {Number}    {}
  \"     { yybegin(STRING); }
  \'     { yybegin(QSTRING); }
  "#"    { yybegin(SCOMMENT); }
+
+ {Unary_op_req_lookahead} / \W    {
```
tarzanek: if it's a noop why is it here?

tarzanek: I think @kahatlen experimented a bit with some lookaheads in jflex, but not sure how good they worked ...

test/org/opensolaris/opengrok/analysis/sh/sample.sh
```diff
@@ -0,0 +1,270 @@
...
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+#
+# Copyright 2009 Sun Microsystems, Inc.  All rights reserved.
```
tarzanek: :)

tarzanek:
lgtm,
I just don't get those empty lookaheads ... what is the intent?

src/org/opensolaris/opengrok/analysis/sh/ShSymbolTokenizer.lex
idodeclare: The handler is a noop, but the match is not. E.g. this match consumes characters so they aren't later processed by the {Identifier} rule; I wanted to prevent unary ops from producing symbols.

I haven't seen any problems with lookaheads. They were already used in other lexers — e.g. here in ShXref `"${" / {WhspChar} | {EOL}`, and used in Scala, Perl, Ada, Kotlin and Swift.

idodeclare: It aligns with the same (new) rule in ShXref.

tarzanek: why would they produce symbols?
(when they don't match the identifier rule ? )
can you give an example of what text you're trying to get avoided to be matched as identifier?

tarzanek: ok, I typed the Q too fast ... I see now
so e.g.
-a -f -x 
but ... what about positive case?
isn't there any?

idodeclare: I referenced the Bash getopts documentation, but you're right that ksh supports '+' as an option prefix. I can update Unary_op_req_lookahead to support that syntax too.

tarzanek:
:-D I meant matches that break this concept (but then I cannot think of any)
so a case where you'd want something prefixed with "-" (or "+") to be detected as identifier
but since I cannot think of such use case/example I think we just let it be and merge this ... or if you can make up an example, then share it

idodeclare:
Nope I can’t conceive of such an example.

tarzanek:
let's merge
and thank you for all the inputs

PR #1904, opened 2017-11-30
Feature/python tests

idodeclare:
Hello,

Please consider for integration this patch to add Python tests and to fix the exposed bugs.

Thank you.


tarzanek:
lgtm, merging
thank you Chris!

PR #1905, opened 2017-12-01
Feature/vb tests

idodeclare:
Hello,

Please consider for integration this patch to add VB tests and to fix the exposed bugs.

Thank you.


src/org/opensolaris/opengrok/analysis/vb/Consts.java
```diff
@@ -19,54 +19,193 @@
...
-                line = line.trim();
-                lline = line.toLowerCase(Locale.US);
-                if (line.charAt(0) != '#') {
-                    set.add(line);
-                    set.add(lline);
-                }
-            }
-        }
-    }
+    private static void populateKeywordSet(Set<String> kwd) {
```
tarzanek: we're loosing precision here ... so this is not a showstopper, but I am just thinking aloud ...
I think we will want to have case sensitivity soon, new lucene has no problems with saving the index in case sensitive format and query it insensitively now ... (with 7.x this will be even more obvious)
now we will loose some of the case sensitivity info ...

tarzanek:
let's merge, we can deal with case sensitivity later

src/org/opensolaris/opengrok/analysis/vb/Consts.java
idodeclare: This patch does not change the values that get stored as symbols — see the unchanged `setAttribs()` call. It changes the lexer to be strictly case-insensitive to accord with the language and changes the `Consts` to reflect this. VB symbols that were actually keywords but not previously found in the case-sensitive set are now properly identified as keywords.

(But yes, if the actual search needs to be case-insensitive, then I've seen the existing ideas on this from the historical Issues list.)

tarzanek: ok, so my concerns are not valid, this is good, so if VB spec expects lowercase, then this fix is perfect, thank you

Issue #394, opened 2013-05-22
convert user-specified keywords in src code  into custom URLs (Bugzilla #1058)


vladak:
status NEW severity _enhancement_ in component _analyzer_ for _---_
Reported in version _unspecified_ on platform _ANY/Generic_
Assigned to: Lubos Kosco

On 2008-04-09 07:35:49 +0000, Hemang Lavana wrote:

> Currently the history page allows us to specify custom URLs for author and bugid. Similarly, provide a mechanism for the admin to specify custom URLs for specific keywords or patterns found in source code.
> 
> For example, the bugid may be listed under code comments in the source code and I would like to see it as clickable link on xref pages.
> 
> Hemang.

On 2010-03-24 14:04:16 +0000, Lubos Kosco wrote:

> **\* Bug 12856 has been marked as a duplicate of this bug. ***

On 2010-03-24 14:07:15 +0000, Lubos Kosco wrote:

> we just spoke with Vladimir,
> that this could be a nice idea,
> we'd get rid of bug and arc patterns and urls
> and just have an array of them pattern1...X and respective url1..X
> 
> another thing is to have proper matcher which would replace the matches in url
> according to groups/back reference
> 
> e.g. the pattern would have more groups, which would then be filled into the
> url
> 
> this way we'd get a universal solution for this (and duplicate) bug ...
> 
> now only just to get it done ;)

On 2010-03-24 14:12:33 +0000, Vladimir Kotal wrote:

> The idea is to be able to do things like this in read-only configuration:
> 
>   <void property="pattern"> 
>    <string>\b([12456789][0-9]{6})\b<string>
>    <string>http://buglookup/showbug.cgi?cr=\1</string> 
>   </void>
>   <void property="pattern"> 
>    <string>({FW,PS}ARC)[\b/](\d{4}/\d{3})\b</string>
>    <string>http://cases/\1/\2/</string> 
>   </void>
> 
> Or something like that (the XML structure should probably be better).
> 
> The second regexp is probably wrong but you get the idea.

PR #1907, opened 2017-12-01
Feature/reuse htmlize

idodeclare:
Hello,

Please consider for integration this patch to condense the disparate htmlize() implementations.

- Add Util.prehtmlize() for the mode which an Xref might need.
- Otherwise, update the Util method to also quote &amp;quot; and &amp;apos;, and update users and tests. Util.diffline() is updated to htmlize() all output -- not just when different.
- Update search/Results to htmlize() where comments had questioned "htmlize ???" — that does seem the correct handling there.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/JFlexXref.java
+++ b/src/org/opensolaris/opengrok/analysis/JFlexXref.java
@@ -575,9 +575,7 @@ public abstract class JFlexXref extends JFlexStateStacker {
      * @return String with escaped html characters
      */
     protected String htmlize(String raw) {
-        return raw.replace("&", "&amp;").replace("<", "&lt;")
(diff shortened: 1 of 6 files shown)


tarzanek:
looks good, I am not sure if quotes are in need of escaping, but let's go with it
I have a small concern of slowing down the code with the if (pre) block, but then I hope jvm will optimize this ;)

idodeclare:
>   I am not sure if quotes are in need of escaping

@tarzanek, no it's not strictly required for HTML presentation, but I have some ideas about supporting post-processing an XREF for features such as issue #394 ("convert user-specified keywords in src code into custom URLs"), and being able to easily distinguish the underlying content from the HTML tagging — i.e., having well-formed XHTML for an XREF — would be valuable.

PR #1911, opened 2017-12-04
Bugfix/symlinks2

idodeclare:
Hello,

Please consider for integration this patch with fixes for symbolic link handling.

The principle addition is `PathUtils.getRelativeToCanonical(...)` and a new explicit `ForbiddenSymlinkException`.

I know PR #1846 is in progress, but my testing showed it still had problems with history processing and also could not support nested symbolic links. I first cherry-picked most of #1846's contents but then revised its versions of `RuntimeEnvironment.getPathRelativeToSourceRoot(...)` and `HistoryGuru.getRepository(...)` to instead use the new `PathUtils`.

I also kept #1846's new tests and added some more.

This patch also contains a new method, `Repository.getRepoRelativePath(...)`, to replace (pre-existing) problematic logic in the various repository implementations (e.g., Git, Mercurial, SVN, etc.) that was lopping off the front of paths assuming they started with "source root". E.g., on macOS, /var is a symlink to /private/var, and the logic was chopping off the number of characters in "/private/var/..." from "/var/...", resulting in paths mangled in the middle of segments.

(I detected the problem while testing scenarios of symbolic-linked project layouts, but unit tests otherwise did not detect this bug.)

TESTING SCENARIOS:

1) Basic case with no symlinks under SRC_ROOT — this behaves as before.
2) Direct symlinks under SRC_ROOT. Here of course `Indexer` "automatically allow symlinks that are directly in source root". `HistoryGuru` and history classes, though, are now fixed to properly work in this case.
3) Add another symlink inside a directory referenced by a symlink under SRC_ROOT. `Indexer` will not have allowed this second symlink, so the new logic properly detects this as disallowed. This repo will not be indexed and will not show up in the UI.
4) Update 3) to add an OPENGROK_READ_XML_CONFIGURATION file with the secondary symlink added to allowedSymlinks. `Indexer` will now allow this repo; it is handled by `HistoryGuru` et al; and the repo will show up in the UI.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -105,6 +105,7 @@ import org.opensolaris.opengrok.history.HistoryGuru;
 import org.opensolaris.opengrok.history.HistoryReader;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.search.QueryBuilder;
+import org.opensolaris.opengrok.util.ForbiddenSymlinkException;
 import org.opensolaris.opengrok.util.IOUtils;
 import org.opensolaris.opengrok.web.Util;
 
@@ -383,10 +384,12 @@ public class AnalyzerGuru {
      * @param fa The analyzer to use on the file
      * @param xrefOut Where to write the xref (possibly {@code null})
      * @throws IOException If an exception occurs while collecting the data
+     * @throws ForbiddenSymlinkException if symbolic-link checking encounters
+     * an ineligible link
      */
     public void populateDocument(Document doc, File file, String path,
             FileAnalyzer fa, Writer xrefOut)
-            throws IOException {
+            throws IOException, ForbiddenSymlinkException {
 
         String date = DateTools.timeToString(file.lastModified(),
                 DateTools.Resolution.MILLISECOND);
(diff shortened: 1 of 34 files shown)


idodeclare:
Closing due to conflicted file — will reopen after a rebase and after a revision to try-remove directories in serial, depth first (after files deleted still in parallel).

tarzanek:
I thought github can rebase within the same PR ...

idodeclare:
> I thought github can rebase within the same PR ...

Maintainers do see, if configured, "rebase and merge" or "squash and merge". Is that what you mean?

I perceive that @vladak uses "squash and merge" (and it saddens me because my carefully crafted commits are squashed). I haven't perceived any OpenGrok maintainer use "rebase and merge." —
 But Github can't resolve conflicts automatically and will advise that it cannot proceed "due to conflicts."

vladak:
I don't use squash universally, actually I tend to use rebase for most of
PRs, at least I think so.

Dne 15. 12. 2017 3:24 odp. napsal uživatel "C Fraire" <
notifications@github.com>:

idodeclare:
> I don't use squash universally, actually I tend to use rebase for most of
PRs, at least I think so.

That's good to hear, but...

E.g., PR #1844 — I added a failing test and then in a separate commit fixed it — but that got squashed.
E.g., PR #1842 — I re-saved web.xml in NetBeans in one commit so the cache-control changes I added in the next commit would be very clear — but that got squashed.

Those are the only two PRs of mine you approved and merged.

vladak:
I see. It was just coincidence.

Dne 15. 12. 2017 4:10 odp. napsal uživatel "C Fraire" <
notifications@github.com>:

PR #1912, opened 2017-12-04
Add Ruby magics and path-less magic for Escript, Perl, PHP, Python

idodeclare:
Hello,

Please consider for integration this patch to add Ruby magics and path-less magic for Escript, Perl, PHP, and Python, which is seen in some unit test files for those languages. Also the patch removes empty MAGICS for e.g. CSharp, Haskell, etc.

Thank you.


vladak:
thanks !

PR #1913, opened 2017-12-04
Feature/golang tests

idodeclare:
Hello,

Please consider for integration this patch to add Go language tests and to fix the exposed bugs.

This also updates Go Consts to include "Predeclare identifiers" and the "Blank identifier".

Thank you.


tarzanek:
lgtm, merging

PR #1914, opened 2017-12-04
Feature/clojure tests

idodeclare:
Hello,

Please consider for integration this patch to add Clojure tests and to fix the exposed bugs.

This patch also revises the Ctags class to use — if Universal ctags is available — the built-in Clojure support for namespace and function tagging in order to avoid redundant tags due to OpenGrok regexes.

Thank you.


LICENSE-eclipse.txt
```diff
@@ -0,0 +1,195 @@
+Clojure is licensed by the Eclipse Public License, below.
```
tarzanek: we should definitely create a directory licenses to stop "polluting" root dir ...

tarzanek:
thank you for this, merging

PR #1915, opened 2017-12-05
Feature/erlang tests

idodeclare:
Hello,

Please consider for integration this patch to add Erlang tests and to fix the exposed bugs. The patch also adds to Erlang Consts for "Module Attributes".

Thank you.


tarzanek:
nice, thank you

PR #1916, opened 2017-12-05
Feature/rust tests

idodeclare:
Hello,

Please consider for integration this patch to add Rust tests and to fix the exposed bugs. The patch also enhances the lexers to recognize the full Rust syntax for numeric literals; raw-, byte- and raw-byte strings; nested comments; and the language's uses of apostrophes for non-quoting reasons.

This patch also revises the Ctags class to use — if Universal ctags is available — the built-in Rust tagging support in order to avoid redundant tags due to OpenGrok regexes. (OpenGrok Rust regexes for tagging beyond what Universal Ctags supports are preserved.)

Thank you.


tarzanek:
thank you sir! merging

PR #1917, opened 2017-12-05
Introduce OPENGROK_DATA_ROOT for better customization

tuxillo:

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -37,6 +37,7 @@
 #   - OPENGROK_DISTRIBUTION_BASE  Base Directory of the OpenGrok Distribution
 #   - OPENGROK_INSTANCE_BASE      Base Directory of the OpenGrok User Data Area
 #   - OPENGROK_SRC_ROOT		  directory with files to be indexed
+#   - OPENGROK_DATA_ROOT	  directory where index and cache files are generated
 #   - OPENGROK_CTAGS              Full path to Exuberant or Universal CTags binary
 #   - OPENGROK_CTAGS_OPTIONS_FILE Full path to file with extra command line
 #                                 options for CTags program (for its --options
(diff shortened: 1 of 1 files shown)


tarzanek:
hmm, how much testing went into this?
I am a bit concerned that we hardcode the assumption that data root is under instance base in java code as well ... 
can you please verify no java code changes are needed?

tuxillo:
Hi,

I've been using it since the beginning because it's needed for my scripts to use the 'OpenGrok' script instead of calling the jar with parameters myself.

About the Java changes needed I am not sure where to check. I have grep'ed the source tree and DATA_ROOT mostly shows up in these Java files, in comments all of them:

    ./src/org/opensolaris/opengrok/web/SearchHelper.java:     * is assumed (i.e. DATA_ROOT/index will be used instead of possible
    ./src/org/opensolaris/opengrok/web/SearchHelper.java:     * multiple DATA_ROOT/$project/index). If the set contains projects
    ./src/org/opensolaris/opengrok/configuration/Configuration.java:     * The name of the eftar file relative to the <var>DATA_ROOT</var>, which
    ./src/org/opensolaris/opengrok/configuration/Configuration.java:     * The name of the file relative to the <var>DATA_ROOT</var>, which should
    ./src/org/opensolaris/opengrok/configuration/Configuration.java:     * The name of the file relative to the <var>DATA_ROOT</var>, which should
    ./src/org/opensolaris/opengrok/configuration/Configuration.java:     * The name of the file relative to the <var>DATA_ROOT</var>, which should
    ./src/org/opensolaris/opengrok/configuration/Configuration.java:     * The name of the file relative to the <var>DATA_ROOT</var>, which should

But the logic I used here was: If the OpenGrok script is building a DATA_ROOT itself from  the base dir, why not allow modifying it from the environment? All the script does, as far as I can tell, is to pass -d ${DATA_ROOT} to the jar.

Regards,
Antonio Huete

tarzanek:
thank you, your OCA is on file, so I am just going to merge this

PR #1920, opened 2017-12-05
Feature/lua tests

idodeclare:
Hello,

Please consider for integration this patch to add Lua tests and to fix the exposed bugs. The patch also adds support for Lua long-bracket quoting syntax.

Thank you.


tarzanek:
cool, moar of teh syntactic goodness :-D , merging

PR #1921, opened 2017-12-05
Feature/joint lexer tests

idodeclare:
Hello,

Please consider for integration this patch to extract `JFlexJointLexer` for use by the Ada, Perl, and Ruby analyzers to eliminate some redundancy there. Also, those lexers are revised to use the `disjointSpan()` support introduced recently in `JFlexXref` instead of writing `<span>` elements directly.

The patch also fixes too-greedy gobbling of URLs in some cases for Perl and Ruby (using the "collateral capture" handling introduced recently and used in most recent languages' "tests" PRs).

Thank you.


tarzanek:
lgtm, merging

PR #1924, opened 2017-12-06
Avoid two SEVERE log messages for normal startup,stop events

idodeclare:
Hello,

Please consider for integration this patch to avoid two spurious SEVERE log messages on startup when data/ is empty, or upon stop.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/messages/StatsMessage.java
+++ b/src/org/opensolaris/opengrok/configuration/messages/StatsMessage.java
@@ -17,11 +17,13 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.configuration.messages;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


tarzanek:
will you fix the tests too? @tulinkry any inputs?

idodeclare:
> will you fix the tests too?

Yep, sorry about that.

I did not find any usage for `StatsMessage` besides `StatsMessageTest`. I revised the former's API slightly so that it only resets the configuration's `statisticsFilePath` if a reload  for a message-indicated file succeeds.

tarzanek:
@tulinkry can you have a look and merge?
to me it looks OK, but then I just did a very quick look (and I lack history with stats code unlike you)

src/org/opensolaris/opengrok/configuration/messages/StatsMessage.java
```diff
@@ -39,11 +41,38 @@
...
+
     @Override
     protected byte[] applyMessage(RuntimeEnvironment env) throws IOException, ParseException {
+        String statsFilePath = getStatisticsFilePath();
+
         if (getText().equalsIgnoreCase("reload")) {
-            env.loadStatistics();
+            if (statsFilePath != null) {
+                env.loadStatistics(new File(statsFilePath));
+                env.getConfiguration().setStatisticsFilePath(statsFilePath);
```
vladak: Having the `RuntimeEnvironment` singleton is bad by itself when objects query the global state. This actually makes it worse by setting global state values from an object. Setting configuration value is expected in the `config` message however here I'd say it's completely unexpected so it violates the concept of least surprise.

idodeclare:
OK I'll withdraw.

vladak:
There's nothing wrong with fixing the exceptions, just there has to be a
better way.

Dne 7. 12. 2017 15:37 napsal uživatel "C Fraire" <notifications@github.com>:

idodeclare:
I want no exception at startup, and that test from the author of `StatsMessage` does want one. So I'll ignore it.

PR #1926, opened 2017-12-06
Bugfix/writer commit

idodeclare:
Hello,

Please consider for integration this patch with a fix not to swallow exceptions for `IndexWriter` `commit()` or `close()`.

Also, this fixes the exit code of the OpenGrok script and also returns non-zero for bad options (i.e., for `ParseException`).

Thank you.


tarzanek:
I assume this is work in progress ... or it's done?
(e.g. the locking var in OpenGrok is not used, is it?)

idodeclare:
This is a complete proposed patch. There is a `—lock` option already supported by `Indexer` and I confirmed that the new script env var does result in a different lock factory being used.

Only “on” and “off” are supported (to toggle `SimpleFSLockFactory`. I wonder if I should revise the switch to “off”, “simple”, “native” to support also the superior (for non-NFS DATA_ROOT) `NativeFSLockFactory`?

tarzanek:
fwiw - more history - I originally made changes to make this lockless when lucene was ready to support it ... since we had huge problems with locking ...

idodeclare:
Oh can you expand on that? I do my own version of SimpleFSLockFactory — i.e. a private file whose existence prevents another instance of my wrapper script from running. Since the —lock switch already existed I hoped to revise to use that.

NativeFSLockFactory interests me more so that I won’t have to worry about orphan lock files, as I do and would do with my own lock strategy or with SimpleFSLockFactory.

idodeclare:
Oops I bungled this with some other commits accidentally pushed here. Closing.

PR #1931, opened 2017-12-06
Bugfix/ruby end interp

idodeclare:
Hello,

Please consider for integration this patch to fix an arcane Ruby problem where colon-quote — i.e., :&quot; — following the end-token of a Ruby string interpolation was being misinterpreted together with the end-token as the start of an interpolated symbolic quoting operator (i.e., as an erroneous {Symquo} match) and interfering with the correct interpretation of the end-token.

(This would have applied to other quoting operator tokens following the heuristic {WxSigils} match and is fixed for those too.)

Thank you.


tarzanek:
thank you, let's merge

PR #1934, opened 2017-12-07
Feature/js tests

idodeclare:
Hello,

Please consider for integration this patch to add JavaScript tests and to fix the exposed bugs.

Thank you.


idodeclare:
My pleasure, @tarzanek.

PR #1939, opened 2017-12-10
Feature/haskell tests

idodeclare:
Hello,

Please consider for integration this patch to add Haskell tests and to fix the exposed bugs.

This also sets `preserveLastModified="true"` for the regex fixups of the special `run-jflex-and-disable-buffer-expansion` target so those files aren't continually regenerated on every build (they'll still regenerate when .lex or .lexh changes).

Thank you.


build.xml
```diff
@@ -250,20 +250,24 @@ Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
...
                                match="(zzFinalHighSurrogate = 1;)(\r?\n)"
-                               replace="\1\2          if (numRead == 1) { return true; }\2"/>               
+                               replace="\1\2          if (numRead == 1) { return true; }\2"
+                               preserveLastModified="true"/>               
                 <!-- also revert 0 character check that got in with 1.6.1 : https://github.com/jflex-de/jflex/blob/master/jflex/examples/zero-reader/README.md -->
                 <replaceregexp file="${src.generatedsrc.dir}/@{dir}/@{name}.java"
                                match="[ \t]*/\* not supposed to occur according to specification of java\.io\.Reader \*/\s+if \(numRead == 0.*?\}[ \t]*\r?\n"
-                               replace="" flags="s" />
+                               replace="" flags="s"
+                               preserveLastModified="true"/>
```
tarzanek: good catch, we need to do this patching due to behaviour of lucene on token limits ... (and to avoid memory leaks)

tarzanek:
thank you and good catch on the regexp replacer, forgot to mention it when you were doing the include of lex files, merging now

PR #1940, opened 2017-12-11
Feature/kotlin tests

idodeclare:
Hello,

Please consider for integration this patch to add Kotlin tests and to fix the exposed bugs.

Thank you.

PR #1941, opened 2017-12-11
Feature/kotlin tests

idodeclare:
Hello,

Please consider for integration this patch to add Kotlin tests and to fix the exposed bugs.

Thank you.

(PR #1940 was the first version; this new version of the patch is rebased on master.)


tarzanek:
thank you, merging

PR #1943, opened 2017-12-12
Feature/tcl tests

idodeclare:
Hello,

Please consider for integration this patch to add Tcl tests, to fix the exposed bugs, and to support Tcl substitutions syntax.

Thank you.


tarzanek:
thank you very much, this is good stuff, merging, have no comment :)

PR #1950, opened 2017-12-14
add auth options and --non-interactive to getHistoryGet()

vladak:
I don't have a setup to test this on, however it conforms to the `svn cat` usage.


tarzanek:
if getAuthCommandLineParams will be empty, will it work ?

vladak:
If the env vars are not present it will return empty array so `addAll()` will do the right thing. After all, this is how it is used already in `getHistoryLogExecutor()`.

PR #1949, opened 2017-12-13
Feature/sql tests

idodeclare:
Hello,

Please consider for integration this patch to add [PL]SQL tests and to fix the exposed bugs.

Thank you.

PR #1951, opened 2017-12-14
Feature/sql tests

idodeclare:
Hello,

Please consider for integration this patch to add [PL]SQL tests and to fix the exposed bugs.

(This is the second try of #1949 after rebasing on `master`).

Thank you.


tarzanek:
lgtm, merging, thank you

PR #1955, opened 2017-12-18
Add Eiffel analysers

idodeclare:
Hello,

Please consider for integration this patch to support Eiffel analysis.

Thank you.


tarzanek:
very nice, thank you Chris!
we're definitely way over 2-3 features per release ... so we should definitely release OpenGrok asap!

PR #1956, opened 2017-12-18
HistoryGuru needs to be constructed only after thread-local config is…

### LINKED ISSUE #1942: OPENGROK_SCAN_DEPTH not userful, long path no history
echo command, 
MinimalInvocation    -r on -H  -D off  -z 6  -C -c /usr/bin/ctags-exuberant -a on  -U localhost:2424    -W /var/data/opengrok/660/etc/configuration.xml -S -P -s /var/data/660 -d /var/data/opengrok/660/data -w 660

OpenGrok  1.1-rc16

vladak:
… set in place

fixes #1942


tarzanek:
merging, thank you

PR #1958, opened 2017-12-19
Feature/symbol matcher

idodeclare:
Hello,

Please consider for integration this refactoring that continues to bring the language xrefers and symbol tokenizers closer together by revising the language tokenizers not to inherit from `JFlexTokenizer` (nor from Lucene `Tokenizer`) but instead to subclass the same parent as `JFlexXref`: i.e. `JFlexStateStacker`.

`JFlexTokenizer` is revised and simplified as a non-abstract class that now uses each of the language-specific tokenizers, listening to their publishing of symbols.

As a consequence of this refactoring, `CommonTokenizer.lexh` and `CommonXref.lexh` are merged, and the language xrefers and symbol tokenizers now include the same, new `CommonLexer.lexh`.

The `getLineNumber()` and `setLineNumber(int x)` overrides in every xrefer are removed in favor of methods implemented in the common base class.

Following this PR, I have a branch ready to go that makes use of this PR's publicizing of line numbers and extends it so that (physical) lines-of-code counts are determined as well and made available in the index and in the UI: https://1drv.ms/i/s!Anl22C9Mx4VFgRqzmGmJ3Lzszyt6.
 
Thank you.


tarzanek:
ok, this is actually the same (or very similar) design that lucene does with its analyzers
I would try to learn from their experience and ev. take any design decisions from them
I am sure that for example the patch I am doing for jflex to keep in limits with lucene could be then moved to matchers directly, resp. we wouldn't need the patch, since analyzer would limit the output before passing on to lucene (which wasn't possible until this change) ...

tarzanek:
another design is how solr or elasticsearch implement their analyzers, which I didn't see yet ... but I'd take that as a good guide, since they are fast

idodeclare:
Closing, as this conflicts with `feature/more_tests` which I just raised.

PR #1962, opened 2017-12-21
fix a bug in getPathRelativeToSourceRoot when there is symlink

autoandshare:
Problem: When there is a symlink, getPathRelativeToSourceRoot will return the path relative to symlink.

Expected: The path relative to source root is returned.
Solution: Add the missing part.

Example:
/var/opengrok/src/project1 -> /root/project1
getPathRelativeToSourceRoot("/root/project1/web") => "/web"
It should be "/project1/web" instead.


vladak:
There is a PR #1952 in progress which is third incarnation of a fix for this (and other, symlink related) problem, thus closing in its favor.

Issue #1966, opened 2018-01-03
escape character does not work


tanglianggithub:
use loop\\(\\) to search  'loop()', but opengrok cannot find the result.

vladak:
I think the parenthesis characters are stripped away in the process of indexing.

tarzanek:
I am wondering if OpenGrok indexes "("
definitely worth talking about if this is a viable use case and worthwhile to improve tokenizers (at least the full plain one)

idodeclare:
> definitely worth talking about if this is a viable use case and worthwhile to improve tokenizers (at least the full plain one)

@tarzanek , I would say it's viable, and in fact I have a branch that uses my `JFlexSymbolMatcher` event model and `JFlexTokenizer` refactoring to do just what you’re suggesting w.r.t. `PlainFullTokenizer`.

But with the Oracle repo grab and not seeing @tarzanek or @vladak or @tulinkry in https://github.com/orgs/oracle/people — and moreover with reviews at a dead halt — I'm thinking of closing all my PRs here. I've forked FreeGrok and am waiting to see what's happening here within the next short interval.

tarzanek:
@idodeclare sorry, we're stuck on some legal talks, but we have a resolution, the repo transfer was a bit unfortunate and premature, but solution is in progress

me & @vladak can still accept PRs, I will try get back to your PRs, bear with us please

deeparora29:
I am also facing the similar issue where I search for M\{ and it returns false positives. FTR, I am using version 0.9.0

tarzanek:
@deeparora29 can you get to 1.2.x please?
0.9 is really old and Chris had his fixes in 1.x I think

PR #1971, opened 2018-01-05
Feature/symbol matcher

idodeclare:
(This replaces and extends PR #1958, which I closed due to minor conflicts with `feature/more_tests`.)

Hello,

Please consider for integration this refactoring that succeeds in bringing the language xrefers and symbol tokenizers together by revising them to inherit from the _same class_: `JFlexSymbolMatcher`.

`JFlexTokenizer` is revised and simplified as a non-abstract class that now uses each of the language-specific tokenizers, listening to their publishing of symbols. (As mentioned in issue #1966, `JFlexTokenizer` can afterward also be enhanced to listen to non-symbolic content to allow e.g. `PlainFullTokenizer` to produce more tokens — even overlapping tokens — from the same source to allow more comprehensive full search.)

`JFlexXref` is likewise revised and simplified as a non-abstract class that now uses each of the language-specific xrefers, listening to their publishing of complete symbolic and non-symbolic content.

The two very non-standard xrefers, `TroffXref` and `MandocXref`, are revised to inherit from `JFlexNonXref`, which is a slimmed-down version of the former, abstract incarnation of `JFlexXref`.

`JFlexXref` is enhanced to comprehensively URL- and HTML-encode content (when necessary and with no additional, transient buffers), with xrefers no longer having to do e.g. `writeUnicodeChar()`, `offerNonword()`, or `offerUnicode()`, etc. `writeSymbol()` is enhanced to include quote characters for parameters which are used for Lucene querying when their arguments contain a character that is not ASCII alpha-numeric and not underscore (e.g. `?defs=%22Set-and-update-format-list%22` for `sample.tcl`).

This new model will allow a future revision to retire `JFlexJointLexer` and to move e.g. `PerlLexHelper`, `RubyLexHelper` into language-specific sub-classes of `JFlexSymbolMatcher`. If desired (which I think it will be to eliminate redundancy), the other existing languages can be revised to share language-specific JFlex rules (à la `RubyProductions.lexh`).

As a consequence of this refactoring, `CommonTokenizer.lexh` and `CommonXref.lexh` are merged, and the language xrefers and symbol tokenizers now include the same, new `CommonLexer.lexh`. The `getLineNumber()` and `setLineNumber(int x)` overrides in every xrefer are removed in favor of methods implemented in `CommonLexer.lexh`.

Following this PR, I have a branch ready to go that makes use of this PR's publicizing of line numbers and extends it so that (physical) lines-of-code counts are determined as well and made available in the index and in the UI: https://1drv.ms/i/s!Anl22C9Mx4VFgRqzmGmJ3Lzszyt6.

@tarzanek , regarding your responses to PR #1958, I wasn't sure if there were specific changes you were asking for. Please would you clarify.

Thank you.


tarzanek:
hi, no, it was more like a design decision, since this looks like the design they have, so it was more about learning from others mistakes ;)
but let's give this a shot, I just had a quick look and it looks OK, code always improves in iterations and I like this iterations improvement over the old approach

perhaps one thing ... we're complicating the analyzers and the original howto and design docs are harder to follow, so it would make sense to write a story about what you do and where do you want to go ... what do you think?
life is all about stories, so let's add more to original @chandanbn stories

tarzanek:
thank you, let's get it rolling! :)

idodeclare:
> it would make sense to write a story about what you do and where do you want to go ... what do you think?
life is all about stories, so let's add more to original @chandanbn stories

Yes I can start on such a story.

PR #1972, opened 2018-01-05
Fix to use CVSRepository getRepoCommand() in test

idodeclare:
Hello,

Please consider for integration this tiny patch to fix `CVSRepositoryTest` to use `CVSRepository getRepoCommand()`, so that the test uses a developer's environment override of `cvs` command.

Thank you.

### DIFF
--- a/test/org/opensolaris/opengrok/history/CVSRepositoryTest.java
+++ b/test/org/opensolaris/opengrok/history/CVSRepositoryTest.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


tarzanek:
weird, I thought repo init will init the SCM command from -d variables ...
(but perhaps cvs needs a fix in this regard)

I can merge this(didn't look at how we do it for other repos and whether the logic to isolate overrides from within repos works well for all SCM providers), so I hope my trust here is OK and you did verify the others ;)

I know that for e.g. git or hg this override works well (since I override them on windows dev env with -d to fit custom binaries on windows)

tarzanek:
merging, let's see if I am wrong or not :)

idodeclare:
Yes it was just this test that was running a non-overridden command. Thank you, Lubos!

Issue #1801, opened 2017-10-03
RESTful API for Messages


noorflex:
is there a json API which gives me list of all projects which i can display in my interface.

vladak:
The existing web services (https://github.com/OpenGrok/OpenGrok/wiki/OpenGrok-web-services) are more tailored for searching. You can use `Messages` script to get the list of projects remotely however exposing the listening port to anyone is not a wise choice as this can be used to write configuration. Marking as enhancement - feel free to submit a pull req.

vladak:
Ideally the code in `ProjectMessage.java` should be refactored and used for web services API.

vladak:
Also, it's not clear to me whether currently the web services API is going through authorization code. @tulinkry ?

vladak:
The other thing to solve is how to authorize the project (non-search) related requests.

tulinkry:
It does the authorization

https://github.com/OpenGrok/OpenGrok/blob/7ccbd27804c2eaf410b0350e5a385587c941bb65/src/org/opensolaris/opengrok/web/JSONSearchServlet.java#L119-L123

and

https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/search/SearchEngine.java#L259-L296

noorflex:
Right, Existing web services are for searching, but to limit the scope of search to a subset of project (e.g one set of Branch out of many branches) will give more control to search.

PR #1975, opened 2018-01-11
Mirroring scripts

vladak:
This set of changes adds a bunch of Python scripts to synchronize SCM repositories (git, Mercurial, CVS, SVN, Teamware are supported) with their parents/upstream. Plus pre-requisite changes to Messages.

The main `mirror.py` script is supposed to be run per-project via the already existing `sync.py` script.


tools/Messages
```diff
@@ -98,6 +100,10 @@ Usage()
     echo "       - \"list-indexed\" - list indexed projects"
     echo "       - \"get-repos\" - get list of repositories in the form of relative paths to source root for given project(s)"
     echo "       - \"get-repos-type\" - get repository type(s) for given project(s)"
+    echo "    repository:"
+    echo "     - repository paths relative to source root are specified as <tags>"
+    echo "     - command is specified in message <text>:"
+    echo "       - \"get-repo-type\" - returns type of the repository or \"N/A\""
```
tulinkry: Update the main README as well

vladak: fixed

vladak:
I found a Pythonista in the company to do additional (sic!) review for the python scripts, still need someone to look at the Messages changes. Also, I will probably switch our production instance to these scripts to give it additional testing.

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -209,6 +211,53 @@ protected void invokeSetter(Configuration config, String field, String value) th
...
 
+    /**
+     * Invokes a getter of a property on the configuration object.
+     * 
+     * @param config configuration object
+     * @param field string with field name
+     * @return string representation of the field value
+     * @throws java.io.IOException 
+     */
+    protected String invokeGetter(Configuration config, String field) throws IOException {
```
tulinkry: this may be a candidate for a util function along with invokeSetter

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2017, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copyright

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
```diff
@@ -252,6 +301,7 @@ public void validate() throws Exception {
         if (toInteger(hasTag("setconf"))
                 + toInteger(hasTag("getconf"))
                 + toInteger(hasTag("auth"))
+                + toInteger(hasTag("get"))
                 + toInteger(hasTag("set")) > 1) {
             throw new Exception("The message tag must be either setconf, getconf, auth or set");
```
tulinkry: + get

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
```diff
@@ -0,0 +1,94 @@
...
+public class RepositoryMessage extends Message {
+    
+    @Override
+    protected byte[] applyMessage(RuntimeEnvironment env) throws Exception {
+        String ret = null;
+        String msgtext = getText();
+        
+        switch (msgtext) {
+            case "get-repo-type":
+                List<String> types = new ArrayList<>();
```
tulinkry: maybe a linked list? or array list with known capacity? to reduce the unnecessary overhead

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
```diff
@@ -0,0 +1,94 @@
...
+                            types.add(ri.getType());
+                            found = true;
+                            break;
+                        }
+                    }
+                    if (!found) {
+                        types.add("N/A");
+                    }
+                }
+                ret = types.stream().collect(Collectors.joining("\n")).toString();
```
tulinkry: just some thoughts..

this gives for a message with tags
`-t opengrok -t opengrok-master -t nothing`
the result
```
git
git
N/A
```

isn't something like
```
opengrok: git
opengrok-master: git
nothing: N/A
```
better?
or some json?

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Validate the message.
+     * Tag is repository path, text is command.
+     * @throws Exception 
+     */
+    @Override
+    public void validate() throws Exception {
+        String command = getText();
+        Set<String> allowedText = new TreeSet<>(Arrays.asList("get-repo-type"));
```
tulinkry: allowedTexts

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
```
tulinkry: 18

test/org/opensolaris/opengrok/configuration/messages/RepositoryMessageTest.java
```diff
@@ -0,0 +1,144 @@
...
+    
+    RuntimeEnvironment env;
+
+    private static TestRepository repository = new TestRepository();
+    
+    @Before
+    public void setUp() throws IOException {
+        Assume.assumeTrue(new MercurialRepository().isWorking());
+        Assume.assumeTrue(new SubversionRepository().isWorking());
+        Assume.assumeTrue(new GitRepository().isWorking());
```
tulinkry: I think I included the repeateable conditional rule for this, can you try it?

test/org/opensolaris/opengrok/configuration/messages/RepositoryMessageTest.java
```diff
@@ -0,0 +1,144 @@
...
+
```
tulinkry: do you need git and subversion in this test?

vladak:
For the record, here are the python comments:

(11:25:01) timf: vladimir.kotal: platform/solaris/ips/create.sh looks pretty antiquated. You really should replace that with a single p5m manifest and pkgsend publish
(11:25:15) timf: rather than publishing via repeated pkgsend add actions
(11:37:09) timf: tools/sync/hook.py looks a bit weird too. There's a 'cwd' parameter for popen call, so you're kinda going the long way around here.
(11:37:46) jan.parcel left the room.
(11:41:13) timf: tools/sync/mirror.py line 140, no need for the '+' at the end of the line
(11:41:28) timf: string literals continued across lines are automatically concatenated.
(11:42:17) timf: line 151, and elsewhere I'd use double quotes rather than single quotes - generally, just pick a style and stick with it.
(11:43:25) timf: line 184, no spaces between the name, the equals sign and the value. eg. good: parm_name=param_value, bad: param_name = param_value
(11:45:18) timf: line 237 should probably be log.error rather than log.debug
(11:49:38) timf: tools/sync/teamware.py a bit of extra safety might be to verify that os.path.isdir(command) before proceeding.
(11:50:19) timf: line 43, copy/paste coding s/hg_command/bringover_command/
(11:53:48) timf: tools/sync/utils.py def which(..) can be replaced by shutil.which(..) if you're guaranteed to have python3.
(11:53:48) timf: >>> import shutil
>>> shutil.which("cc")
'/opt/solarisstudio12.4/bin/cc'
>>> 

(11:58:28) timf: line 28, os.mkdir should perhaps be os.makedirs, so as to make any intermediate dirs along the way
(12:00:42) timf: the get_dict_val function seems a bit weird, not wrong, just.. odd.  (you know you can do dictionary.get(key) and it'll return None if the key doesn't exist, or dictionary.get(key, default) and it'll return default if key doesn't exist)
(12:00:55) timf: ok, I'm done.  Generally, nothing too weird there.

tarzanek:
looking forward to try it out once merged, thank you Vlad :)

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
vladak: fixed, hopefully we won't have to deal with this anymore soon.

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
vladak: you mean moving this to `org.opensolaris.opengrok.util` as a utility class ? or possibly to `Configuration.java` ?

tulinkry: t thought more about a new utility class but you decide both is better than have it here

tulinkry: or maybe if you extract this to the ClassUtils and make it more generic (for any object and not just for configuration) and then you can delegate these calls from here. How about that?

vladak: yep, that's what I did.

src/org/opensolaris/opengrok/configuration/messages/ConfigMessage.java
vladak: fixed by rewriting this test

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
vladak: used specified initial capacity

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
vladak: fixed

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
vladak: fixed

src/org/opensolaris/opengrok/configuration/messages/RepositoryMessage.java
vladak: okay, will use the colon separated output.

I will save JSON output for #1801.

test/org/opensolaris/opengrok/configuration/messages/RepositoryMessageTest.java
vladak: these are not needed, removed.

vladak:
@tarzanek I am a bit worried about the Python scripts having hardcoded Python 3.4 so might not run on anything else (3.5) even if it could.

tarzanek:
d'oh ... so once again the history repeats and the same language is not compatible with its own old versions?

vladak:
Well, python 2 and python 3 are different languages. This is not C. That
said, I can probably make the script headers use Python 3 or even generic
Python (since the script have embedded version checks) and hope it will run
on all 3.x.

Dne 16. 1. 2018 6:04 odp. napsal uživatel "Lubos Kosco" <
notifications@github.com>:

tulinkry:
how about /usr/bin/env python3 ? this should lead to the recent version of python3 on the system, or?

vladak:
I was thinking about `python3` however that would not work on Solaris. And the most recent version guarantee does not universally hold true.

tulinkry:
you can add a check in the code that the version is atleast 3 and throw an exception?

vladak:
The version check is already present however this is still not going to work on a system which has both Python 2 and 3 but `/usr/bin/python` points to Python 2.

tulinkry:
Ok I didn't see it

vladak:
I will just bite the bullet and use `python3` at the expense of having to create `python3` symlink on systems that do not have it.

tarzanek:
I agree, since it seems most OSes have python3 symlink, no ?

thnx
L

vladak:
- Ubuntu has the symlink
- macOS has the symlink if Python 3.x was installed from Homebrew
- Solaris does not have it

tulinkry:
yes I am for python3

tarzanek:
windows also doesn't have it, but I am OK with Ubuntu(Debian) and macOS :)

PR #1976, opened 2018-01-11
Optimize for a single listener. Support --profiler.

idodeclare:
Hello,

Please consider for integration this patch to optimize for a single `SymbolMatchedPublisher` listener. (If we possibly do need multiple listeners in the future, I think a new fan-out mediator would suffice.)

I also added a `--profiler` switch to `Indexer` and `OPENGROK_PROFILER` env var support to the `OpenGrok` script — which is a convenience to pause the process in order to allow attaching a profiler or debugger. I used this to profile this optimization in NetBeans.

Thank you.

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -115,13 +115,15 @@
 #   - OPENGROK_SUBVERSION_PASSWORD password of the user that should be used for
 #                                 fetching the history from subversion
 #
+#   - OPENGROK_PROFILER           Pause to await user input for profiling.
+#
 # Notes:
 #   (*) Any Non-Empty String will enable these options
 #
 
 #
 # Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
-# Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+# Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
 #
 
 #
@@ -866,6 +868,7 @@ CommonInvocation()
         ${READ_XML_CONF}                                        	\
         ${WEBAPP_CONFIG}						\
         ${WEBAPP_CONTEXT}						\
+        ${OPENGROK_PROFILER:+--profiler}				\
         "${@}"
 }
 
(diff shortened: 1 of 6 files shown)


tarzanek:
I generally do it with jdwp agent and suspend server, so it might make sense to add some debug option to OpenGrok script which will turn on jdwp
so just adding something like 
-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=9009
to java execution ...

tarzanek:
so would it make sense to rewrite what you do to rather support enabling jdwp options easily on some known port ?

idodeclare:
> so would it make sense to rewrite what you do to rather support enabling jdwp options easily on some known port ?

@tarzanek , I would argue no. This patch just supports collaborating with ProcessAttach (by literally pausing until the user instructs to continue) which does not have the security implications of opening a JDWP port to allow remote code execution without authentication.

tarzanek:
ok, you're right with security, but then usually grok runs on big remote machines, where you don't usually run X and IDE ... 
let me check the rest of this pull request and approve eventually

idodeclare:
Yes, I mean the `--profiler` switch to be a developer convenience that is harmless to be in the same script as used in production (where it probably will not be used).

tarzanek:
I am OK with merging this - the only thing I am not sure about is WHETHER you want the changes in matchers that are here?
(I hope you don't have them in some other PR)
If you're OK with merging them, then let's go 
@idodeclare ?

idodeclare:
@tarzanek , I'm ok with this PR getting merged. Thank you.

vladak:
Merging. We should have a Debugging wiki where such stuff can be documented.

idodeclare:
> We should have a Debugging wiki where such stuff can be documented.

That does sound useful. I did also later notice existing support in the `OpenGrok` script for a `JAVA_DEBUG` environment variable with a comment for an example setting. (Such a variable is not redundant with this new `--profiler` switch because NetBeans profiler can ProcessAttach without JWDP open though the debugging ProcessAttach cannot do.)

vladak:
I created https://github.com/oracle/opengrok/wiki/Debugging , please update it with the profiler info.

idodeclare:
OK

PR #1986, opened 2018-01-24
Fix PowerShell break|continue <label> handling

idodeclare:
Hello,

Please consider for integration this fix for PowerShell label handling broken in `feature/symbol_matcher`. I added a `break <label>` statement to sample.psm1.

Thank you.


tarzanek:
merging, thnx

Issue #591, opened 2013-05-22
xref directory pages should include LOC column (Bugzilla #356)


vladak:
status ACCEPTED severity _enhancement_ in component _webapp_ for _---_
Reported in version _unspecified_ on platform _Other_
Assigned to: Trond Norbye

On 2008-01-11 19:18:55 +0000, Hemang Lavana wrote:

> The directory listing under xref currently shows the file name, date, and file size info. For example, pls. see:
> 
> http://src.opensolaris.org/source/xref/onnv/onnv-gate/usr/src/
> 
> It should also display number of "lines" or LOC (lines of code) for each file. Author name and revision number for each file would be a nice addition as well.
> 
> Hemang.

On 2008-01-16 14:18:22 +0000, Trond Norbye wrote:

> Calculating this each time a user browse the directory might be slow if we have a lot of files (or large files) in the directory.
> 
> Since the parent directory will be updated if a file is updated we can cache this information and generate it during the "offline" index-run.

On 2008-01-16 14:36:17 +0000, Hemang Lavana wrote:

> It should certainly not be done on the web-side. The line count value for each file is already available in the analyzer code while it is generating the xref html files (since it inserts the line number in the html file). So if these line counts are somehow propagated to the directory level, that should get us the desired feature.

PR #1987, opened 2018-01-25
Feature/LOC

idodeclare:
Hello,

Please consider for integration this patch to count per-file physical lines-of-code ([LOC](http://csse.usc.edu/TECHRPTS/2007/usc-csse-2007-737/usc-csse-2007-737.pdf);
 [Wiki](https://en.wikipedia.org/wiki/Source_lines_of_code)) and to store LOC and total number of lines.

`DirectoryListing` is updated to [present the counts](https://1drv.ms/i/s!Anl22C9Mx4VFgRqzmGmJ3Lzszyt6) while browsing.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/index/IndexDatabase.java
+++ b/src/org/opensolaris/opengrok/index/IndexDatabase.java
@@ -111,6 +111,8 @@ public class IndexDatabase {
     private static final Comparator<File> FILENAME_COMPARATOR =
         (File p1, File p2) -> p1.getName().compareTo(p2.getName());
 
+    private final Object INSTANCE_LOCK = new Object();
+
     private Project project;
     private FSDirectory indexDirectory;    
     private IndexWriter writer;
@@ -272,7 +274,7 @@ public class IndexDatabase {
 
     @SuppressWarnings("PMD.CollapsibleIfStatements")
     private void initialize() throws IOException {
-        synchronized (this) {
+        synchronized (INSTANCE_LOCK) {
             RuntimeEnvironment env = RuntimeEnvironment.getInstance();
             File indexDir = new File(env.getDataRootFile(), INDEX_DIR);            
             if (project != null) {
@@ -1151,8 +1153,8 @@ public class IndexDatabase {
      */
     public Set<String> getFiles() throws IOException {
         IndexReader ireader = null;
-        TermsEnum iter;
-        Terms terms = null;
+        TermsEnum iter = null;
+        Terms terms;
         Set<String> files = new HashSet<>();
 
         try {
@@ -1161,8 +1163,8 @@ public class IndexDatabase {
             if (numDocs > 0) {
                 Fields uFields = MultiFields.getFields(ireader);//reader.getTermVectors(0);
                 terms = uFields.terms(QueryBuilder.U);
+                iter = terms.iterator(); // init uid iterator
             }
-            iter = terms.iterator(); // init uid iterator
             while (iter != null && iter.term() != null) {
                 files.add(Util.uid2url(iter.term().utf8ToString()));
                 BytesRef next = iter.next();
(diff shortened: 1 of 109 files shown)


idodeclare:
I should note that this resolved issue #591.

tulinkry:
I would personally prefer if all the methods mentioning LOC would have it expanded to LinesOfCode.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -272,7 +274,7 @@ public void run() {
 
     @SuppressWarnings("PMD.CollapsibleIfStatements")
     private void initialize() throws IOException {
-        synchronized (this) {
```
tulinkry: does this make any difference?

idodeclare: Just that synchronizing on public objects is suspect as it breaks encapsulation — so I revised it.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -805,7 +805,7 @@ private boolean accept(File file) {
         return !RuntimeEnvironment.getInstance().isIndexVersionedFilesOnly();
     }
 
-    boolean accept(File parent, File file) {
+    private boolean accept(File parent, File file) {
```
tulinkry: is this relevant? or just a cleanup?

idodeclare:
> I would personally prefer if all the methods mentioning LOC would have it expanded to LinesOfCode.

I was motivated to have it quite terse for the methods that are used in analyzers, since I had to edit hundreds of lines in them. Moreover since LOC is almost orthogonal to the bulk of logic in the xrefers, I wanted a tiny function name — `chkLOC()` — rather than seeing `checkLinesOfCode()` over and over (and also having to reformat possibly hundreds of newly too-long lines).

I do think "LOC" is well-known enough acronym in source code analysis; if analyzers are ever updated to support logical lines-of-code, then "LLOC" also will be well-known.

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: Just a cleanup

ChristopheBordieu:
Excellent feature!!!
Will all the total lines and slocs be aggregated at repository root level ?

vladak:
@idodeclare pls use 'fixes #591' in one the commit comments then.

idodeclare:
> Will all the total lines and slocs be aggregated at repository root level ?

That is certainly feasible in the future. Right now, only files have document data in the Lucene index. To store aggregated num-lines and LOC, it will be necessary to store Lucene data for directories — but without conflicting with the existing queries that all assume only file-level data.

One straight-forward idea would be to add Lucene "documents" for directories with wholly-separate fields so that these new documents never appear in current queries. Of course it would also be necessary to write an aggregator that runs at the appropriate stages.

vladak:
This also brings a question about how to use these counts for restricting
searches.

Dne 26. 1. 2018 5:10 odp. napsal uživatel "C Fraire" <
notifications@github.com>:

idodeclare:
> This also brings a question about how to use these counts for restricting
searches.

Yes, indeed.

Relatedly, I did not update to show the counts in search results, as I did not have any idea how best to present them.

tarzanek:
I am merging this as is, thank you Chris, it looks good

idodeclare:
Thanks, @tarzanek !

vladak:
I wonder how this will cope with incremental indexing. Will pre-existing files have LOC set to 0 ? Or is full reindex required ?

idodeclare:
Pre-existing documents will have blank LOC/#Lines — the same as for files which do not get an xref (e.g. `Ignored/FileAnalyzer`) or for files that do not have a counting xref (e.g. `UuencodeAnalyzer`).

Incrementally re-indexed files can get LOC/#Lines. Full re-index is required for completeness.

PR #1988, opened 2018-01-25
Feature/help details

idodeclare:
Hello,

Please consider for integration this patch to support a `help --detailed` switch for `OpenGrok` script and for `Indexer`. These following additional sections (excerpts) will be shown in the terminal:

AnalyzerGuru prefixes:
---
	MAKEFILE*  : Shell script

AnalyzerGuru extensions:
---
	*.ada     : Ada
	*.adb     : Ada
	*.ads     : Ada
	*.afm     : IgnorantAnalyzerFactory
	...

AnalyzerGuru magic strings:
---
	BZh                     : Bzip(2)
	\{7f}ELF                : ELF
	#!/usr/bin/env escript  : Erlang
	...

AnalyzerGuru magic matchers:
---
	Jar         : PK\{3}\{4} magic signature with 0xCAFE Extra Field ID
	Zip         : PK\{3}\{4} magic signature
	Mandoc      : heuristic matcher
	...

Configuration examples:
---
	
	<java version="1.8.0_25" class="java.beans.XMLDecoder">
	 <object class="org.opensolaris.opengrok.configuration.Configuration"/>
	</java>

	  
	  <void property="allowedSymlinks">
	   <void method="add">
	    <string>user-defined-element</string>
	   </void>
	  </void>
 
	  
	  <void property="bugPage">
	   <string>user-specified-value</string>
	  </void>

and lastly, the block comment in `OpenGrok` is revised as a `cat` that displays with `--detailed`:

	Supported Environment Variables:
	  - OPENGROK_CONFIGURATION      User Configuration (Shell Script)
	  - OPENGROK_NON_INTERACTIVE    Suppress Progress and Warnings Messages (*)
	  - OPENGROK_STANDARD_ENV       Run Time Shell Environment (Shell Script)
	...

Some addiitonal revisions to support their self-describability:
- `JarAnalyzerFactory` gets its own specific `Matcher`, and `ZipAnalyzerFactory`'s `Matcher` is revised to only apply to `ZipAnalyzerFactory`.
- Add `FileAnalyzerFactory.Matcher.forFactory()` method.
- Only return unmodifiable collections from `AnalyzerGuru`.

Thank you.


src/org/opensolaris/opengrok/configuration/ConfigurationHelp.java
```diff
@@ -0,0 +1,288 @@
...
+     * Gets sample content for a configuration XML file.
+     * @return a defined instance
+     * @throws RuntimeException if an error occurs producing the sample
+     */
+    public static String getSamples()
+            throws RuntimeException {
+
+        Configuration conf = new Configuration();
+        Class klass = conf.getClass();
+        klass.getDeclaredMethods();
```
vladak: This seems to throw away the returned array of Method objects. Is this necessary ?

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -529,39 +542,62 @@ public void setUrlPrefix(String urlPrefix) {
...
+     * N.b. the value is not mediated to {@link Configuration}.
      *
-     * @param value the mandoc program to use or {@code null}
+     * @param value a defined value or {@code null} to reset to use the
+     * {@link Configuration#getMandoc()} fallbacks
+     * @see #getMandoc()
      */
     public void setMandoc(String value) {
-        threadConfig.get().setMandoc(value);
+        mandoc = value;
```
vladak: use this.mandoc to be consistent with setCtags() ?

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -371,8 +387,13 @@ public static WebAddress parseWebAddress(String webAddr) {
...
             parser.on("-?", "-h", "--help", "Display this usage summary.").Do( v -> {
-                parser.help();
-                System.exit(status);
+                help = true;
+                helpUsage = parser.getUsage();
+            });
+
+            parser.on("--detailed",
+                "Display additional help with -h,--help.").Do(v -> {
+                helpDetailed = true;
```
vladak: What should happen if `--detailed` is used without `-h` ? Trigger an error or imply `help = 1` ? (in the latter case `--detailed-help` might be better)

test/org/opensolaris/opengrok/analysis/executables/JarAnalyzerFactoryTest.java
```diff
@@ -0,0 +1,55 @@
...
+public class JarAnalyzerFactoryTest {
+
+    /**
+     * Tests a JAR file.
+     * @throws IOException
+     */
+    @Test
+    public void testJarWrtAnalyzerGuru() throws IOException {
+        InputStream res = getClass().getClassLoader().getResourceAsStream(
+            "org/opensolaris/opengrok/analysis/executables/javajar.bin");
```
vladak: there is already a jar file in `testdata/sources/java/main.jar`

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -1110,7 +1131,7 @@ private static void pauseToAwaitProfiler() {
         String in;
         do {
             System.out.print("Start profiler. Continue (Y/N)? ");
-            in = scan.nextLine().toLowerCase();
+            in = scan.nextLine().toLowerCase(Locale.ROOT);
```
vladak: maybe one day we will have localized help messages ? :-)

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -529,39 +542,62 @@ public void setUrlPrefix(String urlPrefix) {
...
+     * <p>
+     * N.b. the value is not mediated to {@link Configuration}.
      *
-     * @param ctags the ctags program to use
+     * @param ctags a defined value or {@code null} to reset to use the
+     * {@link Configuration#getCtags()} fallbacks
+     * @see #getCtags()
      */
     public void setCtags(String ctags) {
-        threadConfig.get().setCtags(ctags);
```
vladak: What's the logic behind this change ? Also I wonder what will be the effect w.r.t. `Configuration` serialization - will the XML form still contain custom ctags value ?

OpenGrok
```diff
@@ -18,116 +18,12 @@
...
-# Notes:
-#   (*) Any Non-Empty String will enable these options
+# Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
+# Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
 #
 
 #
-# Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
-# Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
+# OpenGrok Wrapper (initial setup and cron job updating)
```
tulinkry: can you add something like "Please refer to the Usage section for supported environment variables." ?

src/org/opensolaris/opengrok/configuration/ConfigurationHelp.java
idodeclare: OK deleted

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
idodeclare: OK changed

src/org/opensolaris/opengrok/index/Indexer.java
idodeclare: It would be ignored, which I thought was OK.

test/org/opensolaris/opengrok/analysis/executables/JarAnalyzerFactoryTest.java
idodeclare: I did copy an existing OpenGrok jar. I thought it was OK to have a test file close at hand like all the other tests which use embedded resources, and I thought it OK to indicate via filename that it is not being matched by name.

vladak: I get the point about name matching (even though it can be renamed while the test is running) however `opengrok.jar` weighs some 2 megs which is probably too much for matching first couple of bytes.

src/org/opensolaris/opengrok/index/Indexer.java
vladak: I think we can do better than that, especially if it costs one line of code.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
idodeclare: (There are no non-test users of `RuntimeEnvironment setCtags(String)`.)

Right now, `Configuration` has a non-static default value for `ctags` (from the env var, `org.opensolaris.opengrok.analysis.Ctags`) so `OpenGrok` has the non-intuitive behavior that if you specify `-c,--ctags ...` on the command-line when you have the environment variable set, then no value will be written to `configuration.xml` because the serializer determines that `Configuration` matches a default-constructed instance.

This just moves the reading of the env var to `RuntimeEnvironment` as a runtime concern vs a configuration concern and so that `configuration.xml` will always have a defined element when you specify `-c,--ctags ...` on the command-line.

OpenGrok
idodeclare: OK added

src/org/opensolaris/opengrok/index/Indexer.java
idodeclare: Hmm. It's documented as "additional help with -h,--help". Why raise an exception if `-h,--help` is not used?

To compare with another option, `-V,--version` just "Print[s] version and quit" — but there is no exhaustive checking that `-V,--version` is not used with any other options. Likewise with `--man` to "Generate OpenGrok XML manual page [and quit]".

test/org/opensolaris/opengrok/analysis/executables/JarAnalyzerFactoryTest.java
idodeclare: :) I picked a 1000-byte jar.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: Makes sense. Will this make code like

```
private static final String CTAGS_PROP =
60          "org.opensolaris.opengrok.analysis.Ctags";
...

env.setCtags(System.getProperty(CTAGS_PROP, "ctags"));
```

in `JavaAnalyzerFactoryTest.java` redundant ?

idodeclare: Yes, I think so. Let me push a cleanup of those — unless you think to wait.

src/org/opensolaris/opengrok/index/Indexer.java
vladak: I was actually advocating for the reverse - pretend that `-h` is set if `--detailed` stands alone. The current option parsing framework does not give the flexibility to state incompatible options so feel free to leave as is.

test/org/opensolaris/opengrok/analysis/executables/JarAnalyzerFactoryTest.java
vladak: sounds better. I am sure people still behind ADSL will appreciate it :-)

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: Go ahead, I think it makes to have the cleanup done in the same PR.

src/org/opensolaris/opengrok/index/Indexer.java
idodeclare: OK

vladak:
Looks good now. Please resolve the conflicts and I will merge.

Issue #1993, opened 2018-01-29
indexer fails with 'Too many open files' since 1.1-rc20


vladak:
Since 1.1-rc20 the indexing of "prepped" (i.e. unpacked with patches applied) version of https://github.com/oracle/solaris-userland gives me:

```
2018-01-29 09:03:01.638+0100 SEVERE t37 Indexer$1.run: An error occured while updating index
java.nio.file.FileSystemException: /opengrok/data/index/userland-default-prepped: Too many open files
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
        at sun.nio.fs.UnixFileSystemProvider.newDirectoryStream(UnixFileSystemProvider.java:427)
        at java.nio.file.Files.newDirectoryStream(Files.java:457)
        at org.apache.lucene.store.FSDirectory.listAll(FSDirectory.java:215)
        at org.apache.lucene.store.FSDirectory.listAll(FSDirectory.java:234)
        at org.apache.lucene.store.FilterDirectory.listAll(FilterDirectory.java:57)
        at org.apache.lucene.index.IndexFileDeleter.refresh(IndexFileDeleter.java:429)
        at org.apache.lucene.index.IndexWriter.rollbackInternalNoCommit(IndexWriter.java:2249)
        at org.apache.lucene.index.IndexWriter.rollbackInternal(IndexWriter.java:2193)
        at org.apache.lucene.index.IndexWriter.shutdown(IndexWriter.java:1129)
        at org.apache.lucene.index.IndexWriter.close(IndexWriter.java:1171)
        at org.opensolaris.opengrok.index.IndexDatabase.update(IndexDatabase.java:487)
        at org.opensolaris.opengrok.index.Indexer$1.run(Indexer.java:1046)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

[...]

PR #1994, opened 2018-01-30
Fix to try-with-resources spotted by @tulinkry

idodeclare:
Hello,

Please consider integration of this bugfix patch.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/index/PendingFileCompleter.java
+++ b/src/org/opensolaris/opengrok/index/PendingFileCompleter.java
@@ -30,7 +30,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


vladak:
can you make the `fixes #1993` to appear in the commit comments ?

idodeclare:
No, I can't prove it fixes this issue.

vladak:
Let me take the diff and deploy on the production (sic!) server first so that we can see if it fixes the problem or not.

tarzanek:
anyways I'd merge this and we can look for more places like this ...

tarzanek:
thank you @idodeclare  , let's look for more

vladak:
Could you just hold on for a minute ? It does not take much time to test
this.

Dne 31. 1. 2018 8:19 napsal uživatel "Lubos Kosco" <notifications@github.com

tulinkry:
It doesn't, however closing a stream is a good practice anyway.

vladak:
It's good that you're so gung-ho over fixing bugs. My preference is to have the issue number in the history log.

vladak:
confirmed that this fixed #1993.

PR #1998, opened 2018-02-02
Fix to also check for boolean-"is" methods for Help

idodeclare:
Hello,

Please consider for integration this fix so that `ConfigurationHelp` also checks for "is" methods.

Thank you.


tarzanek:
merging, good fallback

PR #2001, opened 2018-02-02
Close stream. Buffer directly from File. Fix arg.

idodeclare:
Hello,

Please consider for integration this patch to use try-with-resources for a `FileInputStream` in `FileHistoryCache`, to revise so `Buffered...` objects are directly above `File`, and to fix a wrong arg.

For the "wrong arg", the fix has no true effect because the argument is ignored anyway by `FileHistoryCache.getHistory(File, boolean, boolean)`, but it was counterintuitive without the fix.

Thank you.


tarzanek:
thank you for the quick fix, merging

PR #2003, opened 2018-02-05
Fix #431 : option -w should die

### LINKED ISSUE #431: option -w should die (Bugzilla #13985)
status NEW severity _minor_ in component _webapp_ for _---_
Reported in version _unspecified_ on platform _ANY/Generic_
Assigned to: Trond Norbye

On 2010-01-20 08:05:25 +0000, Lubos Kosco wrote:

> the price for getting rid of -w is bigger xrefs, which need to generate proper paths instead of just /source/<servlet> ....
> what we gain is, that we can name the webapp anyhow without the need to reindex the sources
> 
> fix is quite easy - adding a generator of proper URL path pointer for generated xref links ( ../../../ ), the question is, whether it will be worth it, since the webapp you just name once, but the xref index size increase for this will be like 10-20% (which means more traffic and more disk space usage)
> 
> I tried finding another solution, like generate a base meta header, but that will not help here

idodeclare:
Hello,

Please consider for integration this patch to kill option `-w,--web` and to handle instead a JFlex-powered, dynamic rewriting of `href="/source/` to target `request.getContextPath()`.

Thank you.

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -125,10 +125,6 @@ Supported Environment Variables for configuring the default setup:
                                 (by default 3 directories from SRC_ROOT)
   - OPENGROK_WEBAPP_CFGADDR     Web app address to send configuration to
                                 (use "none" to avoid sending it to web app)
-  - OPENGROK_WEBAPP_CONTEXT     Context URL of the OpenGrok webapp
-                                (by default /source)
-                                - FULL reindex is needed once this is used
-                                (old already indexed files won't be refreshed)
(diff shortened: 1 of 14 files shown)


idodeclare:
Users will not have to re-index unless they want to relocate and were previously using `-w,--web` with a non-standard context path.

idodeclare:
Oops sorry this will have a conflict with `feature/LOC`. I'll re-open this subsequently.

PR #2011, opened 2018-02-13
Fix to handle AlreadyClosedException specially

idodeclare:
Hello,

Please consider for integration this patch so that `AlreadyClosedException` is detected to terminate a run early. This happens after e.g. `OutOfMemoryException` has occurred and the Lucene `IndexWriter` has aborted. This way, an indexing run will terminate ASAP without doing pointless work that cannot be committed anyway.

Thank you.


tarzanek:
merging, thank you, good guard

PR #2012, opened 2018-02-14
Fix to consistently default to UTF-8 for SRCROOT readers

idodeclare:
Hello,

Please consider for integration this patch to consistently use UTF-8 as a fallback for readers against `SRCROOT`. This is a fix that helps to prepare for later integration of the Lucene `UnifiedHighlighter` (a2a76a2).

I encountered problems with inconsistent charsets when testing OpenGrok on JVMs on macOS and on OmniOS. On macOS, the JDK uses UTF-8 by default as the system charset; while on OmniOS, I determined it was Latin-1.

I think UTF-8 is a more sensible default for OpenGrok because it eliminates some unexpected inconsistency and aligns with the default encoding of Universal ctags. I am using `StandardCharsets.UTF_8` instead of a literal `"UTF-8"` as a way of tagging code that I reviewed. If it was decided later to support specifying an explicit, alternate default charset (for OpenGrok and ctags), I think having used `StandardCharsets` will help identify places for revision.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -32,6 +32,7 @@ import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
 import java.lang.reflect.InvocationTargetException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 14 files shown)


tarzanek:
ah, thank you for this, to finish reviewing places/streams that needed explicit encoding was something that is long due
were you able to check all of them? I think I've had on helius findbugs/checkstyle or in sonar a trigger to validate this ... @vladak can you check if after merging this we get the default encoding streams check to 0 hits?

thank you
L

PR #2014, opened 2018-02-14
Fix not to add out-of-order tokens for Jar, JavaClass

idodeclare:
Hello,

Please consider for integration this patch to fix `JarAnalyzer` and `JavaClassAnalyzer` not to add out-of-order tokens. This is a prerequisite for the upcoming Lucene `UnifiedHighlighter`; and, as Lubos found, for Lucene 7 in PR #1871.

Thank you.


tarzanek:
lovely, thank you, merging

PR #2015, opened 2018-02-14
Fix not to reuse a Definitions.Tag within an xref line

idodeclare:
Hello,

Please consider for integration this patch to account for `Tag` use during xref creation so as not to reuse a definition within a single line — i.e. so `hasDefinitionAt(String, int, ...)` only returns `true` once for a given symbol and line.

Some source code files define a symbol and then reference the same symbol name on the same line. Previously, both of those usages in a single line would be marked up as definitions. With this patch, the usage of `Tag` instances is tracked to avoid this specific re-use.

The new field is `transient`, so it will not affect existing binary serializations.

Please see all the affected tests below that demonstrate the corrected behavior.

Thank you.


tarzanek:
lgtm

PR #2016, opened 2018-02-14
Fix to ignore IOException from mandoc errstream

idodeclare:
Hello,

Please consider for integration this patch to silence an unhelpful mandoc error regarding its normally-closed error stream.

Thank you.


tarzanek:
straightforward, merging, thank you!

Issue #51, opened 2013-04-19
Search results of complex queries are not highlighted correctly


cnst:
If the search query has quotes and an `OR` operator, the results are not highlighted correctly &mdash; only the last term from within the quotes is highlighted.

For example, http://bxr.su/s?q=%22Robert%20Watson%22%20OR%20rwatson, where the query is, `"Robert Watson" OR rwatson`, produces highlighting like `Robert <b>Watson</b>`, e.g. only the second/last term from within the quotes is highlighted.  Similarly, http://bxr.su/s?q=%22Constantine+A.+Murenin%22+OR+cnst highlights as `Constantine A. <b>Murenin</b> &lt;<b>cnst</b>`.

I can reproduce this with OpenGrok 0.10, some 0.11-rc2 from an old trunk, as well as the recent master, so, it probably never worked right.

idodeclare:
Hi. This will be fixed before too long with OpenGrok using the modern Lucene `UnifiedHighlighter`. E.g., please see this [screen cap](https://1drv.ms/i/s!Anl22C9Mx4VFgRsa1l5ouRlELltC) from a working, development version.

PR #2017, opened 2018-02-14
Store a SHA-256 digest with all documents

idodeclare:
Hello,

Please consider for integration this patch to store a SHA hash for documents.

This is a prerequisite for including the Lucene `UnifiedHighlighter` as that highlighter requires strictly affirming that a Lucene `Document` and the underlying source text are in accordance. I.e., since OpenGrok rightly does not store full, original source text in the Lucene index, a SHA hash will allow `Context` to affirm that the source text is either unchanged (so that `UnifiedHighlighter` can be used) or that a temporary fallback to the old `PlainLineTokenizer` method should be done.

(This latter fallback could happen transiently e.g. when `git update` has been executed but before `OpenGrok index` has completed.)

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis;
 
@@ -52,6 +52,7 @@ import org.apache.lucene.document.Field;
 import org.apache.lucene.document.Field.Store;
 import org.apache.lucene.document.FieldType;
 import org.apache.lucene.document.SortedDocValuesField;
+import org.apache.lucene.document.StoredField;
 import org.apache.lucene.document.StringField;
 import org.apache.lucene.document.TextField;
 import org.apache.lucene.util.BytesRef;
@@ -465,13 +466,24 @@ public class AnalyzerGuru {
             doc.add(npstring);
         }
 
+        StreamSource src = StreamSource.fromFile(file);
+        byte[] digest;
+        try (DigestedInputStream digestStream = src.getSHA256stream()) {
+            digest = digestStream.digestAll();
+        } catch (IOException ex) {
+            throw ex;
+        } catch (Exception ex) {
+            throw new IOException("DigestedInputStream close()");
+        }
+        doc.add(new StoredField(QueryBuilder.SHA256, digest));
+
         if (fa != null) {
             Genre g = fa.getGenre();
             if (g == Genre.PLAIN || g == Genre.XREFABLE || g == Genre.HTML) {
                 doc.add(new Field(QueryBuilder.T, g.typeName(),
 		        string_ft_stored_nanalyzed_norms));
             }
-            fa.analyze(doc, StreamSource.fromFile(file), xrefOut);
+            fa.analyze(doc, src, xrefOut);
 
             String type = fa.getFileTypeName();
             doc.add(new StringField(QueryBuilder.TYPE, type, Store.YES));
--- a/src/org/opensolaris/opengrok/analysis/StreamSource.java
+++ b/src/org/opensolaris/opengrok/analysis/StreamSource.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis;
 
@@ -27,6 +28,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
 /**
  * This class lets you create {@code InputStream}s that read data from a
@@ -39,6 +43,9 @@ import java.io.InputStream;
  * with each other.
  */
 public abstract class StreamSource {
+
+    protected static final String SHA256 = "SHA-256";
+
     /**
      * Get a stream that reads data from the input source. Every call should
      * return a new instance so that multiple readers can read from the source
@@ -48,6 +55,16 @@ public abstract class StreamSource {
      * @throws IOException if an error occurs when opening the stream
      */
     public abstract InputStream getStream() throws IOException;
+
+    /**
+     * Gets a stream that reads data from the input source and produces a
+     * SHA-256 digest. Every call should return a new instance so that multiple
+     * readers can read from the source without interfering with each other.
+     *
+     * @return a defined instance
+     * @throws IOException if an error occurs when opening the stream
+     */
+    public abstract DigestedInputStream getSHA256stream() throws IOException;
 
     /**
      * Helper method that creates a {@code StreamSource} instance that
@@ -62,6 +79,58 @@ public abstract class StreamSource {
             public InputStream getStream() throws IOException {
                 return new BufferedInputStream(new FileInputStream(file));
             }
+
+            @Override
+            public DigestedInputStream getSHA256stream() throws IOException {
+                return getSHA256stream(getStream());
+            }
         };
+    }
+
+    protected static DigestedInputStream getSHA256stream(InputStream under)
+            throws IOException {
+        MessageDigest md;
+        try {
+            md = MessageDigest.getInstance(SHA256);
+        } catch (NoSuchAlgorithmException ex) {
+            /*
+             * NOT EXPECTED because "Every implementation of the Java
+             * platform is required to support ... SHA-256 ...."
+             */
+            throw new IOException("digest failed");
+        }
+        DigestInputStream dis = new DigestInputStream(under, md);
+
+        return new DigestedInputStream() {
+            @Override
+            public InputStream getStream() {
+                return dis;
+            }
+
+            @Override
+            public MessageDigest getMessageDigest() {
+                return dis.getMessageDigest();
+            }
+
+            @Override
+            public byte[] digestAll() throws IOException {
+                return StreamSource.digestAll(this);
+            }
+
+            @Override
+            public void close() throws Exception {
+                dis.close();
+            }
+        };
+    }
+
+    protected static byte[] digestAll(DigestedInputStream dis)
+            throws IOException {
+        byte[] buf = new byte[64 * 1024];
+        InputStream iss = dis.getStream();
+        while (iss.read(buf) != -1) {
+            // iterate
+        }
+        return dis.getMessageDigest().digest();
     }
 }
--- a/src/org/opensolaris/opengrok/analysis/DigestedInputStream.java
+++ b/src/org/opensolaris/opengrok/analysis/DigestedInputStream.java
@@ -1,0 +1,39 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
+ */
+
+package org.opensolaris.opengrok.analysis;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.MessageDigest;
+
+/**
+ * Represents an API to allow associating a {@link MessageDigest} with an input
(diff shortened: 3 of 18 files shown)


vladak:
Also, what's the plan w.r.t. full reindexing. This only adds a new field to the index so I believe should be fine however the upcoming highlighter change will need to consume it which will require full reindex, correct ?

src/org/opensolaris/opengrok/analysis/StreamSource.java
```diff
@@ -62,6 +79,58 @@ public static StreamSource fromFile(final File file) {
...
+            }
+        };
+    }
+
+    protected static byte[] digestAll(DigestedInputStream dis)
+            throws IOException {
+        byte[] buf = new byte[64 * 1024];
+        InputStream iss = dis.getStream();
+        while (iss.read(buf) != -1) {
+            // iterate
```
tarzanek: let's see what will be the performance impact,
but since we traverse the same file maybe like for the 4th time, I guess it'll be OK

tarzanek:
from my side it's OK, as long as Vladak is OK, feel free to merge ... (so @vladak feel free to merge once you are OK)

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
```diff
@@ -465,13 +466,24 @@ public void populateDocument(Document doc, File file, String path,
...
+        StreamSource src = StreamSource.fromFile(file);
+        byte[] digest;
+        try (DigestedInputStream digestStream = src.getSHA256stream()) {
+            digest = digestStream.digestAll();
+        } catch (IOException ex) {
+            throw ex;
+        } catch (Exception ex) {
+            throw new IOException("DigestedInputStream close()");
+        }
+        doc.add(new StoredField(QueryBuilder.SHA256, digest));
```
vladak: `StreamSource.SHA256` or better `StreamSource.getDigestName()` ?

vladak:
Also, I wonder if the index should store digest algorithm identification so that it can be used to incrementally upgrade if the algorithm changes.

vladak:
It would be also nice to have some digest test vectors in the tests to match against known output (or possibly do this on AnalyzerGuru level.

src/org/opensolaris/opengrok/analysis/StreamSource.java
```diff
@@ -62,6 +79,58 @@ public static StreamSource fromFile(final File file) {
...
+            }
+
+            @Override
+            public void close() throws Exception {
+                dis.close();
+            }
+        };
+    }
+
+    protected static byte[] digestAll(DigestedInputStream dis)
```
vladak: I wonder if this should be part of DigestedInputStream (make it abstract class rather than interface) to avoid exposing MessageDigest in order to avoid anyone getting the value before the digest is fully computed.

idodeclare:
> the SHA256 seem to be spread around too much. Can this be made more generic ?

> DigestedInputStream ... make it abstract class ...

@vladak , from both of these quotes, I am inferring that you think it should be feasible to declare and define the logic in one place, but that is not tractable because of the pre-existence of abstract `StreamSource` and its three primary implementations where two are transforming subclasses: `GZipAnalyzer` and `BZip2Analyzer`.

The comment to `DigestedInputStream` clarifies this:
```
/**
 * Represents an API to allow associating a {@link MessageDigest} with an input
 * stream where OpenGrok can re-wrap the stream while carrying forward the
 * digest of the underlying data.
 */
```
To execute the SHA calculation in single place would mean at best overlaying the `InputStream` coming from a `StreamSource` instance (again, either `StreamSource.fromFile()`, `GZipAnalyzer` and `BZip2Analyzer`). That would mean incorrectly digesting bytes from `GZipAnalyzer` and `BZip2Analyzer` to produce a hash different from the hash computed from the underlying file via `StreamSource.fromFile()` — i.e. it would not "carry forward the digest of the underlying data."

If you study this patch, you'll see the bulk of logic is installed into `StreamSource` where it is reused as appropriate by `StreamSource` itself as well as by `GZipAnalyzer` and `BZip2Analyzer`. (As well as by the three test classes which for test purposes have their own anonymous subclasses of `StreamSource`; I deleted six anonymous subclasses in tests that were superfluous.)

Thus, with this patch, you can `StreamSource.fromFile().getSHA256stream().digestAll()` and that is the same value as for e.g.: 1) `var bzdis = BZip2Analyzer.wrap(StreamSource).getSHA256stream()`, 2) reading all unzipped bytes from `bzdis.getStream()`; 3) `bzdis.getMessageDigest().digest()`.

src/org/opensolaris/opengrok/analysis/StreamSource.java
idodeclare: I argue no in my previous note.

Also, how would a third-party affirm that an `InputStream` is fully digested without possibly altering the stream unless a mediator was overlayed? That seems of dubious value.

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
idodeclare: This is a key name into the Lucene document, which I believe could be argued is the purview of `QueryBuilder` (for historical reasons). Extracting an entirely new static class from `QueryBuilder` is one future idea, but it seems questionable whether it would be a good idea to disperse key names around the source code as with e.g. `StreamSource.SHA256` or `StreamSource.getDigestName()`.

vladak:
I meant something weaker than that - basically change the code so that replacing digest algorithm is almost one liner. (e.g. `getSHA256stream` can be named `getDigestStream`, the `SHA256` string can be moved to completely to `StreamSource` etc.)

src/org/opensolaris/opengrok/analysis/StreamSource.java
idodeclare: [README](https://github.com/oracle/opengrok/blob/master/README.md) does advise that "Source base should be available locally for OpenGrok to work efficiently"; so maybe we count on ARC or the Linux equivalent to let additional reads be minimally more costly.

idodeclare:
> replacing digest algorithm is almost one liner

This is a hash constrained logically within an OpenGrok project, representing files within an SCM repository, and _derived_ from underlying data. It's not e.g. a hash for ZFS blocks where the block is the source of truth (i.e. not derived), might survive for decades, and where blocks with heterogenous algorithms might need to coexist.

There can be no rational feasibility that OpenGrok would move away from SHA-256.

idodeclare:
> It would be also nice to have some digest test vectors

Let me add a couple of tests.

vladak:
> There can be no rational feasibility that OpenGrok would move away from SHA-256.

The digest algorithm ecosystem is moving in time so I'd not dare to say something like that. Eventually SHA256 might be phased out in favour of SHA-3 or something completely different. Alternatively look at it purely from readability perspective. It is not necessary to expose that the SHA256 is used.

src/org/opensolaris/opengrok/analysis/StreamSource.java
vladak: The ZFS ARC can be big beast no doubt however what I see in the production (on what can be considered now a modest machine with 256 GB of RAM) is that JVM memory pressure forces ARC to shrink during indexing quite drastically, especially with parallel project indexing. Granted, I could tune `zfs_arc_min` to avoid that however that would hamper the indexer(s). You can call it bad capacity planning :-)

src/org/opensolaris/opengrok/analysis/StreamSource.java
vladak: I was more after what happens if someone calls `getMessageDigest().digest()` before `digestAll()`. Seems like API deficiency to me.

idodeclare:
> It is not necessary to expose that the SHA256 is used.

It is though or else we have to introduce a metadata system as you are asking for where the algorithm name has to go along with the `byte[]`. I am saying that it is pointless to make a derived data system like OpenGrok be able to track the evolution of hashes.

> SHA256 might be phased out in favour of SHA-3

You seem to assert that projects should change their hash algorithms for fashionable reasons. On the contrary, without a compelling reasons to do so, hash algorithms usage should be stable and follow the true usage. E.g., this patch is not using SHA-256 for cryptographic security over streams of external data (as for HMAC), nor for object identity in an expanding database (as for Git which is moving to SHA-3-256), but to detect whether a file of a given name has the same content now as compared to the content of the same file at a previous point in time. Again, it's not rational to think that SHA-256 may not suffice for that _in perpetuity_.
 
Moreover, you're going to require people to reindex quite often to add fields or to upgrade Lucene. If you wanted to change hashes, piggy back it on that.

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
vladak: Can the key name be named in generic way then ? (`digest` ?)

src/org/opensolaris/opengrok/analysis/StreamSource.java
idodeclare: No consumer would call both of those because `getMessageDigest().digest()` is only used when there is a true reader of the data, and `digestAll()` is only used where the data is thrown away and only the digest is important.

vladak:
> You seem to assert that projects should change their hash algorithms for fashionable reasons.

No, I meant that it might be simply removed from future JDK (e.g. in case someone finds a security problem in SHA-256). I just don't see why the methods need to have SHA256 in their name.

idodeclare:
> I just don't see why the methods need to have SHA256 in their name.

It's due to our different notions. I am baking SHA-256 into this (as per my previous notes) while you want metadata about the hash (e.g., `ALGORITHM` and `DIGEST`).

> It might be simply removed from future JDK (e.g. in case someone finds a security problem in SHA-256).

That's neutral to our points of view. E.g., my key name documents what the `byte[]` value represents (though you suggested to rename the key to `DIGEST`).

idodeclare:
> It might be simply removed from future JDK

Also, really are you conceiving that it would be removed even if a security problem were found?

src/org/opensolaris/opengrok/analysis/StreamSource.java
idodeclare: In this case I just mean ARC enough for a sequence of file reads (sequential in OpenGrok; parallel in Lucene) during any single `addFile()` call. I.e., not long-term ARC for these files.

vladak:
That's probably true that the algs will stay for a long time. The reason why I am making so much fuss is that I got burned myself by selecting SHA-1 as the only alg however that was in different area (a piece of operating system internals that need to last tens of years without disrupting backward compatibility).

src/org/opensolaris/opengrok/analysis/StreamSource.java
vladak: Still, could you maybe do some quick benchmark (say with thousands of files changed) to see if the performance impact is indeed negligible ?

idodeclare: It's imperceptible. The best wipe/reindex of three for `feature/doc_sha` (f0ad75794):

```
2018-02-20 08:32:43.462-0600 FINE t40 PendingFileCompleter.complete: renamed 1,329 file(s)
2018-02-20 08:32:43.462-0600 FINE t40 IndexDatabase.finishWriting: completed 1,329 file(s)
2018-02-20 08:32:43.464-0600 INFO t40 IndexDatabase.optimize: Optimizing the index ... 
2018-02-20 08:32:44.572-0600 INFO t40 IndexDatabase.optimize: done
...
2018-02-20 08:32:44.583-0600 INFO t1 Statistics.report: Done ... (took 27.869 seconds)
```

... was a quarter-second faster than for `master` (efc7f934900):

```
2018-02-20 08:29:47.553-0600 FINE t40 PendingFileCompleter.complete: renamed 1,329 file(s)
2018-02-20 08:29:47.553-0600 FINE t40 IndexDatabase.finishWriting: completed 1,329 file(s)
2018-02-20 08:29:47.555-0600 INFO t40 IndexDatabase.optimize: Optimizing the index ... 
2018-02-20 08:29:48.882-0600 INFO t40 IndexDatabase.optimize: done
...
2018-02-20 08:29:48.896-0600 INFO t1 Statistics.report: Done ... (took 28.173 seconds)
```

vladak:
Leaving the implementation details aside, could you maybe explain how exactly will the digest be used esp. w.r.t. the UnifiedHighlighter ? Do I understand it correctly that the indexer will store the digest in the document and when search is performed the file will be hashed again to figure out if it is consistent with the index ?

idodeclare:
> when search is performed the file will be hashed again ...

Almost — but actually when search *results* are being presented, the source text is read again to provide it to `UnifiedHighlighter` for synthesizing with Lucene postings (i.e. offsets and positions) for highlighting. The digest is recalculated during this re-read to affirm consistency with the index, as you summarized.

(The current `Context`/`PlainLineTokenizer` also re-reads source text to present search results, but `PlainLineTokenizer` also re-analyzes the text to determine highlights. This current re-analysis is not truly consistent with the original indexing, particularly with complex queries; thus installing `UnifiedHighlighter` will resolve many presentation problems such as issue #51).

An OpenGrok Lucene `Document` is associated with a source file via a constant `QueryBuilder.PATH` value, which is why I am arguing that it is unreasonable to conceive that SHA-256 will not suffice in perpetuity for this specific purpose.

tarzanek:
btw. could the timestamp be used as an alternative to hashing?
after all we use timestamps to detect if the file needs to be reindexed as well, so perhaps the highlighter could use it as a less costly operation than hash

idodeclare:
> btw. could the timestamp be used as an alternative to hashing?

I think you are right, Lubos, that timestamp may suffice as a proxy for identity within `SRC_ROOT`. When I was incorporating `UnifiedHighlighter` the wording regarding identity of source text is stern enough that I immediately incorporated the SHA-256.

(It's a bit different from how timestamp is used in indexing [or in e.g. source code builds] where timestamp is used not as an identity-proxy but as a broad signal to redo all work w.r.t. a document regardless of its content.)

tarzanek:
(well I'd still think this PR would be useful for later index verification purposes or some sanity check ... I guess we need some tag for PRs that should be kept handy for later ... )

PR #2022, opened 2018-02-21
Fix to call createSlider() with revised signature

idodeclare:
Hello,

Please consider for integration this patch to use in `history.jsp` too the revised version of `createSlider()` which was changed in c29e0463794e (Lucene 7).

I'm not sure if it's a Tomcat weirdness, but after deployment I was seeing `NoSuchMethodError` related to `createSlider()` whose `size` parameter was changed from `int` to `long`.

This now matches the change already made to `search.jsp`.

Thank you.

### DIFF
--- a/web/history.jsp
+++ b/web/history.jsp
@@ -19,8 +19,8 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
-
 Portions Copyright 2011 Jens Elkner.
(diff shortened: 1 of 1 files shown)


tarzanek:
thank you, this is weird, I am sure I've built the web app, too !

tarzanek:
so this means that the build for webapp doesn't properly check the deps... weird, this should be investigated and fixed ...

tarzanek:
I filed #2023 ... it seems autoboxing was in play and tomcat had problems redeploying, anyway the fix is a good thing

thank you

PR #2031, opened 2018-03-05
add pep8 checks

vladak:
On Travis this currently passes because the Ubuntu there has much older pep8 version. On my Ubuntu laptop this fails on `tools/sync/test/test_commands.py`. I will leave it there as a canary.

This also disables the `install` phase that triggered `./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` in Travis which is most probably not needed as everything is done in the `script:` section (the `mvnw` is good candidate for removal anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ cache:
   - $HOME/.m2
 before_install:
  - sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


tarzanek:
thank you, merging

PR #2034, opened 2018-03-06
Fix search matching first line has a line number 0

### LINKED ISSUE #2033: Search matching first line of a file has a link to line 0 instead of 1
You can see the problem on the attached screenshot.

screen shot 2018-03-06 at 09 20 39 2

Orviss:
fixes #2033 

Testing done: manual verification

Other lines were working correctly because they were repaired by this line: `markedLine = yyline+1;`

I hope I did not break anything else.

Thanks for review :)

### DIFF
--- a/src/org/opensolaris/opengrok/search/context/PlainLineTokenizer.lex
+++ b/src/org/opensolaris/opengrok/search/context/PlainLineTokenizer.lex
@@ -54,11 +54,11 @@ import org.opensolaris.opengrok.analysis.Scopes.Scope;
    * first line part of the matching region.
    */
   private final StringBuilder markedContents = new StringBuilder();
-  int markedPos=0;
-  int curLinePos=0;
-  int matchStart=-1;
-  int markedLine=0;
-  int rest=0;
+  int markedPos = 0;
+  int curLinePos = 0;
+  int matchStart = -1;
+  int markedLine = 1; // lines are indexed from 1
+  int rest = 0;
   boolean wait = false;
   boolean dumpRest = false;
   Writer out;
@@ -115,10 +115,10 @@ import org.opensolaris.opengrok.analysis.Scopes.Scope;
         wait = false;
         dumpRest = false;
         rest = 0;
-        markedPos=0;
-        curLinePos=0;
-        matchStart=-1;
-        markedLine=0;
+        markedPos = 0;
+        curLinePos = 0;
+        matchStart = -1;
+        markedLine = 1;
         yyline = 1;
         this.out = out;
         this.url = url;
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/search/context/PlainLineTokenizer.lex
```diff
@@ -54,11 +54,11 @@ import org.opensolaris.opengrok.analysis.Scopes.Scope;
    * first line part of the matching region.
    */
   private final StringBuilder markedContents = new StringBuilder();
-  int markedPos=0;
-  int curLinePos=0;
-  int matchStart=-1;
-  int markedLine=0;
-  int rest=0;
+  int markedPos = 0;
```
idodeclare: In a project this old — and especially one which you did not found — I think it can be disrespectful to overlay your own style preferences. I.e., IMHO it's wrong to efface personal traces of previous fellows with your own edits that have no effective difference.

The previous version was committed by Trond Norbye in Nov 2006, and possibly written originally by ChandanB — and now it seems changed at a whim for personal stylistic reasons.

It would be different if the edit was not just for style. E.g., if you edited to remove redundant assignment to zero.

tarzanek: guys, seriously, quit it, it's not worth it

fwiw what I'd love to see is to bundle a style with the project (and let e.g. Idea auto format based based on this style)

that said, we have some checkstyle rules in place (which are also obsolete)

and on top of that, what my lame lazy ass is doing is that I usually just use auto format on whole (yay!) document from Netbeans (which can be seen in my commits and yes, it does hinder the readability of changes a bit :( )

but ... let's be honest, since Netbeans is almost dead, we could go for Idea, where I bet 100,-CZK the default styling will be different
(but I'd bet another hundred, that it will be closer and more up to date with java rules from the standard AND from Google (which is defacto the java standard looking at android ecosystem, Sun Micro did really miss the opportunity back then) )

so where is the truth? ;-)
(and yes, I am biased by Google)

tarzanek:
other than that I think we can merge and start the styling debate elsewhere ...
is everybody OK with that progress?

src/org/opensolaris/opengrok/search/context/PlainLineTokenizer.lex
Orviss: You're not betting much :D

tarzanek: :-D
I am a poor guy from a poor country, so we can just play pennys.
But seriously, checkstyle would be good, enforcing it to a degree too ...

PR #2035, opened 2018-03-06
Store TABSIZE in a supplementary document for each IndexDatabase

idodeclare:
Hello,

Please consider for integration this patch to store project tabsize (if defined) at the time a document is indexed.

The source text presented to `UnifiedHighlighter` must accord with the text as it was when the indexing occured. A non-zero tab size results in transformed source text via `ExpandTabsReader`, so any change in tab size must be recognized.

In addition to `UnifiedHighlighter` checking tab size, this patch also updates `IndexDatabase` so that it detects a changed tab size in order to reindex files automatically. (This fields-check is augmented in later patches [e16d4b92e351] with additional versioning metadata for analyzers and for `AnalyzerGuru` itself so that re-indexing can occur automatically *as necessary* when OpenGrok is updated.)

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/index/IndexDatabase.java
+++ b/src/org/opensolaris/opengrok/index/IndexDatabase.java
@@ -989,8 +989,8 @@ public class IndexDatabase {
                          * If the file was not modified, probably skip to the
                          * next one.
                          */
-                        if (uidIter != null && uidIter.term() != null &&
-                            uidIter.term().bytesEquals(buid)) {
+                        if (uidIter != null && uidIter.term() != null
+                                && uidIter.term().bytesEquals(buid)) {
                             boolean chkres = chkFields(file, path);
                             if (!chkres) removeFile(false);
 
@@ -1522,8 +1522,8 @@ public class IndexDatabase {
              * ignore the check so that no extra work is done. After a re-index,
              * the TABSIZE check will be active.
              */
-            int reqTabSize = project != null ? project.hasTabSizeSetting() ?
-                project.getTabSize() : 0 : 0;
+            int reqTabSize = project != null && project.hasTabSizeSetting() ?
+                project.getTabSize() : 0;
             IndexableField tbsz = doc.getField(QueryBuilder.TABSIZE);
             int tbszint = tbsz != null ? tbsz.numericValue().intValue(): 0;
             if (tbsz != null && tbszint != reqTabSize) {
--- a/src/org/opensolaris/opengrok/index/IndexDatabase.java
+++ b/src/org/opensolaris/opengrok/index/IndexDatabase.java
@@ -61,12 +61,10 @@ import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.IndexWriterConfig.OpenMode;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.index.MultiFields;
-import org.apache.lucene.index.PostingsEnum;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.queryparser.classic.ParseException;
-import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TopDocs;
@@ -108,17 +106,15 @@ public class IndexDatabase {
     private static final Comparator<File> FILENAME_COMPARATOR =
         (File p1, File p2) -> p1.getName().compareTo(p2.getName());
 
-    private static final Set<String> CHECK_FIELDS;
-
     private final Object INSTANCE_LOCK = new Object();
 
     private Project project;
     private FSDirectory indexDirectory;
     private IndexReader reader;
     private IndexWriter writer;
+    private IndexAnalysisSettings settings;
     private PendingFileCompleter completer;
     private TermsEnum uidIter;
-    private PostingsEnum postsIter;
     private IgnoredNames ignoredNames;
     private Filter includedNames;
     private AnalyzerGuru analyzerGuru;
@@ -159,11 +155,6 @@ public class IndexDatabase {
         this.project = project;
         lockfact = NoLockFactory.INSTANCE;
         initialize();
-    }
-
-    static {
-        CHECK_FIELDS = new HashSet<>();
-        CHECK_FIELDS.add(QueryBuilder.TABSIZE);
     }
 
     /**
@@ -395,8 +386,8 @@ public class IndexDatabase {
 
         reader = null;
         writer = null;
+        settings = null;
         uidIter = null;
-        postsIter = null;
 
         IOException finishingException = null;
         try {
@@ -439,6 +430,10 @@ public class IndexDatabase {
 
                 String startuid = Util.path2uid(dir, "");
                 reader = DirectoryReader.open(indexDirectory); // open existing index
+                settings = readAnalysisSettings();
+                if (settings == null) {
+                    settings = new IndexAnalysisSettings();
+                }
                 Terms terms = null;
                 int numDocs = reader.numDocs();
                 if (numDocs > 0) {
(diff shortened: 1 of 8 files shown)


src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -969,14 +985,19 @@ private void indexDown(File dir, String parent, IndexDownArgs args)
...
 
-                        // If the file was not modified, skip to the next one.
-                        if (uidIter != null && uidIter.term() != null
-                                && uidIter.term().bytesEquals(buid)) {
-                            BytesRef next = uidIter.next(); // keep matching docs
-                            if (next == null) {
-                                uidIter = null;
-                            }
-                            continue;
+                        /**
```
Orviss: Just a silly thing, why are you using javadoc style of comments instead of ordinary ones?

I believe it would be better to use simple comments because:
```
/*
 *
 */
```

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -969,14 +985,19 @@ private void indexDown(File dir, String parent, IndexDownArgs args)
...
+                         * If the file was not modified, probably skip to the
+                         * next one.
+                         */
+                        if (uidIter != null && uidIter.term() != null &&
+                            uidIter.term().bytesEquals(buid)) {
```
Orviss: Wrong indentation, code in condition might be mistaken for the code inside it.

It should be:
```java
if (uidIter != null && uidIter.term() != null && 
        uidIter.term().bytesEquals(buid)) {
```

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -969,14 +985,19 @@ private void indexDown(File dir, String parent, IndexDownArgs args)
...
+                            boolean chkres = chkFields(file, path);
+                            if (!chkres) removeFile(false);
```
Orviss: Missing curly braces after `if`, it is a good practice to use them always. Good code:

```
if (!chkres) {
    removeFile(false);
}
```

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -969,14 +985,19 @@ private void indexDown(File dir, String parent, IndexDownArgs args)
...
+
+                            BytesRef next = uidIter.next();
+                            if (next == null) uidIter = null;
```
Orviss: Missing braces.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -969,14 +985,19 @@ private void indexDown(File dir, String parent, IndexDownArgs args)
...
+
+                            if (chkres) continue; // keep matching docs
```
Orviss: Missing braces.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -1483,6 +1504,44 @@ private void finishWriting() throws IOException {
...
+        while (postsIter.nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
+            ++n;
+            // Read a limited-fields version of the document.
+            Document doc = reader.document(postsIter.docID(), CHECK_FIELDS);
+            if (doc == null) {
+                LOGGER.log(Level.FINER, "No Document: {0}", path);
+                continue;
+            }
+
+            /**
```
Orviss: Please use ordinary comments.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -1483,6 +1504,44 @@ private void finishWriting() throws IOException {
...
+             * Verify TABSIZE, or return a value to indicate mismatch.
+             * For an older OpenGrok index that does not yet have TABSIZE,
+             * ignore the check so that no extra work is done. After a re-index,
+             * the TABSIZE check will be active.
+             */
+            int reqTabSize = project != null ? project.hasTabSizeSetting() ?
```
Orviss: Nested ternary operator, it's very bad for readability, maybe rewrite like this?

```java
 project != null && project.hasTabSizeSetting() ? project.getTabSize() : 0;
```

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: I am not using Javadoc style of comments, I am using a begin-comment delimiter with two asterisks. NetBeans conveniently auto-formats that, so I'll stick to that.

Orviss:
I went over the code very quickly, I have writtencomments to some parts that I do not like and checkstyle would probably pick (we do have to set it up properly).

Hope you're not mad about the comments, I know some of them are quite silly but I still like to do it properly :) 

I will look at the code more closely tomorrow, I'm super tired right now...

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: I disagree that it's bad practice for terse, single-line if-statements. This is a style discussion the project will have to have.

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: This kind of review feedback has zero value.

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: fixed

src/org/opensolaris/opengrok/index/IndexDatabase.java
idodeclare: fixed

tarzanek:
:-D
glad you guys had fun

tarzanek:
I actually have a very stupid Q ... @kahatlen might chime in here ...
but originally the tabsize was per whole opengrok
THEN we decided it will be per project (and I hope it is now, not 100% sure ;) )
and now we will save it per document ...

So what is the sweet spot?
(I said it will be a stupid Q ;) )

tarzanek:
after reading #2036 I can see that I am really asking a stupid Q :-D

idodeclare:
> very stupid Q .... the ideal to fit all is that we will save it per document .... sweet spot?

No, Lubos, your instinct is right. It's just that it's ultra-convenient to have this metadata transactionally stored in Lucene, but up to now OpenGrok hasn't stored any aggregate or summary data in the index; i.e. right now every `Document` is one-to-one w.r.t. a `SRC_ROOT` file. Storing a single project tab size would necessarily be in a second instantiation of `Document` for a project/index summary or "top-level" data.

I have a draft implementation of something similar I was working on to store another type of orthogonal data: aggregate Line# and LOC counts for directories as a follow on to PR #1987 and mentioned on that PR.

For this current PR, the idea is easier because a new instantiation of `Document` for tabsize (and conveniently as well for the upcoming `ANALYZER_GURU_VERSION` and `ANALYZER_VERSIONS` map in 2513c431f907b04c5) does not have to share any fields at all with any current `SRC_ROOT` `Document`. I.e. it's easier to affirm that the presence of a new `Document` for project top-level data won't mess up the existing code search. Also, in this case, it's just a blob of metadata and not truly indexed as is required for the Line#/LOC aggregate counts.

Let me repurpose my other draft, push, and get thoughts.

tarzanek:
btw. I actually like the idea of this PR, so let me walk over it (I am especially curious how you detect the tab size per document)

idodeclare:
> detect the tab size per document

No detection — just store and compare

idodeclare:
> let me walk over it

OK let me know when done, and I'll push the revision.

tarzanek:
feel free to push the revision, the { } brackets would be good to have it's just good code hygiene (and after all most IDEs do it automagically)
the comment is OK with me
I will enable checkstyle in travis, I think this is the best approach to such nitpicks (and yes, I do hate it too when my builds in work fail me on silly things such as unused imports or missing {} :-D )

idodeclare:
OK thank you for the heads-up.

tarzanek:
sorry for a late reply, let's merge this ...

Issue #1374, opened 2017-01-31
extra spaces after pointers (asterisks) in the "Scopes Window"


jfriedel:
We could consider removing the extra blanks between pointer sign (asterisk) and the argument name in the "Scopes Window".

![pointers](https://cloud.githubusercontent.com/assets/5879789/22473847/6c1bd3cc-e7da-11e6-995d-f15feda03bfe.jpg)

vladak:
I think that's sort of controversial. Some users prefer to have the argument names there.

vladak:
For the record, this is already present in the generated code:

```
<span class='scope-signature'>mkcert(X509 ** x509p,EVP_PKEY ** pkeyp,int bits,int serial,int days)</span>
```

mrate:
I was unable to reproduce it on my machine using universal-ctags and exuberant-ctags (in both cases mkcert signature in scope window looks good). The content of <span class='scope-signature'> is taken directly from ctags output.

idodeclare:
I confirm this behavior of ctags (as recently as universal-ctags/ctags@1620c633). I encountered it while working on my feature for OpenGrok to use the current Lucene `UnifiedHighlighter` which demands postings (i.e., `DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS`).

For DEFS, I needed to cross-reference ctags's output with source, and this spacing change was an obstacle to work around.

Anyway, after that feature is PRed, probably we can enhance `CtagsReader` so that `Definitions.Tag.signature` reflects true source w.r.t. asterisks and includes a single space after a comma (regardless of true source).

PR #2036, opened 2018-03-06
DEFS stores true offsets from enhanced CtagsReader

idodeclare:
Hello,

Please consider for integration this patch to enhance `CtagsReader` to store true posting for `DEFS` for use by the `UnifiedHighlighter` — i.e. so that a document is not required to be re-analyzed to present `DEFS` search matches.

Because of limitations in ctags results (e.g. partial patterns or truncated patterns) and limitations in the ctags format itself (e.g., having line numbering but no character offsets as required for Lucene postings), `CtagsReader` is enhanced to supplement the tags *when necessary* with additional analysis of the source text.

E.g., in `sample.cc`/`sampletags_cc`, the function definition (excluding body) for `definitionContains()` spans seven lines in source code:
```
int
Ancestor::definitionContains(const fru_regdef_t *def,
				const fru_regdef_t *parent_def,
				Str element,
				uint32_t offset,
				Ancestor *ant,
				Str path)
```
 but Universal Ctags only produces a pattern for one line of the definition:
```
/^Ancestor::definitionContains(const fru_regdef_t *def,$/;"
```
and ctags presents a `signature` with modified cstyle (note the different whitespace around asterisks and commas):
```
(const fru_regdef_t * def,const fru_regdef_t * parent_def,Str element,uint32_t offset,Ancestor * ant,Str path)
```

`CtagsReader` is enhanced to analyze the text in various ways to arrive at the best Lucene postings given the ctags variances.

`CtagsReader` is also enhanced to be project tabsize-aware or else postings would be mistakenly shifted when a non-zero project tab size is defined.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/analysis/Ctags.java
+++ b/src/org/opensolaris/opengrok/analysis/Ctags.java
@@ -24,6 +24,7 @@
 package org.opensolaris.opengrok.analysis;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -38,13 +39,14 @@ import java.util.logging.Logger;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.util.IOUtils;
+import org.opensolaris.opengrok.util.SourceSplitter;
 
 /**
  * Provides Ctags by having a running subprocess of ctags.
  *
  * @author Chandan
  */
-public class Ctags {
+public class Ctags implements Resettable {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(Ctags.class);
 
@@ -57,6 +59,7 @@ public class Ctags {
     //default: setCtags(System.getProperty("org.opensolaris.opengrok.analysis.Ctags", "ctags"));
     private String binary;
     private String CTagsExtraOptionsFile = null;
+    private int tabSize;
 
     private boolean junit_testing = false;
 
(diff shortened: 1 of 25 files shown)


vladak:
I wonder if this will fix issues like  #1374.

Dne 6. 3. 2018 9:01 odp. napsal uživatel "C Fraire" <
notifications@github.com>:

idodeclare:
> I wonder if this will fix issues like  #1374.

I wrote on that issue last month that this PR will provide an avenue to address the problem.

tarzanek:
thank you Chris, this is a good start, I have finally read it, 
I am not sure on few things (like if we need all those tokenstream wrappers, I was under the impression lucene is trying to simplify this and get rid of custom Tokenstreams, but in our case this is a bit of problem)
let's merge it
I have also some concerns about performance impact (both with tabs), but then all the info we gain is vital, so let's give it a shot

thank you
L

idodeclare:
Thank you for the review, Lubos.

> if we need all those tokenstream wrappers, I was under the impression lucene is trying to simplify this and get rid of custom Tokenstreams

I am not sure if Lucene will get rid of custom token streams or make it even easier to use `Tokenizer`-based classes (e.g. `JFlexTokenizer`) that take `Reader` instances as input. For a token enumeration that uses alternate source data (as for this patch that sources ctags `Definitions`), I still think from studying the Lucene API docs that a custom `TokenStream` subclass is appropriate.

As for performance, I am concerned as well, but the only optimization path short of replacing ctags is external to OpenGrok: i.e. enhancing Universal ctags to augment the tags data to support more granular and more comprehensive metadata in tags (namely character offsets, full multi-line patterns, and perhaps "`true_signature`" that reflects actual source code).

Anyway, I am glad because having this and #2035 merged allows me to raise the PR for the incorporation of `UnifiedHighlighter`.

idodeclare:
(BTW I have two small concerns with Github "Rebase and merge": 1] the commits go in without log text referring to the original PR#; and 2] for a multi-commit branch, it would make any git-revert [thankfully rare] more difficult since there would not be a single merge-commit to revert.)

tarzanek:
AH!
sorry for that, I will try to avoid rebase and merge, my fault
github recently switched the merge UI , so I will try to find a better way (resp. stick to original way which preserved everything) , it might involve a bit of testing, so I hope you don't mind me using you as guinea pig ;)

tarzanek:
regarding universal ctags - I think those guys do a very good job, it makes sense to file issues with them, they are very responsive and helpful

PR #2038, opened 2018-03-10
Fix comment

idodeclare:
Hello,

Please consider for integration this patch with a simple comment fix.

Thank you.


tarzanek:
thnx, merging (resp. trying a rebase and merge, let's see how github will handle this)

PR #2039, opened 2018-03-10
Use en dash for Results count. Fix to specify UTF-8 for jsps.

idodeclare:
Hello,

Please consider for integration this patch to use an en-dash in the results item range instead of a hyphen.

With this tiny revision, it became apparent that many jsps did not have their encoding set to UTF-8. It's required or else — in the presence of a non-Latin code point — a conversion of content to ISO-8859-1 is warned and then executed. That happens even if a jsp has an `include` of a file specifying UTF-8 (e.g., `web/httpheader.jspf`).

Thank you.


tarzanek:
thank you, let's have a look at utf-8 , merging

PR #2040, opened 2018-03-10
Log some additional places where !accept()ed

idodeclare:
Hello,

Please consider for integration this patch with a minor, FINER logging of some files that aren't `accept()`ed.

Thank you.


tarzanek:
merging, thank you

PR #2043, opened 2018-03-15
Properly set LOGGER_CONF_SOURCE for binary distributions

sjaensch:
I'm seeing a warning when using the OpenGrok script to start an index run using opengrok 1.1-rc21:

    Loading the default instance configuration ...
      Creating default /tmp/opengrok/logging.properties ...
    WARNING: Can't find distribution logging configuration (/nail/home/sjaensch/pg/services/opengrok/opengrok-1.1-rc21/bin/../lib/../doc/) to install as default logging configuration (/tmp/opengrok/logging.properties)

LOGGER_CONF_SOURCE is supposed to point to the logging.properties file. This simple PR fixes the issue.

I do accept the terms of the OCA.

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -221,7 +221,7 @@ DefaultInstanceConfiguration()
         else
             # Handle Binary Distributions
             OPENGROK_DISTRIBUTION_BASE="${SCRIPT_DIRECTORY}/../lib"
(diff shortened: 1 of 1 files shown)


vladak:
Thanks !

PR #2044, opened 2018-03-17
Enable conditional test runs by default

Orviss:
I think it might be counterintuitive if `mvn clean test` command fails on a completely correct repository just because some SCM tools are not installed. User can skip uninstalled SCM tools by adding `-Djunit-force-all=false` but this option is mentioned at the very bottom of the `README.md` file which makes it very easy to miss. (I for one installed BitKeeper and CVS to make the tests working and did not know about this option until I started to investigate it further...)

Therefore, I think it might be wise to skip some of those tests by default and check them all via Travis.

Thanks for considering this PR :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,10 @@ env:
    #   via the "travis encrypt" command using the project repo's public key
    - secure: "O_cda5pWDBAP-O3_0nG5RQ"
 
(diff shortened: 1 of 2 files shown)


vladak:
I was bitten by this too, so +1.

tarzanek:
cool thank you sir! merging L

PR #2045, opened 2018-03-20
Remove redundant changeRequests field from HistoryEntry

Orviss:
Removed `changeRequests` field from `HistoryEntry`. This field was not used at all. Some data were inserted from `Razor` SCM but they were not consumed.

Thanks for considering this PR :)


tarzanek:
if it is not consumed/used at all, then let's clean up, thank you!

PR #2048, opened 2018-03-26
introduce projadm script for automating project data management

vladak:
Simple wrapper script of couple of steps used internally for project management.


tarzanek:
lgtm, please merge :)

Issue #673, opened 2013-10-17
`/more/` should include context for the matching lines


cnst:
Since the `/more/` page is already somewhat relaxed, providing more in-context results compared to the original search results from which it is linked from, it might as well include some context to the lines found &mdash; perhaps a standard 3-lines before, 3-lines after, would nice.

For example, in the nginx source, I want to figure out what diagnostic messages are available with each log level, for <a href="http://nginx.org/r/error_log">`error_log`</a>.  However, `/more/trunk/src/http/ngx_http_request.c?refs=ngx_log_error` page only lists the actual levels, but doesn't list the text message of the warnings, making the grokking difficult.

Issue #638, opened 2013-06-10
Improve highlighter code, ev. get rid of summarizer


tarzanek:
http://blog.mikemccandless.com/2012/12/a-new-lucene-highlighter-is-born.html

PR #2052, opened 2018-03-28
Feature/uhighlighter

idodeclare:
Hello,

Please consider for integration this patch that integrates the Lucene `UnifiedHighlighter` into OpenGrok's `Context` [see `getContext2()`] while preserving for now the original `getContext()` for falling back when `UnifiedHighlighter` cannot be used (e.g. when source text has been updated via `git-update` but Lucene `Document` postings are not yet updated).

`more.jsp` and `search.jsp` both use of `getContext2()` so their presentations are fully consistent.

OpenGrok's Lucene indexing settings are modified (see `OGKTextField`) to enable storage of "postings" (i.e., `IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS`) for use by `UnifiedHighlighter` so that documents do not have to be re-lexed (as done by `PlainLineTokenizer`) in order to be highlighted. My testing with some large repos showed only a tiny increase in index size because of this change.

Additionally, for DEFS, tokens must now be indexed with stored term vectors (see `OGKTextVecField`) or else my testing showed that DEFS multi-term queries (MTQs e.g., prefixes or wildcards) did not operate correctly. This is due to Lucene `UnifiedHighlighter` currently falling back to `ANALYSIS` in the presence of MTQs when term vectors are not available. For DEFS, re-analysis from source text does not accord whatsoever with ctags `Definitions`; but the presence of term vectors makes this work perfectly. (See my long comment in `OGKUnifiedHighlighter` regarding this issue.)

Since DEFS are a small subset of index data relative to FULL and REFS fields, my testing showed again only a tiny increase in index size due to storage of DEFS term vectors.

This patch also enables optionally the display of additional surrounding lines for each match (as requested in issues #673 and #638). `Configuration` `<contextSurround/>` (default 0) specifies the number of lines of leading and trailing context to display surrounding each match.

A new `Configuration` `<contextLimit/>` (default 10) specifies the total number of context lines to display for each matching file.

Please see the following new classes that collaborate to produce OpenGrok's style of context under `UnifiedHighlighter`'s coordination:

* `StrictLineBreakIterator` : represents a subclass of Java `BreakIterator` that breaks at standard OpenGrok EOL -- namely `\r\n`, `\n`, or `\r`, which is needed since OpenGrok's context is always w.r.t. whole source-code lines (which may be truncated to fit within the context "window").
* `PhraseHighlight`: represents a highlighted phrase within a line.
* `LineHighlight`: represents a collection of metadata related to highlighting a single line of code.
* `PassageConverter`: represents an object that can translate Lucene `Passage` instances into indexed `LineHighlight` instances, taking into account a configurable number of leading and trailing lines of context for each match.
* `ContextFormatter`: represents a subclass of Lucene's `PassageFormatter` that uses OpenGrok's `PassageConverter`.
* `FormattedLines`:  represents structured results from `ContextFormatter` that can be merged with other instances. `UnifiedHighlighter` returns results separated by field, and `OGKUnifiedHighlighter` merges them together to return a coherent result for OpenGrok presentation.
* `OGKUnifiedHighlighter`: represents a subclass of Lucene `UnifiedHighlighter` with customizations for OpenGrok.

There are new tests of this functionality. At the highest level, new `SearchAndContextFormatterTest` tests the indexing, search, and format of the results. (There are many tests at lower levels.)

Thank you.

### DIFF
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -478,6 +478,22 @@ Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-highlighter</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-join</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-memory</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queries</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-queryparser</artifactId>
         </dependency>
         <dependency>
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -19,7 +19,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
-Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
 
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,26 @@ Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
             </dependency>
             <dependency>
                 <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-highlighter</artifactId>
+                <version>${lucene.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-join</artifactId>
+                <version>${lucene.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-memory</artifactId>
+                <version>${lucene.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-queries</artifactId>
+                <version>${lucene.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
                 <artifactId>lucene-queryparser</artifactId>
                 <version>${lucene.version}</version>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
+Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
 
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -213,6 +213,13 @@ public class AnalyzerGuru {
 
     private static final Map<String, String> fileTypeDescriptions = new TreeMap<>();
 
+    /**
+     * Maps from {@link FileAnalyzer#getFileTypeName()} to
+     * {@link FileAnalyzerFactory}
+     */
+    private static final Map<String, FileAnalyzerFactory> FILETYPE_FACTORIES =
+            new HashMap<>();
+
     /*
      * If you write your own analyzer please register it here. The order is
      * important for any factory that uses a FileAnalyzerFactory.Matcher
@@ -338,6 +345,9 @@ public class AnalyzerGuru {
         }
         matchers.addAll(factory.getMatchers());
         factories.add(factory);
+
+        FileAnalyzer fa = factory.getAnalyzer();
+        FILETYPE_FACTORIES.put(fa.getFileTypeName(), factory);
     }
 
     /**
@@ -383,6 +393,17 @@ public class AnalyzerGuru {
      */
     public static FileAnalyzer getAnalyzer() {
         return DEFAULT_ANALYZER_FACTORY.getAnalyzer();
+    }
+
+    /**
+     * Gets an analyzer for the specified {@code fileTypeName} if it accords
+     * with a known {@link FileAnalyzer#getFileTypeName()}.
+     * @param fileTypeName a defined name
+     * @return a defined instance if known or otherwise {@code null}
+     */
+    public static FileAnalyzer getAnalyzer(String fileTypeName) {
+        FileAnalyzerFactory factory = FILETYPE_FACTORIES.get(fileTypeName);
+        return factory == null ? null : factory.getAnalyzer();
     }
 
     /**
--- a/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
+++ b/src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis;
 
@@ -531,6 +531,9 @@ public class AnalyzerGuru {
         args.setProject(project);
 
         FileAnalyzer analyzer = factory.getAnalyzer();
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        analyzer.setScopesEnabled(env.isScopesEnabled());
+        analyzer.setFoldingEnabled(env.isFoldingEnabled());
         analyzer.writeXref(args);
     }
 
--- a/src/org/opensolaris/opengrok/analysis/CtagsReader.java
+++ b/src/org/opensolaris/opengrok/analysis/CtagsReader.java
@@ -378,8 +378,8 @@ public class CtagsReader {
      * Adds a tag to a {@code Definitions} instance.
      */
     private void addTag(Definitions defs, int lineno, String symbol,
-        String type, String text, String namespace, String signature,
-        int lineStart, int lineEnd) {
+            String type, String text, String namespace, String signature,
+            int lineStart, int lineEnd) {
         // The strings are frequently repeated (a symbol can be used in
         // multiple definitions, multiple definitions can have the same type,
         // one line can contain multiple definitions). Intern them to minimize
@@ -397,9 +397,7 @@ public class CtagsReader {
      * syntax.
      * @return a defined instance
      */
-    private CpatIndex bestIndexOfTag(int lineno, String whole,
-        String str) {
-
+    private CpatIndex bestIndexOfTag(int lineno, String whole, String str) {
         if (whole.length() < 1) {
             return new CpatIndex(lineno, 0, 1, true);
         }
--- a/src/org/opensolaris/opengrok/analysis/OGKTextField.java
+++ b/src/org/opensolaris/opengrok/analysis/OGKTextField.java
@@ -1,0 +1,82 @@
+/*
+ * This work is licensed under the Creative Commons Attribution-ShareAlike 3.0
+ * United States License. To view a copy of this license, visit
+ * http://creativecommons.org/licenses/by-sa/3.0/us/ or send a letter to
+ * Creative Commons, PO Box 1866, Mountain View, CA 94042, USA.
+ *
+ * Copyright (c) 2012, https://stackoverflow.com/users/1270457/amas
+ * Copyright (c) 2012, https://stackoverflow.com/questions/11945728/how-to-use-termvector-lucene-4-0
+ * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
+ */
+
+package org.opensolaris.opengrok.analysis;
+
+import java.io.Reader;
+import org.apache.lucene.analysis.TokenStream;
+import org.apache.lucene.document.Field;
+import org.apache.lucene.document.FieldType;
+import org.apache.lucene.index.IndexOptions;
+
+/**
+ * Represents an OpenGrok-customized tokenized text field to centralize settings
+ * across all the analyzers.
+ */
+public class OGKTextField extends Field {
+
+    /** Indexed, tokenized, not stored. */
+    public static final FieldType TYPE_NOT_STORED = new FieldType();
+
+    /** Indexed, tokenized, stored. */
+    public static final FieldType TYPE_STORED = new FieldType();
+
+    static {
+        TYPE_NOT_STORED.setIndexOptions(
+            IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS);
+        TYPE_NOT_STORED.setTokenized(true);
+        TYPE_NOT_STORED.freeze();
+
+        TYPE_STORED.setIndexOptions(
+            IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS);
+        TYPE_STORED.setTokenized(true);
+        TYPE_STORED.setStored(true);
+        TYPE_STORED.freeze();
+    }
+
+    /**
+     * Creates a new un-stored instance with Reader value.
+     * @param name
+     * @param reader
+     */
+    public OGKTextField(String name, Reader reader) {
+        super(name, reader, TYPE_NOT_STORED);
+    }
+
+    /**
+     * Creates a new instance with {@link Reader} value.
+     * @param name
+     * @param reader
+     * @param store
+     */
+    public OGKTextField(String name, Reader reader, Store store) {
+        super(name, reader, store == Store.YES ? TYPE_STORED : TYPE_NOT_STORED);
+    }
+
+    /**
+     * Creates a new instance with {@code String} value.
+     * @param name
+     * @param value
+     * @param store
+     */
+    public OGKTextField(String name, String value, Store store) {
+        super(name, value, store == Store.YES ? TYPE_STORED : TYPE_NOT_STORED);
+    }
+
+    /**
+     * Creates a new un-stored instance with {@link TokenStream} value.
+     * @param name
+     * @param stream
+     */
+    public OGKTextField(String name, TokenStream stream) {
+        super(name, stream, TYPE_NOT_STORED);
+    }
+}
--- a/src/org/opensolaris/opengrok/analysis/OGKTextField.java
+++ b/src/org/opensolaris/opengrok/analysis/OGKTextField.java
@@ -43,9 +43,7 @@ public class OGKTextField extends Field {
     }
 
     /**
-     * Creates a new un-stored instance with Reader value.
-     * @param name
-     * @param reader
+     * Creates a new un-stored instance with {@link Reader} value.
      */
     public OGKTextField(String name, Reader reader) {
         super(name, reader, TYPE_NOT_STORED);
@@ -53,9 +51,6 @@ public class OGKTextField extends Field {
 
     /**
      * Creates a new instance with {@link Reader} value.
-     * @param name
-     * @param reader
-     * @param store
      */
     public OGKTextField(String name, Reader reader, Store store) {
         super(name, reader, store == Store.YES ? TYPE_STORED : TYPE_NOT_STORED);
@@ -63,9 +58,6 @@ public class OGKTextField extends Field {
 
     /**
      * Creates a new instance with {@code String} value.
-     * @param name
-     * @param value
-     * @param store
      */
     public OGKTextField(String name, String value, Store store) {
         super(name, value, store == Store.YES ? TYPE_STORED : TYPE_NOT_STORED);
@@ -73,8 +65,6 @@ public class OGKTextField extends Field {
 
     /**
      * Creates a new un-stored instance with {@link TokenStream} value.
-     * @param name
-     * @param stream
      */
     public OGKTextField(String name, TokenStream stream) {
         super(name, stream, TYPE_NOT_STORED);
--- a/src/org/opensolaris/opengrok/analysis/archive/TarAnalyzer.java
+++ b/src/org/opensolaris/opengrok/analysis/archive/TarAnalyzer.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis.archive;
 
@@ -26,13 +27,14 @@ import java.io.IOException;
 import java.io.Writer;
 import java.util.ArrayList;
 import org.apache.lucene.document.Document;
-import org.apache.lucene.document.TextField;
 import org.apache.tools.tar.TarEntry;
 import org.apache.tools.tar.TarInputStream;
 import org.opensolaris.opengrok.analysis.FileAnalyzer;
 import org.opensolaris.opengrok.analysis.FileAnalyzerFactory;
 import org.opensolaris.opengrok.analysis.IteratorReader;
 import org.opensolaris.opengrok.analysis.StreamSource;
+import org.opensolaris.opengrok.analysis.OGKTextField;
+import org.opensolaris.opengrok.search.QueryBuilder;
 import org.opensolaris.opengrok.web.Util;
 
 /**
@@ -62,6 +64,6 @@ public class TarAnalyzer extends FileAnalyzer {
             }
         }
 
-        doc.add(new TextField("full", new IteratorReader(names)));
+        doc.add(new OGKTextField(QueryBuilder.FULL, new IteratorReader(names)));
     }
 }
--- a/src/org/opensolaris/opengrok/analysis/archive/ZipAnalyzer.java
+++ b/src/org/opensolaris/opengrok/analysis/archive/ZipAnalyzer.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis.archive;
 
@@ -28,11 +29,12 @@ import java.util.ArrayList;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import org.apache.lucene.document.Document;
-import org.apache.lucene.document.TextField;
 import org.opensolaris.opengrok.analysis.FileAnalyzer;
 import org.opensolaris.opengrok.analysis.FileAnalyzerFactory;
 import org.opensolaris.opengrok.analysis.IteratorReader;
 import org.opensolaris.opengrok.analysis.StreamSource;
+import org.opensolaris.opengrok.analysis.OGKTextField;
+import org.opensolaris.opengrok.search.QueryBuilder;
 import org.opensolaris.opengrok.web.Util;
 
 /**
@@ -62,6 +64,6 @@ public class ZipAnalyzer extends FileAnalyzer {
             }
         }
 
-        doc.add(new TextField("full", new IteratorReader(names)));
+        doc.add(new OGKTextField(QueryBuilder.FULL, new IteratorReader(names)));
     }
 }
--- a/src/org/opensolaris/opengrok/analysis/document/MandocAnalyzer.java
+++ b/src/org/opensolaris/opengrok/analysis/document/MandocAnalyzer.java
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opensolaris.opengrok.analysis.document;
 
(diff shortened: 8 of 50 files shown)


tarzanek:
what is the difference between CDDL and http://creativecommons.org/licenses/by-sa/3.0/us/ ?
(is CC similar like BSD?)

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
```diff
@@ -50,6 +53,9 @@
  */
 public class PlainAnalyzer extends TextAnalyzer {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(
```
tulinkry: unused?

src/org/opensolaris/opengrok/analysis/OGKTextField.java
```diff
@@ -0,0 +1,82 @@
...
+        TYPE_STORED.setIndexOptions(
+            IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS);
+        TYPE_STORED.setTokenized(true);
+        TYPE_STORED.setStored(true);
+        TYPE_STORED.freeze();
+    }
+
+    /**
+     * Creates a new un-stored instance with Reader value.
+     * @param name
```
Orviss: Could you please add comments for the parameters? I know that they are mostly self explaining so you could remove them if you prefer to (that's what I would do). It's just that if they have this format then they do not provide any additional info than the parameters already in code. :)

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
```diff
@@ -134,4 +140,51 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut)
             }
         }
     }
+
+    private void tryAddingDefs(Document doc, Definitions defs, StreamSource src,
+        String fullpath) throws IOException {
```
Orviss: Could you please add 4 additional spaces? The same you do in other places :)

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
```diff
@@ -134,4 +140,51 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut)
...
+    /**
+     * Identical to {@link #getReader(java.io.InputStream)} but overlaying an
+     * existing stream.
+     * @see #getReader(java.io.InputStream)
+     */
+    private Reader wrapReader(Reader reader) {
+        return ExpandTabsReader.wrap(reader, project);
+    }
+
+    private static boolean bytesEqual(byte[] a, byte[] b) {
```
Orviss: Maybe unused? :) If it is used then you can consider using `Arrays.equals()`.

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -395,6 +397,13 @@ public void update(IndexerParallelizer parallelizer)
             IndexWriterConfig iwc = new IndexWriterConfig(analyzer);
             iwc.setOpenMode(OpenMode.CREATE_OR_APPEND);
             iwc.setRAMBufferSizeMB(env.getRamBufferSize());
+            /**
+             * Most data in OpenGrok is indexed but not stored, so use the best
+             * compression on the minority of data that is stored, since it
+             * should not have a detrimental impact on overall throughput.
+             */
+            iwc.setCodec(new Lucene70Codec(
+                Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
```
tarzanek: lucene usually has some "latest" links pointing to latest available codecs and formats, does it have it for those two too?
the idea I have is that upon every lucene upgrade everything (possible) should get upgraded automatically to latest greatest
(I know, downside is that you need to index from scratch and all the bugs get automatically picked up, but engine upgrade is transparent)

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -191,8 +195,12 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
                 out.write("<span class=\"important-note important-note-rounded\" data-messages='" + messages + "'>!</span>");
                 out.write("</a>");
             }
+
+            int tabSize = sh.getTabSize(p);
+
             out.write("</td></tr>");
-            for (Document doc : entry.getValue()) {
+            for (int docId : entry.getValue()) {
```
tarzanek: nice catch and optimalization :) 
thank you

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -263,6 +254,46 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
...
+
+        boolean didPresentNew = searcHelper.sourceContext.getContext2(env,
+            searcHelper.searcher, docId, out, xrefPrefix, morePrefix, true,
+            tabSize);
+
+        if (!didPresentNew) {
+            /**
+             * Fall back to the old view, which re-analyzes text using
+             * PlainLinetokenizer. E.g., when source code is updated (thus
+             * affecting SHA hashes) but re-indexing is not yet complete.
```
tarzanek: old comment? do you use SHA hashes?

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -263,6 +254,46 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
         }
     }
 
+    private static void printPlain(SearchHelper searcHelper,
+        RuntimeEnvironment env, Document doc, int docId, Writer out,
```
Orviss: Could you please consider somehow combining the method parameters? This violates current checkstyle parameter number setting which is set to default value, right now it is `7`.

src/org/opensolaris/opengrok/search/SearchEngine.java
```diff
@@ -377,9 +410,32 @@ private int search(List<Project> projects, File root) {
...
+     * @return a defined instance if a query succeeded, or {@code null}
+     */
+    public ScoreDoc[] scoreDocs() {
+        return hits;
+    }
+
+    /**
+     * Gets the document of the specified {@code docId} from
+     * {@code search(...)} if it was called.
+     * @param docId
```
Orviss: Could you please add comment to the `@param` and `@throws` tags? :) Or remove them if you prefer to.

src/org/opensolaris/opengrok/search/context/Context.java
```diff
@@ -64,34 +73,164 @@
...
+     * {@code query} -- which might be {@code null} and result in
+     * {@link #isEmpty()} returning {@code true}.
      * @param query the query to generate the result for
-     * @param queryStrings map from field names to queries against the fields
+     * @param qbuilder required builder used to create {@code query}
      */
-    public Context(Query query, Map<String, String> queryStrings) {
+    public Context(Query query, QueryBuilder qbuilder) {
+        if (qbuilder == null) {
+            throw new IllegalArgumentException("`qbuilder' is null");
```
Orviss: Maybe use the same apostrophes? :)

src/org/opensolaris/opengrok/search/context/Context.java
```diff
@@ -64,34 +73,164 @@
...
+     * {@link Document}, and output according to the parameters.
+     * @param env required environment
+     * @param searcher required search that produced the document
+     * @param docId document ID for producing context
+     * @param dest required target to write
+     * @param urlPrefix prefix for links
+     * @param morePrefix optional link to more... page
+     * @param limit a value indicating if the number of matching lines should be
+     * limited. N.b. unlike
+     * {@link #getContext(java.io.Reader, java.io.Writer, java.lang.String, java.lang.String, java.lang.String, org.opensolaris.opengrok.analysis.Definitions, boolean, boolean, java.util.List, org.opensolaris.opengrok.analysis.Scopes)},
```
Orviss: Maybe break the line so it won't be that long? I do not think that it will break the javadoc.

src/org/opensolaris/opengrok/search/context/Context.java
```diff
@@ -64,34 +73,164 @@
...
+        OGKUnifiedHighlighter uhi = new OGKUnifiedHighlighter(env,
+            searcher, anz);
+        uhi.setBreakIterator(() -> new StrictLineBreakIterator());
+        uhi.setFormatter(formatter);
+        uhi.setTabSize(tabSize);
+
+        try {
+            List<String> fieldList = qbuilder.getContextFields();
+            String[] fields = new String[fieldList.size()];
+            fieldList.toArray(fields);
```
Orviss: Could you please consider using this code: `String[] fields = fieldList.toArray(new String[fieldList.size()]);` ? I think this might be a little dangerous since it relies heavily on the size of the `fields` array. If someone would change it then it would break.

src/org/opensolaris/opengrok/search/context/ContextFormatter.java
```diff
@@ -0,0 +1,404 @@
...
+        if (moreUrl != null) {
+            bld.append("<a href=\"");
+            bld.append(moreUrl);
+            bld.append("\">");
+            bld.append(MORE_LABEL);
+            bld.append("</a>");
+            bld.append(HtmlConsts.BR);
+            bld.append("\n");
+            res.setFooter(bld.toString());
+            bld.setLength(0);
```
Orviss: `bld` is local variable so it is eligible for garbage collection so this line is redundant, right? :)

src/org/opensolaris/opengrok/search/context/FormattedLines.java
```diff
@@ -0,0 +1,142 @@
...
+        this.limited = limited;
+    }
+
+    /**
+     * Removes the highest line from the instance.
+     * @return a defined value
+     * @throws NoSuchElementException if the instance is empty
+     */
+    public String pop() {
+        String value = lines.remove(lines.lastKey());
```
Orviss: Return could be simplified: `return lines.remove(lines.lastKey());` :)

src/org/opensolaris/opengrok/search/context/FormattedLines.java
```diff
@@ -0,0 +1,142 @@
...
+        return res;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder bld = new StringBuilder();
+        for (String line : lines.values()) {
+            bld.append(line);
+        }
+        String f = footer;
```
Orviss: Maybe remove the variable `f`? I think it is nicer to have it just as:
```
if (footer != null && limited) {
    bld.append(footer);
}
```

idodeclare:
> what is the difference between CDDL and http://creativecommons.org/licenses/by-sa/3.0/us/ ?
(is CC similar like BSD?)

Yes, CC is a permissive license like BSD which is compatible for use in a CDDL project because CDDL licenses each file individually so that its files can be kept separate from other files with their licenses (permissive or even proprietary) — and neither CDDL nor CC (nor BSD, Apache etc.) restricts the license of the executable version (as does GPL).

CC was applicable to those two new classes, `OGKTextField` and `OGKTextVecField`, which I imported in a slightly-transformed way from the original source. CC requires attribution, a license link, an indication of any revision, and redistribution of source must be under CC — but otherwise there are no restrictions.

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
```diff
@@ -531,6 +531,9 @@ public static void writeXref(FileAnalyzerFactory factory, Reader in,
         args.setProject(project);
 
         FileAnalyzer analyzer = factory.getAnalyzer();
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
```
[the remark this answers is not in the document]
idodeclare: While testing the new highlighter, I found that scopes were not working when looking at historical versions of files or when annotating a file — i.e. where this `writeXref()` is called by `list.jsp` instead of its using the stored .gz version.

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -92,6 +92,13 @@
      * value.
      */
     private static final String NEGATIVE_NUMBER_ERROR = "Invalid value for \"%s\" - \"%s\". Expected value greater or equal than 0";
+    /**
+     * Error string for non-positive numbers (could be int, double, long, ...).
+     * First argument is the name of the property, second argument is the actual
+     * value.
+     */
+    private static final String NONPOSITIVE_NUMBER_ERROR =
+        "Invalid value for \"%s\" - \"%s\". Expected value greater than 0";
```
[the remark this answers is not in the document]
idodeclare: I agree with the original author that `%s` is more forgiving for error strings. E.g., if the field were changed to a boxed primitive, the error string would still work (printing `null` if necessary) but `%d` might throw an exception.

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
idodeclare: yes, removed

src/org/opensolaris/opengrok/analysis/OGKTextField.java
idodeclare: ok removed

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
idodeclare: ok removed

src/org/opensolaris/opengrok/search/Results.java
idodeclare: ok fixed

src/org/opensolaris/opengrok/search/Results.java
idodeclare: ok revised to use a private container object for the arguments to `printPlain()` that do not change in the calling loop.

src/org/opensolaris/opengrok/search/context/Context.java
idodeclare: ok revised

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1288,7 +1301,64 @@ public boolean getListDirsFirst() {
...
+     * <p>
+     * N.b. the value is not mediated to {@link Configuration}.
+     * @param value a defined value or {@code null} to reset to use the
+     * {@link Configuration#getContextSurround()}
+     * @throws IllegalArgumentException if {@code value} is not positive
+     */
+    public void setContextLimit(Short value)
+            throws IllegalArgumentException {
+        if (value < 1) {
+            throw new IllegalArgumentException("`value' is not positive");
```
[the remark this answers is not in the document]
vladak: I would too prefer to use the same character for quoting.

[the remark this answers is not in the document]
tarzanek: what if we don't use any quotes there? ;)

src/org/opensolaris/opengrok/index/IndexAnalysisSettingsAccessor.java
```diff
@@ -70,20 +86,24 @@ public IndexAnalysisSettings read(IndexReader reader) throws IOException {
...
 
-        Document doc = searcher.doc(top.scoreDocs[0].doc);
-        IndexableField objser = doc.getField(QueryBuilder.OBJSER);
-        try {
-            return objser == null ? null : IndexAnalysisSettings.deserialize(
-                objser.binaryValue().bytes);
-        } catch (ClassNotFoundException ex) {
-            // This is not expected, so translate to RuntimeException.
-            throw new RuntimeException(ex);
+        int nres = top.totalHits > n ? n : (int)top.totalHits;
```
tarzanek: thank you for this update, it will make our life easier with new releases of lucene, where they prefer this new approach to fetching results

src/org/opensolaris/opengrok/index/IndexDatabase.java
tarzanek: ok, we shall stick with this exact matching then, if I will find a way around I will change it to generic to be able to stick to latest lucene (yes, it breaks backwards compatibility, but this is a developer tool, latest greatest is of the essence)

src/org/opensolaris/opengrok/search/context/FormattedLines.java
tarzanek: if there are concerns about race condition could it be solved otherwise?
e.g. sync on setter?
I was under the impression this object would be thread local when used ...

src/org/opensolaris/opengrok/search/context/LineHighlight.java
```diff
@@ -0,0 +1,208 @@
...
+     * @return zero or greater
+     */
+    public int getRelide() {
+        return relide;
+    }
+
+    /**
+     * Sets the right elide value.
+     */
+    public void setRelide(int value) {
```
tarzanek: I am wondering if we should move the project towards scala or kotlin to make the data objects less verbose

src/org/opensolaris/opengrok/search/context/OGKUnifiedHighlighter.java
```diff
@@ -0,0 +1,289 @@
...
+             * postings should be sufficient in the comment for
+             * shouldHandleMultiTermQuery(String): "MTQ highlighting can be
+             * expensive, particularly when using offsets in postings."
+             *     DEFS are stored with term vectors to avoid this problem.
+             *     FULL should be approximately fine with re-analysis using an
+             * on-the-fly PlainAnalyzer.
+             *     REFS should be approximately fine with re-analysis using an
+             * on-the-fly PlainAnalyzer. It might not accord with the true
+             * language symbol tokenizer, but it should not be wildly
+             * divergent.
```
tarzanek: can we inject the language symbol tokenizer somehow?

tarzanek: (fwiw we already do that when doing analysis, it's a bit hacky since lucene defaulted to just one default analyzer ( see https://github.com/oracle/opengrok/blob/master/src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java#L98 ) )

src/org/opensolaris/opengrok/web/HtmlConsts.java
```diff
@@ -44,4 +45,10 @@
     public static final String NUMBER_CLASS = "n";
     public static final String STRING_CLASS = "s";
 
+    public static final String B = "<b>";
+    public static final String ZB = "</b>";
+
+    public static final String BR = "<br/>";
+
+    public static final String HELLIP = "&hellip;";
```
tarzanek: :-D

src/org/opensolaris/opengrok/web/SearchHelper.java
```diff
@@ -520,7 +536,7 @@ public SearchHelper prepareSummary() {
             return this;
         }
         try {
-            sourceContext = new Context(query, builder.getQueries());
+            sourceContext = new Context(query, builder);
```
tarzanek: especially curious if regexp queries will work

tarzanek:
thank you very much, i don't have any other comments or musings now
so I'd say +1
@vladak @tulinkry @Orviss is it OK to merge?

src/org/opensolaris/opengrok/search/context/FormattedLines.java
idodeclare: Well my point is that my making a local copy is rationalized — i.e. it's a pattern I do recommend generally when nullably-mutable fields are involved so that you can have objects that do not use heavyweight synchronization but also that are not inconsistent intra-method.

A local copy is also valuable in some synchronized settings — e.g., where you have field data that strictly evolves lazily from `null` to non-`null` (as in e.g. a singleton) and a `volatile` field is properly used. In that case too, it's better to make a local copy of the field so that for the predominant case happening after initialization (i.e., when non-`null`), only a single atomic read is required w.r.t. to field.

src/org/opensolaris/opengrok/search/context/OGKUnifiedHighlighter.java
idodeclare: OK after studying the Lucene `UnifiedHighlighter` implementation, I pushed an arcane override of `getIndexAnalyzer()` when OpenGrok's own `highlightFieldsUnion()` is executing so that an OpenGrok `TYPE`-specific analyzer is returned when re-`ANALYSIS` is being done. (I also updated those block comments.)

src/org/opensolaris/opengrok/search/context/LineHighlight.java
tarzanek: come on, less typing ;)
(but then one needs to get ready for some code readability changes - for good or for bad)

src/org/opensolaris/opengrok/search/context/FormattedLines.java
tarzanek: ok, makes sense

tarzanek:
@idodeclare could you please ping me a mail to tarzanek at gmail.com  with your email address? Since @Orviss  will be updating the build we will start using Idea for development, folks from Jetbrains gave us licenses for ultimate, so if you are interested you can have one license ... (the plan is to have the maven used from Idea for development)

opengrok-indexer/pom.xml
```diff
@@ -19,7 +19,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

pom.xml
```diff
@@ -19,6 +19,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/AnalyzerGuru.java
```diff
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/CtagsReader.java
```diff
@@ -378,8 +378,8 @@ private static String cutPattern(String tagLine, int startTab, int endTab) {
      * Adds a tag to a {@code Definitions} instance.
```
tulinkry: copy 2018

src/org/opensolaris/opengrok/analysis/archive/TarAnalyzer.java
```diff
@@ -19,20 +19,22 @@
 
 /*
  * Copyright (c) 2005, 2013, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/archive/ZipAnalyzer.java
```diff
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2013, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/document/MandocAnalyzer.java
```diff
@@ -19,20 +19,20 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/document/TroffAnalyzer.java
```diff
@@ -19,21 +19,21 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/executables/ELFAnalyzer.java
```diff
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/executables/JarAnalyzer.java
```diff
@@ -30,11 +30,11 @@
 import java.util.zip.ZipInputStream;
```
tulinkry: old copyright

src/org/opensolaris/opengrok/analysis/executables/JavaClassAnalyzer.java
```diff
@@ -25,7 +25,6 @@
```
tulinkry: old copyright

src/org/opensolaris/opengrok/analysis/plain/PlainAnalyzer.java
```diff
@@ -27,19 +27,22 @@
 import java.io.InputStream;
```
tulinkry: old copyright

src/org/opensolaris/opengrok/analysis/plain/XMLAnalyzer.java
```diff
@@ -19,21 +19,22 @@
 
 /*
  * Copyright (c) 2005, 2014, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/analysis/uue/UuencodeAnalyzer.java
```diff
@@ -19,21 +19,21 @@
 
 /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -17,7 +17,7 @@
  * CDDL HEADER END
  */
 
- /*
+/*
  * Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -136,7 +136,20 @@
     private int messagesInTheSystem = 0;
```
tulinkry: old copyright

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -50,6 +50,8 @@
 import java.util.zip.GZIPOutputStream;
```
tulinkry: old copyright

src/org/opensolaris/opengrok/search/QueryBuilder.java
```diff
@@ -30,6 +30,7 @@
 import java.security.NoSuchAlgorithmException;
```
tulinkry: old copyright

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -82,9 +82,11 @@ private Results() {
      * @throws CorruptIndexException
```
tulinkry: old copyright

src/org/opensolaris/opengrok/search/SearchEngine.java
```diff
@@ -19,6 +19,7 @@
 
  /*
  * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/search/context/Context.java
```diff
@@ -19,8 +19,8 @@
 
 /*
  * Copyright (c) 2005, 2015, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

src/org/opensolaris/opengrok/web/SearchHelper.java
```diff
@@ -20,14 +20,15 @@
 /*
  * Copyright (c) 2011, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

test/org/opensolaris/opengrok/search/context/ContextTest.java
```diff
@@ -83,32 +83,32 @@ public void testIsEmpty() throws ParseException {
```
tulinkry: old copyright

test/org/opensolaris/opengrok/search/context/SearchAndContextFormatterTest.java
```diff
@@ -0,0 +1,183 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: wrong copyright

test/org/opensolaris/opengrok/search/context/SearchAndContextFormatterTest2.java
```diff
@@ -0,0 +1,262 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: wrong copyright

web/more.jsp
```diff
@@ -19,8 +19,8 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 CDDL HEADER END
 
 Copyright (c) 2010, 2017, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: copy

idodeclare:
> copyright issues

Hi @tulinkry , I do not work for Oracle, so if I were to modify Oracle's notices, it would violate the following part of CDDL section 3.3:

> You may
    not remove or alter any copyright, patent or trademark notices
    contained within the Covered Software, or any notices of licensing
    or any descriptive text giving attribution to any Contributor or
    the Initial Developer.

The proscription about altering copyright is also in effect when a CDDL-licensed file is copied and altered, as I did with `SearchAndContextFormatterTest.java` from `SearchEngineTest.java` (with a JavaDoc comment "Derived from Trond Norbye's SearchEngineTest"). The copyright notices that are not my own are identical between those files.

All the files altered or created by me in this patch include a notice satisfying another part of CDDL section 3.3:

> You must include a notice in each of Your Modifications that
    identifies You as the Contributor of the Modification

I don't update that manually. I use a script I have, `update_cddl.pl`, as in e.g.:

```
    $ git diff master --name-only | xargs update_cddl.pl -P
```
(`-P` to use "Portions Copyright" language as appropriate.)

If you like I can post it for use by others. Oracle contributors would run as e.g.:

```
    $ git diff master --name-only | \
        CDDL_ORG_IDENTITY="Oracle and/or its affiliates" xargs update_cddl.pl -P -r
```
(`-r` additionally because Oracle's notices include "All rights reserved.")

tarzanek:
thank you Chris, merging, let's play with this and file issues as we go

PR #2057, opened 2018-04-06
document and handle InvalidPathException in history parsers

vladak:
With this change it is possible to index internal Solaris source code repository (that contains UTF-8 file names) on a system with `LC_ALL="C"` and `LANG=""`.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -96,6 +96,7 @@ import org.opensolaris.opengrok.web.Statistics;
 import org.opensolaris.opengrok.web.Util;
 
 import static java.nio.file.FileVisitResult.CONTINUE;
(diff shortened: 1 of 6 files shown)


src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -96,6 +96,7 @@
 import org.opensolaris.opengrok.web.Util;
 
 import static java.nio.file.FileVisitResult.CONTINUE;
+import java.nio.file.InvalidPathException;
```
Orviss: Breaks checkstyle's import order option: <http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderOption.html#UNDER>

I will have to work on enabling that checkstyle...

Orviss:
I'm fine with the changes, if you want to you can wait for @tarzanek or merge it yourself :)

PR #2061, opened 2018-04-10
add OS X Travis build

vladak:
Enables Travis builds on OS X as well as Linux.

I had to move the javadoc check out of OS X because Maven was failing to find the jrcs dependency for some reason - it was searching for it on Maven central as opposed to locally (most likely because of the way `javadoc` target is invoked - quite possibly this could be done better; I never liked the descending into module directories and invoking the target there).

Also, I got rid of the `junit-force-all` option. There is a bit of risk of losing the coverage however on OS X I don't see how we can make all the tests to be run (especially for those exotic SCMs).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,33 @@
-dist: trusty
-sudo: required
 language: java
-jdk:
-  - oraclejdk8
 cache:
   directories:
   - $HOME/.m2
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: required
+      jdk: oraclejdk8
+    - os: osx
+      osx_image: xcode7.3
 before_install:
- - sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


tarzanek:
I don't have a problem with this change request, however we will then need to fix the javadoc maven target to be able to run it properly (instead of the quick per module hack).
Otherwise we will get build failures whenever someone forgets some silly javadoc change and waste time with such trivial issues ...

tarzanek:
@Orviss will the new maven fixes be able to fix javadoc too?

tarzanek:
anyway, +1, please merge from my side

vladak:
the javadoc check is still run after this change, only for linux builds,
not osx.

tarzanek:
d'oh!
ok, I skipped the block with version verifying without noticing the check ;)
in such case please do merge now! :)

vladak:
Ok, merging from tram..

Dne st 11. 4. 2018 7:53 uživatel Lubos Kosco <notifications@github.com>
napsal:

PR #2065, opened 2018-04-13
OpenGrok help uses bashism

kahatlen:
On systems where /bin/sh is dash (like Debian), invoking OpenGrok help
will print the following error message along with its output:

./OpenGrok: 930: [: unexpected operator

Also, the --detailed option won't print all the detailed help.

This patch replaces a bashism (the == operator) with standard shell
syntax (the = operator).

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -927,7 +927,7 @@ UpdateDescriptionCache()
 
 OpenGrokUsage()
 {
-    [ "$1" == "--detailed" ] && helpargs=$1
+    [ "$1" = "--detailed" ] && helpargs=$1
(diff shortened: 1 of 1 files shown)


tarzanek:
Thank you sir!
How are the fjords? ;)

tarzanek:
merging, thank you :)

tarzanek:
fwiw - wondering if [[ ]] would work in dash ... (to be safe on string comparison)

kahatlen:
Thanks for merging the fix, Ľuboš!

It doesn't look like dash supports [[ ]]:
```
$ [[ a = b ]]
dash: 1: [[: not found
```

Issue #1296, opened 2016-12-05
opengrok does't jump with function name with prefix "__" in C language


cheneydeng:
hi guys,
   I'm using OpenGrok to read one project which written with C language,everything is perfect except the function name with "__" as prefix will not jump to the definition when clicked.I read the html code and find that the marker has no "name" item with such functions.How can i fix it quickly as I'm really rely on OpenGrok.Thanks as advance.

vladak:
What ctags implementation (+version) do you use ?

cheneydeng:
@vladak This is my ctags version:

`
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Sep 13 2015, 03:28:50
  Addresses: <dhiebert@users.sourceforge.net>, http://ctags.sourceforge.net
  Optional compiled features: +wildcards, +regex
`

vladak:
and OpenGrok version ?

One thing you can try is to replace Exuberant ctags with Universal ctags ?

Also, could you try to isolate the issue by shortening the file to smallest possible size where the problem still occurs ?

tulinkry:
0.13-RC5
[main.c.txt](https://github.com/OpenGrok/OpenGrok/files/631307/main.c.txt)

tulinkry:
The problem is here: [JFlexXref.java#L577-L587](https://github.com/OpenGrok/OpenGrok/blob/7ae7d63284bc15a7584a62552ceeea7f6a4f01a6/src/org/opensolaris/opengrok/analysis/JFlexXref.java#L577-L587)

cheneydeng:
@tulinkry @vladak Thanks for your reply.I've modified the code and rebuild it,when deploy it said:

> ➜  OpenGrok git:(master) ✗ OPENGROK_TOMCAT_BASE=/usr/local/Cellar/tomcat/8.5.3/libexec ./OpenGrok deploy
Loading the default instance configuration ...
Installing /Users/cheneydeng/software/OpenGrok/dist/source.war to /usr/local/Cellar/tomcat/8.5.3/libexec/webapps ...

> Start your application server (Tomcat),  if it is not already
  running, or wait until it loads the just installed web  application.

> OpenGrok should be available on <HOST>:<PORT>/source
  where HOST and PORT are configured in Tomcat.

Actually my tomcat is running on localhost:8080 as before so i restarted tomcat and found my modification seems not works,maybe the deploy failed?

[...]

PR #2067, opened 2018-04-13
Use HTML5 DOCTYPE, as XHTML was not at all confirming anyway

idodeclare:
Hello,

Please consider for integration this patch to specify an HTML5 DOCTYPE and not XHTML, as the documents were not at all confirming anyway.

This patch also updates the ID support for HTML5 as advised by @tulinkry in #1296.

Thank you.


test/org/opensolaris/opengrok/analysis/ada/AdaXrefTest.java
```diff
@@ -86,13 +86,10 @@ private int writeAdaXref(InputStream iss, PrintStream oss)
...
     private String getHtmlBegin() {
-        return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
-            "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\"\n" +
-            "    \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">\n" +
-            "<html xmlns=\"http://www.w3.org/1999/xhtml\"" +
-            " xml:lang=\"en\" lang=\"en\"\n" +
-            "      class=\"xref\">\n" +
+        return "<!DOCTYPE html>\n" +
+            "<html>\n" +
             "<head>\n" +
```
tulinkry: here you discarded .xref class, are you sure that any javascript does not depend on this?

tulinkry: Also why removing lang attribute?

idodeclare: Hi, @tulinkry . For the test classes and data, I left the `<html>` element simple since that part of the documents is not actually under test (e.g. it is static content not truly coming from `httpheader.jspf` — where that header sets a dynamic `class`).

I removed `lang` because it did not seem correct to assert a principal natural language of `en` for OpenGrok pages since the primary content by far are `xref` pages on documents of unknown languages. Instead I thought to leave `lang` unknown to reflect this indeterminacy. Let me know if you think `en` should be restored.

tulinkry:
> class

Ok I was blind

> lang 

I would say that English prevails in the source code + the keywords are always in English. Let the others to say some words on this.

@vladak @tarzanek

tulinkry:
Plus also the page itself is in English only.

tulinkry:
Let’s go

PR #2070, opened 2018-04-13
Travis release

vladak:
This set of changes makes it possible to produce a release via Travis by pushing a new tag to the repository.

Drawbacks:
  - I had to restructure `plugins` build to avoid a problem with auto-generated class names by the maven javadoc plugin (basically it generated `plugins/target/site/apidocs/packages` file - instead of ` plugins/target/site/apidocs/package-list` - that contained classes with `test-classes` in their name on which javadoc check barfed with `illegal class name`). As a result the structure is more native to Maven (`src` and `test` directories) however Netbeans project config was lost in the process. I might add it later. On the other hand, javadoc check now works even for the authorization plugins.
  - it seems that a release is created even if the verify phase fails. I am not entirely sure when the release is made given the build matrix. Given that a release will be produced only after the previous changeset passed the build this is not a big problem.
  - Solaris packages are no more. The `tools/dorelease.py` script can be updated so that it just uploads the package files to existing release later on.


plugins/src/opengrok/auth/plugin/ldap/LdapServer.java
```diff
@@ -176,7 +176,7 @@ private synchronized LdapContext connect() {
      * @param cons controls for the LDAP request
      * @return LDAP enumeration with the results
      *
-     * @throws NamingException
+     * @throws NamingException naming exception
```
Orviss: This kind of javadoc comments is redundant. I assume it was done to remove warnings while generating javadoc? If this is part of the public API then more reasonable comment is preferable. If not then I would remove the whole line.

vladak: yep, it was sort of a desperate effort to clear the errors. I don't think the line can be removed.

plugins/pom.xml
```diff
@@ -27,44 +27,30 @@ Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
...
-        <version>1.1-SNAPSHOT</version>
+        <artifactId>opengrok-top</artifactId>
+        <version>1.1-rc22</version>
     </parent>
 
     <build>
-        <sourceDirectory>.</sourceDirectory>
-	<!-- Note that the plugins/UserPlugin/test/ directory is added via the build-helper plugin below -->
-        <testSourceDirectory>LdapPlugin/test</testSourceDirectory>
+        <sourceDirectory>src</sourceDirectory>
```
Orviss: I think we should adhere to standard project structure, e.g. `src/main/java`.

vladak:
Persinally I find src/main a bit strange. Also, the src/org/... matches the
OpenGrok structure.

Dne so 14. 4. 2018 13:35 uživatel Adam Hornáček <notifications@github.com>
napsal:

Orviss:
It might seem strange at first but it is a standard (default) so many programmers are used to it. It is also good that it provides you with a nice tree structure for source files, e.g.:
* src/main/java
* src/main/resources
* src/main/webapp
* src/test/java
* src/test/resources
* etc.

And in my PR I already changed the OpenGrok structure to this standard. (I was actually working on it yesterday but I had some problems with package renaming -> actually plugins were the only thing broken because of the `testplugins.jar` test resource. However, I think it will be better to wait till this is integrated since I will have quite a few conflicts because of the `pom.xml` changes.)

vladak:
Okay, I will leave like it is to be consistent and then you can change it
all. My only worry is how specific is this structure to Maven (in case we
transition to Gradle or such).

Dne so 14. 4. 2018 14:00 uživatel Adam Hornáček <notifications@github.com>
napsal:

Orviss:
I believe Gradle has the same structure, at least its Java Plugin: <https://docs.gradle.org/current/userguide/java_plugin.html>

plugins/pom.xml
tarzanek: I think this change will be inevitable ... (look at gradle, sbt, ... )

tarzanek:
I didn't read the changes yet, though I think you can merge and we will continue with Adams changes ...

PR #2075, opened 2018-04-19
run Jacoco and publish the results to Coveralls.io

vladak:
Enables Jacoco proper, publishes results to [Coveralls.io](https://coveralls.io/repos) (I enabled the oracle/opengrok repo there).

Used resources:
  - https://angus.nyc/2015/continuous-integration-for-java-github
  - https://stackoverflow.com/questions/12269558/maven-jacoco-plugin-error
  - https://github.com/onelogin/java-saml/issues/74

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
 script: "mvn -B -V verify"
 
 after_success:
- - mvn clean test jacoco:report coveralls:report
+ - mvn clean test jacoco:report org.eluder.coveralls:coveralls-maven-plugin:report
 
 addons:
   coverity_scan:
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,8 @@ env:
 # in 'script' no matter if an individual command fails or not. This kludgy
 # looking shell incantation takes care of that.
 #
-script: "mvn -B -V verify"
-
-after_success:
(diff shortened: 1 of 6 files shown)


tarzanek:
looks good, let's merge ... I guess coveralls will be part of checks now?

vladak:
It will display the coverage nicely for each build, the actual threshold
enforcement is up to us. In an ideal world the difference for each pull
request would be checked automatically however that would require some sort
of API from Coveralls + associated scripting. Not worth doing now I guess.

Dne čt 19. 4. 2018 15:40 uživatel Lubos Kosco <notifications@github.com>
napsal:

tarzanek:
we have more such tools in https://github.com/oracle/opengrok/wiki/How-to-build-OpenGrok-from-source
coverity is there too .. (but not sure if it is up to date)

tarzanek:
isn't the plugin you use a bit out of date ?
per https://www.versioneye.com/user/projects/54b7a584050646ca5c0001aa
?

vladak:
Possibly. That might be true for other plugins as well.

Dne čt 19. 4. 2018 17:33 uživatel Lubos Kosco <notifications@github.com>
napsal:

vladak:
Yep. Not sure the rest can be integrated as well as Jacoco, though.

Dne čt 19. 4. 2018 17:29 uživatel Lubos Kosco <notifications@github.com>
napsal:

PR #2076, opened 2018-04-20
Update JaCoCo to 0.8.1

Orviss:
JaCoCo updated to the newest version because the previous version had problems with Java 10. I also replaced `\t` chars to ordinary spaces. Thanks :)


tarzanek:
merging, thank you sir!

tarzanek:
(and I know ant is not updated, but let's worry about that later (or never)

Issue #2069, opened 2018-04-13
indexing with history broken on Windows


tarzanek:
15:43:49 SEVERE: Problem updating lucene index database:
java.nio.file.InvalidPathException: Illegal char <:> at index 50: C:\var\opengrok\source\OpenGrok\testdata\sources\C:\var\opengrok\source\OpenGrok
        at java.base/sun.nio.fs.WindowsPathParser.normalize(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPath.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(Unknown Source)
        at java.base/java.nio.file.Paths.get(Unknown Source)
        at org.opensolaris.opengrok.history.HistoryGuru.getRepository(HistoryGuru.java:858)
        at org.opensolaris.opengrok.history.HistoryGuru.ensureHistoryCacheExists(HistoryGuru.java:819)
        at org.opensolaris.opengrok.index.IndexDatabase.update(IndexDatabase.java:429)

put OpenGrok clone into C:\var\opengrok\source\ and index with git setup for parsing ...

tarzanek:
seems the cause is #1989

vladak:
Too bad Travis does not support Windows yet (travis-ci/travis-ci#2104)

tarzanek:
I am actually surprised there are 2 paths in the string ... @idodeclare is there some weird repo paths merging going on?

vladak:
Also, if you run OpenGrok tests on your Windows box, does any of them fail ?

tarzanek:
I run only subsets, it's very hard to set up junit tests on Windows (especially the ones that use hardcoded names for executables in ants build.xml such as git, cvs ... ), but partial tests can be run
I hope this situation will improve with maven

idodeclare:
> This might be some weird interaction of getRelativeToCanonical() with getRepository()

You're right. The return value is not discriminated properly as already done in `RepositoryInfo`. I'll push a fix for that. But the other possible problem is if there is a `c:`/`C:` casing difference (which the fix will not address), since `getRelativeToCanonical()` (OpenGrok in general) is not specifically handling case-insensitive naming as allowed on NTFS. (And HFS too generally but perhaps less sensitively since drive letters are not involved).

[...]

PR #2077, opened 2018-04-22
Properly handle result of getRelativeToCanonical()

idodeclare:
Hello,

Please consider for integration this patch to fix result handling from `getRelativeToCanonical()`. Existing tests cover this change for Linux and Mac. This may or may not be the solution to issue #2069 on Windows, as I could not test on that environment.

Thank you.


tarzanek:
I just verified on windows and it works, merging
Thank you!

PR #2079, opened 2018-04-23
Do not run Jacoco/Coveralls for forks

vladak:
Should still work for pull requests.


tarzanek:
lovely, works now!
purrfect! thank you @vladak !  :-D
( https://travis-ci.org/tarzanek/OpenGrok works! )

PR #2081, opened 2018-04-24
add the OPENGROK_DEFAULT_PROJECTS env var

vladak:
passthru for the -p option. I don't like that the shell script is a filter to `Indexer` however until it is rewritten such changes need to happen.

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -111,6 +111,9 @@ Supported Environment Variables for configuring the default setup:
   - OPENGROK_ENABLE_PROJECTS    Enable projects, where every directory in
                                   SRC_ROOT is considered a separate project.
                                   on|off (default on) (^)
+  - OPENGROK_DEFAULT_PROJECTS   List of projects (relative to source root,
+                                starting with slash) separated with comma to be
+				selected by default in the UI.
   - OPENGROK_IGNORE_PATTERNS    Set ignored patterns for indexer to skip.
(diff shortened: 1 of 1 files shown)


OpenGrok
```diff
@@ -311,6 +314,13 @@ DefaultInstanceConfiguration()
         *) ENABLE_PROJECTS="-P" ;;
     esac
 
+    if [ -n "${ENABLE_PROJECTS}" ]; then
+        DEFAULT_PROJECTS=""
+            for proj in $( IFS=","; echo ${OPENGROK_DEFAULT_PROJECTS} ); do
```
tulinkry: non blocking but wrong indentation

tulinkry:
The only point for this is if this is really usefull because (as far as I remember - this might have changed) this default projects are used only when we don't have any cookies which is:

1. first application start
2. user cleaned cookies

Neither of this is very often.

OpenGrok
vladak: fixed

vladak:
True, this only works when there are no cookies. I guess it is a way how to guide first time users a bit.

vladak:
added some documentation to https://github.com/oracle/opengrok/wiki/Webapp-configuration#default-projects

tulinkry:
Yes it is a bit more user friendly 😀

PR #2083, opened 2018-04-25
Search result background color for odd rows fix

### LINKED ISSUE #2082: background color of highlighted documents in search result change inconsistently
Issue can be seen on the screenshot:

screen shot 2018-04-24 at 10 41 54

It just seems weird to me that the color changes only when multiple documents are found in the same directory. I think it would be better if the color would change for every row or the same color would be at the start of the directory.

Orviss:
fixes #2082

I was not able to resolve this by css :(

Thanks for review :) 

For comparison: (almost the same part as described in issue)
screen shot 2018-04-25 at 17 07 44

### DIFF
--- a/src/org/opensolaris/opengrok/search/Results.java
+++ b/src/org/opensolaris/opengrok/search/Results.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2018, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright 2011 Jens Elkner.
  * Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
  */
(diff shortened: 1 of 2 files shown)


vladak:
Looks good enough to me.

src/org/opensolaris/opengrok/search/Results.java
```diff
@@ -253,6 +247,20 @@ public static void prettyPrint(Writer out, SearchHelper sh, int start,
                 out.write("</tt></td></tr>\n");
             }
         }
+        out.write("</tbody>");
```
tarzanek: really?
(I am just confused it wasn't there before ... isn't it close somewhere else? )

Orviss: I was surprised as well. It seems the browser adds it there by itself.

Simple test: (when I `Inspect Element` the `<tbody>` is there)
```html
<!doctype html>
<html>
<head>
</head>
<body>
	<table>
		<tr><td>test1</td></td>
		<tr><td>test2</td></td>
	</table>
</body>
</html>
```

tarzanek:
thank you, closing

PR #2087, opened 2018-04-27
check Lucene index version before reindex

vladak:
This only captures when there is Lucene major version discrepancy. The `check()` method can be updated later to cope with internal index changes.

I had to add Lucene backward codecs so that the segment file can actually be opened. Otherwise it will result in `IllegalArgumentException` as noted in the issue.

For now no data deletion is done, this is up to the user.


vladak:
The `oldindex.zip` (~ 4 KB) contains Lucene 6.5 index (produced by OpenGrok 1.0) of the `foo` respository.

PR #2093, opened 2018-05-03
Mirror timeout

vladak:
Some of the internal repositories recently started exhibiting strange traffic patterns (either connect was lagging or the data was painfully slow to receive from the other side. For Mercurial for instance this means that commands like `hg incoming` were hanging for hours) which made me realize there is no timeout mechanism in the mirroring scripts.

Also, this fixes bunch of places where logger is not properly passed onto the created objects so some logging was missing.

### DIFF
--- a/tools/sync/command.py
+++ b/tools/sync/command.py
@@ -95,7 +95,7 @@ class Command:
                         self.logger.info("Terminating command {} with PID {} "
                                          "after timeout of {} seconds".
                                          format(p.args, p.pid, self.timeout))
-                        self.popen.terminate()
+                        p.terminate()
                         self.exception = TimeoutException("Command {} with pid"
                                                           " {} timed out".
                                                           format(p.args,
--- a/tools/sync/command.py
+++ b/tools/sync/command.py
@@ -93,8 +93,8 @@ class Command:
                     if not self.condition.wait(self.timeout):
                         p = self.popen
                         self.logger.info("Terminating command {} with PID {} "
-                                         "after timeout".
-                                         format(p.args, p.pid))
+                                         "after timeout of {} seconds".
+                                         format(p.args, p.pid, self.timeout))
                         self.popen.terminate()
                         self.exception = TimeoutException("Command {} with pid"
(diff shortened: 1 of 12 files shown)


vladak:
Once this is merged, I will update the docs on https://github.com/oracle/opengrok/wiki/Repository-synchronization#mirrorpy

tools/sync/command.py
```diff
@@ -61,6 +71,41 @@ def execute(self):
...
+                                         format(p.args, p.pid, self.timeout))
+                        self.popen.terminate()
+                        self.exception = TimeoutException("Command {} with pid"
+                                                          " {} timed out".
+                                                          format(p.args,
+                                                                 p.pid))
+                    else:
+                        return None
+
+            def getexception(self):
```
Orviss: Maybe rename to `get_exception` or `getException`? (Not sure what Python style is.)

tools/sync/command.py
```diff
@@ -199,3 +270,6 @@ def getoutput(self):
 
     def getstate(self):
         return self.state
+
+    def getpid(self):
```
Orviss: Maybe `get_pid` or `getPid`?

tools/sync/command.py
tulinkry: `get_exception`

tools/sync/command.py
```diff
@@ -26,6 +26,13 @@
 import subprocess
 import string
 import threading
+import time
+
+
+class TimeoutException(Exception):
+    """
+    Exception returned when command exceeded its timeout.
+    """
```
tulinkry: \+ `pass`

vladak: I think the block comment does the same service however for readability it does not hurt so adding.

tools/sync/command.py
vladak: ok

tools/sync/command.py
vladak: I'd have to change all the other get* methods, maybe some other time.

PR #2104, opened 2018-05-10
Feature/tokenizer mode

idodeclare:
Hello,

Please consider for integration this patch to add a mode switch `OPENGROK_ALL_NONWHITESPACE` (`--allNonWhitespace [on|off]` [default off]) to index all non-whitespace for `FULL` queries.

With the mode activated, the set of tokens currently indexed by the current pre-release 1.1-rc26 will still return results, but additionally all strings of contiguous non-whitespace will be multiply tokenized to allow flexible but precise matching of text fragments that can include e.g. punctuation (any non-whitespace really).

As an example, the source text `(the "license")` in release 1.1-rc26 will only tokenize the following:
```
the
license
```
so the user cannot home in on queries that include the punctuation parentheses or quotes.

With `--allNonWhitespace on`, the same source text `(the "license")` will be indexed with the following tokens (and position offsets):
```
(the	|1
the	|0
"license	|1
"license"	|0
"license")	|0
license	|0
license"	|0
license")	|0
```

to support the same previous tokens but additionally a number of very precise queries indexing true full-text.

The non-whitespace runs are broken on specific, most-commonly found places in programming languages: e.g. word/non-word boundaries, open- and closing- punctuation boundaries, and quoting character boundaries (with new support for recognizing English contractions so that words such as `that'll` will not be broken).

The cost, of course, is an index that consumes more space. In release 1.1-rc26 I find that indexes are approximately 75% of the size of source code. With `--allNonWhitespace on`, indexes are at least 100% of source code size and often higher. When I index a sample of ten large, common open source projects, the total index size is 125% of source code size.

I find, however, that some repos (depending on runs of non-whitespace) can even have indexes 200% the size of source code. Despite this, I find that the utility of producing matches on a multitude of fragments of non-whitespace is invaluable.

Since this patch will support queries with punctuation, it also updates `QueryBuilder` to avoid re-escaping punctuation when the user includes explicit Lucene escape codes. E.g., `QueryBuilder` will continue to support a user submitting a `DEFS` query `method1:forValue:` and escape the colons for Lucene. OpenGrok will also now support a query `method1\:forValue\:` where the user includes explicitly the Lucene escapes and no longer throw an error.

Thank you.


Orviss:
Hello,

maybe I am missing something but why not just add `(`, `"`, etc. as terms and then just use `PhraseQuery` (e.g. `"(the"`)? Just a handful of new terms would be added to the index which is negligible. Regarding the performance, I'm not completely sure but I think that your way will be faster for these lookups but some other queries (e.g. wildcard queries or regexp queries will be slowed down by the factor of the changed lucene index size).

Also, how does this perform on minified code, e.g. `*.min.js` files?

Regardless, to me this feels like twisting Lucene. If we truly want to provide full-text search then I believe that it should be done differently: e.g. https://swtch.com/~rsc/regexp/regexp4.html or  https://livegrep.com/search/linux 

Regards,
Adam

tarzanek:
I like that this is "experimental" :)
I am also curious what it will do on minified code, but then I hope the jflex token length limiter will kick in there.
Let's finish the review (I did like 20%, so need more time, especially on how do you preserve backwards compatibility, since the analysers are written in the way that they ignore special chars on purpose)
and I'd be happy for other reviews too (and thanks Adam, this is an interesting idea with PhraseQuery, I think OpenGrok has a huge debt in using queries properly and in most effective way (for user of course) )

idodeclare:
> curious what it will do on minified code

@tarzanek , I pushed a test of minified css and a revision to ensure that default plain-full tokenization is used when new word-breaking operations reach the const limits — which will happen for minified code.

tarzanek:
looking ... 
we need to fix the javadoc problem so travis behaves again ... <sigh>

tarzanek:
ah ...
I see ...
[ERROR] /home/travis/build/oracle/opengrok/test/org/opensolaris/opengrok/util/CustomAssertions.java:112: error: reference not found
[ERROR]      * {@link #assertSymbolStream(java.lang.Class, java.io.InputStream, java.util.List, boolean)}
[ERROR]               ^

Orviss:
Actually, `PhraseQuery`'s text is tokenized. Therefore, queries like `"class.member"` or `"class . member"` are the same. However, you can easily override this behaviour in `QueryParserBase#newFieldQuery` -> just override this method in OpenGrok's `CustomQueryParser`. Therefore, there would be no problem with query `"(the"` and treating the `(` and `the` as separate terms.

idodeclare:
@tarzanek, thank you for spotting that. I was baffled because NetBeans "Generate Javadoc (opengrok)" was running without error. I guess it does not generate for `test/` classes, unlike the travis build.

idodeclare:
@Orviss , that would break support for languages with punctuation in their symbols names, such as Objective-C with names like `sendAction:toObject:forAllCells:` or for Forth with names such as `ram!16`.

Orviss:
Your provided examples still has to be escaped in normal query then why not escape them in `PhraseQuery` as well? However, let's suppose that there is a token which contains punctuation that will be detected by `PhraseQuery` but not the file analyzer (analyzer treats it as a single token) and is not reserved query symbol (not !, :, *, etc. ). (something like `t#t` (just an example for illustration – no idea if that applies to our case))

Different approaches that comes to my mind:
* we could tokenize `PhraseQuery`'s text based on the selected `type` field. (E.g. if Objective-C is selected then `:` would not be a separate token) (however, this is quite constraining)
* we could use your approach but specifically tailored for the tokens which contain punctuation that is deemed as a separate token by the `PhraseQuery`'s analyzer (I think there would not be as much new terms as with your approach)
* do nothing, let the users escape them -> with suggester implementation there will be no problem because suggester can escape it for them (it knows it is a single term) (suggester works with PhraseQuery as well)

I, personally, like the third option.

Why I prefer using `PhraseQuery` to your approach:
* whitespaces are ignored; therefore, no difference between `methodName (param)` and `methodName(param)` whereas your approach will find documents with only the later example
* nearly no additional tokens; therefore, almost no storage impact and negligible performance impact on some queries and the upcoming suggester (the performance of which for some queries depends on the term count)

vladak:
There does not seem to be universal agreement on this one so closing.

PR #2108, opened 2018-05-15
various repository methods should use interactive timeout

vladak:
This change introduces interactive timeout that will be used when:
  - setting new configuration (happens e.g. when webapp is starting or when new configuration is set via `Message` or when `Search` is done using CLI)
  - getting content of a file in given revision
  - getting repository metadata

The first item basically triggers `invalidateRepositories()` in `HistoryGuru` that will in turn call `getRepositories()`. Thus, all the `Repository` calls inside need to be made "interactive".

Basically, I've modified all repositories to use `Executor` as opposed to `Runtime` or `ProcessBuilder` directly. I don't like the way how `Executor` sets up the timeout too much (it should not know about `RuntimeEnvironment` or unset `this.timeout` when it is already set. Maybe it should also have getter/setter for timeout.), however that's for next time.

Also, maybe all the repository annotation/tag parsers should be an interface/abstract class however there are subtle difference how these are implemented for various repositories so for the time being I mirrored the style already used by Bitkeeper repository.

After the change there will be 3 types of timeout related behavior:
  - no timeout (full reindex, i.e. no revision specification)
  - command timeout (incremental reindex, i.e. revision specification or number of entries - in case of Subversion)
  - interactive timeout (as described above)

Tested with unreachable Subversion repository + webapp start and also full reindex.

### DIFF
--- a/src/org/opensolaris/opengrok/history/BazaarRepository.java
+++ b/src/org/opensolaris/opengrok/history/BazaarRepository.java
@@ -23,28 +23,22 @@
  */
 package org.opensolaris.opengrok.history;
 
-import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.Reader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 import org.opensolaris.opengrok.util.Executor;
 
 /**
  * Access to a Bazaar repository.
- *
  */
 public class BazaarRepository extends Repository {
 
@@ -52,7 +46,7 @@ public class BazaarRepository extends Repository {
 
     private static final long serialVersionUID = 1L;
     /**
-     * The property name used to obtain the client command for thisrepository.
+     * The property name used to obtain the client command for this repository.
      */
     public static final String CMD_PROPERTY_KEY
             = "org.opensolaris.opengrok.history.Bazaar";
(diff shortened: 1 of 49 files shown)


src/org/opensolaris/opengrok/history/GitRepository.java
```diff
@@ -544,64 +505,32 @@ private TagEntry buildTagEntry(File directory, String tags) throws HistoryExcept
...
+        Executor executor = new Executor(argv, directory,
+                RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
+        GitTagParser parser = new GitTagParser(tags);
+        executor.exec(true, parser);
+        return parser.getEntries().first();
     }
 
     @Override
-    protected void buildTagList(File directory) {
+    protected void buildTagList(File directory, boolean interactive) {
```
idodeclare: Wondering whether the new `interactive` parameter should possibly be used in this method.

src/org/opensolaris/opengrok/history/BazaarRepository.java
```diff
@@ -295,57 +263,23 @@ boolean hasFileBasedTags() {
      * @param directory Directory where we list tags
      */
     @Override
-    protected void buildTagList(File directory) {
+    protected void buildTagList(File directory, boolean interactive) {
```
idodeclare: Wondering if new `interactive` should be used in this method.

src/org/opensolaris/opengrok/history/MercurialRepository.java
```diff
@@ -613,70 +561,23 @@ boolean hasFileBasedTags() {
     }
 
     @Override
-    protected void buildTagList(File directory) {
+    protected void buildTagList(File directory, boolean interactive) {
```
idodeclare: Should new `interactive` be used in the method?

src/org/opensolaris/opengrok/history/GitRepository.java
vladak: good catch, yes. Also, I've just hit this when testing full reindex on a busy machine where I deliberately set the interactive timeout to 3 seconds.

src/org/opensolaris/opengrok/history/BazaarRepository.java
vladak: ditto

src/org/opensolaris/opengrok/history/MercurialRepository.java
vladak: ditto

vladak:
In the end I decided to properly propagate the interactive flag from `readConfiguration()` so that the interactive mode is not implicit. This resulted in some methods of the `Repository` abstract class to be redefined.

vladak:
As for the `Messages` handling, specifically `ProjectMessage` it might be considered interactive since one usually emits them via command line and expects quick response however given this will be actually exposed via REST API, I am leaving them as non-interactive (i.e. the `getRepositoriesInDir()/getRepository()` calls)

PR #2109, opened 2018-05-15
Messages modularization

Orviss:
OpenGrok has modules and some other modules might be added (e.g. suggester which I'm working on). These other modules might need to listen to some events -> e.g. new project was added, data were reindexed etc. However, this is not easily possible to do right now without adding dependency to the main module in which messages are located. Thus, this PR.

Main features: 
* messages logic moved from `RuntimeEnvironment` to `MessageListener`
* it is now possible to register custom handlers for messages
* messages are now immutable -> so no handler can modify them
* there was a problem with immutability -> used XML encoders/decoders were for JavaBeans, therefore they needed setters so I moved completely from XML to JSON (I think @vladak already experimented with this so I hope it is okay)

I tried to leave all the tests logic unmodified so we might be sure that it works as before.

@tulinkry could you please take a look when you are free? I believe as original author you will have the best insight into it.

Also, I would like to change JSON message conversions in `Util` from `Json Simple` to `Gson` if that's okay with you guys.

**Notice:** 
* I believe this PR is a step forward but I'm not sure you will agree with all of my changes. So please consider it a work in progress.
* Ant is not supported in this. I am not sure if we truly want to support it anymore. If yes, I will add it.

Thanks :)


src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1407,7 +1398,10 @@ public void writeConfiguration(InetAddress hostAddr, int port) throws IOExceptio
     }
 
     protected void writeConfiguration() throws IOException {
-        writeConfiguration(configServerSocket.getInetAddress(), configServerSocket.getLocalPort());
+        if (messageListener == null) {
+            throw new IllegalStateException("Cannot write configuration because ");
```
tulinkry: because what?

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1893,111 +1719,65 @@ public void refreshDateForLastIndexRun() {
...
-                }
-            }, "configurationListener");
-            configurationListenerThread.start();
-        } catch (UnknownHostException ex) {
-            LOGGER.log(Level.WARNING, "Problem resolving sender: ", ex);
-        } catch (IOException ex) {
-            LOGGER.log(Level.WARNING, "I/O error when waiting for config: ", ex);
-        }
+    private void addDefaultMessageHandlers() {
+        messageListener.addHandler(AbortMessage.class, m -> {
```
tulinkry: why not use `addMessageHandler` ?

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,435 @@
...
+            output.write(Message.DeliveryStatus.ERROR.getStatusCode());
+            output.write(ex.getMessage().getBytes());
+            return;
+        }
+
+        LOGGER.log(Level.FINER, "Message received: {0}", m.getTags());
+        LOGGER.log(Level.FINER, "Messages in the system: {0}", getMessagesInTheSystem());
+
+        output.write(Message.DeliveryStatus.OK.getStatusCode());
+        output.write(gson.toJson(response).getBytes());
```
vladak: can you abstract the encoding/decoding ? `message.getEncoded()` ?

Orviss: What do you mean exactly? Messages encoding and decoding is in different part of the code. Either in `Message` class or in `MessageReader/MessageWriter`. This part just sends back the generic `Response`.

vladak: I mean - could the `gson` part be constrained just to `Message` class ? i.e. is it necessary to call gson directly in this place ?

vladak:
Could you post some example of encoded Message ? I'd like Messages to be an API in the future so encoding will matter. Also, in my set of changes, I had a test for idempotency of the encode/decode operations.

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
Orviss: Yes, because I am marshalling `Response` and not `Message`. To clear things up, you just want me to add method `getEncoded()` to `Response` class so there would not have to be `gson` field in the `MessageListener`, right?

vladak: I.e. to make it easily replaceable in the future with different encoding mechanism (once JSON ceases being hip).

Orviss:
Default `NormalMessage`:
```json
{"cssClass":"info","tags":["main"],"created":{"seconds":1526989567,"nanos":865846000},"expiration":{"seconds":1526990167,"nanos":865846000},"type":"org.opensolaris.opengrok.configuration.messages.NormalMessage"}
```

Also, the test in `MessageTest` for idempotency is not enough?

src/org/opensolaris/opengrok/configuration/messages/Response.java
```diff
@@ -0,0 +1,64 @@
...
+/*
+ * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opensolaris.opengrok.configuration.messages;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class Response {
```
vladak: add comment pls about what it is and how it is used

Orviss: I'm not a big fun of comments for no library-like code. Comment like 
```java
/**
 * Generic message response.
 */
```
is useless because it can be immediately seen from the package and the class name.

However, if I were to write
```java
/**
 * Generic message response. It allows to combine multiple String responses into one because...
 */
```
then I'm going into details which are out of scope of the class.

vladak:
Yep.

Dne út 22. 5. 2018 13:41 uživatel Adam Hornáček <notifications@github.com>
napsal:

vladak:
What does "cssClass" stand for ? Also, the time stamps are not particularly user friendly. Could you make them human readable ? Also, nanosecond precision is definitely not needed.

vladak:
In my case the JSON looked like this:

```json
{"text":"foo","className":"Normal","tags":["tag1","tag2"],"created":"May 22, 2018 2:25:04 PM","expiration":"May 22, 2018 2:35:04 PM"}
```

Orviss:
`cssClass` is just css class which should be added to the element if they are shown on the web. I don't know why it is used this way. I did not like it either but I let it there for the sake of preserving functionality for the time being.

The problem with the timestamps: 

Previously there was `Date` class which had millisecond precision but the class is pretty old and most of it is deprecated. 

However, the problem is with the interpretation. Why there should be a date ? It is just used for checking how long the message should stay in the system and then it should be deleted. We do not care if it should end on May 1st. It is irrelevant. Therefore, as stated in `Instant` it represents `An instantaneous point on the time-line` when it was created and when it should be expired. 

Also, why should JSON be user friendly? It is just an interchange format.

Also, changing precision in `JSON` format makes the decoded message not equal to the message which was encoded. I could look up some other class with lesser precision.

With date there is a lot of hidden issues, e.g. timezones.

Other solution which makes sense is for `createdDate` to be <https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html> and change `expirationDate` to <https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html>. However, we will not save much since `LocalDateTime` has nanoseconds precision as well.

Other solution is to `strip` the nanoseconds precision in `Message` class but it seems weird.

Orviss:
But with your example the decoded message will not be equal to the previous one because of the `Date`'s milliseconds precision.

vladak:
Exactly, that's why I had to override the equality check for date.

vladak:
As for human readability, my goal is to make `Message` an API which implies user friendly (take a look at https://developer.github.com/v3/ for example - the API can be easily comprehended by a human, some of it can be typed by hand even).

Orviss:
If the user friendliness is the goal then maybe we could set `created` as `transient`, it should be server centric anyway. (However, I'm not sure about backwards compatibility) And change expiration to <https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html>?

Or if you prefer then I can add specific encoder/decoder for `Instant` and modify the `equals()` method.

vladak:
I am fine with either method. The latter is what I used in my branch, it is not particularly nice.

Orviss:
I like the solution with `Duration` better because you can send the same message multiple times over some period of time and you would not have to modify the dates each time. And also we would not have to care about issues if `created` was in the future or way in the past.

Orviss:
New message format:
```json
{"cssClass":"info","tags":["main"],"duration":"PT10M","type":"org.opensolaris.opengrok.configuration.messages.NormalMessage"}

```

tulinkry:
> cssClass is just css class which should be added to the element if they are shown on the web. I don't know why it is used this way. I did not like it either but I let it there for the sake of preserving functionality for the time being.

You can rethink this in any way you want. It's like this because we originaly started with normal and abort message just to display some info on the web page and then the messages kinda took over the opengrok flow.

So it's very specific content for normal and abort message and you can freely redo the design.

tulinkry:
> If the user friendliness is the goal then maybe we could set created as transient, it should be server centric anyway. (However, I'm not sure about backwards compatibility) And change expiration to https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html?

When I wrote this, the created wasn't really important field.

tulinkry:
It's funny but for

> New message format:
> {"cssClass":"info","tags":["main"],"duration":"PT10M","type":"org.opensolaris.opengrok.confi..

you need to study other resources on how to properly write the duration string. For me for example the unix timestamp is way more intuitive. But let's stick to this new format.

Orviss:
I would like more `10m` or `10s` as well but this is how guys at `ISO` declared it :D

src/org/opensolaris/opengrok/configuration/messages/Response.java
tulinkry: but the purpose for the opengrok and the usage in the opengrok is not self-evident

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+            // if closed then it is okay (socket was closed while blocked in accept())
+            if (!configServerSocket.isClosed()) {
+                LOGGER.log(Level.SEVERE, "Socket error", e);
+            }
+        } catch (IOException e) {
+            LOGGER.log(Level.SEVERE, "Error reading config file: ", e);
+        } catch (RuntimeException e) {
+            LOGGER.log(Level.SEVERE, "Error parsing config file: ", e);
+        } catch (ValidationException e) {
+            LOGGER.log(Level.WARNING, "Received invalid message", e);
```
tulinkry: if there is unhandled exception it will cause that opengrok will need to be restarted, isn't it better to check for Exception or Throwable?

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+     * Handle incoming message.
+     *
+     * @param m message
+     * @param output output stream for errors or success
+     */
+    private void handleMessage(final Message m, final OutputStream output) throws IOException, ValidationException {
+        m.validate();
+
+        if (isMessageLimitExceeded()) {
+            LOGGER.log(Level.WARNING, "Message dropped: {0} - too many messages in the system", m.getTags());
```
tulinkry: here you use tags and below you use the message itself. ensure consistency

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+            output.write(Message.DeliveryStatus.OVER_LIMIT.getStatusCode());
+            return;
+        }
+
+        Response response;
+        try {
+            response = processMessage(m);
+        } catch (Exception ex) {
+            LOGGER.log(Level.WARNING, String.format("Message dropped: %s - message error", m), ex);
```
tulinkry: here

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+            output.write(Message.DeliveryStatus.ERROR.getStatusCode());
+            output.write(ex.getMessage().getBytes());
+            return;
+        }
+
+        LOGGER.log(Level.FINER, "Message received: {0}", m.getTags());
```
tulinkry: and here

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+        Response finalResponse = Response.empty();
+        synchronized (messageHandlers) {
+            List<MessageHandler> handlers = messageHandlers.get(messageType);
+            if (handlers != null) {
+                for (MessageHandler handler : handlers) {
+                    try {
+                        Response response = processMessage(handler, message);
+                        finalResponse = finalResponse.combine(response);
+                    } catch (Exception e) { // exception in handler should not stop notifying other handlers
+                        LOGGER.log(Level.WARNING, "Exception while processing message", e);
```
tulinkry: add some message info

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+        return finalResponse;
+    }
+
+    private Response processMessage(final MessageHandler handler, final Message message) {
+        Response response;
+        try {
+            response = handler.handle(message);
+        } catch (MessageHandler.HandleException e) {
+            response = Response.of(e.getMessage());
+            LOGGER.log(Level.WARNING, "Exception while processing message", e);
```
tulinkry: add some message info

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -282,113 +188,214 @@ public boolean equals(Object obj) {
      * @param port port number
```
tulinkry: this is not XML anymore

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -282,113 +188,214 @@ public boolean equals(Object obj) {
...
+        private Duration duration;
+
+        public Builder(final Class<T> cl) {
+            messageClass = cl;
+        }
+
+        public T build() {
+            T result = createMessage(messageClass);
+
+            ((Message) result).tags.addAll(tags);
```
tulinkry: can you avoid the casting so many times?

src/org/opensolaris/opengrok/configuration/messages/Messages.java
```diff
@@ -201,8 +201,8 @@ private static final void a_usage() {
         System.err.println("Id");
         System.err.println("-i <id>              set the id of the message (default is auto-generated)");
```
tulinkry: is this used?

src/org/opensolaris/opengrok/configuration/messages/util/MessageWriter.java
```diff
@@ -0,0 +1,56 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+ /*
+ * Copyright (c) 2016, 2018 Oracle and/or its affiliates. All rights reserved.
```
tulinkry: is this 2016?

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -57,95 +57,48 @@
  */
 public abstract class Message implements Comparable<Message> {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(Message.class);
+    public static final String DELIMITER = "\0";
```
tulinkry: is it safe that the delimiter can not occur in the middle of the json?

src/org/opensolaris/opengrok/configuration/messages/handler/ProjectMessageHandler.java
```diff
@@ -0,0 +1,330 @@
...
+     * because it does not have access to the currently used RuntimeEnvironment.
+     */
+    private void validateMore(RuntimeEnvironment env, Message message) throws ValidationException {
+        String command = message.getText();
+
+        if (!env.isProjectsEnabled()) {
+            throw new ValidationException("Projects have to be enabled in order to use this message");
+        }
+
+        // The same check can be done for the "delete" command however it is better
```
tulinkry: same as what?

src/org/opensolaris/opengrok/configuration/messages/handler/ProjectMessageHandler.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    private void addProjects(final Set<String> projectNames) {
+        for (String projectName : projectNames) {
+            File srcRoot = env.getSourceRootFile();
+            File projDir = new File(srcRoot, projectName);
+
+            if (!env.getProjects().containsKey(projectName)) {
+                Project project = new Project(projectName, "/" + projectName);
+                project.setTabSize(env.getConfiguration().getTabSize());
```
tulinkry: what about `project.completeWithDefaults(env.getConfiguration());`

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -1234,17 +1234,17 @@ public static void printMessages(Writer out, SortedSet<Message> set, boolean lim
      * @return json array containing the set of messages
      */
     @SuppressWarnings("unchecked")
-    public static JSONArray messagesToJson(SortedSet<Message> set) {
+    public static JSONArray messagesToJson(SortedSet<AcceptedMessage> set) {
         JSONArray array = new JSONArray();
         SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z");
-        for (Message m : set) {
+        for (AcceptedMessage m : set) {
```
tulinkry: aren't we using gson now?
call your encode/decode on the message?

tulinkry: but keep in mind this will break the ui as it needs to be adjusted too

test/org/opensolaris/opengrok/configuration/messages/ConfigMessageTest.java
```diff
@@ -22,234 +22,192 @@
...
-        byte[] out = m.apply(env);
-        String newconfStr = new String(out);
-        Assert.assertEquals(configStr, newconfStr);
-    }
+        builder.clearTags();
+        builder.setText(null);
+        builder.addTag("getconf");
 
-    @Test(expected = IOException.class)
-    public void testApplySetInvalidMethod() throws Exception {
```
tulinkry: why are these and some other tests not needed anymore?

tulinkry: nevermind, i found them

tulinkry:
And to sum it up to add a new message following must be taken:
1. implement a new message class (possibly on the validation)
2. implement a new message handler for this class

is this correct?

src/org/opensolaris/opengrok/configuration/messages/Messages.java
```diff
@@ -18,16 +18,19 @@
  */
 
 /*
- * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2018, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: update wiki https://github.com/oracle/opengrok/wiki/Messages#examples if anything changed

tools/Messages
```diff
@@ -53,13 +53,12 @@ Usage()
     echo ""
```
tulinkry: update https://github.com/oracle/opengrok/tree/master/tools/sync if any of them use `-e or --expire`

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+
+    private int messagesInTheSystem = 0;
+
+    private int messageLimit = DEFAULT_MESSAGE_LIMIT;
+
+    private Timer expirationTimer;
+
+    public void addMessageHandler(final Class<? extends Message> messageType, final MessageHandler handler) {
+        if (messageType == null) {
+            throw new IllegalArgumentException("Cannot add handler for null message type");
```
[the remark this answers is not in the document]
Orviss: https://stackoverflow.com/questions/3881/illegalargumentexception-or-nullpointerexception-for-a-null-parameter

src/org/opensolaris/opengrok/configuration/messages/handler/ProjectMessageHandler.java
Orviss: I don't know, this copied is copied :D

tools/Messages
Orviss: I was looking for that but I didn't find any. Message calls are in sync config.

src/org/opensolaris/opengrok/configuration/messages/Messages.java
Orviss: I will as soon as this is integrated :)

Orviss:
> To sum it up to add a new message following must be taken:
> 
> implement a new message class (possibly on the validation)
> implement a new message handler for this class
> is this correct?
yes

src/org/opensolaris/opengrok/web/Util.java
Orviss: Okay, I will try to do it :)

src/org/opensolaris/opengrok/configuration/messages/handler/ProjectMessageHandler.java
Orviss: Again, copied code, I will look into it.

src/org/opensolaris/opengrok/configuration/messages/Message.java
Orviss: <https://www.json.org> I don't see `\0` there...

tulinkry: perfect

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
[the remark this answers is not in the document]
tulinkry: ok

src/org/opensolaris/opengrok/configuration/messages/handler/ProjectMessageHandler.java
tulinkry: then delete it or find it out 🍡

src/org/opensolaris/opengrok/web/Util.java
tulinkry: you can start with using gson (to replace the json simple) and the ui changes can come in another pr

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,497 @@
...
+     * @param endpoint The socket address to listen on
+     * @throws IOException if the endpoint is not available
+     */
+    public void start(SocketAddress endpoint) throws IOException {
+        if (isListening()) {
+            throw new IllegalStateException("Cannot start when already listening for messages");
+        }
+        bindToAddress(endpoint);
+        new Thread(() -> {
+            while (!configServerSocket.isClosed()) {
```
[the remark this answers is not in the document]
Orviss: There is no explicit reference to this thread so nobody can call `interrupt()` explicitly. However, I'm not sure how to do this properly. If you think that it should be added I can look into it in more detail.

src/org/opensolaris/opengrok/configuration/messages/Messages.java
Orviss: Nope, I will remove it, nice catch.

src/org/opensolaris/opengrok/configuration/messages/util/MessageWriter.java
Orviss: I started refactoring the previous implementation and forgot about this part :D I will change it to 2018 since it is nothing alike the previous one and there is no history...

src/org/opensolaris/opengrok/configuration/messages/Message.java
Orviss: I can use another variable (of `Message` type -> I actually had it like that at first) but that seemed even worse because it was hiding the intent. Also I would like to avoid casting to `T` since that would cause a warning.

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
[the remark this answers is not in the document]
tulinkry: seems ok, socket is interupted with SocketException
I was more confused about parent thread exiting and possibly interupting this child thread but it appears that it doesn't happen

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: And this won't work? I see that it kinda tweaks the approach that one changes some variable which isn't returned at the end of the function but it avoids the casting which is not necessary here to perform it 10 times.

```java
T result = createMessage(messageClass);
Message resultMessage = ((Message) result);

resultMessage.tags.addAll(tags);

if (resultMessage.tags.isEmpty()) {
    Set<String> defaultTags = result.getDefaultTags();
    if (defaultTags != null) {
        resultMessage.tags.addAll(defaultTags);
    }
}

if (text != null) {
    resultMessage.text = text;
}

if (cssClass != null) {
    resultMessage.cssClass = cssClass;
} else {
    resultMessage.cssClass = result.getDefaultCssClass();
}
if (duration != null) {
    resultMessage.duration = duration;
}
return result;
```

Orviss: This is how I had it before but I didn't like it that much because it is weird a little bit :D if you look at the end it seems weird that you are returning `result` but updating `resultMessage`

Not sure what is better though...

Maybe someone else could decide? :D

tulinkry: @lkosco

tulinkry: Or maybe you can introduce a method with parameter Message and fill the fields inside and call this method here and it does not seem that awful suddenly :-D

Orviss: Yeah, that will work :D

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -339,29 +339,32 @@ public Builder(final Class<T> cl) {
 
         public T build() {
             T result = createMessage(messageClass);
+            fillData(result);
```
tulinkry: I wonder if now it works without the cast why it didn't before when you had it inline

Orviss: Now it works without cast because `T` is covariant type to `Message`. However, before was the problem that Java somehow could not deduce that the modifications to `private` members were done inside the enclosing class which has the private members. I don't know the reason why :D

tulinkry: ok :D seems fun

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,503 @@
...
+            output.write(Message.DeliveryStatus.ERROR.getStatusCode());
+            output.write(ex.getMessage().getBytes());
+            return;
+        }
+
+        LOGGER.log(Level.FINER, "Message received: {0}", m);
+        LOGGER.log(Level.FINER, "Messages in the system: {0}", getMessagesInTheSystem());
+
+        output.write(Message.DeliveryStatus.OK.getStatusCode());
+        output.write(gson.toJson(response).getBytes());
```
vladak: `response.encode().getBytes()` ?

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,503 @@
...
+    /**
+     * Immediately remove all messages in the application.
+     */
+    public void removeAllMessages() {
+        tagMessages.clear();
+        messagesInTheSystem = 0;
+    }
+
+    /**
+     * Remove all messages containing at least on of the tags.
```
vladak: on -> one

test/org/opensolaris/opengrok/configuration/messages/MessageTestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
+import org.opensolaris.opengrok.configuration.messages.MessageListener.AcceptedMessage;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.time.Duration;
+
+public class MessageTestUtils {
```
vladak: I gather this class is used for testing only ?

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -260,135 +176,238 @@ public boolean equals(Object obj) {
...
+        public Duration read(final JsonReader reader) throws IOException {
+            if (reader.peek() == JsonToken.NULL) {
+                reader.nextNull();
+                return null;
+            }
+            return Duration.parse(reader.nextString());
+        }
+    }
+
+    public static class Builder<T extends Message> {
```
vladak: I wonder what is the advantage of this builder compared to having a message factory class ?

src/org/opensolaris/opengrok/configuration/messages/Message.java
```diff
@@ -260,135 +176,238 @@ public boolean equals(Object obj) {
...
+                if (defaultTags != null) {
+                    message.tags.addAll(defaultTags);
+                }
+            }
+
+            if (text != null) {
+                message.text = text;
+            }
+
+            if (cssClass != null) {
```
vladak: can you rename the `css` property so it is more descriptive ?

src/org/opensolaris/opengrok/configuration/messages/Response.java
```diff
@@ -0,0 +1,64 @@
...
+    public static Response empty() {
+        return new Response(Collections.emptyList());
+    }
+
+    public Response combine(final Response other) {
+        if (other == null) {
+            throw new IllegalArgumentException("Cannot combine with null response");
+        }
+
+        List<String> combinedMessages = new ArrayList<>(data.size() + other.data.size());
```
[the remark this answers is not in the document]
tulinkry: maybe more `combinedData` with respect to the class fields

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: but that would be misleading, one could thing that can pass arbitrary css selectors into this property

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    /**
+     * Start a thread to listen on a socket to receive new messages.
+     * The messages can contain various commands for the webapp, including
+     * upload of new configuration.
+     * @param endpoint The socket address to listen on
+     * @throws IOException if the endpoint is not available
+     */
+    public void start(SocketAddress endpoint) throws IOException {
```
vladak: `startListenerThread` ?

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: but that would be misleading, one could thing that can pass arbitrary css selectors into this property

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: I like this approach more with the builder

src/org/opensolaris/opengrok/configuration/messages/Message.java
Orviss: Do you have any suggestion? Before there was just `class`. `cssClassAddedToWebMessageWindow`? :D

tulinkry: best would be to rethink this and hide it for example in message.extras (encoded as json and translated to map)

src/org/opensolaris/opengrok/configuration/messages/NormalMessage.java
```diff
@@ -18,37 +18,38 @@
...
 package org.opensolaris.opengrok.configuration.messages;
 
-import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
+import java.util.Collections;
+import java.util.Set;
 
 /**
  * @author Kryštof Tulinger
  */
 public class NormalMessage extends Message {
```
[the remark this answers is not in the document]
Orviss: This will break `Messages` usage. No problem with that? Now there will have to be `-n info` option. Maybe because it is default there won't be problem with that?

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: which would be generic and wouldn't kinda full the other messages with non-sence css class property

test/org/opensolaris/opengrok/configuration/messages/MessageTestUtils.java
Orviss: Yes, it's in `test` directory.

src/org/opensolaris/opengrok/configuration/messages/Message.java
Orviss: I can add it to `NormalMessage`. Maybe it is also used by `AbortMessage`? Not sure though. I will check.

src/org/opensolaris/opengrok/configuration/messages/NormalMessage.java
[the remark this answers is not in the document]
tulinkry: I'm for this definitely.

src/org/opensolaris/opengrok/configuration/messages/MessageListener.java
Orviss: `Listener` is already in class name. That it starts different thread or process or something else should not be visible outside the class IMHO. To assure consistency I would rather rename `stopListenerThread` to just `stop`. However, if you prefer `startListenerThread`, I will add it there.

src/org/opensolaris/opengrok/configuration/messages/Message.java
tulinkry: Yes this is just for normal and abort.

src/org/opensolaris/opengrok/configuration/messages/Message.java
Orviss: Well, Google Protobufs use Builder as well and I kind of liked it when I was using those. 

However, I like Builder more because now you could create a lot of new message types. Each a little different and you would not have to have one big factory class. 

However, each have pros and cons. If you prefer factory class I can look into it in more detail.

src/org/opensolaris/opengrok/configuration/messages/Response.java
Orviss: Could you please help me out? I'm still not sure what exactly is missing because for me as an author it is self-evident.

tulinkry: And I'm not kidding this happened to me - I saw a Response and I had to go and search where it's used, what kind of data it transports and what to expect from it when looking into this PR.

src/org/opensolaris/opengrok/configuration/messages/Message.java
vladak: I like the Builder pattern too, just am not aware of its pros and cons. The one-line usage looks nice, as is using reflection to get the constructor which saves space and is more flexible compared to the factory class.

test/org/opensolaris/opengrok/configuration/messages/MessageTestUtils.java
vladak: oh, okay :-)

Orviss:
I think this PR does not provide much improvement since we are planning to move to REST anyway. I even think it might be a little counterproductive. Therefore, closing :)

vladak:
Actually, I liked it as it would provide good foundation.

Dne po 4. 6. 2018 18:11 uživatel Adam Hornáček <notifications@github.com>
napsal:

Orviss:
I looked at the <https://github.com/jersey/jersey> and the implementation was really smooth. I actually tested it quickly out (configuration worked nicely). However, many things in here would not apply to the Rest API, there is no need for `MessageListener`, `handlers`, `Messages`, `validation`, `Builder`, etc.

vladak:
Okay. I thought you needed the Message handlers for your suggester code.

PR #2117, opened 2018-05-18
Add link to demo on gh-pages

Orviss:
Changes:
* updated the template from <https://github.com/orderedlist/minimal> (I couldn't do merge because of no common history and with rebase I couldn't create PR)
* changed html formatting because it was a mess at some places (default formatting by Idea)
* removed images
* removed link for OpenGrok users and added link to wiki installations (more easily maintainable)
* added link to demo
* moved releases link to be more visible

Final look:
screen shot 2018-05-18 at 16 52 19

I am no designer so if you have any suggestions for improvements I will be more than glad to implement them :) 

Thanks :)


tarzanek:
I actually did like the picture ... a picture is worth a thousand words ...
can we get it back (obviously some newer screenshot)
please?

Orviss:
Isn't demo enough?

Also, now we do not have to change the picture if some UI changes occur.

Regardless, I think the UI is not OpenGrok's strongest feature.

vladak:
Yeah, screenshots are so 90's :-) These days we should hire a youtuber :-)

Dne po 21. 5. 2018 18:13 uživatel Adam Hornáček <notifications@github.com>
napsal:

PR #2120, opened 2018-05-21
Add support for Tomcat webapp context in deploy command

Orviss:
Partly fixes <https://github.com/oracle/opengrok/issues/2113>

This was needed for demo site. Therefore, I created just a quick fix for Tomcat. I could not find how it should be done for GlassFish or Resin. If you know then please let me know :) 

I tried to conform to this text in Tomcat documentation:
`
If the context path is the empty string then the base name will be ROOT (always in upper case) otherwise the base name will be the context path with the leading '/' removed and any remaining '/' characters replaced with '#'
`

If `OPENGROK_WEBAPP_CONTEXT` variable is not set then `source` is used by default. Otherwise I try to follow the Tomcat instructions specified above.

I program in shell very occasionally so there will be probably some mistakes. So please look at it carefully :)

Also, I do not know what are the conventions for variable names or `then` position in `if` statements. Are there even any?

**Note:**
I had to use `! -z ${OPENGROK_WEBAPP_CONTEXT+x}` construct instead of `-n ${OPENGROK_WEBAPP_CONTEXT+x}` because the later did not work as needed.

Testing:
I tried some basic contexts and they seemed to work. Not sure if some error proof constructs should be added there or not. If yes, let me know :) 

Thanks :)

### DIFF
--- a/OpenGrok
+++ b/OpenGrok
@@ -983,8 +983,30 @@ DeployWar()
         FatalError "Missing Deployment Directory ${warTarget}"
     fi
 
+    APP_CONTEXT="source"
+    FINAL_WAR_NAME="source"
+    if [ "$applicationServer" = "Tomcat" ]
+    then
+        if [ ! -z ${OPENGROK_WEBAPP_CONTEXT+x} ]
+        then
+            APP_CONTEXT="${OPENGROK_WEBAPP_CONTEXT}"
+
+            # strip leading /
+            case ${APP_CONTEXT} in
+                /*) APP_CONTEXT=`echo ${APP_CONTEXT} | sed 's|^/||'`
+            esac
+
+            FINAL_WAR_NAME=`echo ${APP_CONTEXT} | sed 's|/|#|g'`
+
+            if [ -z "${FINAL_WAR_NAME}" ]
+            then
+                FINAL_WAR_NAME="ROOT"
+            fi
+        fi
+    fi
+
     Progress "Installing ${OPENGROK_DIST_WAR} to ${warTarget} ..."
-    ${DO} cp -p "${OPENGROK_DIST_WAR}" "${warTarget}/"
+    ${DO} cp -p "${OPENGROK_DIST_WAR}" "${warTarget}/${FINAL_WAR_NAME}.war"
     if [ $? != 0 ]
     then
         FatalError "Web Application Installation FAILED"
@@ -1011,7 +1033,7 @@ DeployWar()
         if [ -n "${EXTRACT_COMMAND}" ]
         then
             cd "${warTarget}"
-            eval "${EXTRACT_COMMAND} ${warTarget}/source.war WEB-INF/web.xml"
+            eval "${EXTRACT_COMMAND} ${warTarget}/${FINAL_WAR_NAME}.war WEB-INF/web.xml"
             if [ "${OPENGROK_INSTANCE_BASE}" != '/var/opengrok' ]
             then
                 sed -e 's:/var/opengrok/etc/configuration.xml:'"$XML_CONFIGURATION"':g' \
@@ -1028,7 +1050,7 @@ DeployWar()
 		mv "${warTarget}/WEB-INF/web.xml.tmp" \
 		    "${warTarget}/WEB-INF/web.xml"
             fi
-            eval "${COMPRESS_COMMAND} ${warTarget}/source.war WEB-INF/web.xml"
+            eval "${COMPRESS_COMMAND} ${warTarget}/${FINAL_WAR_NAME}.war WEB-INF/web.xml"
             rm -rf "${warTarget}/WEB-INF"
         fi
     fi
@@ -1039,7 +1061,7 @@ DeployWar()
     Progress "running, or wait until it loads the just installed web " \
         "application."
     Progress
-    Progress "OpenGrok should be available on <HOST>:<PORT>/source"
+    Progress "OpenGrok should be available on <HOST>:<PORT>/${APP_CONTEXT}"
     Progress "  where HOST and PORT are configured in ${applicationServer}."
     Progress
 }
--- a/OpenGrok
+++ b/OpenGrok
@@ -18,7 +18,7 @@
 # CDDL HEADER END
 
 #
-# Copyright (c) 2008, 2017, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


OpenGrok
```diff
@@ -995,8 +995,30 @@ DeployWar()
         FatalError "Missing Deployment Directory ${warTarget}"
     fi
 
+    APP_CONTEXT="source"
+    FINAL_WAR_NAME="source"
+    if [ "$applicationServer" = "Tomcat" ]
+    then
+        if [ ! -z ${OPENGROK_WEBAPP_CONTEXT+x} ]
```
tarzanek: what is x ?

vladak:
The style of added code should match the neighboring code, that's the only rule we have for now.

OpenGrok
vladak: I think you can do just

```shell
if [ -n "${OPENGROK_WEBAPP_CONTEXT}" ]
```

Orviss:
Style is quite hard to determine since multiple styles are used, e.g. `if [ ... ]; then` or `if [] \n then`, also `${VAR}` or ${myVar}.

vladak:
That's why I said "neighboring" :-) Anyway, this will be rewritten into Python one day so do not bother with style too much.

OpenGrok
Orviss: `if [ -n "${OPENGROK_WEBAPP_CONTEXT}" ]` did not work :(

Probably it will be good to add a comment there.

vladak:
That's strange. What is your shell ?

Dne po 21. 5. 2018 21:17 uživatel Adam Hornáček <notifications@github.com>
napsal:

OpenGrok
Orviss: Sorry, I noticed that you did not include parameter expansion.

This code check if the variable has zero length, not if it is set.
```bash
if [ -n "${OPENGROK_WEBAPP_CONTEXT}" ]
```

I would like to distinguish two cases:
`OPENGROK_WEBAPP_CONTEXT` is not set -> use `source` context path
`OPENGROK_WEBAPP_CONTEXT` is empty -> use root context path -> `/`

vladak:
While we are there, I remember some discussion about having to reindex when the app context changed. The source code of `Indexer.java` and the help in the `OpenGrok` shell script contain comments in this regard however I wonder if this is still true.

OpenGrok
```diff
@@ -995,8 +995,30 @@ DeployWar()
...
+        then
+            APP_CONTEXT="${OPENGROK_WEBAPP_CONTEXT}"
+
+            # strip leading /
+            case ${APP_CONTEXT} in
+                /*) APP_CONTEXT=`echo ${APP_CONTEXT} | sed 's|^/||'`
+            esac
+
+            FINAL_WAR_NAME=`echo ${APP_CONTEXT} | sed 's|/|#|g'`
```
vladak: how does this interact with the code below ?

```shell
    # OPTIONAL: Context URL of the OpenGrok webapp                              
    #           (default is /source)                                            
    WEBAPP_CONTEXT=""                                                           
    if [ -n "${OPENGROK_WEBAPP_CONTEXT}" ]; then                                
        WEBAPP_CONTEXT="-w ${OPENGROK_WEBAPP_CONTEXT}"                          
    fi
```

OpenGrok
vladak: Fair enough. I don't like how both the `Configuration` and the shell script contain the `source` default however I cannot think of anything better for now.

Orviss:
At a first glance, everything seems to be working when I tried different context without reindexing.

OpenGrok
Orviss: What do you mean exactly? I used different variable names just to be safe. Also, the check
```bash
if [ -n "${OPENGROK_WEBAPP_CONTEXT}" ]; then 
```
does not seem to set the variable to empty value. Otherwise it would not work, right? Because this code is always executed.

vladak: I was curious about the non-initial slash substitution. Should the parsing done in `Indexer#parseOptions()` be in line with the substitutions done in the script ? i.e. what happens if `OPENGROK_WEBAPP_CONTEXT` is set to `foo/bar` ?

vladak:
The question can be rephrased: do the xrefs contains hard-coded context path ?

Orviss:
Unfortunately they do :( <http://demo.opengrok.org/xref/opengrok/src/org/opensolaris/opengrok/analysis/JFlexXrefUtils.java?r=a9f20c1f#290>

vladak:
Okay, no worries, I was just unsure if anything has changed there.

OpenGrok
Orviss: `Indexer#parseOptions()` does not seem to do much -> only adds `/` before and after the `OPENGROK_WEBAPP_CONTEXT` if needed and then uses the value like `${value}s?` for creating `href`s

tarzanek:
well I am not sure if they really do contain hardcoded stuff ... I thought per discussion with @splitbrain in
https://github.com/OpenGrok/docker/pull/9
that they don't contain anything hardcoded ... @idodeclare had a patch for this, I think we even merged it ...

Orviss:
Sample line from my OpenGrok index:
```html
<a class="l" name="34" href="#34">34</a><span class='fold-space'>&nbsp;</span>@<a href="/source/s?defs=SuppressWarnings&amp;project=OpenGrok" class="intelliWindow-symbol" data-definition-place="undefined-in-file">SuppressWarnings</a>(<span class="s">&quot;PMD.AvoidThrowingRawExceptionTypes&quot;</span>)
```
`/source/s?` is there. Maybe I have a few month old index though... Not sure since I'm always going between my Maven version and others :D (and have not run indexer for a while)

idodeclare:
@tarzanek, I hadn’t raised that PR again, which was waiting for the now-completed LOC feature. I’ll look to do it tonight.

The file system-cached xref content will remain the same—I.e. with `/source/?—but will be able to be served from arbitrary web app deployments. And thus option `-w` will be killed per the 20 Jan 2010 request of some fellow named Lubos.

tarzanek:
ah ... sorry for the merges, can you resolve the conflicts @Orviss ?
we can merge after that

tia
L

Orviss:
Rebased. Added back `OPENGROK_WEBAPP_CONTEXT` variable description for correct deployment to Tomcat.

Thanks :)

tarzanek:
;) thank you
let's see if this will work for  @splitbrain fully
merging

PR #2126, opened 2018-05-28
wait for output thread to finish reading

vladak:
Lately, the mirroring on our production instance started failing with strange conditions such as `Messages` not being able to retrieve source root or because a branch of a Mercurial repository could not be determined. These commands exited with 0 however the output was empty. It might be because the server is more loaded than usual due to full reindex running in background so thread scheduling is not going in its usual lines.

I think this is caused by an outstanding TODO item (another reason to get rid of these comments and file bugs for each one of them) in `command.py` - if a process exits too quickly, the main thread will not wait for the output thread to read the full output. As a result these single line cases mentioned above will complete with empty output.

### DIFF
--- a/tools/sync/command.py
+++ b/tools/sync/command.py
@@ -168,7 +168,7 @@ class Command:
                 return
 
         timeout_thread = None
-        sema = asyncio.Semaphore(value=1)
+        sema = asyncio.Semaphore(value=0)
         output_thread = OutputThread(sema, self.logger)
         try:
             start_time = time.time()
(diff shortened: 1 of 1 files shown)


tools/sync/command.py
```diff
@@ -163,7 +168,8 @@ def close(self):
                 return
 
         timeout_thread = None
-        output_thread = OutputThread()
+        sema = asyncio.Semaphore(value=0)
```
idodeclare: The completion of an `OutputThread` seems more strictly Boolean than countable; can perhaps `threading.Event` be used here versus a counting `Semaphore`?

vladak: I think so. It seems the semantics of set() vs. wait() should be the same as for semaphores - if the internal flag is set, then wait() will not block at all.

PR #2128, opened 2018-05-28
Add search statistics

Orviss:
Added search statistics:
* average number of search hits

These stats could be useful for the upcoming suggester.

I also changed a few concepts:
* divided statistics saving and statistics reporting -> these are in my opinion 2 different things, one saves the Statistics object and all its inner fields and the other creates reports based on this data
* now it will be easier to add different report formats -> e.g. `Prometheus`
* replaced the JSON Simple with Gson

`Statistics` methods were a little counter-intuitive because the same request could be added multiple times and there were quite big differences between `addRequest()`, `addRequest(String)` and `addRequestTime(String)`. Request counters were used for ordinary counting (not just requests).

**Note:**
Backwards compatibility is not preserved. Statistics will be reset if they could not be loaded. 

Thanks :)


src/org/opensolaris/opengrok/web/AuthorizationFilter.java
```diff
@@ -35,6 +35,7 @@
 import javax.servlet.http.HttpServletResponse;
 import org.opensolaris.opengrok.configuration.Project;
 import org.opensolaris.opengrok.logger.LoggerFactory;
+import org.opensolaris.opengrok.web.stats.StatisticsFilter;
```
tulinkry: ?

src/org/opensolaris/opengrok/web/WebappListener.java
```diff
@@ -50,11 +48,6 @@
         implements ServletContextListener, ServletRequestListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WebappListener.class);
-    private static final String ENABLE_AUTHORIZATION_WATCH_DOG = "enableAuthorizationWatchDog";
```
tulinkry: where did this go?

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
```diff
@@ -943,48 +944,42 @@ public void testCanAcceptMessage() throws Exception {
...
-            + "\"requests_per_minute_max\":3,"
+            + "\"requestsPerMinuteMax\":3,"
             + "\"timing\":{"
                 + "\"*\":2288,"
                 + "\"xref\":53,"
                 + "\"root\":2235"
             + "},"
             + "\"minutes\":756,"
-            + "\"timing_min\":{"
+            + "\"timingMin\":{"
```
tulinkry: TIMING_MIN does not indicate this, am I missing something?

src/org/opensolaris/opengrok/web/AuthorizationFilter.java
Orviss: It was added by Idea because of the refactoring. You can see there is a java doc without full name. Better to change the javadoc. I will repair it.

src/org/opensolaris/opengrok/web/WebappListener.java
Orviss: It was not used at all.

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
tulinkry: i guess there is a magic in the gson stuff?

Orviss: This is encoded Statistics which reflect the attribute names in the class.

Orviss: This is encoded Statistics which reflect the attribute names in the class.

Orviss: No magic, reflection :D

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: but it should have been, when did this disappear?

src/org/opensolaris/opengrok/web/AuthorizationFilter.java
tulinkry: ok

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
tulinkry:
```java
            obj.add(REQUEST_CATEGORIES, context.serialize(stats.getRequestCategories()));
+            obj.add(TIMING, context.serialize(stats.getTiming()));
+            obj.add(TIMING_MIN, context.serialize(stats.getTimingMin()));
+            obj.add(TIMING_MAX, context.serialize(stats.getTimingMax()));
+            obj.add(TIMING_AVG, context.serialize(stats.getTimingAvg()));
```

I thought that these are the json fields. But this goes only for the output or what?

test/org/opensolaris/opengrok/authorization/AuthorizationFrameworkReloadTest.java
```diff
@@ -34,7 +34,7 @@
 import org.opensolaris.opengrok.configuration.Project;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.web.DummyHttpServletRequest;
-import org.opensolaris.opengrok.web.Statistics;
+import org.opensolaris.opengrok.web.stats.Statistics;
```
tulinkry: again the javadoc import?

Orviss: Yeah, I ignored those changes :D

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
Orviss: Yes, there is difference between saving the statistics (all the raw data) vs reporting (processing raw data to have meaningful output -> e.g. averages).

test/org/opensolaris/opengrok/configuration/messages/StatsMessageTest.java
```diff
@@ -67,7 +61,7 @@ public void testValidate() {
         Assert.assertNull(m.getClassName());
         m.setTags(new TreeSet<>());
         Assert.assertTrue(MessageTest.assertValid(m));
-        Assert.assertEquals(new TreeSet<>(), m.getTags());
+        assertEquals(new TreeSet<>(), m.getTags());
```
tulinkry: I dont know about this but we use I'd say in all of the tests the `Assert.assert ...` so you better change your Idea settings or redo the whole test style so it goes only one way.

Orviss: Why? Static import for asserts in tests is the most common static import I've ever seen. And the tests are much more readable.

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
tulinkry: aha ok,

but what if user wants to store current statistics in a file and reload them afterwards to reset to grok state after let's say some testing or something like that?

That's not possible now, is it?

test/org/opensolaris/opengrok/configuration/messages/StatsMessageTest.java
tulinkry: I'm not against but do it in the whole file so it's not mixed

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
tulinkry: And I'm sorry I didn't read the PR message :-D

Orviss: Well, all the save/load methods are still in the `RuntimeEnvironment`. However, `StatsMessage` so far provides only `clean` or `reload` but not `save` commands.

tulinkry: My usecase was something like
```bash
$ Messages -n stats get > stats.json
$ # doing some requests ...
$ mv stats.json /var/opengrok/etc/statistics.json
$ Messages -n stats reload
```

But it's not a common use case for statistics so maybe this is the way to go. @vladak ?

Orviss: Well, this is probably for testing and requires somehow knowledge of the inner workings. Therefore, you can just turn off/restart the webapp, copy `/var/opengrok/etc/statistics.json`, then do requests and copy it back.

Orviss: But we can add `save` command to `StatsMessage`...

src/org/opensolaris/opengrok/web/WebappListener.java
tulinkry: Oh I see, the code is still there, but the thread and the constants were moved to different file. Ok

src/org/opensolaris/opengrok/web/stats/Statistics.java
```diff
@@ -0,0 +1,299 @@
...
+    private long[] dayHistogram = new long[24];
+    private long[] monthHistogram = new long[31];
+    private long requests = 0;
+    private long minutes = 1;
+    private long requestsPerMinute = 0;
+    private long requestsPerMinuteMin = Long.MAX_VALUE;
+    private long requestsPerMinuteMax = Long.MIN_VALUE;
+
+    private AtomicLong searchRequests = new AtomicLong();
+    private AtomicLong zeroHitSearchCount = new AtomicLong();
```
tulinkry: I think this is doubling the value in `requestCategories['empty_search']`

src/org/opensolaris/opengrok/web/stats/Statistics.java
```diff
@@ -0,0 +1,299 @@
...
+    private Map<String, Long> timingMax = new TreeMap<>();
+    private long[] dayHistogram = new long[24];
+    private long[] monthHistogram = new long[31];
+    private long requests = 0;
+    private long minutes = 1;
+    private long requestsPerMinute = 0;
+    private long requestsPerMinuteMin = Long.MAX_VALUE;
+    private long requestsPerMinuteMax = Long.MIN_VALUE;
+
+    private AtomicLong searchRequests = new AtomicLong();
```
tulinkry: I think this is doubling the value in `requestCategories['search']`

Orviss: Yes, it is. I missed `addRequest` call in `addRequestTime`...

src/org/opensolaris/opengrok/authorization/AuthorizationFramework.java
```diff
@@ -762,46 +764,42 @@ private boolean checkAll(HttpServletRequest request, String cache, Nameable enti
...
-                    String.format("authorization_%s", overallDecision ? "positive" : "negative"),
-                    time);
-            stats.addRequestTime(
-                    String.format("authorization_%s_of_%s", overallDecision ? "positive" : "negative", entity.getName()),
-                    time);
-            stats.addRequestTime(
-                    String.format("authorization_of_%s", entity.getName()),
-                    time);
-        }
+        Duration checkTime = Duration.between(checkStart, Instant.now());
```
tulinkry: I wonder if this is faster than substracting two longs..

Orviss: It is definitely not faster, but I prefer readability and cleaner code than to save a few instructions.

src/org/opensolaris/opengrok/web/stats/Statistics.java
```diff
@@ -206,45 +295,29 @@ public double getRequestsPerMinuteAvg() {
...
-    public long getZeroHitSearchCount() {
-        return zeroHitSearchCount.get();
+        synchronized (lock) {
+            return Arrays.copyOf(monthHistogram, monthHistogram.length);
+        }
     }
 
     public long getAverageSearchHits() {
-        long searchRequestsCount = getSearchRequests();
+        long searchRequestsCount = getCount("search");
```
tulinkry: now when this string repeats it should be a constant probably

Orviss:
I think it would be better if we used some metrics library, e.g. https://github.com/dropwizard/metrics It provides nice API for our purposes and also different output formats. Therefore, closing :)

PR #2131, opened 2018-05-29
Update Servlet API to 3.1

Orviss:
Basic functionality seems to work on Tomcat 9. I tried search and navigating through project files.

Added methods were generated automatically by Idea. 

I also used fully qualified class name for `javax.servlet.http.HttpSessionContext` so I could remove `import` which caused deprecation warning.

fixes https://github.com/oracle/opengrok/issues/2089

Thanks :)

### DIFF
--- a/plugins/test/opengrok/auth/plugin/util/DummyHttpServletRequestLdap.java
+++ b/plugins/test/opengrok/auth/plugin/util/DummyHttpServletRequestLdap.java
@@ -18,37 +18,44 @@
  */
 
 /*
- * Copyright (c) 2016, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2018, Oracle and/or its affiliates. All rights reserved.
  */
 package opengrok.auth.plugin.util;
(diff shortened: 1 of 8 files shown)


plugins/test/opengrok/auth/plugin/util/DummyHttpServletRequestLdap.java
```diff
@@ -83,8 +90,9 @@ public int getMaxInactiveInterval() {
             return 3600;
         }
 
+        @Override
         @SuppressWarnings("deprecation")
```
tulinkry: and so can you remove the deprecation?

Orviss: No, the method is still deprecated. Both method and the `HttpServletContext` class are deprecated. Maybe adding `@Deprecated` will solve the issue? However, not sure if one or the other is better.

vladak:
Also pls update the README.md file (feel free to move the stuff to wiki).

PR #2132, opened 2018-05-30
Update jQuery UI to 1.12.1

Orviss:
For suggester `Autocomplete` widget from jQuery UI is needed. So I added it and also upgraded the version when I was at it.

Basic functionality seems to work. I also tried dragging and that worked too.

Thanks :)


tulinkry:
maybe also check the diff jumper, I don't remember if it has the dependency

Orviss:
Jumper seems to work properly as well.

tulinkry:
And I assume console doesn't show any warnings .. ?

Orviss:
Does not seems so 
screen shot 2018-05-30 at 09 35 34

PR #2134, opened 2018-05-30
per project renamed file handling

vladak:
Introduces per-project tunable to enable override of renamed file handling (this is handy e.g. when doing full reindex when renamed file handling is enabled by default and some projects have too rich history that is not needed. Or vice versa.). After some trial and error I decided to push this information down to `Repository` level so that it is cached.

As a test I ran the indexer with the following read-only config (supplied to the -R option) and without `--renamedHistory` (it will be off by default):

```xml
<java version="1.8.0_60" class="java.beans.XMLDecoder">

  <void property="projects">
   <void method="put">
    <string>foo-renamed</string>
      <boolean>true</boolean>
     </void>
    </object>
   </void>
  </void>

 </object>
</java>
```

and checked renamed files for this project have complete history.

Then used the same, reversed the condition to `false`, cloned the project and ran the indexer without `--renamedHistory on`, checked the first project does not have renamed history and the second has.

Also ran the indexer without projects (without the -P option) as a sanity check.


vladak:
Once this is merged in, I will update https://github.com/oracle/opengrok/wiki/Per-project-configuration

src/org/opensolaris/opengrok/configuration/Project.java
```diff
@@ -68,6 +68,12 @@
      */
     private boolean navigateWindowEnabled = false;
 
+    /**
+     * This flag sets per-project handling of renamed files.
+     */
+    private enum propState { YES, NO, UNSET };
+    private propState handleRenamedFiles = propState.UNSET;
```
idodeclare: Hi. Perhaps `Boolean` (i.e. a nullable object wrapping `boolean` and therefore tri-state) would do here just as well?

vladak: good idea!

src/org/opensolaris/opengrok/configuration/Project.java
```diff
@@ -229,6 +234,20 @@ public void setNavigateWindowEnabled(boolean navigateWindowEnabled) {
         this.navigateWindowEnabled = navigateWindowEnabled;
     }
 
+    /**
+     * @return true if this project handles renamed files.
+     */
+    public boolean isHandleRenamedFiles() {
+        return handleRenamedFiles;
```
tulinkry: this cause potential NPE when `handleRenamedFiles` is null, doesn't it?

vladak: yes, did not mean to expose `null` outside.

PR #2141, opened 2018-06-04
allow to enable/disable history per project

vladak:
This basically removes the notion that `Repository` presence constitutes history creation. With these changes it is possible to have history disabled globally and enabled on project level and vice versa.

Also, I believe this is one of the last pieces where configuration is stored outside of `Configuration` object.


vladak:
Once this is merged, I will update https://github.com/oracle/opengrok/wiki/Per-project-management

idodeclare:
LGTM

PR #2143, opened 2018-06-05
report more information during webapp startup

vladak:
2 lines added to log during webapp startup, e.g.:

```
05-Jun-2018 10:19:48.811 INFO [localhost-startStop-1] org.opensolaris.opengrok.web.WebappListener.contextInitialized Starting webapp with version 1.1-rc22 (f8c14b49af83451078552967dac5d22d949f2a09)
05-Jun-2018 10:19:48.811 INFO [localhost-startStop-1] org.opensolaris.opengrok.configuration.Configuration.read Reading configuration from /opengrok/etc/configuration.xml
```

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Configuration.java
+++ b/src/org/opensolaris/opengrok/configuration/Configuration.java
@@ -1340,6 +1340,7 @@ public final class Configuration {
     }
 
     public static Configuration read(File file) throws IOException {
+        LOGGER.log(Level.INFO, "Reading configuration from {0}", file.getCanonicalPath());
         try (FileInputStream in = new FileInputStream(file)) {
             return decodeObject(in);
         }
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -1340,6 +1340,7 @@ public void encodeObject(OutputStream out) {
     }
 
     public static Configuration read(File file) throws IOException {
+        LOGGER.log(Level.INFO, "Reading configuration from {0}", file.getCanonicalPath());
```
idodeclare: Perhaps just logging `file` would be good enough, since `getCanonicalPath()` can throw `IOException`.

vladak: I wanted to resolve any symlinks however on my Linux system getCanonicalPath() does not do that and reports the path without resolving the symlink.

vladak: Even if it threw an exception while resolving the path, it would be caught in the same block in `contextInitialized()` so the reporting would be the same.

PR #2145, opened 2018-06-06
Add demo link to README

Orviss:
If you would like to change the order then let me know :) I had no idea where to put it so I appended it to the end.

Thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Copyright (c) 2006, 2018 Oracle and/or its affiliates. All rights reserved.
 9.  [Tuning OpenGrok for large code bases](#9-tuning-opengrok-for-large-code-bases)
 10. [Authors](#10-authors)
 11. [Contact us](#11-contact-us)
+12. [Demo](#12-demo)
 
 ## 1. Introduction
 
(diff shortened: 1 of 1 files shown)


vladak:
I wonder if there is a button similar to coverage/build status.

Orviss:
I was looking at it and I was not able to find "Demo" one. However, it is customizable so it is easy to create it.

Orviss:
<http://shields.io/#your-badge> However, I'm not exactly sure what to specify as a status. Maybe online/offline ? Still, this would probably require some file hosting where this badge would be accessible and we could swap it by periodically checking the server? Or am I just complicating things and it could be done in a different manner?

tulinkry:
By the way, the demo has some problems: http://demo.opengrok.org/xref/opengrok/web/js/utils-0.0.22.js?r=38f77573 (no content)

Orviss:
Should be working now, xref was not created somehow. I have a pipeline in Jenkins which clones the project, builds it, runs `sync.py`, gets new config and deploys the new webapp. However, this does not work always as it should...

PR #2149, opened 2018-06-08
Cut off diffbar fix

Orviss:
fixes https://github.com/oracle/opengrok/issues/2133

Tested on Safari and Google Chrome. I'm not sure what exactly is broken but this should be equivalent to the desired functionality...

Thanks :)


vladak:
Got a screenshot ?

PR #2153, opened 2018-06-11
Project message set/get

vladak:
Introduces set/get commands for Project message. Both the command name and its arguments have to reside in message text since the tags are already used for project names.


src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -319,6 +398,10 @@ public void validate() throws Exception {
             throw new Exception("The message must contain a tag (project name(s))");        
         }
 
+        if (command.compareTo("get") == 0 && getTags().size() != 1) {
```
Orviss: `command.equals("get")`?

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -248,6 +273,60 @@ private void validateMore(RuntimeEnvironment env) throws Exception {
...
+                                matcher.group(1), matcher.group(2)).getBytes();
+                    } else {
+                        LOGGER.log(Level.WARNING, "cannot find project " +
+                               projectName + " to set a property");
+                    }
+                }
+                break;
+            case "get":
+                for (String projectName : getTags()) {
+                    Project project;
```
Orviss: This seems more readable to me:
```java
if (project != null) {
...
```

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -248,6 +273,60 @@ private void validateMore(RuntimeEnvironment env) throws Exception {
...
+                if (!matcher.find()) {
+                    // invalid pattern
+                    throw new IOException(
+                            String.format("The pattern \"%s\" does not match \"%s\".",
+                                    VARIABLE_PATTERN.toString(),
+                                    getText()));
+                }
+                        
+                for (String projectName : getTags()) {
+                    Project project;
```
Orviss: Same as below.

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -102,13 +115,25 @@ private void validateMore(RuntimeEnvironment env) throws Exception {
...
     @Override
     protected byte[] applyMessage(RuntimeEnvironment env) throws Exception {
-        String command = getText();
+        String command = getCommand();
 
         validateMore(env);
-
-        switch (command) {
+        
+        switch (getCommand()) {
```
Orviss: Why call `getCommand()` for the second time? The result is already in `command` variable.

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -248,6 +273,60 @@ private void validateMore(RuntimeEnvironment env) throws Exception {
...
+                                // set the property
+                                ClassUtil.invokeSetter(
+                                        repo,
+                                        matcher.group(1), // field
+                                        matcher.group(2) // value
+                                );
+                            }
+                        }
+                        
+                        return String.format("Variable \"%s\" set to \"%s\".",
```
tulinkry: I'm worried that this will set just the first project in `getTags()`.

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
```diff
@@ -248,6 +273,60 @@ private void validateMore(RuntimeEnvironment env) throws Exception {
...
+                        LOGGER.log(Level.WARNING, "cannot find project " +
+                               projectName + " to set a property");
+                    }
+                }
+                break;
+            case "get":
+                for (String projectName : getTags()) {
+                    Project project;
+                    if ((project = env.getProjects().get(projectName)) != null) {
+                        return ClassUtil.invokeGetter(project, getText().substring(4)).getBytes();
```
tulinkry: I'm worried that this will set just the first project in getTags().

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
tulinkry: And a test case for this would be great

src/org/opensolaris/opengrok/configuration/messages/ProjectMessage.java
Orviss: There is a check that `get` can only have one project specified.

tulinkry: I didn't see that, ok.

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -495,4 +497,53 @@ public void testGetRepos() throws Exception {
...
+        m.setText("add");
+        for (String proj : projects) {
+            m.addTag(proj);
+        }
+        m.apply(env);
+        Assert.assertEquals(2, env.getProjectList().size());
+        
+        // Change their property.
+        m = new ProjectMessage();
+        m.setText("set handleRenamedFiles = false");
```
tulinkry: can you try doing this test for `= true`? I'm still not conviced that this works for both of the projects and `false` seems to be the default value.

tulinkry: Based on the `applyMessage() -> set -> return`

vladak: I have not committed the fix yet to see that the test is really failing. Coming in a minute..

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -495,4 +497,53 @@ public void testGetRepos() throws Exception {
         out = new String(m.apply(env));
```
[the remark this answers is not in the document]
vladak: yep

test/org/opensolaris/opengrok/configuration/messages/ProjectMessageTest.java
```diff
@@ -495,4 +494,53 @@ public void testGetRepos() throws Exception {
...
         Assert.assertEquals("Mercurial", out);
     }
+    
+    @Test
+    @ConditionalRun(RepositoryInstalled.GitInstalled.class)
```
tulinkry: nb but this is now a duplicate

tools/Messages
```diff
@@ -99,6 +99,15 @@ Usage()
     echo "       - \"list-indexed\" - list indexed projects"
     echo "       - \"get-repos\" - get list of repositories in the form of relative paths to source root for given project(s)"
     echo "       - \"get-repos-type\" - get repository type(s) for given project(s)"
+    echo "       - \"set\" - sets a particular project property (only primitive types)"
+    echo "                   in the webapp."
+    echo "                      Examples:"
+    echo "                          tabSize = 8"
+    echo "                          handleRenamedFiles = true"
```
tulinkry: and also I think there is a wiki page listing the tunables so it should be also added there

vladak: already there: https://github.com/oracle/opengrok/wiki/Per-project-configuration

PR #2157, opened 2018-06-14
make the config set/setconf message use the interactive flag

vladak:
When uploading new configuration via Message this could make the webapp refresh configuration slowly in the presence of stale repositories. This change makes the set configuration process use interactive flag to shorten the timeout with the only exception - reindex.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
+++ b/src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
@@ -787,7 +787,7 @@ public final class RuntimeEnvironment {
 
     /**
      * Add repositories to the list.
-     * @param repositories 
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1867,7 +1883,7 @@ public void applyConfig(Message m, boolean reindex) {
      * @param reindex is the message result of reindex
      *
      */
-    public void applyConfig(Configuration config, boolean reindex) {
+    public void applyConfig(Configuration config, boolean reindex, boolean interactive) {
```
Orviss: Did you forget to use `interactive` param? Otherwise it seems like YAGNI.

Issue #2156, opened 2018-06-12
JarAnalyzer hangs at indexing


Nurm78:
After upgrading from opengrok 1.1-rc3 to rc28, the indexing did not finished. After some tries, I recognized, that the indexer always "hangs" at the same file selenium-server-standalone-2.43.1.jar. The indexing of this file does not finish within 30 minutes on my machine, no matter if the file is indexed alone or in a project. 

> D:\source_playgroundforimprovements>java -jar D:\opengrok-1.1-rc28\lib\opengrok.jar -W D:\opengrok-1.1-rc28\configuration_debug.xml -c d:\ctags58\ctags.exe --leadingWildCards on -P -v -H -r off -s D:\bugreport -d d:\grokdata_rc28_debug -i d
> :doc -i *.p12
> Logging filehandler pattern: %h/java%u.log
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.index.Indexer main
> INFO: Indexer version 1.1-rc28 (dbcad1993409e04b41712fdcf0b2d90534e45c9a)
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.index.Indexer prepareIndexer
> INFO: Generating history cache for all repositories ...
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.history.HistoryGuru createCacheReal
> INFO: Creating historycache for 0 repositories
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.util.Statistics report
> INFO: Done historycache for all repositories (took 0)
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.index.Indexer prepareIndexer
> INFO: Done...
> Jun 12, 2018 1:33:39 PM org.opensolaris.opengrok.index.Indexer doIndexerExecution
> INFO: Starting indexing
> Jun 12, 2018 1:33:40 PM org.opensolaris.opengrok.index.DefaultIndexChangedListener fileAdd
> INFO: Add: /TeamAb/ABCTestautomation_XYZ/sw/TestAutomationXYZ/libs/selenium-server-standalone-2.43.1.jar (JarAnalyzer)

The process is started with a batch script 
```
set ogrokdir=D:\opengrok-1.1-rc28
set ogrokdatadir=d:\grokdata_rc28_debug
set sourcedir=D:\bugreport

SET JAVA_OPTS="-Xms2048m -Xmx2048"
SET OPENGROK_SCAN_DEPTH=2
[...]

PR #2158, opened 2018-06-15
Issue #2156 : speed up JarAnalyzer

idodeclare:
Hello,

Please consider for integration this patch to speed up `JarAnalyzer` by reusing the same string builders for an entire jar.

Thank you.


tarzanek:
it looks ok, merging, thank you

PR #2163, opened 2018-06-16
Per project history test data fix

Orviss:
When running per project history tests a directory `dataForPerProjectHistoryTest{number}` is created under `opengrok-indexer`. This can be achieved simply by running `mvn test`.

It can look like this:
screen shot 2018-06-16 at 19 10 37

However, these directories are not removed when running `mvn clean`.

This PR fixes the problem.

Thanks :)


vladak:
I noticed this during my testing but then forgot about it. Thanks !

PR #2167, opened 2018-06-19
Index version check partial

vladak:
Today I tried to create some partal indexes (with `indexpart` sub-command of the `OpenGrok` script) by hand and got an error that different project than what was being indexed had old index version. This change should fix that.

### DIFF
--- a/src/org/opensolaris/opengrok/index/IndexVersion.java
+++ b/src/org/opensolaris/opengrok/index/IndexVersion.java
@@ -24,6 +24,7 @@ package org.opensolaris.opengrok.index;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.List;
 import org.apache.lucene.index.IndexNotFoundException;
 import org.apache.lucene.index.SegmentInfos;
 import org.apache.lucene.store.Directory;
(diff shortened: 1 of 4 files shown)


vladak:
Tested by unpacking `test/org/opensolaris/opengrok/index/oldindex.zip` into `/var/opengrok/data/index/foo3` and running `OPENGROK_READ_XML_CONFIGURATION=/var/opengrok/etc/configuration.xml ./OpenGrok indexpart <project>`  for both `foo3` and different project with current Lucene index.

src/org/opensolaris/opengrok/index/Indexer.java
```diff
@@ -206,11 +193,24 @@ public static void main(String argv[]) {
             // Assemble the unprocessed command line arguments (possibly
             // a list of paths). This will be used to perform more fine
             // grained checking in invalidateRepositories().
-            for (int optind=0; optind< argv.length; optind++) {
+            for (int optind = 0; optind < argv.length; optind++) {
```
Orviss: Maybe use enhanced-for loop?

src/org/opensolaris/opengrok/index/IndexVersion.java
```diff
@@ -52,16 +53,25 @@ public IndexVersionException(String s) {
...
+     * @param subFilesList list of paths. If non-empty, only projects matching these paths will be checked.
      * @throws Exception otherwise
      */
-    public static void check(Configuration cfg) throws Exception {
+    public static void check(Configuration cfg, List<String> subFilesList) throws Exception {
         File indexRoot = new File(cfg.getDataRoot(), IndexDatabase.INDEX_DIR);
-        if (cfg.isProjectsEnabled()) {
-            for (String projectName : cfg.getProjects().keySet()) {
+        
+        if (subFilesList.size() > 0) {
```
Orviss: I think `isEmpty()` is more preferable to this construct.

src/org/opensolaris/opengrok/index/Indexer.java
vladak: I'd leave it for now - it's a code I did not change (other than whitespace).

PR #2168, opened 2018-06-20
Don't select line numbers when highlighting code

thecodeboss:
When you copy and paste code results from OpenGrok, it will no longer copy the line numbers with this change.

Will save us the trouble of having to view the raw file before copying.

### DIFF
--- a/web/default/style.css
+++ b/web/default/style.css
@@ -24,6 +24,10 @@ div#src0 a.l.selected:after, div#src0 a.hl.selected:after{
 div[id^='src'] a.l.target, div[id^='src'] a.hl.target {
     background: orange;
     color:yellow;
+    -webkit-user-select: none;  /* Chrome all / Safari all */
+    -moz-user-select: none;     /* Firefox all */
+    -ms-user-select: none;      /* IE 10+ */
+    user-select: none;          /* Likely future */
 }
 
 a:link {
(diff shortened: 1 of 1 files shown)


tulinkry:
I think that the line numbers are there by design. If you need the code directly you go to raw.

vladak:
It is definitely intention to have the line numbers to be copied to the buffer, at least in the normal view mode. Perhaps change is targeting some other mode (diff, search) ? (have not have the time to try it out yet)

thecodeboss:
> I think that the line numbers are there by design. If you need the code directly you go to raw.

This doesn't seem like a very good design. On GitHub if you copy code it never includes the line numbers, and it's the same with basically every code editor or code viewer out there.
Having to go to the raw mode to copy code is especially annoying because you no longer have line numbers to refer to, and it's hard to find the code you were trying to copy. You also lose the syntax highlighting.
If there's still disagreement, maybe it is worth adding a checkbox that says something along the lines of "Include line numbers when copying" so power users can copy without having to go to raw mode.

vladak:
Having the ability to select the line numbers is still important in many deployments where code is pasted say to a bug tracking system which then includes comments referencing the line numbers. I can still see this being used heavily, at least in our environment.

There is actually a 'Line#' link at the top of the page which enables/disables line numbers, does this do the trick for you ?

thecodeboss:
> There is actually a 'Line#' link at the top of the page which enables/disables line numbers, does this do the trick for you ?

Ah yes actually, this is a nice trick. Maybe it should say "Toggle line numbers" instead of just "Line#" to make it more obvious what it does.

vladak:
I'd like to keep it short. The UI should be minimalistic I think. There is a help bubble that explains what it does if you do mouse over.

PR #2169, opened 2018-06-20
configurable directory for web app include files

vladak:
This change makes the location of include files used in web application configurable. By default they are searched for in data root, if include root is specified they are searched for there.

This change also refreshes the contents of these files when new configuration is set in the web application.

### DIFF
--- a/src/org/opensolaris/opengrok/configuration/Configuration.java
+++ b/src/org/opensolaris/opengrok/configuration/Configuration.java
@@ -143,6 +143,10 @@ public final class Configuration {
     private Set<Group> groups;
     private String sourceRoot;
     private String dataRoot;
+    /**
+     * directory with include files for web application (header, footer, etc.)
+     */
+    private String includeRoot;
     private List<RepositoryInfo> repositories;
     /**
      * @deprecated This is kept around so not to break object deserialization
@@ -804,6 +808,18 @@ public final class Configuration {
         this.dataRoot = dataRoot;
     }
 
+    /**
+     * If {@link includeRoot} is not set, {@link dataRoot} will be returned.
+     * @return web include root directory
+     */
+    public String getIncludeRoot() {
+        return includeRoot != null ? includeRoot : dataRoot;
+    }
+    
+    public void setIncludeRoot(String newRoot) {
+        this.includeRoot = newRoot;
+    }
+    
     public List<RepositoryInfo> getRepositories() {
         return repositories;
     }
@@ -1175,12 +1191,13 @@ public final class Configuration {
     /**
      * Get the contents of the footer include file.
      *
+     * @param force if true, reload even if already set
      * @return an empty string if it could not be read successfully, the
      * contents of the file otherwise.
      * @see #FOOTER_INCLUDE_FILE
      */
-    public String getFooterIncludeFileContent() {
-        if (footer == null) {
+    public String getFooterIncludeFileContent(boolean force) {
+        if (footer == null || force) {
             footer = getFileContent(new File(getDataRoot(), FOOTER_INCLUDE_FILE));
         }
         return footer;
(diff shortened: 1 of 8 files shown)


vladak:
Once this is merged in, I will update https://github.com/oracle/opengrok/wiki/Webapp-configuration

vladak:
Tested by adding this snippet into read-only configuration, creating `/var/opengok/etc/header_include` with some content, reindexing and verifying that the web page contains the content in the header.

```xml
  <void property="includeRoot">
     <string>/var/opengrok/etc</string>
  </void>
```

src/org/opensolaris/opengrok/configuration/Configuration.java
```diff
@@ -143,6 +143,10 @@
     private Set<Group> groups;
     private String sourceRoot;
     private String dataRoot;
+    /**
+     * directory with include files for web application (header, footer, etc.)
+     */
+    private String includeRoot;
```
Orviss: Wouldn't be better to use `Path`? However, I'm not sure if the serialization will work...

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1552,6 +1561,11 @@ public void setConfiguration(Configuration configuration, List<String> subFileLi
         // The invalidation of repositories above might have excluded some
         // repositories in HistoryGuru so the configuration needs to reflect that.
         configuration.setRepositories(new ArrayList<>(histGuru.getRepositories()));
+        
+        configuration.getBodyIncludeFileContent(true);
```
Orviss: I, personally, would get rid of the `force` param and would add method like `resetIncludeFiles` or something similar. I think the intent would be more clear.

test/org/opensolaris/opengrok/configuration/IncludeFilesTest.java
```diff
@@ -0,0 +1,103 @@
...
+    static final String LINE_SEP = System.getProperty("line.separator");
+    
+    @BeforeClass
+    public static void setUpClass() throws IOException {
+        includeRoot = Files.createTempDirectory("include_root");
+        env.getConfiguration().setIncludeRoot(includeRoot.toString());
+    }
+    
+    private void writeStringToFile(File file, String str) throws IOException {
+        try (BufferedWriter bw = new BufferedWriter(new FileWriter(file.toString()))) {
```
Orviss: No need to call `file.toString()` -> `FileWriter` can take `File` in the constructor as well.

test/org/opensolaris/opengrok/configuration/IncludeFilesTest.java
```diff
@@ -0,0 +1,103 @@
...
+ * Test include file functionality for web application.
+ * 
+ * @author Vladimir Kotal
+ */
+public class IncludeFilesTest {
+    static Path includeRoot;
+    static final String CONTENT_1 = "foo";
+    static final String CONTENT_2 = "bar";
+    static RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+    static final String LINE_SEP = System.getProperty("line.separator");
```
Orviss: Since Java 8: `System.lineSeparator()`

test/org/opensolaris/opengrok/configuration/RuntimeEnvironmentTest.java
```diff
@@ -119,6 +119,28 @@ public void testDataRoot() throws IOException {
...
 
+    @Test
+    public void testIncludeRoot() throws IOException {
+        RuntimeEnvironment instance = RuntimeEnvironment.getInstance();
+        assertNull(instance.getIncludeRootPath());
+        
+        // set data root
+        File f = File.createTempFile("dataroot", null);
+        String path = f.getCanonicalPath();
+        assertTrue(f.exists());
```
Orviss: I think the `f` will exist always. From documentation: 

> IOException - If a file could not be created

src/org/opensolaris/opengrok/configuration/Configuration.java
vladak: I will rather keep it the same as sourceRoot/dataRoot.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
vladak: Somehow I thought the force flag was quite elegant, not having to duplicate the calls to `getFileContent()`. What I could do is to move these 4 calls from `setConfiguration()` into a private method.

PR #2172, opened 2018-06-21
refactor annotate code once again

vladak:
This change refactors the annotate JSP code once again, the top-level decision making is now checking only:
  - directory
  - revision
  - otherwise (redirect)

### DIFF
--- a/web/list.jsp
+++ b/web/list.jsp
@@ -285,7 +285,8 @@ Click <a href="<%= rawPath %>">download <%= basename %></a><%
                                 // SRCROOT is read with UTF-8 as a default.
                                 r = IOUtils.createBOMStrippedReader(in,
                                     StandardCharsets.UTF_8.name());
-                                AnalyzerGuru.writeXref(
+                                AnalyzerGuru.writeDumpedXref(
+                                        request.getContextPath(),
                                         a, r, out,
                                         defs, annotation, project);
                             } else if (g == Genre.IMAGE) {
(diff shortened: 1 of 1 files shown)


vladak:
After this change the code looks like this:

```java
if (cfg.isDir()) {
  // directory listing
} else if (rev.length() != 0) { // this is true if r parameter is on
  // requesting a revision
  if (cfg.isLatestRevision(rev)) {
    if (cfg.annotate()) {
      // annotate latest revision of the file
    } else {
      // dump the file
    }
  } else {
    // fetch historic revision and dump it (with possible annotation)
  }
} else {
  // redirect to URI with r? parameter
}
```

I believe that in order to dump the historical (annotated) version of the file it is needed to use `AnalyzerGuru.writeXref()` (as opposed to `AnalyzerGuru.writeDumpedXref()`)

web/list.jsp
```diff
@@ -282,8 +285,8 @@ Click <a href="<%= rawPath %>">download <%= basename %></a><%
                                 // SRCROOT is read with UTF-8 as a default.
                                 r = IOUtils.createBOMStrippedReader(in,
                                     StandardCharsets.UTF_8.name());
-                                AnalyzerGuru.writeDumpedXref(
-                                        request.getContextPath(), a, r, out,
+                                AnalyzerGuru.writeXref(
```
idodeclare: This was not reverted after you found the true problem.

vladak: Yeah, I was sort of waiting for you to chime in :-) Need to test if the fix still works with `writeDumpedXref()`.

vladak: I can see that `writeDumpedXref()` is just a wrapper of `writeXref()` so it should work fine.

PR #2177, opened 2018-06-22
make the tests pass in Wercker

vladak:
I am submitting this pull req mainly to experiment with Wercker, namely trying to get the `PageConfig` tests to pass in that environment.

Initially, there are only bunch of tests failing:

```
[ERROR] Failures: 
[ERROR]   PageConfigTest.canProcessHistory:110->assertCanProcess:480 expected:<> but was:<null>
[ERROR]   PageConfigTest.canProcessXref:130->assertCanProcess:480 expected:<> but was:<null>
[ERROR]   PageConfigTest.testCheckSourceRootExistence4:438 This should throw an exception when the file is not readable
[ERROR]   PageConfigTest.testGetAnnotation:343
[ERROR]   PageConfigTest.testGetLatestRevisionValid:258 expected:<aa35c258> but was:<null>
[INFO] 
[ERROR] Tests run: 717, Failures: 5, Errors: 0, Skipped: 58
```

so that's not too bad.

Originally I thought that's because of missing pre-requisites however it seems to boil down to the `testCheckSourceRootExistence4` test that removes permission read bit(s) from a directory and verifies `canRead()` returns false which it fails to do, therefore the test is failing. Possibly, the other tests fail as a result of cleanup not being completed.

### DIFF
--- a/test/org/opensolaris/opengrok/web/PageConfigTest.java
+++ b/test/org/opensolaris/opengrok/web/PageConfigTest.java
@@ -444,7 +444,8 @@ public class PageConfigTest {
         assertTrue(temp.exists());
         set = Files.getPosixFilePermissions(temp.toPath());
         System.out.println("XXX: after " + temp.toString() + " " + PosixFilePermissions.toString(set));
-        assertFalse(temp.canRead());
+        File newtemp = new File(temp.toString());
+        assertFalse(newtemp.canRead());
         RuntimeEnvironment.getInstance().getConfiguration().setSourceRoot(temp.getAbsolutePath());
         try {
             cfg.checkSourceRootExistence();
--- a/test/org/opensolaris/opengrok/web/PageConfigTest.java
+++ b/test/org/opensolaris/opengrok/web/PageConfigTest.java
@@ -435,6 +435,7 @@ public class PageConfigTest {
         File temp = Files.createTempDirectory("opengrok-src-root4").toFile();
         temp.deleteOnExit();
         temp.setReadable(false);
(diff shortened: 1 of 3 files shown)


test/org/opensolaris/opengrok/web/PageConfigTest.java
```diff
@@ -64,7 +64,10 @@ public static void setUpClass() throws Exception {
         repository = new TestRepository();
         repository.create(
                 HistoryGuru.class.getResourceAsStream("repositories.zip"));
-        RuntimeEnvironment.getInstance().setRepositories(repository.getSourceRoot());
+        
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        env.setHistoryEnabled(true);
```
idodeclare: Since `RuntimeEnvironment` is shared between all runs of tests, usually I have noticed a reversion of original settings — e.g. `env.setHistoryEnabled(originalIsHistoryEnabled)` in `@AfterClass`. I see tests passing but wonder if it should be done anyway for correctness.

vladak: Yeah, could do that. However, if all the test classes ran in the same JVM instance I think we would have failures all over the place.

vladak:
The problem is that the tests run under `root` user (https://github.com/wercker/wercker/issues/229) + there might be some problems with default locale.

vladak:
reopening to try new Maven plugin

vladak:
seems to be working, closing again.

PR #2178, opened 2018-06-23
Improve URI-ending pushback for comma and odd right paren

idodeclare:
Hello,

Please consider for integration this patch to improve the URI-ending pushback to recognize trailing commas or dangling right parentheses.

(I did not consider this change significant enough to bump `FileAnalyzer.getVersionNo()` to force automatic re-analysis.)

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/util/StringUtils.java
+++ b/src/org/opensolaris/opengrok/util/StringUtils.java
@@ -116,7 +116,7 @@ public final class StringUtils {
     private static final Pattern URI_CHARS_STARTSMATCH =
         Pattern.compile("^" + URI_CHARS_PAT);
 
-    /** Private to enforce singleton */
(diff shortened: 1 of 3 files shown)


src/org/opensolaris/opengrok/util/StringUtils.java
```diff
@@ -323,4 +330,20 @@ public static int countPushback(String capture, Pattern pat) {
         }
         return 0;
     }
+
+    private static int countURIUnmatchedRightParen(String value, int ridx) {
```
vladak: this method returns either 0 or 1 and is not specific to URI (other than used for checking it) so perhaps name it `isUnmatchedRightParen` and return boolean and increment in the caller accordingly ?

idodeclare:
ok

PR #2179, opened 2018-06-23
Subsume --checkIndexVersion into regular run

idodeclare:
Hello,

Please consider for integration this patch to fold `--checkIndexVersion` handling into a regular run of opengrok.jar.

When `--checkIndexVersion` was activated a couple of months ago, it added an inconvenience to `JAVA_DEBUG` because a developer had to attach to two, serial runs of opengrok.jar.

Thank you.


vladak:
Cool, I wanted to do this myself however could not find the time.

idodeclare:
> this message should be more granular if running per-project reindex - possibly just the projects specified by the subFilesList have old indexes.

I copied literally the same message you wrote for all failures of `CheckIndexVersion`.

vladak:
merged in 927782eeaf14ffef966e647afe8eeaa8682fbcc7, will fix the partial index message myself.

Issue #2092, opened 2018-05-02
accessing xref modified before reindex should return its old version


vladak:
With 1.1-rc21 I see the following behavior: when a source file is modified and the indexer has not yet run, the webapp will display empty content (the standard header and footer are displayed normally, there is just not anything in between). The corresponding xref file exist, it is just stale. The content (albeit stale) should be displayed or a warning message. Also, as far as I remember, xrefs could be generated on demand.

vladak:
For reference, `opensnoop` displays the xref to be read so there is probably some logic in the webapp that prevents the contents from appearing.

This behavior seems to be shared with other files updated before reindex. The xrefs for non-updated files can be displayed normally.

PR #2180, opened 2018-06-24
Re-gen xref if missing. Fix list.jsp w.r.t. isGenerateHtml().

idodeclare:
Hello,

Please consider for integration this patch that re-generates an xref when it is detected as missing or as stale (e.g. for issue #2092 when source code is updated but not yet re-indexed). This patch also fixes handling of `isGenerateHtml()` in list.jsp.

Thank you.

### DIFF
--- a/src/org/opensolaris/opengrok/index/IndexDatabase.java
+++ b/src/org/opensolaris/opengrok/index/IndexDatabase.java
@@ -88,6 +88,7 @@ import org.opensolaris.opengrok.analysis.Ctags;
 import org.opensolaris.opengrok.analysis.Definitions;
 import org.opensolaris.opengrok.analysis.FileAnalyzer;
 import org.opensolaris.opengrok.analysis.FileAnalyzer.Genre;
+import org.opensolaris.opengrok.analysis.FileAnalyzerFactory;
 import org.opensolaris.opengrok.configuration.Project;
 import org.opensolaris.opengrok.configuration.RuntimeEnvironment;
 import org.opensolaris.opengrok.configuration.messages.Message;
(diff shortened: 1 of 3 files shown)


src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -1633,7 +1641,9 @@ private void finishWriting() throws IOException {
      * @param path the source file path
      * @return {@code false} if a mismatch is detected
      */
-    private boolean chkSettings(File file, String path) throws IOException {
+    private boolean chkSettings(boolean[] outIsXrefWriter, File file,
```
vladak: the javadoc needs update

src/org/opensolaris/opengrok/index/IndexDatabase.java
```diff
@@ -1732,6 +1753,24 @@ private IndexAnalysisSettings2 readAnalysisSettings() throws IOException {
...
     }
 
+    private boolean checkXrefExistence(String path) {
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        boolean chkres = whatXrefFile(path, env.isCompressXref()).exists();
+
+        if (!env.isGenerateHtml()) {
+            if (chkres) {
+                LOGGER.log(Level.FINEST, "Extraneous {0}", path);
+                removeXrefFile(path);
```
vladak: I'd just leave the file alone. Firstly, it does not go well with the name of the method ("check existence" should do just what it says in the name), secondly it can be a result of current mis-configuration.

vladak:
Rebase and address comments ?

vladak:
Replaced by #2319.

PR #2181, opened 2018-06-24
Fix #2169 for altered sig of getForbiddenIncludeFileContent()

idodeclare:
Hello,

Please consider for integration this patch to fix a JSP for a method that changed signature.

Thank you.


vladak:
Thanks !

PR #2182, opened 2018-06-25
add parameter to Command constructor to disable stderr redirection

### LINKED ISSUE #2176: projadm sometimes creates invalid configuration.xml
When running `/opengrok/dist/bin/projadm.py -b /opengrok -R /opengrok/etc/readonly_configuration.xml -r -u` it created `configuration.xml` that started with:

```
Jun 22, 2018 3:30:52 PM org.opensolaris.opengrok.configuration.Configuration read
INFO: Reading configuration from /opengrok/etc/readonly_configuration.xml
Jun 22, 2018 3:30:52 PM org.opensolaris.opengrok.configuration.Configuration read
INFO: Reading configuration from /tmp/tmpjth79saq
<java version="1.8.0_172" class="java.beans.XMLDecoder">
  <void property="CTagsExtraOptionsFile">
   <string>/opengrok/etc/ctags.config</string>
  </void>
```

which obviously the web app could not parse.

vladak:
Basically adds a parameter to `Command` constructor to specify whether `stderr` should be redirected to `stdout` as it used to be until now. If not, a new thread consuming `stderr` will be created. New method `geterroutput()` can be used to gather this output.

### DIFF
--- a/tools/sync/command.py
+++ b/tools/sync/command.py
@@ -49,7 +49,8 @@ class Command:
     TIMEDOUT = "timed out"
 
     def __init__(self, cmd, args_subst=None, args_append=None, logger=None,
-                 excl_subst=False, work_dir=None, env_vars=None, timeout=None):
+                 excl_subst=False, work_dir=None, env_vars=None, timeout=None,
+                 redirect_stderr=True):
         self.cmd = cmd
         self.state = "notrun"
         self.excl_subst = excl_subst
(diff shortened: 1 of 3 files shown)


vladak:
Of course, this could be fixed by saving the output in `ConfigMerge` to a file via command line parameter however I believe most would run it by redirecting stdout to a file on command line so they would not have this problem. Also, the added flexibility to `Command` could be used elsewhere.

PR #2184, opened 2018-06-26
REST API support

Orviss:
This is an attempt at REST API support. Features:
* Jersey – nice, annotation based API
* clean API -> no need to have e.g. `expiration/duration` param option if it makes no sense
* API that is easily modifiable -> no need to affect all the other places

Tests done:
* unit tests rewritten
* manual testing -> e.g. run Indexer and check that the new configuration was sent
 
Caveats:
* My changes to Python scripts are a little tedious (my first Python code attempt :D) Please feel free to change or write me how to do it properly. (probably mainly @vladak )
* Some Python scripts rely on `messages` -> this will need to be resolved but I'm not sure how -> @vladak ?
* Java 9/ Java 10 exception that JAXB is required (because of the modules) -> can be resolved by `--add-modules` or by adding JAXB as dependency to `pom.xml` -> however, it still seems to work, so not sure if I should deal with it right now
* url to webapp now needs to be specified with protocol and context path -> e.g. `http://localhost:8080/source`

Fixes https://github.com/oracle/opengrok/issues/1801

Please take it as a proof of concept :) I'm willing to make any changes that you notice.

Thanks :)


tulinkry:
First is it possible to add a versioned api for the future? `/source/api/v1/...`

Orviss:
It is possible to change `@ApplicationPath(/api)` to `@ApplicationPath(/api/v1)`.

src/org/opensolaris/opengrok/web/api/RestApp.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.glassfish.jersey.server.ResourceConfig;
+
+import javax.ws.rs.ApplicationPath;
+
+@ApplicationPath("/api/v1")
+public class RestApp extends ResourceConfig {
+
+    public RestApp() {
+        packages("org.opensolaris.opengrok.web.api.controller", "org.opensolaris.opengrok.web.api.filter");
```
tulinkry: I'd strongly be for versioning also in the package names. `web.api.v1.controller` and so on. This makes the future very simple

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final String DEFAULT_DURATION = "PT10M";
+    private static final String DEFAULT_TAGS = MessagesContainer.MESSAGES_MAIN_PAGE_TAG;
+    private static final String DEFAULT_CSS_CLASS = "info";
+
+    private final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @PUT
+    @Consumes(MediaType.TEXT_PLAIN)
+    public void addMessage(
+            @QueryParam("duration") @DefaultValue(DEFAULT_DURATION) final String durationStr,
```
tulinkry: I think this goes against rest design principles. This should be `POST` and the query params should be the body in json.

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
```diff
@@ -0,0 +1,84 @@
...
+    private Duration parseDuration(final String duration) {
+        try {
+            return Duration.parse(duration);
+        } catch (DateTimeParseException e) {
+            throw new WebApplicationException(e, Response.Status.BAD_REQUEST);
+        }
+    }
+
+    @DELETE
+    public void removeMessagesWithTag(@QueryParam("tags") final Set<String> tags) {
```
tulinkry: this is comma separated list of tags or?

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
```diff
@@ -0,0 +1,84 @@
...
+        env.removeAnyMessage(tags);
```
tulinkry: what is the output of this? 200 OK or?

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+}
```
tulinkry: no need for `GET` ?

src/org/opensolaris/opengrok/web/api/controller/ProjectsController.java
```diff
@@ -0,0 +1,333 @@
...
+@Path("/projects")
+public class ProjectsController {
+
+    private static final Logger logger = LoggerFactory.getLogger(ProjectsController.class);
+
+    private RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void addProject(final List<String> projects) {
```
tulinkry: `addProjects` or maybe do `addProject` for a single project and fire it multiple times for multiple projects?

tulinkry: overall the second is what is preferable for web services - simple usage, no complex scenarios, if clients want more he can issue more requests

src/org/opensolaris/opengrok/web/api/controller/ProjectsController.java
```diff
@@ -0,0 +1,333 @@
...
+                            // since {@code getReposFromString()} inside
+                            // {@code removeCache()} will return nothing.
+                            return "";
+                        }
+                    }).collect(Collectors.toSet()));
+        }
+    }
+
+    @POST
+    @Path("markIndexed")
```
tulinkry: `PUT` and path should be only `indexed`

src/org/opensolaris/opengrok/web/api/controller/ProjectsController.java
```diff
@@ -0,0 +1,333 @@
...
+                        .map(RepositoryInfo::getDirectoryNameRelative)
+                        .collect(Collectors.toList()));
+            }
+        }
+
+        return repos;
+    }
+
+    @GET
+    @Path("repositoriesType")
```
tulinkry: `project/repositories/type`

src/org/opensolaris/opengrok/web/api/controller/ConfigurationController.java
```diff
@@ -0,0 +1,88 @@
...
+            ClassUtil.invokeSetter(env.getConfiguration(), field, value);
+        } catch (IOException e) {
+            throw new WebApplicationException(e, Response.Status.BAD_REQUEST);
+        }
+
+        // apply the configuration - let the environment reload the configuration if necessary
+        env.applyConfig(env.getConfiguration(), false, true);
+    }
+
+    @POST
```
tulinkry: `PATCH` or `PUT`

src/org/opensolaris/opengrok/web/api/controller/StatsController.java
```diff
@@ -0,0 +1,60 @@
...
+    public String get() {
+        return Util.statisticToJson(env.getStatistics()).toJSONString();
+    }
+
+    @DELETE
+    public void clean() {
+        env.setStatistics(new Statistics());
+    }
+
+    @POST
```
tulinkry: `PATCH` or `PUT`

src/org/opensolaris/opengrok/web/api/controller/SystemController.java
```diff
@@ -17,30 +17,29 @@
...
+    private final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
 
-    @Override
-    public void validate() throws Exception {
-        if (getTags().isEmpty()) {
-            getTags().add(RuntimeEnvironment.MESSAGES_MAIN_PAGE_TAG);
-        }
-        super.validate();
+    @POST
+    @Path("/refresh")
```
tulinkry: `PATCH` or `PUT`

src/org/opensolaris/opengrok/web/api/controller/StatsController.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public String get() {
+        return Util.statisticToJson(env.getStatistics()).toJSONString();
+    }
+
+    @DELETE
+    public void clean() {
+        env.setStatistics(new Statistics());
```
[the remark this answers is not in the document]
Orviss: https://stackoverflow.com/questions/25970523/restful-what-should-a-delete-response-body-contain

src/org/opensolaris/opengrok/web/api/controller/StatsController.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @POST
+    @Path("reload")
+    public void reload() throws IOException, ParseException {
+        env.loadStatistics();
```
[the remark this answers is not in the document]
Orviss: Is it a problem? It will return 204 so it is okay. And it cannot be `GET` because this changes stuff and would go against the REST principles.

tulinkry:
Overall thing which I don't like here is that this violates REST design principles (while it just copies the functionality we had here).

Mostly this covers:
 - doing complex scenarios instead of a single scenario (action for multiple projects while there should be for one)
 - using different HTTP methods than they should be
 - describing resources with verbs (while this is non-blocking here for our use case)

So I'm giving this to discussion if this is important for us or not. @vladak @tarzanek

tulinkry:
Eventhough this is a great step forward so keep moving !

src/org/opensolaris/opengrok/web/api/controller/StatsController.java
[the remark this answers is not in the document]
tulinkry: and also `PATCH` should be here
it's not a problem but before I think we returned the statistics for reload and clean

src/org/opensolaris/opengrok/web/api/controller/StatsController.java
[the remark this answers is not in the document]
tulinkry: yes, but droping existing feature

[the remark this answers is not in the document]
Orviss: In this case I think it is redundant to always return `{}`. And as you mentioned, it should be as clean as possible; therefore, 2 requests one for delete and second for get would be more clear. Otherwise it does 2 things whereas it should do one.

src/org/opensolaris/opengrok/web/api/filter/LocalhostFilter.java
```diff
@@ -0,0 +1,72 @@
...
+                localAddresses.add(inetAddress.getHostAddress());
+            }
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Could not get localhost addresses", e);
+        }
+    }
+
+    @Override
+    public void filter(final ContainerRequestContext context) {
+        if (!localAddresses.contains(request.getRemoteAddr())) {
```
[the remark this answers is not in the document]
Orviss: From servlet API:

> Returns the Internet Protocol (IP) address of the client or last proxy that sent the request. For HTTP servlets, same as the value of the CGI variable REMOTE_ADDR.

So it seems it should be safe. However, I'm no security expert :D

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
Orviss: This does this: `tags=one&tags=two`.

Here they mention that there is no standard how to do this:
https://stackoverflow.com/questions/6243051/how-to-pass-an-array-within-a-query-string

I think the Spring needs `test[]` params the same as PHP. But Jersey can deal with this by itself somehow...

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
Orviss: 204 No content

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
Orviss: There was no support for that in `Messages` as far as I know so I tried to avoid adding functionality. But I will add it :)

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
Orviss: I was actually reading about this because I was not sure and I was not able to find any proper explanation. They say that in body should be data. However, it was not completely clear what they mean by that. I chose this approach only because `duration` and `cssClass` have default values which are not changed often so the ease of use is better here...

However, I have no problem with rewriting it.

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
tulinkry: this is ok, but maybe the keyword should be `tag` instead of `tags`

Orviss: I was thinking that as well for `projects` but decided to leave it. I will change it.

src/org/opensolaris/opengrok/web/api/controller/MessagesController.java
tulinkry: Query parameters are usef for `GET` queries for filtering and searching.
Here duration and the others are state of the entitity which are being created thus falls into the data in the request body.

Orviss:
As for `PATCH` there is quite a problem. Jersey does not support it properly... There is no `patch()` method. There is a workaround <https://stackoverflow.com/questions/22355235/patch-request-using-jersey-client>:
```java
target("stats")
                .property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true) // this needs to be added
                .path("reload")
                .request()
                .method("PATCH", Entity.text("")); // patch needs to be written like this
```

However, there is also a comment which mentions that it does not work with `https`...

Maybe we should use different client? This will add another dependency though.

tulinkry:
Then I'm for `PUT` as it signals an update, eventhough it may break the idempotency rule.

tools/sync/reindex-project.ksh
```diff
@@ -67,7 +67,7 @@ function reindex_one
 		exit 1
 	fi
 
-	$binary_base/Messages -n config -t getconf > "$config_xml"
+	curl "${OPENGROK_WEBAPP_CFGADDR}/${OPENGROK_WEBAPP_CONTEXT}/api/v1/configuration" > "$config_xml"
```
vladak: the whole script should be really rewritten into Python so that we don't introduce additional dependency on `curl`.

Orviss: Do you volunteer? :D I'm not skilled in Python at all. However, if I have to do it then I will :D

Orviss: Or is it better to use `wget`? Or some other tool?

tulinkry: I think curl will do the job for now

src/org/opensolaris/opengrok/web/api/RestApp.java
```diff
@@ -30,7 +30,7 @@
 public class RestApp extends ResourceConfig {
 
     public RestApp() {
-        packages("org.opensolaris.opengrok.web.api.controller", "org.opensolaris.opengrok.web.api.filter");
+        packages("org.opensolaris.opengrok.web.api.v1.controller", "org.opensolaris.opengrok.web.api.filter");
```
tulinkry: `api.v1.filter` also?

src/org/opensolaris/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -0,0 +1,308 @@
...
+        HistoryGuru histGuru = HistoryGuru.getInstance();
+
+        // There is no need to perform the work of invalidateRepositories(),
+        // since addRepositories() calls getRepository() for each of
+        // the repos.
+        return new ArrayList<>(histGuru.addRepositories(new File[]{projDir}, env.getIgnoredNames()));
+    }
+
+    @DELETE
+    public void deleteProject(@QueryParam("project") final String projectName)
```
tulinkry: the proejct should be a part of the path so `DELETE projects/example1`

src/org/opensolaris/opengrok/web/api/RestApp.java
Orviss: Why? I thought that this filter will probably be used later as well. Until we come up with different security approach. 

Either way, I think that it won't be a problem to rename it.

src/org/opensolaris/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -0,0 +1,308 @@
...
+@Path("/projects")
+public class ProjectsController {
+
+    private static final Logger logger = LoggerFactory.getLogger(ProjectsController.class);
+
+    private RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @PUT
+    @Consumes(MediaType.TEXT_PLAIN)
+    public void addProject(final String projectName) {
```
tulinkry: now this should be a POST (creating a new entity or semantically recreating)
and return should be 202 CREATED

src/org/opensolaris/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -0,0 +1,308 @@
...
+                        // {@code removeCache()} will return nothing.
+                        return "";
+                    }
+                }).collect(Collectors.toSet()));
+    }
+
+    @PUT
+    @Path("indexed")
+    @Consumes(MediaType.TEXT_PLAIN)
+    public void markIndexed(final String projectName) throws Exception {
```
tulinkry: `PUT projects/example1/indexed`

src/org/opensolaris/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -0,0 +1,308 @@
...
+        env.maybeRefreshIndexSearchers(Collections.singleton(projectName));
+
+        env.refreshDateForLastIndexRun();
+    }
+
+    @PUT
+    @Path("property/{field}")
+    public void set(
+            @PathParam("field") final String field,
+            @QueryParam("project") final String projectName,
```
tulinkry: `PUT projects/{project}/property/{field}`

src/org/opensolaris/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -0,0 +1,308 @@
...
+            }
+        } else {
+            logger.log(Level.WARNING, "cannot find project {0} to set a property", projectName);
+        }
+    }
+
+    @GET
+    @Path("property/{field}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Object get(@PathParam("field") final String field, @QueryParam("project") final String projectName)
```
tulinkry: `GET projects/{project}/property/{field}`

src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
```diff
@@ -30,34 +30,22 @@
...
-            for (RepositoryInfo ri : env.getRepositories()) {
-                if (ri.getDirectoryNameRelative().equals(tag)) {
-                    types.add(tag + ":" + ri.getType());
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) {
-                types.add(tag + ":N/A");
+    public String getType(@QueryParam("repository") final String repository) {
```
tulinkry: `GET repositories/{repository}/type`

src/org/opensolaris/opengrok/web/api/v1/controller/MessagesController.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.opensolaris.opengrok.web.messages.MessagesContainer.MESSAGES_MAIN_PAGE_TAG;
+
+@Path("/messages")
+public class MessagesController {
+
+    private final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void addMessage(@Valid final Message message) {
```
tulinkry: `POST` and `202 CREATED` is more appropriate here

src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
```diff
@@ -0,0 +1,51 @@
...
+
+@Path("/repositories")
+public class RepositoriesController {
+
+    private RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @GET
+    @Path("/type")
+    @Produces(MediaType.APPLICATION_JSON)
+    public String getType(@QueryParam("repository") final String repository) {
```
tulinkry: `GET repositories/{repository}/type`

src/org/opensolaris/opengrok/web/api/RestApp.java
tulinkry: I didn't study the framework but this creates the app looking for controllers in v1.controller and it should be looking into v1.filter for filters to remain consistent.

If anyone wants to create other filter in other version there would be v2 and one can extend v1 filters or create new ones.

tulinkry:
My last main question is, can the api be documented somehow? I mean for example with annotations stating some simple description of the route and if it can generate the static doc out of it accessible for example at `/source/api/v1/doc`?

tulinkry:
It doesn't need to be super cool or something. Swagger looks nice.

Orviss:
Doesn't have to be but can be :D

tools/sync/reindex-project.ksh
vladak: It does not make too much sense to rewrite the script until the `OpenGrok` script is rewritten too so it's okay to use `curl` for now.

tools/sync/sync.py
```diff
@@ -98,8 +98,8 @@ def worker(base):
                         help='config file in JSON format')
     parser.add_argument('-I', '--indexed', action='store_true',
                         help='Sync indexed projects only')
-    parser.add_argument('-m', '--messages',
-                        help='path to the Messages binary')
+    parser.add_argument('-u', '--url',
```
vladak: add a sane default based on `localhost` ?

tools/sync/projadm.py
```diff
@@ -197,8 +202,8 @@ def project_delete(doit, logger, project, messages):
                         help='OpenGrok instance base directory')
     parser.add_argument('-R', '--roconfig',
                         help='OpenGrok read-only configuration file')
-    parser.add_argument('-m', '--messages',
-                        help='path to the Messages binary')
+    parser.add_argument('-u', '--url',
```
vladak: default

tools/sync/mirror.py
```diff
@@ -80,8 +80,8 @@
                         help='Enable debug prints')
     parser.add_argument('-c', '--config',
                         help='config file in JSON/YAML format')
-    parser.add_argument('-m', '--messages',
-                        help='path to the Messages binary')
+    parser.add_argument('-u', '--url',
```
vladak: default

Orviss:
There is a minor problem: repositories often contain slashes -> problem with `repositories/{repository}/type`. So URL encoding should solve it, right? However, Tomcat returns `HTTP 400 Invalid URI: noSlash`. The solution might be to add `-Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true`... But it is still bad to force that :/

tools/sync/opengrok.py
```diff
@@ -22,81 +22,108 @@
...
-    return get_first_line(logger, [messages_file, '-n', 'config', '-t',
-                          'get', name])
+    idx = line.rfind(":")
+    return line[idx + 1:]
 
 
-def get_repo_type(logger, path, messages_file):
-    """
-    Get repository type for given path relative to sourceRoot.
+def get_configuration(logger, host):
```
vladak: the `host` param should probably be named `uri` ?

vladak: same for other functions

tools/sync/projadm.py
```diff
@@ -168,13 +174,12 @@ def project_delete(doit, logger, project, messages):
...
-                 [messages, '-n', 'project', '-t', project, 'delete'],
-                 "deletion of the project failed")
 
-    src_root = exec_command(True, logger,
-                            [messages, '-n', 'config', '-t', 'get',
-                             'sourceRoot'], "cannot get config")
+    if doit:
+        delete_project(logger, project, host)
+
+    src_root = get_config_value(logger, 'sourceRoot', host)
```
vladak: should probably check the return value afterwards given the function can return `None`

In general it is not clear to me what are the failure scenarios for `requests`, e.g. what if:
  - the URI is specified incorrectly
  - the HTTP request fails
  - the service returns error

vladak:
The python scripts should probably have their `__version__` bumped.

vladak:
As for the API documentation, there is this company called Apiary that Oracle bought.. I was thinking about having a special API endpoint (e.g. `/api/v1/repositories/help`) that would display some HTML-ized/Markdown help content.

Orviss:
Why `repositories` in `/api/v1/repositories/help`? Why not just `/api/v1/help`?

vladak:
Dumping the whole help thing might be too much content to swallow at once.

tools/sync/projadm.py
Orviss: If the service returns `3xx`, `4xx`, `5xx` then `if r:` returns `False`.

On error `ConnectionError` is thrown.

vladak:
I mean - `/api/v1/projects` would display just the help for project message, `/api/v1/repositories` for repository message etc.

tools/sync/sync.py
Orviss: What is a proper default? `localhost:8080`, `localhost:8080/source`, `localhost`, `localhost/source`?

tulinkry:
I believe this is what modern tools do when doing automated documentation of an api endopoints.

tulinkry:
> There is a minor problem: repositories often contain slashes -> problem with repositories/{repository}/type. So URL encoding should solve it, right? However, Tomcat returns HTTP 400 Invalid URI: noSlash. The solution might be to add -Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true... But it is still bad to force that :/

Heh. Then maybe we can live with the violation in repositories.
But won't `GET repositories/type?repository=abc/ded` be also offensive?

tools/sync/opengrok.py
vladak: I am all for renaming.

vladak:
Any configuration of Tomcat (-D) should be avoided IMHO.

Orviss:
> Heh. Then maybe we can live with the violation in repositories.
But won't GET repositories/type?repository=abc/ded be also offensive?

Nope, seems to work fine :)

tools/sync/sync.py
Orviss: Also there is problem with the schema. If I use just `localhost` then `requests` don't work. `http://` or `https://` needs to be specified :/

vladak:
As for the documentation, I think it is better to be figured out separately as there are many choices there and this pull req is already big enough.

tulinkry:
> As for the documentation, I think it is better to be figured out separately as there are many choices there and this pull req is already big enough.

Yes.

tulinkry:
> >    Heh. Then maybe we can live with the violation in repositories.
    But won't GET repositories/type?repository=abc/ded be also offensive?

> Nope, seems to work fine :)

Then ok, stay with this but just for the repositories.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @PostConstruct
+    public void init() {
+        try {
+            localAddresses.add(InetAddress.getLocalHost().getHostAddress());
+            for (InetAddress inetAddress : InetAddress.getAllByName("localhost")) {
+                localAddresses.add(inetAddress.getHostAddress());
+            }
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Could not get localhost addresses", e);
```
vladak: this means the filter will not work. Is there any way how to bail out and cancel the deploy ?

Orviss: `System.exit(1)`? :D Or just API?

vladak: I wonder what if the exception is left alone, possibly that would do the job ?

distribution/assembly.xml
```diff
@@ -62,7 +62,6 @@
           <outputDirectory>bin</outputDirectory>
           <includes>
             <include>ConfigMerge</include>
-            <include>Messages</include>
```
tulinkry: Any other tool in the plan? To wrap the most common calls.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
Orviss: If it throws an exception then:
```
28-Jun-2018 10:59:25.009 WARNING [RMI TCP Connection(2)-127.0.0.1] org.glassfish.jersey.internal.Errors.logErrors The following warnings have been detected: WARNING: Unknown HK2 failure detected:
MultiException stack 1 of 2
java.io.IOException: test
	at org.opensolaris.opengrok.web.api.v1.filter.LocalhostFilter.init(LocalhostFilter.java:55)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
...
```

And when I try to access the API then I get:
```
avax.servlet.ServletException: Servlet.init() for servlet [org.opensolaris.opengrok.web.api.v1.RestApp] threw exception
...
```

So `RestApp` won't work.

opengrok-indexer/pom.xml
```diff
@@ -530,11 +534,43 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
...
+            <artifactId>jersey-bean-validation</artifactId>
+            <version>${jersey.version}</version>
+        </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
```
tulinkry: maybe I missed it but how is this used?

Orviss: To check if expiration timer works for messages:

`await().atMost(2, TimeUnit.SECONDS).until(() -> container.getMessages("info").isEmpty());`

because it is a bad practice to use `Thread.sleep(...)` in tests.

distribution/assembly.xml
Orviss: As far as I'm concerned I am not a big fan of this because with large API it might be impossible to maintain (huge switch, lot of options, etc.). And most of the calls are 1 liners and can be performed by common tools e.g. `curl, wget, ...`.

opengrok-indexer/pom.xml
tulinkry: Ok. Just missed it.
Isn't there something like a dependency only for tests?

Orviss: `<scope>test</scope>` take care of that.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
tulinkry: Isn't it better to log the exception but not break the api and then check if the host list is not empty? so if this fails and it is empty 401 will be still returned

distribution/assembly.xml
tulinkry: ok

opengrok-indexer/pom.xml
tulinkry: I'm blind sorry

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
Orviss: This is how it works now, right? At first the set is empty so:
```java
if (!localAddresses.contains(request.getRemoteAddr())) {
            context.abortWith(Response.status(Response.Status.UNAUTHORIZED).build());
 }
```

will return `UNAUTHORIZED`

src/org/opensolaris/opengrok/web/api/filter/LocalhostFilter.java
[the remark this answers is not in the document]
vladak: `getRemoteAddr()` is just an interface so it will depend on each application server how it will be implemented. I believe for Tomcat this will eventually bubble down to socket implementation so it should do the right thing.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
vladak: Aha, should be fine then.

distribution/assembly.xml
vladak: as long as the API is properly documented (pls file a new issue to track this) it should be fine to live without the CLI tool.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
tulinkry: :-D sorry. It's morning

vladak:
I am okay with merging this now. https://github.com/oracle/opengrok/wiki/Messages needs to be updated after integration.

Orviss:
Wouldn't be better to remove the `Messages` wiki page and add information to <https://github.com/oracle/opengrok/wiki/Web-services>?

vladak:
Makes sense.

Dne čt 28. 6. 2018 15:22 uživatel Adam Hornáček <notifications@github.com>
napsal:

tulinkry:
I'm also ok.

PR #2186, opened 2018-06-27
reduce test coverage gaps

vladak:
This is mainly to reduce intermittent code coverage loss as reported e.g. in https://coveralls.io/builds/17674160/source?filename=src%2Forg%2Fopensolaris%2Fopengrok%2Fhistory%2FRepositoryFactory.java#L116.


vladak:
For some reason, Coveralls is not updating this pull request so we will see the benefits only once it integrates or I will submit brand new pull request for this.

vladak:
Aaa, there we go.

PR #2189, opened 2018-06-28
Search endpoint filtered by localhost filter fix

Orviss:
I totally forgot to exclude `search` from `LocalhostFilter`. Therefore, this fix.

Thanks :)


vladak:
Ideally this should be covered by a test case.

Dne čt 28. 6. 2018 16:49 uživatel Adam Hornáček <notifications@github.com>
napsal:

Orviss:
Yes, I was thinking about it. However, even if I load the whole `RestApp` in `JerseyTest` then `@Context` variable is not initialized -> that means it will be always `null` -> I was reading on stackoverflow that it is like that and they do not intend to change it soon. (This would not cover different IPs than localhost though). So I was thinking about adding `Mockito` and mocking the request which should work nice.

PR #2190, opened 2018-06-29
Add LocalhostFilter test and other REST API improvements

Orviss:
Features:
* added `LocalhostFilterTest`
* added possibility to retrieve all messages in the system
* fixed search endpoint -> it was not working properly because I misinterpreted hits
 -> now `maxresults` specifies maximum number of documents that may be retrieved
 -> data are grouped by document `path`
* delete messages -> now only one `tag` can be specified (instead of multiple)
* repositories endpoint fix -> repositories type now returns `TEXT/PLAIN` instead of `APPLICATION/JSON`

I wanted to create `SearchControllerTest` as well but there are some problems with `@Context` injection and I was not sure how to add test repository properly and did not want to spend a lot of time with it. So I will look into it later.

Thanks :)

### DIFF
--- a/src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
+++ b/src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
@@ -38,7 +38,7 @@ public class RepositoriesController {
 
     @GET
     @Path("/type")
-    @Produces(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.TEXT_PLAIN)
     public String getType(@QueryParam("repository") final String repository) {
         for (RepositoryInfo ri : env.getRepositories()) {
             if (ri.getDirectoryNameRelative().equals(repository)) {
--- a/test/org/opensolaris/opengrok/web/api/v1/controller/MessagesControllerTest.java
+++ b/test/org/opensolaris/opengrok/web/api/v1/controller/MessagesControllerTest.java
@@ -237,4 +237,17 @@ public class MessagesControllerTest extends JerseyTest {
         assertEquals(1, allMessages.size());
     }
 
+    @Test
+    public void multipleMessageAndTagsTest() {
+        addMessage("test1", "tag1", "tag2");
+        addMessage("test2", "tag3", "tag4");
+
+        List<AcceptedMessageModel> allMessages = target("messages")
+                .queryParam("tag", "tag3")
+                .request()
+                .get(messagesType);
+
+        assertEquals(1, allMessages.size());
+    }
+
 }
--- a/test/org/opensolaris/opengrok/web/api/v1/controller/MessagesControllerTest.java
+++ b/test/org/opensolaris/opengrok/web/api/v1/controller/MessagesControllerTest.java
@@ -24,15 +24,17 @@ package org.opensolaris.opengrok.web.api.v1.controller;
 
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.test.JerseyTest;
-import org.junit.AfterClass;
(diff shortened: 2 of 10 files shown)


src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
```diff
@@ -38,7 +38,7 @@
 
     @GET
     @Path("/type")
-    @Produces(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.TEXT_PLAIN)
```
tulinkry: I know that it's been like this before but please no for the future :D use json

test/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilterTest.java
```diff
@@ -0,0 +1,113 @@
...
+        when(info.getPath()).thenReturn(path);
+
+        when(context.getUriInfo()).thenReturn(info);
+
+        return context;
+    }
+
+    @Test
+    public void localhostTest() throws Exception {
+        LocalhostFilter filter = mockWithRemoteAddress("127.0.0.1");
```
tulinkry: what about 0:0:0:0:0:0:0:1?

test/org/opensolaris/opengrok/web/api/v1/controller/MessagesControllerTest.java
```diff
@@ -186,4 +199,42 @@ private void setText(final Message m, final String text) throws Exception {
...
+
+        assertEquals(1, messages.size());
+        assertEquals("text", messages.get(0).message.getText());
+
+        assertThat(messages.get(0).message.getTags(), contains("info"));
+    }
+
+    @Test
+    public void multipleTagsTest() {
+        addMessage("test", "info", "main");
```
tulinkry: I would also add a test for two messages, both multiple distinct tags, and selecting one tag and expecting one message in the output.

src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
Orviss: Well, I don't know what is exactly the rule. But if I'm returning just `String` which has no formatting then is it not better to return `TEXT/PLAIN`?

test/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilterTest.java
Orviss: `LocalhostFilter` works with it. I can add test case for it, though.

src/org/opensolaris/opengrok/web/api/v1/controller/RepositoriesController.java
tulinkry: Yes that is correct, but I mean in some future PR convert the API to json results.

test/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilterTest.java
tulinkry: If it doesn't kill you :-)

tulinkry:
> Should I delete the localhost IPv6 test or do you have any suggestion how to resolve this?

I wouldn't delete it, it happened to me in the past that I wasn't able to access something with just ipv6 localhost address. 

I would add to the already discovered addresses also some hardcodes like `0:0:0:0:0:0:0:1` or `localhost`. To catch these cases.

tulinkry:
Btw that is exactly the reason I wanted you to add the test for ipv6 :D

PR #2192, opened 2018-06-29
Main tag fix

vladak:


src/org/opensolaris/opengrok/search/Summary.java
```diff
@@ -75,7 +75,7 @@ protected static String htmlize(String q) {
 
     private static final Fragment[] FRAGMENT_PROTO = new Fragment[0];
 
-    /** Adds a fragment to a summary.*/
+    /* Adds a fragment to a summary.*/
```
Orviss: Why remove javadoc formatting? I think in this place it makes sense.

vladak: javadoc complains about missing @param and the comment is self-explaining.

PR #2199, opened 2018-07-02
sync.py needs to use RESTful API calls

vladak:
`sync.py` needs to be able to call RESTful API for `Messages`. This change complements the sync config file so that it is possible to do by RESTful calls next to executing programs. This changes makes this possible by replacing `%PROJECT%` in both URL and body by project name passed to `sync.py`.

Tested with:

```json
{
   "commands": [
                ["http://localhost:8080/source/api/v1/messages", "POST",
                 { "cssClass" : "info", "duration" : "PT1H",
                   "tags" : ["%PROJECT%"], "text" : "resync + reindex in progress"
                 }
                ]
               ]
}
```

by running `sync.py` like this:

```shell
./sync.py -D -P opengrok-master -c /var/tmp/sync-config-API.json
```

and checking that the project received the exclamation mark and message in the UI.

and then removing it with:

```json
{
   "commands": [
                ["http://localhost:8080/source/api/v1/messages?tag=%PROJECT%", "DELETE", ""]
               ]
}
```

and checking the flag is gone in the UI.

In my original testing had to suppress proxy (by running the command as `htp_proxy="" ./sync.py ...` since I have it set in my environment (`http_proxy` and `https_proxy` env vars) and our proxies refuse (logically) to make requests to `localhost`. I solved this by instructing `requests` not to use proxies upon seeing request to localhost. This method will keep using the proxies from environment if the request is not to localhost.

https://github.com/oracle/opengrok/wiki/Repository-synchronization will be updated once this is in.


vladak:
Also, in order to supply HTTP headers, I had to use older `requests` API since this is what our production runs with.

tools/sync/commands.py
```diff
@@ -67,35 +70,68 @@ def __init__(self, base):
...
+                        continue
+                    data = None
+                else:
+                    headers = {'Content-Type': 'application/json'}
+                    json_data = json.dumps(data).replace(PROJECT_SUBST,
+                                                         self.name)
+                    self.logger.debug("JSON data: {}".format(json_data))
+
+                if verb == 'PUT':
+                    put(self.logger, uri, headers=headers, data=json_data)
```
Orviss: `requests` have `json` param which will do conversion to json for you. Not sure if this is what you want since you are doing string replacement there... Just a remark :)

tools/sync/commands.py
```diff
@@ -67,35 +70,68 @@ def __init__(self, base):
...
-                                     format(cmd))
-                    self.run_cleanup()
+            if is_web_uri(command[0]):
+                uri = command[0].replace(PROJECT_SUBST, self.name)
+                verb = command[1]
+                data = command[2]
+                if len(data) == 0:
+                    # PUT/POST need the data.
+                    if verb in ['PUT', 'POST']:
+                        self.logger.error('Empty data for PUT/POST')
```
Orviss: I think there are some cases when `PUT` and `POST` can have an empty body in our REST impl. E.g. marking project as indexed : `/api/v1/{project}/indexed`...

Maybe just log a warning? Or if you prefer you can change the REST API but I did not know how :/

tools/sync/commands.py
```diff
@@ -67,35 +70,68 @@ def __init__(self, base):
...
-
-            # If a command fails, terminate the sequence of commands.
-            retcode = cmd.getretcode()
-            if retcode != 0:
-                if retcode == 2:
-                    self.logger.info("command '{}' requested break".
-                                     format(cmd))
-                    self.run_cleanup()
+            if is_web_uri(command[0]):
+                uri = command[0].replace(PROJECT_SUBST, self.name)
```
tulinkry: maybe python builtin format method would be better

`'Coordinates: {latitude}, {longitude}'.format(latitude='37.24N', longitude='-115.81W')`

tools/sync/commands.py
Orviss: However, if you use it then it won't be needed to use custom headers yet. (I hope it sets the `application/json` header by itself).

tools/sync/commands.py
tulinkry: We can return something in the REST API - the updated state of the entity or some message in json? Not that important though.

tools/sync/opengrok.py
```diff
@@ -43,21 +43,37 @@ def delete(logger, uri, params=None):
         return None
 
 
-def post(logger, uri, data=None):
+def post(logger, uri, headers=None, data=None):
+    rv = None
     try:
-        return requests.post(uri, data=data)
+        if headers:
+            s = requests.Session()
```
tulinkry: why? can't you pass the headers directly? This doesn't seem neccessary to me

tools/sync/commands.py
vladak: I know about the `json` param. However, the project needs to be substituted first. Also, the version of `requests` I am working with does not have the param yet.

tools/sync/commands.py
vladak: Aha, removed.

tools/sync/opengrok.py
vladak: yes, as mentioned in the pull req comment the version of `requests` I am working with does not have the param yet.

tools/sync/commands.py
tulinkry: I looked in the doc and it is there http://docs.python-requests.org/en/v1.0.4/user/quickstart/ and the 1.0.4 is from the 2012, which kind of version do you use?

tools/sync/opengrok.py
tulinkry: I looked in the doc and it is there http://docs.python-requests.org/en/v1.0.4/user/quickstart/ and the 1.0.4 is from the 2012, which kind of version do you use?

tools/sync/commands.py
tulinkry: Aha. Nice

tools/sync/opengrok.py
vladak: Hmm, I tried again and it works, not sure where I tested this originally. Let me rewrite it.

tools/sync/opengrok.py
```diff
@@ -168,3 +179,22 @@ def delete_project(logger, project, uri):
...
     return '/'.join(s.strip('/') for s in uri_parts)
+
+
+def is_localhost_url(url):
+    """
+    Check if given URL is based on localhost.
+    """
+
+    o = urlparse(url)
+    return o.hostname == 'localhost'
```
Orviss: Maybe add loopback addresses? Or no need for that?

vladak: can do.

PR #2203, opened 2018-07-03
Authorize search REST endpoint

Orviss:
Authorize the search endpoint. This will be necessary when whole REST API will be made an exempt from web authorization.

Thanks :)

### DIFF
--- a/src/org/opensolaris/opengrok/web/api/v1/controller/SearchController.java
+++ b/src/org/opensolaris/opengrok/web/api/v1/controller/SearchController.java
@@ -24,6 +24,8 @@ package org.opensolaris.opengrok.web.api.v1.controller;
 
 import org.opensolaris.opengrok.search.Hit;
 import org.opensolaris.opengrok.search.SearchEngine;
+import org.opensolaris.opengrok.web.api.v1.filter.AuthorizationFilter;
+import org.opensolaris.opengrok.web.api.v1.filter.Authorized;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.DefaultValue;
@@ -49,6 +51,7 @@ public class SearchController {
     private static final int MAX_RESULTS = 1000;
 
     @GET
+    @Authorized
     @Produces(MediaType.APPLICATION_JSON)
     public SearchResult search(
             @Context final HttpServletRequest req,
@@ -58,7 +61,7 @@ public class SearchController {
             @QueryParam("path") final String path,
             @QueryParam("hist") final String hist,
             @QueryParam("type") final String type,
-            @QueryParam("projects") final List<String> projects,
+            @QueryParam(AuthorizationFilter.PROJECTS_PARAM) final List<String> projects,
             @QueryParam("maxresults") @DefaultValue(MAX_RESULTS + "") final int maxResults,
             @QueryParam("start") @DefaultValue(0 + "") final int startDocIndex
     ) {
--- a/src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
+++ b/src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
@@ -1,0 +1,76 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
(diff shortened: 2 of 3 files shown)


src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
```diff
@@ -0,0 +1,76 @@
...
+    private String[] getProjects() {
+        String[] projects = request.getParameterValues(PROJECTS_PARAM);
+        if (projects != null) {
+            return projects;
+        }
+
+        // if no projects are specified then it means that all projects are wanted
+        Map<String, Project> projectsMap = env.getProjects();
+        if (projectsMap != null) {
+            return projectsMap.keySet().toArray(new String[0]);
```
tulinkry: `new String[projectsMap.size()]`?

src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
```diff
@@ -0,0 +1,76 @@
...
+    private final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    @Context
+    private HttpServletRequest request;
+
+    @Override
+    public void filter(final ContainerRequestContext context) {
+        AuthorizationFramework auth = env.getAuthorizationFramework();
+        if (auth != null) {
+            for (String project : getProjects()) {
```
tulinkry: have a look on `ProjectHelper.java` if this can not be optimized

tulinkry: non blocking though

src/org/opensolaris/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -49,6 +51,7 @@
     private static final int MAX_RESULTS = 1000;
 
     @GET
+    @Authorized
```
tulinkry: cool

src/org/opensolaris/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -58,7 +61,7 @@ public SearchResult search(
             @QueryParam("path") final String path,
             @QueryParam("hist") final String hist,
             @QueryParam("type") final String type,
-            @QueryParam("projects") final List<String> projects,
+            @QueryParam(AuthorizationFilter.PROJECTS_PARAM) final List<String> projects,
             @QueryParam("maxresults") @DefaultValue(MAX_RESULTS + "") final int maxResults,
             @QueryParam("start") @DefaultValue(0 + "") final int startDocIndex
```
tulinkry: now this is not behind the localhost filter and is just covered with the authorization filter?

src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
Orviss: I used to do it as well. Only recently IDEA started to suggest:

> There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

src/org/opensolaris/opengrok/web/api/v1/controller/SearchController.java
Orviss: Yep, `LocalhostFilter` has an exemption for the `search` path.

src/org/opensolaris/opengrok/web/api/v1/filter/AuthorizationFilter.java
tulinkry: ok

vladak:
This is actually not needed, the search will be authorized via `ProjectHelper.filterProjects()`

vladak:
search_authz

tulinkry:
ah, good.

PR #2204, opened 2018-07-03
Allow RESTful API in AuthorizationFilter

vladak:
This explicitly allows RESTful API calls to go through `AuthorizationFilter`. It seems this is not strictly necessary (as mentioned in #2203 for the `/search` endpoint) however it is better not to rely on the fact that `config.getProject()` returns `null` in `AuthorizationFilter.doFilter()`.

### DIFF
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -29,6 +29,16 @@
                     <warSourceDirectory>../web</warSourceDirectory>
                 </configuration>
             </plugin>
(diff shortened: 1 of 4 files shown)


vladak:
I also added Tomcat 7 plugin for Maven (works for Tomcat 8 too), ideally the URL should be configurable via `~/.m2/settings.xml` file however even @Orviss could not make that work so leaving it hardcoded for now.

PR #2212, opened 2018-07-10
Suggester

Orviss:
Hello,

running instance can be found at <http://demo.opengrok.org>. I'm sorry that it is in the demo already without being merged but I would like to monitor the usage and notice some errors.

Not yet complete text regarding this change can be found at <https://github.com/Orviss/thesis> (for PDF check releases).

To sum up:
* added suggestions support
* suggestions for multiple queries
  * prefix queries
  * wildcard queries
  * fuzzy queries
  * phrase queries
  * regexp queries
  * range queries
  * grouping queries (AND, OR)
  * sloppy phrase queries
* most popular completion (if some term is searched for many times then it will be higher in the suggestions) -> for this ChronicleMap (<https://github.com/OpenHFT/Chronicle-Map>) library was chosen because its performance was the best
* RAM usage will be higher (data for lookup must be stored in RAM to provide quick access)
* there is a space for a lot of optimizations and extensions

Suggester core functionality is in a new separate module.

I will be thankful for any notes, remarks or questions :) It will help me a lot!

Thanks a lot :)


web/js/utils-0.0.23.js
```diff
@@ -1716,32 +1716,250 @@ function pageReadyMast() {
...
+        source: function(request, response) {
+
+            var caretPos = $('#' + inputId).caret();
+            if (!Number.isInteger(caretPos)) {
+                console.error("Suggest: could not get caret position");
+                return;
+            }
+
+            $.ajax({
+                url: window.contextPath + "/suggest",
```
tulinkry: does it work in the @vladak setup? with the authorization and all these stuff?

web/js/utils-0.0.23.js
```diff
@@ -2038,3 +2266,20 @@ function textInputHasFocus() {
         document.activeElement.nodeName === 'INPUT' &&
         document.activeElement.type === 'text';
 }
+
+var htmlEscapeMap = {
```
tulinkry: can this be a local variable in the escapeHtml?

src/org/opensolaris/opengrok/web/SearchHelper.java
```diff
@@ -373,6 +374,8 @@ public SearchHelper prepareExec(Project project) {
      * @return this instance
      */
     public SearchHelper executeQuery() {
+        SuggesterServiceImpl.getInstance().onSearch(projects, query); // notify suggester that query was searched
```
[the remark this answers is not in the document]
Orviss: yes, I added the same call to `SuggesterController`

src/org/opensolaris/opengrok/web/Util.java
```diff
@@ -1840,4 +1842,45 @@ public static String completeUrl(String url, HttpServletRequest req) {
...
+        }
+
+        String[] pairs = url.getQuery().split("&");
+
+        for (String pair : pairs) {
+            if (pair.isEmpty()) {
+                continue;
+            }
+
+            int idx = pair.indexOf('=');
```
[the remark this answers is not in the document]
Orviss: I will create a test case for that :)

src/org/opensolaris/opengrok/web/suggester/parser/SuggesterQueryDataParser.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.logging.Logger;
+
+/**
+ * Parser for the raw {@link SuggesterQueryData}.
+ */
+public class SuggesterQueryDataParser {
+
+    private static final int IDENTIFIER_LENGTH = 5;
+
+    private static final Logger logger = Logger.getLogger(SuggesterQueryDataParser.class.getName());
```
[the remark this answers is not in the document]
Orviss: No `.class` in `getLogger()` -> https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html

I cannot use OpenGrok's `LoggerFactory.getLogger` since this module has no dependency to it.

src/org/opensolaris/opengrok/web/suggester/provider/service/impl/SuggesterServiceImpl.java
```diff
@@ -0,0 +1,283 @@
...
+                }
+                return null;
+            }).filter(Objects::nonNull).collect(Collectors.toList());
+        } else {
+            String project;
+            if (!projects.isEmpty()) {
+                project = projects.iterator().next();
+            } else {
+                project = DEFAULT_PROJECT_NAME;
+            }
```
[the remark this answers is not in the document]
Orviss: This is somehow last resort, I need to show the user in which project the suggestions were for, I think just the empty String `""` would be better. I will change it.

web/js/utils-0.0.23.js
Orviss: Yes. I will change it.

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
```diff
@@ -1717,6 +1718,7 @@ public void applyConfig(Configuration config, boolean reindex, boolean interacti
         getAuthorizationFramework().reload();
 
         messagesContainer.setMessageLimit(config.getMessageLimit());
+        SuggesterServiceImpl.getInstance().refresh(configuration);
```
tulinkry: Only thing I've got here is that perhaps when you use this approach of design (interface - implementaion) then the implementation should never occur anywhere in the code except for the constructing.

So maybe a factory, or factory method would be more appropriate and the suggester service would be a part of `RuntimeEnvironment`? The point is that the concrete implementation is always hidden to the end user.

If it cannot be done, leave it like this.

src/org/opensolaris/opengrok/web/suggester/parser/SuggesterQueryDataParser.java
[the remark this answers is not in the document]
Orviss: Sorry, it's not :D (I somehow got confused) I will repair it :)

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
Orviss: I use the interface-implementation contract in Jersey (`SuggesterAppBinder`) which allows me to do dependency injection (`@Inject`). I cannot use it here since `RuntimeEnvironment` is used in `indexer` :/ I did not know what would be the best solution. At first, I started only using it in the `suggester` context but then I realized that I need to call it from some other parts as well... 

In the near future, after mavenization is in place, I would like to divide and move relevant parts to `indexer` and `web` modules.

However, still `SuggesterService.getInstance()` would be better (but it's far from what I would prefer) :)

src/org/opensolaris/opengrok/web/Util.java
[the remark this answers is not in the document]
Orviss: I thought that param without `=` is not valid but here they say otherwise <https://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value>

So I repaired it, good catch :)

src/org/opensolaris/opengrok/web/suggester/parser/SuggesterQueryDataParser.java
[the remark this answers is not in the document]
tulinkry: :D

src/org/opensolaris/opengrok/configuration/RuntimeEnvironment.java
tulinkry: So what about `SuggesterServiceFactory.getDefault()` or something like that?

Problem with `SuggesterService.getInstance()` is that it is an interface.

Orviss: What do you mean? You can have `static` methods in Java 8.

tulinkry: Yes but this is a place for a factory not a factory method of an interface. It's definitely better than now but still a way to improve.

Orviss: Done :)

src/org/opensolaris/opengrok/web/suggester/provider/service/SuggesterServiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package org.opensolaris.opengrok.web.suggester.provider.service;
+
+import org.opensolaris.opengrok.web.suggester.provider.service.impl.SuggesterServiceImpl;
+
+public class SuggesterServiceFactory {
+
+    private SuggesterServiceFactory() {
+    }
+
+    public static SuggesterService getDefault() {
```
tulinkry: a javadoc wouldn't kill you :D something like what to expect from the default implementation

Orviss: I was thinking about it but I really hate obvious comments :D 

There will probably never be any other implementation of `SuggesterService` + this class will be probably deleted once the `@Inject` could be used freely

tarzanek:
can we target 1.1?  or 2.0?

tarzanek:
or 1.1 will be called 2.0? ;)

tarzanek:
" Cannot parse '*left': '*' or '?' not allowed as first character in WildcardQuery "
- disabled prefix * by design?

Orviss:
The problem is `allowLeadingWildcard` option in `Configuration` : <http://demo.opengrok.org/xref/opengrok/src/org/opensolaris/opengrok/configuration/Configuration.java?r=c1a21bc0#178>

I will look into that more...

Orviss:
If I override the leading wildcard option in `SuggesterQueryParser` with `setAllowLeadingWildcard(true);` then it works fine. I think this is the wanted behaviour for the suggester. Searches will still behave according to this setting.

web/js/utils-0.0.23.js
vladak: Just tried to deploy on the staging server and seems to work fine w.r.t. authorization (request with faked user SSO headers to `http://localhost:8080/source/suggest?full=&field=full&caret=0&projects%5B%5D=proj` fails with 403 where `proj` is project the user does not have access to)

vladak:
@tarzanek this will be part of 1.1

tarzanek:
just some usability stuff from demo:
if I click into one of the fields it will start suggesting - I think we don't want suggester to kick in for empty fields + it also overflows bottom fields, so it's hard then to click into bottom field if I misclick ;)

if I look at google.com, it will suggest when I click on empty field - my old queries ... (which might be handy later on)

+ when I click on empty field and scroll with mouse with suggester open, it will select first entry (but this is just a nitpick :) )

tulinkry:
> if I click into one of the fields it will start suggesting - I think we don't want suggester to kick in for empty fields + it also overflows bottom fields, so it's hard then to click into bottom field if I misclick ;)

use tab or click somewhere else to close it
on empty click this suggester (if i've got it correctly) will show you the most frequent terms merged with trending queries

> and when I click on empty field and scroll with mouse with suggester open, it will select first entry (but this is just a nitpick :) )

doesn't happen to me, or maybe I'm doing it in a different way

Orviss:
> if I click into one of the fields it will start suggesting - I think we don't want suggester to kick in for empty fields + it also overflows bottom fields, so it's hard then to click into bottom field if I misclick ;)

`esc` should also work, it does not now because of some bug (I will look into it). Or do you have any suggestion how to resolve this? :) 

> if I look at google.com, it will suggest when I click on empty field - my old queries ... (which might be handy later on)

Yes, this is as @tulinkry said. You have a good point but then I would have to remember last n queries in cache. I think this might be a good extension. If someone would not like to have suggestions shown on empty field then they can specify `minChars` in suggester configuration to `1`.

> and when I click on empty field and scroll with mouse with suggester open, it will select first entry (but this is just a nitpick :) )

I don't use a mouse, only trackpad :D I will try to verify when I'm at home :)

Orviss:
>and when I click on empty field and scroll with mouse with suggester open, it will select first entry (but this is just a nitpick :) )

Cannot reproduce. Maybe the mouse just hovered above the suggestion? If yes then it will be focused and the value in the field will be changed. However, this happens ordinarily on mouse hover.

tarzanek:
minChars looks like interesting setting

vladak:
Can you create new wiki page describing the suggester ? (mainly its configuration)

Orviss:
> Can you create new wiki page describing the suggester ? (mainly its configuration)

Done: <https://github.com/oracle/opengrok/wiki/Suggester>

vladak:
Is there a way how to get rid of the `Custom*` classes in the future ? (i.e. file a issue against Lucene to make them expose those private/protected interfaces)

Orviss:
Maybe there is but I don't know of any good way. First of all, I doubt that Lucene will expose them but we can try. From time to time, Lucene changes the inner workings of these classes since it's not a public API (e.g. this happened in transition 7.3.1 -> 7.4.0).

If I were to somehow hack it, e.g. use Lucene's package or use reflection or some other means, it will still be fragile since any Lucene update could break it.

src/org/opensolaris/opengrok/web/api/v1/filter/LocalhostFilter.java
```diff
@@ -47,7 +48,8 @@
 
     private static final Logger logger = LoggerFactory.getLogger(LocalhostFilter.class);
 
-    private static final Set<String> allowedPaths = Collections.singleton("search");
+    private static final Set<String> allowedPaths = new HashSet<>(Arrays.asList(
+            SearchController.PATH, SuggesterController.PATH, SuggesterController.PATH + "/config"));
```
vladak: maybe add a comment why suggester configuration needs to be exposed

src/org/opensolaris/opengrok/web/api/v1/suggester/provider/filter/AuthorizationFilter.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.ws.rs.ext.Provider;
+
+/**
+ * Serves for authorization of specific REST API endpoints.
+ */
+@Provider
+@Authorized
+public class AuthorizationFilter implements ContainerRequestFilter {
+
+    public static final String PROJECTS_PARAM = "projects[]";
```
vladak: should this be taken from elsewhere ? This literal seems to be specific to the request rather than filter.

Orviss: Yes, but I wanted to make the filter more generic. Now, if some other endpoint takes projects in a param it can be authorized by only using the `@Authorized` annotation.

src/org/opensolaris/opengrok/web/api/v1/suggester/provider/service/impl/SuggesterServiceImpl.java
```diff
@@ -0,0 +1,382 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public Suggestions getSuggestions(
+            final Collection<String> projects,
+            final SuggesterQuery suggesterQuery,
+            final Query query
+    ) {
+        List<SuperIndexSearcher> superIndexSearchers = new LinkedList<>();
+        lock.readLock().lock();
```
vladak: why it is necessary to lock this block ?

src/org/opensolaris/opengrok/web/api/v1/suggester/provider/service/impl/SuggesterServiceImpl.java
```diff
@@ -0,0 +1,382 @@
...
+                suggesterConfig.getAllowedFields(),
+                suggesterConfig.getTimeThreshold());
+
+        new Thread(() -> {
+            suggester.init(getAllProjectIndexDirs());
+            scheduleRebuild();
+        }).start();
+    }
+
+    private Collection<NamedIndexDir> getAllProjectIndexDirs() {
```
vladak: this looks a bit like a candidate for some generic utility class

vladak:
when a project is deleted, the suggester data should be removed (see `deleteProject()`).

vladak:
I am assuming that after the integration https://github.com/oracle/opengrok/wiki/Web-services will be updated with the `/suggest` endpoint.

Orviss:
The web services wiki should already be updated. Is it not enough?

src/org/opensolaris/opengrok/web/api/v1/controller/SuggesterController.java
```diff
@@ -0,0 +1,390 @@
...
+                List<String> projects = params.get("project");
+
+                for (String field : QueryBuilder.searchFields) {
+
+                    List<String> fieldQueryText = params.get(field);
+                    if (fieldQueryText == null || fieldQueryText.isEmpty()) {
+                        continue;
+                    }
+                    if (fieldQueryText.size() > 2) {
+                        logger.log(Level.WARNING, "Bad format, ignoring");
```
vladak: log the actual bad data

src/org/opensolaris/opengrok/web/api/v1/controller/SystemController.java
```diff
@@ -35,11 +37,15 @@
...
 
+    @Inject
+    private SuggesterService suggester;
+
     @PUT
     @Path("/refresh")
     @Consumes(MediaType.TEXT_PLAIN)
     public void refresh(final String project) {
         env.maybeRefreshIndexSearchers(Collections.singleton(project));
+        suggester.refresh(project);
```
vladak: I wonder if this is necessary. The refresh here is meant to make the `indexSearcher`s to move to the new index generation. The suggester refresh happens already when project is marked as indexed, no ?

src/org/opensolaris/opengrok/web/api/v1/suggester/provider/service/impl/SuggesterServiceImpl.java
Orviss: Because if suggester is being enabled/disabled or totally refreshed then in `refresh()` method I close the old `suggester`, set it to `null` and then I reinitialize it from scratch. If that were to happen in the middle of the invocation of this method, NPE could be possibly thrown or closed `suggester` could be invoked.

src/org/opensolaris/opengrok/web/api/v1/controller/SuggesterController.java
```diff
@@ -0,0 +1,390 @@
...
+    @POST
+    @Path("/init/queries")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void addSearchCountsQueries(final List<String> urls) {
+        for (String urlStr : urls) {
+            try {
+                URL url = new URL(urlStr);
+                Map<String, List<String>> params = Util.getQueryParams(url);
+
+                List<String> projects = params.get("project");
```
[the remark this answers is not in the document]
Orviss: I only found `QueryBuilder#PROJECT`; however, its meaning is different...

src/org/opensolaris/opengrok/web/api/v1/controller/SystemController.java
Orviss: Well, from the actual implementation in the `Indexer` this is not necessary. This is only called for partial reindex and only after the `markIndexed` call. Therefore, it will probably rebuild the same data structure for the second time. However, the `markIndexed` call also refreshes index searchers so what is the point of calling refresh again via this method? Or did I get it wrong?

From the logical point of view. This is called to notify the webapp that an index changed; therefore, suggester should react, right?

vladak:
> The web services wiki should already be updated. Is it not enough?

I see, the content is fine.

src/org/opensolaris/opengrok/web/api/v1/controller/SystemController.java
vladak: It seems that the refresh done in `main()` via `/refresh` is largely redundant given that the `searcherManager`s are refreshed via `/indexed`. This is not a problem of the suggester, rather than the indexer itself. Filed #2218 to track this.

vladak:
Merging this now, any new problems found will be dealt with as separate issues.

PR #2213, opened 2018-07-11
Allow configuration objects to be set via REST API

Orviss:
Hello,

this is useful for setting POJO objects in configuration. Also useful for suggester (and its config): https://github.com/oracle/opengrok/pull/2212

Example usage:
```bash
curl -d '{"enabled":"false"}' -X PUT http://localhost:8080/api/v1/configuration/suggesterConfig
```

I wanted to write some tests but I was not able to find any POJO class in configuration (`IgnoredNames` does not work and collections (e.g. `Set<Project>`) cannot be converted from JSON because of the type erasure)

I think it makes sense to extend this in the future to allow nested paths (e.g. `/suggesterConfig/enabled`)

Thanks :)

### DIFF
--- a/src/org/opensolaris/opengrok/util/ClassUtil.java
+++ b/src/org/opensolaris/opengrok/util/ClassUtil.java
@@ -45,6 +45,9 @@ public class ClassUtil {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClassUtil.class);
 
+    private ClassUtil() {
+    }
+
     /**
      * Mark all transient fields in {@code targetClass} as @Transient for the
      * XML serialization.
--- a/src/org/opensolaris/opengrok/util/ClassUtil.java
+++ b/src/org/opensolaris/opengrok/util/ClassUtil.java
@@ -33,6 +33,8 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.opensolaris.opengrok.logger.LoggerFactory;
 
 /**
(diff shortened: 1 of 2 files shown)


src/org/opensolaris/opengrok/util/ClassUtil.java
```diff
@@ -165,10 +167,9 @@ public static void invokeSetter(Object obj, String field, String value) throws I
             } else if (paramClass.equals(String.class.getName())) {
                 setter.invoke(obj, value);
             } else {
-                // error unknown type
-                throw new IOException(
-                        String.format("Unsupported type conversion for the name \"%s\". Expecting \"%s\".",
-                                field, paramClass));
+                ObjectMapper mapper = new ObjectMapper();
+                Object objValue = mapper.readValue(value, c);
```
tulinkry: which kind of exceptions does this throw?

tulinkry:
Maybe for this special case of `enable/disable` some features it makes sense to have a dedicated uri; however from the REST point of view the uri should not describe the state of the entitity (thus calling it like in your first example is more appropriate) nor any action which should be taken against the entity (that's why there are HTTP methods).

vladak:
Is there any way this could be tested ?

Orviss:
There needs to a POJO object in the `Configuration`. `IgnoredNames` is not because it extends `Filter` which has `isEmpty()` method but no `empty` field (at least I think that was the problem I tried). I cannot test collections `Set<>` because `Jackson` does not know the type because of type erasure. Only `AuthorizationStack` is left and I was not sure how to create some default instance of it so I left it be...


However, I can test the `ClassUtil#invokeSetter` but not sure if that is sufficient.

vladak:
Testing `ClassUtil#invokeSetter` sounds good enough to me, at least coverage will not be lost.

PR #2221, opened 2018-07-19
add elapsed statistics logs to suggester rebuild

vladak:
Another piece that reports how long does suggester rebuild take. Had to get the methods from the indexer as suggester is independent.


suggester/src/main/java/org/opengrok/suggest/util/StringUtils.java
```diff
@@ -0,0 +1,74 @@
...
+
+package org.opengrok.suggest.util;
+
+/**
+ * Various String utility methods taken from
+ * org.opensolaris.opengrok.util.StringUtils
+ *
+ * @author austvik
+ */
+public final class StringUtils {
```
Orviss: I think we should not duplicate utility methods. I think it would be better to create some separate module which would contain that stuff.

However, there is no need for this method, you can simply use `DurationFormatUtils` (<https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/DurationFormatUtils.html>) which are already on the classpath.

suggester/src/main/java/org/opengrok/suggest/util/Statistics.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.opengrok.suggest.util;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class Statistics {
+
+  private final long startTime;
```
Orviss: Bad indentation, only 2 spaces.

suggester/src/main/java/org/opengrok/suggest/util/Statistics.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public Statistics() {
+      startTime = System.currentTimeMillis();
+  }
+
+  public void report(Logger log, String msg) {
```
Orviss: It is not a good practice to pass loggers as arguments.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -200,11 +202,11 @@ private boolean indexExists(final Path indexDir) throws IOException {
         }
     }
 
-    private void shutdownAndAwaitTermination(final ExecutorService executorService, final String logMessageOnSuccess) {
+    private void shutdownAndAwaitTermination(final ExecutorService executorService, Statistics elapsed, final String logMessageOnSuccess) {
```
Orviss: Maybe simply pass `Instant` as an argument which specifies the start time?

ahornace:
@vladak do you plan to continue working on this? It's been over a year.

suggester/src/main/java/org/opengrok/suggest/util/Statistics.java
vladak: I know however in this case it makes sense - otherwise the log message will have a prefix of the `Statistics` class, not the one of the class using it.

vladak:
In the end I got rid of both previously added classes. Please take a look again.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -64,7 +65,7 @@
 
     private static final String PROJECTS_DISABLED_KEY = "";
 
-    private static final Logger logger = Logger.getLogger(Suggester.class.getName());
+    private static final Logger LOGGER = Logger.getLogger(Suggester.class.getName());
```
ahornace: We in our internal project always use `logger` so I googled a bit: https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

However, I know that other modules use upper case so it's up to you to decide what to do.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -140,23 +141,24 @@ public Suggester(
...
+            LOGGER.log(Level.INFO, "No index directories found, exiting...");
             return;
         }
         if (!projectsEnabled && luceneIndexes.size() > 1) {
             throw new IllegalArgumentException("Projects are not enabled and multiple Lucene indexes were passed");
         }
 
         synchronized (lock) {
-            logger.log(Level.INFO, "Initializing suggester");
+            long startTime = System.currentTimeMillis();
```
ahornace: Why not https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html ?

suggester/src/main/java/org/opengrok/suggest/Suggester.java
vladak: okay, seems like only constants should be upper case.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
vladak: seems too much trouble to convert this to `long` and back.

ahornace: Really? I think `Duration.between(start, Instant.now()).toMillis()` is far more descriptive.

vladak: all the wrapping of `System` seemed too long winded to me, in reality it's not that bad.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -209,11 +211,13 @@ private boolean indexExists(final Path indexDir) throws IOException {
         }
     }
 
-    private void shutdownAndAwaitTermination(final ExecutorService executorService, final String logMessageOnSuccess) {
+    private void shutdownAndAwaitTermination(final ExecutorService executorService, Instant start, final String logMessageOnSuccess) {
         executorService.shutdown();
         try {
             executorService.awaitTermination(awaitTerminationTime.toMillis(), TimeUnit.MILLISECONDS);
-            logger.log(Level.INFO, logMessageOnSuccess);
+            logger.log(Level.INFO, logMessageOnSuccess + " (took {0})",
```
ahornace: Maybe rewrite to avoid another concatenation?
```
logger.log(Level.INFO, "{0} (took {1})", new Object[] {logMessageOnSuccess, DurationFormatUtils...)
```

PR #2225, opened 2018-07-24
Fix checkstyle for suggester module

### LINKED ISSUE #2223: compile in suggester fails due to missing checkstyle/fileheader.txt
When `mvn compile` is run in the `suggester` directory, it fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (checkstyle) on project suggester: Failed during checkstyle configuration: cannot initialize module Header - illegal value 'checkstyle/fileheader.txt' for property 'headerFile' of module Header: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to find: checkstyle/fileheader.txt -> [Help 1]
```

Orviss:
Hello,

fixes #2223

I was able to resolve this without additional Maven module. Seems to work properly on my setup. Could you please verify? :) 

Thanks :)

### DIFF
--- a/checkstyle/style.xml
+++ b/checkstyle/style.xml
@@ -5,7 +5,7 @@
 <module name="Checker">
     <module name="NewlineAtEndOfFile"/>
     <module name="Header">
-        <property name="headerFile" value="checkstyle/fileheader.txt"/>
+        <property name="headerFile" value="${checkstyle.header.file}"/>
     </module>
     <module name="FileLength"/>
     <module name="FileTabCharacter"/>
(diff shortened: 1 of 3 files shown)


vladak:
So the way it works with this change is that checkstyle is run only for the suggester module ?

Orviss:
Yes.

vladak:
Ok, pls create a new issue so that the indexer gets fixed w.r.t. checkstyle.

PR #2235, opened 2018-07-26
Fixes #2231 History not loading Mercurial/Windows

### LINKED ISSUE #2231: History not loading Mercurial/Windows
Expected behaviour: When viewing a source file from a Mercurial project in OpenGrok web the "History" and "Annotate" items on the top menu should be linked to file history and annotate functions, and show Mercurial file history.

Actual behaviour: History and Annotate items are unlinked.

Windows 10, OpenGrok built from source with tag 1.1-rc34, running under Tomcat 8 under JDK 8 64-bit for Windows (1.8.0_172).
Mercurial version is 3.7.3.

Code directory is c:\work\myproject, which is also the Mercurial root directory (i.e. it contains the Mercurial .hg directory).

Built OpenGrok index with:
 c:\jdk8\bin\java -jar opengrok.jar -H -W configuration.xml -c ctags\ctags58\ctags.exe -P -S -v -s c:\work\myproject -d grokdata  

This successfully builds the Mercurial history cache in grokdata\historycache, with Mercurial history file for each controlled source file.

Relevant settings in configuration.xml:
```  
<void property="repositories">
     <void method="add">
    <object class="org.opensolaris.opengrok.history.RepositoryInfo">
     <void property="branch">
      <string>some-branch</string>
     </void>
     <void property="currentVersion">
      <string>2018-07-18 16:37 +1200 35f3e5b06b29 commit-comment</string>
     </void>
     <void property="datePatterns">
      <array class="java.lang.String" length="1">
       <void index="0">
        <string>yyyy-MM-dd hh:mm ZZZZ</string>
       </void>
      </array>
     </void>
     <void property="directoryNameRelative">
      <string></string>
     </void>
     <void property="parent">
      <string>http://server/path/myproject</string>
     </void>
     <void property="type">
      <string>Mercurial</string>
     </void>
    </object>
   </void>
  </void>
  <void property="sourceRoot">
   <string>C:\work\myproject</string>
  </void>
  <void property="verbose">
   <boolean>true</boolean>
  </void>
```
Also tried directoryNameRelative as ```<string>\</string>```, no difference.

Cause: in HistoryGuru.getRepository(), inRootPath that should match is "c:\work\myproject"
but repositories.get(inRootPath) returns null, because the repositories key is c:\work\myproject\

Fix that works for me:
```diff --git a/src/org/opensolaris/opengrok/history/RepositoryInfo.java b/src/org/opensolaris/opengrok/history/RepositoryInfo.java
index 81d72fd5..dd8058a6 100644
--- a/src/org/opensolaris/opengrok/history/RepositoryInfo.java
+++ b/src/org/opensolaris/opengrok/history/RepositoryInfo.java
@@ -26,6 +26,7 @@ package org.opensolaris.opengrok.history;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.Objects;
 import java.util.logging.Level;
@@ -139,8 +140,9 @@ public class RepositoryInfo implements Serializable {
      * @return the name of the root directory
      */
     public String getDirectoryName() {
-        return RuntimeEnvironment.getInstance().getSourceRootPath() +
-                directoryNameRelative;
+        return Paths.get(RuntimeEnvironment.getInstance().getSourceRootPath(),
+                directoryNameRelative).toString();
+
     }
 
     /**

### PR DESCRIPTION
**Pull request by anotherdave**
See issue #2231 for description of fix. I accept the terms of the OCA and have sent in an OCA form.

### DIFF
--- a/src/org/opensolaris/opengrok/history/RepositoryInfo.java
+++ b/src/org/opensolaris/opengrok/history/RepositoryInfo.java
@@ -26,6 +26,7 @@ package org.opensolaris.opengrok.history;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.Objects;
 import java.util.logging.Level;
@@ -139,8 +140,9 @@ public class RepositoryInfo implements Serializable {
      * @return the name of the root directory
      */
     public String getDirectoryName() {
-        return RuntimeEnvironment.getInstance().getSourceRootPath() +
-                directoryNameRelative;
+        return Paths.get(RuntimeEnvironment.getInstance().getSourceRootPath(),
+                directoryNameRelative).toString();
+
     }
 
     /**

### CODE REVIEW

**Comment by Orviss** [comment]
Some tests are broken. Could you please take a look and fix them? The PR can't be merged until it is resolved.

**Comment by vladak** [comment]
The BazaarHistoryParserTest#setup() failed on line 60 due to null pointer:

```java
55      @Before
56      public void setUp() {
57          if (RuntimeEnvironment.getInstance().getSourceRootPath() == null) {
58              RuntimeEnvironment.getInstance().setSourceRoot("");
59          }
60          instance = new BazaarHistoryParser(new BazaarRepository());
61      }
```

so most likely `BazaarRepository()` ended up as `null`.

vladak:
Actually, the problem is more likely in:

```java
57      BazaarHistoryParser(BazaarRepository repository) {
58          this.repository = repository;
59          myDir = repository.getDirectoryName() + File.separator;
60      }
```

vladak:
Also, it is surprising that `BazaarHistoryParserTest` does not use the `@ConditionalRun(RepositoryInstalled.BazaarInstalled.class)`

vladak:
The annotation is not necessary, the test uses stored output.

vladak:
The problem is in the test itself. Source root is not set and is empty string and somehow becomes `null` along the way which makes `Paths.get()` to fail.

vladak:
I will fix the test in `master` branch and then merge this by hand.

vladak:
fixed via 72cbb868f92076386cc1832ca18b504dceffe585. Thanks !

PR #2237, opened 2018-07-28
Add possibility to disable suggester by a cookie

Orviss:
Hello,

some users might not like the suggester functionality. Therefore, there should be an option to disable it for specific users. (Not only for the whole OpenGrok instance).

I thought that cookie was the easiest solution.

How to disable:
* Open a web console on OpenGrok webpage
* Type: `document.cookie = 'suggester.enabled=false;' + document.cookie`
* Refresh webpage
* Suggester should be disabled.

Thanks :)


vladak:
Please document the procedure for disabling on the suggester wiki page after this is merged.

Orviss:
> Please document the procedure for disabling on the suggester wiki page after this is merged.

Sure :)

vladak:
The cookie should probably be named with OpenGrok specific prefix.

The instructions should ideally use some animated gif for users who do not know how to type JavaScript code into web console.

Orviss:
> Or maybe there are browser plugins to manipulate the cookies directly ?

There are some, e.g. <https://chrome.google.com/webstore/detail/editthiscookie/fngmhnnpilhplaeedifhccceomclgfbg> or <https://addons.mozilla.org/en-US/firefox/addon/edit-cookie/>

However, I think that the plugins will have different use across the various web browsers whereas the web console is almost the same. And most of the users are probably programmers so it should not be that hard.

> The instructions should ideally use some animated gif (https://blog.github.com/2018-06-29-GIF-that-keeps-on-GIFing/) for users who do not know how to type JavaScript code into web console

GIF is a must! I will create one :)

tulinkry:
What about some checkbox setting the value. On dedicated settings page or something like that.

Orviss:
> What about some checkbox setting the value. On dedicated settings page or something like that.

That's what I wanted to avoid.

tulinkry:
> That's what I wanted to avoid.

Why?

vladak:
I think per user settings (possibly with its own page) would be nice however we are not there yet. So, I'd file a new issue to provide basis for per-user settings and solve this one ad-hoc.

Orviss:
>  Why?

Because I think it would launch a lot of requests for specific modifications per user. I'm afraid that after a while we would have user-specific configuration.

However, if you prefer that I can do it :)

tulinkry:
> I think per user settings (possibly with its own page) would be nice however we are not there yet. So, I'd file a new issue to provide basis for per-user settings and solve this one ad-hoc.

Ok.

> Because I think it would launch a lot of requests for specific modifications per user. I'm afraid that after a while we would have user-specific configuration.
> However, if you prefer that I can do it :)

No, I was just asking out of curiosity. I think this modifications per user are the future. But not a really close one so this solution is fine.

Orviss:
I can't make the GIF :/

There are several problems:
* I don't know how to make those animations
* If I do screen record and then transform it to GIF it is too large. It has 75MB, if I lower the quality (640x355) then it has 15MB but the resolution is pretty bad and it's hard to recognize the letters...

vladak:
Ok, don't worry about the GIF then.

Incidentally, I created https://github.com/oracle/opengrok/wiki/Cookies , pls update it once this merges.

opengrok-web/src/main/webapp/js/utils-0.0.24.js
```diff
@@ -1716,6 +1716,11 @@ function pageReadyMast() {
 }
 
 function domReadyMenu(minisearch) {
+    if (getCookie('opengrok.suggester.enabled') === 'false') {
```
vladak: maybe this should follow the common naming "scheme" used for pre-existing cookies ? i.e. `OpenGrokSuggester` ?

Orviss: I just used what came first to my mind. However, I found only one cookie in the whole project which has name `OpenGrokProject`.  Maybe there are some others?

For me `OpenGrokSuggester` is harder to read than `opengrok.suggester` and I think it would be worse if we use other modules (e.g. `OpenGrokIndexer`, ...).

But I agree that we should have a common naming. So `OpenGrokSuggester.enabled` it is? Or do you propose something different? :)

vladak: There is another one, see the wiki. `OpenGrokSuggester.enabled` is good enough I guess.

PR #2239, opened 2018-07-28
Enable checkstyle for Indexer module

### LINKED ISSUE #2226: Enable checkstyle for indexer module
OpenGrok has a running checkstyle for suggester module. It should be enabled for indexer module as well and the checkstyle errors should be repaired.

Orviss:
fixes #2226 

Added suppressions for `MethodLength` and `ParameterNumber` because it would require non-trivial refactoring and I think it would be better to do it in a separate patch.

Thanks :)


vladak:
Great to have style enforcement finally. Did you use any automation to make the code compliant ?

Orviss:
Nope, I did it by hand.

vladak:
Good job, thanks !

PR #2245, opened 2018-08-01
tools should have its own Maven module

vladak:
For now this is only for the tests to run separately. In the future this could be used to properly deliver tools and have them consumed properly by the `distribution` module.


Orviss:
Just a suggestion: could we move `.py` files to `src/main/python` directory? I found that Apache Spark does it similarly and it should make the structure more consistent (<https://github.com/apache/spark/tree/master/examples/src/main/python>). 

However, I have never used Maven with other languages so maybe it should be done differently.

vladak:
What advantage this will have w.r.t. Maven ?

Orviss:
> What advantage this will have w.r.t. Maven ?

I'm not sure. I don't know if you need some Maven plugins or something else to make it work properly.

However, for me, it would look nice that the source code would have the same directory structure. Imagine that tools module were to contain some Java code now. Then it would be separated nicely into `src/main/java` and `src/main/python`.

vladak:
Makes sense, also for the tests.

vladak:
Reorganized the structure, had to set the directories in the pom file explicitly. I am not sure how it would work if I added Java/... code there.

tulinkry:
I think eventualy it should be a python package and the distribution version of this package we should ship with each opengrok release.

tools/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+
+    <artifactId>tools</artifactId>
+    <version>1.1-rc35</version>
+    <packaging>pom</packaging>
+
+    <name>OpenGrok tools</name>
+
+    <build>
+        <sourceDirectory>src/main/python</sourceDirectory>
+	<testSourceDirectory>src/test/python</testSourceDirectory>
```
Orviss: Tab character.

PR #2032, opened 2018-03-05
Improve project structure

Orviss:
Features:
* removed IDE specific files (nbproject, .classpath, etc.)
* moved files to correct maven structure (src/main/java)
* moved resource files to resources directory 

There are a few things that I would like to add -> mainly checkstyle plugin and errors on compilation warnings but I think it would be wiser to include them in later PRs.

Thanks for review :)

PR #2250, opened 2018-08-01
enable Sonarqube checker

vladak:
Enables SonarQube checker.

Here are my notes esp. on blockers/constrains:
  - the Sonar plugin is invoked directly in Maven using `sonar:sonar` - after jacoco agent is prepared (which it is by default in our Maven plugin configuration). This direct invocation will download the sonar plugin and execute its goal. Personally I'd feel more comfortable if the sonar plugin was listed in the top-level `pom.xml` file.
  - in order to submit the results it is necessary to run the analysis on the main branch first (usually the main branch is called `master` otherwise further configuration of Sonar will be required - https://groups.google.com/forum/#!msg/sonarqube/obMtswvz0_U/q_2najAnBAAJ). This is probably done so that it has a baseline. For this reason I had to test my `vladak/OpenGrok:sonarqube` branch by pulling the changesets to `vladak/OpenGrok:master` and do the `sonar:sonar` build there first.
  - it is not possible to perform the analysis on external pull requests (because of security reasons as per https://docs.travis-ci.com/user/sonarcloud/#Analysis-of-internal-pull-requests)
  - it can be made to work on the main OpenGrok source tree only after #2032 is integrated, otherwise it will analyse just the `jrcs` and `plugins` projects since it expects to find source code under module directory (in the build it complains with `[WARNING] File '/home/travis/build/vladak/OpenGrok/src/org/opensolaris/opengrok/authorization/AuthControlFlag.java' is ignored. It is not located in module basedir '/home/travis/build/vladak/OpenGrok/opengrok-indexer'.` etc. etc. - for the whole source tree of the `opengrok-indexer` Maven module)
  - the authentication token is already setup on sonarcloud.io and encrypted with my Travis key. Since I am the admin of the `opengrok` organization on sonarcloud.io I guess this will make it to allow the results upload.
  - this checks all Maven modules, and therefore the `jrcs` module as well. We are not so much interested in the module given it is third party module so it creates a bit of a noise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: java
 cache:
   directories:
-  - $HOME/.m2
+  - '$HOME/.m2'
+  - '$HOME/.sonar/cache'
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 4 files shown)


Orviss:
Still, why does Travis fail?

vladak:
Evidently when executing the `sonar:sonar` target, it tries to connect to the standalone server:

```
[ERROR] SonarQube server [http://localhost:9000] can not be reached
```

vladak:
I will try merging this in and we will see what happens.

vladak:
Re-encrypting the Sonarcloud token helped to solve the 'Not authorized' issue. The error message should be more descriptive however Sonar github repositories does not allow to file new Issues.

PR #2256, opened 2018-08-03
Sonar plugin fixes

vladak:
bunch of Sonar bug fixes for plugins

### DIFF
--- a/plugins/src/opengrok/auth/entity/LdapUser.java
+++ b/plugins/src/opengrok/auth/entity/LdapUser.java
@@ -39,6 +39,8 @@ public class LdapUser implements Serializable {
     private Set<String> ou;
     private final Map<String, Set<String>> attrs = new HashMap<>();
 
+    private static final long serialVersionUID = 1L;
+
     public LdapUser(String mail, String uid, Set<String> ou) {
         this.mail = mail;
         this.uid = uid;
--- a/plugins/src/opengrok/auth/entity/LdapUser.java
+++ b/plugins/src/opengrok/auth/entity/LdapUser.java
@@ -22,6 +22,7 @@
  */
 package opengrok.auth.entity;
 
+import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -31,7 +32,7 @@ import java.util.TreeSet;
  *
  * @author Krystof Tulinger
  */
-public class LdapUser {
+public class LdapUser implements Serializable {
 
     private String mail;
     private String uid;
--- a/plugins/src/opengrok/auth/entity/LdapUser.java
+++ b/plugins/src/opengrok/auth/entity/LdapUser.java
@@ -39,7 +39,9 @@ public class LdapUser implements Serializable {
     private Set<String> ou;
     private final Map<String, Set<String>> attrs = new HashMap<>();
 
-    private static final long serialVersionUID = 1L;
+    // Use default serial ID value. If the serialized form of the object
+    // changes, feel free to start from 1L.
+    private static final long serialVersionUID = -8207597677599370334L;
 
     public LdapUser(String mail, String uid, Set<String> ou) {
         this.mail = mail;
--- a/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -45,14 +45,14 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
     protected static final String ATTR_PARAM = "attribute";
     protected static final String FILE_PARAM = "file";
 
-    private final static String SESSION_ALLOWED_PREFIX = "opengrok-attr-plugin-allowed";
-    private String SESSION_ALLOWED = SESSION_ALLOWED_PREFIX;
+    private static final String SESSION_ALLOWED_PREFIX = "opengrok-attr-plugin-allowed";
+    private String sessionAllowed = SESSION_ALLOWED_PREFIX;
 
     private String ldapAttr;
     private final Set<String> whitelist = new TreeSet<>();
 
     public LdapAttrPlugin() {
-        SESSION_ALLOWED += "-" + nextId++;
+        sessionAllowed += "-" + nextId++;
     }
 
     @Override
@@ -78,7 +78,7 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
     @Override
     protected boolean sessionExists(HttpServletRequest req) {
         return super.sessionExists(req)
-                && req.getSession().getAttribute(SESSION_ALLOWED) != null;
+                && req.getSession().getAttribute(sessionAllowed) != null;
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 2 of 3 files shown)


plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -45,7 +45,8 @@
     protected static final String ATTR_PARAM = "attribute";
     protected static final String FILE_PARAM = "file";
 
-    public String SESSION_ALLOWED = "opengrok-attr-plugin-allowed";
+    private final static String SESSION_ALLOWED_PREFIX = "opengrok-attr-plugin-allowed";
```
Orviss: Correct order is `private static final ...`.

plugins/src/opengrok/auth/plugin/LdapFilterPlugin.java
```diff
@@ -44,7 +44,8 @@
     private static final Logger LOGGER = Logger.getLogger(LdapFilterPlugin.class.getName());
 
     protected static final String FILTER_PARAM = "filter";
-    protected String SESSION_ALLOWED = "opengrok-filter-plugin-allowed";
+    private static final String SESSION_ALLOWED_PREFIX = "opengrok-filter-plugin-allowed";
```
Orviss: Correct order is `private static final ...`

plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -45,7 +45,8 @@
...
+    private String SESSION_ALLOWED = SESSION_ALLOWED_PREFIX;
```
Orviss: `SESSION_ALLOWED` should be named `sessionAllowed`.

plugins/src/opengrok/auth/plugin/LdapFilterPlugin.java
```diff
@@ -44,7 +44,8 @@
...
+    private String SESSION_ALLOWED = SESSION_ALLOWED_PREFIX;
```
Orviss: `SESSION_ALLOWED` should be named `sessionAllowed`.

plugins/src/opengrok/auth/entity/LdapUser.java
```diff
@@ -31,13 +32,15 @@
  *
  * @author Krystof Tulinger
  */
-public class LdapUser {
+public class LdapUser implements Serializable {
```
Orviss: Could you please elaborate why use `Serializable`?

vladak: Sonar says:

> If you have no intention of writting an HttpSession object to file, then storing non-serializable objects in it may not seem like a big deal. But whether or not you explicitly serialize the session, it may be written to disk anyway, as the server manages its memory use in a process called "passivation". Further, some servers automatically write their active sessions out to file at shutdown & deserialize any such sessions at startup.

> The point is, that even though HttpSession does not extend Serializable, you must nonetheless assume that it will be serialized, and understand that if you've stored non-serializable objects in the session, errors will result.

plugins/src/opengrok/auth/plugin/LdapFilterPlugin.java
vladak: ok

plugins/src/opengrok/auth/entity/LdapUser.java
vladak: This happens in `updateSession()`:
```
req.getSession().setAttribute(SESSION_ATTR, user);
```

Orviss: Okay. I am wondering, wouldn't be better to set `serialVersionUID` to the default value instead of `1`? In that case if anyone tries to deserialize it it won't break.

vladak: what do you mean by default value ?

vladak: https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it quotes https://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html that says

> If a serializable class does not explicitly declare a serialVersionUID, then the serialization runtime will calculate a default serialVersionUID value for that class based on various aspects of the class, as described in the Java(TM) Object Serialization Specification. However, it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization. Therefore, to guarantee a consistent serialVersionUID value across different java compiler implementations, a serializable class must declare an explicit serialVersionUID value.

vladak: http://www.javapractices.com/topic/TopicAction.do?Id=45 has these guidelines for the ID:


-   always include it as a field, for example:
```
    private static final long serialVersionUID = 7526472295622776147L;
```
-   include this field even in the first version of the class, as a reminder of its importance
-   don't change the value of this field in future versions, unless you are knowingly making changes to the class which will render it incompatible with old serialized objects
-   new versions of Serializable classes may or may not be able to read old serialized objects; it depends upon the nature of the change; provide a pointer to Oracle's guidelines for what constitutes a compatible change, as a convenience to future maintainers

Orviss: Exactly, you did not change any fields; therefore, the class is still backwards compatible. Thus, I would use for `serialVersionUID` the value that is used by default if no `serialVersionUID` is provided. (JVM computes this value based on some class properties)  For instance, IDEA uses this value when I add missing `serialVersionUID` to the `Serializable` class.

vladak: I'd rather heed the recommendation from official Java SE docs and include the ID explicitly.

Orviss: I'm not saying you should not use `serialVersionUID`. I'm saying that you should use value `X` instead of `1` where `X` is the value that would be computed by a JVM.

Orviss: Imagine you have a `Serializable` class with no `serialVersionUID`. JVM computes some value for `serialVersionUID` and stores it in the serialized form. Then you add `serialVersionUID` without making any changes to the class. That means, the class could still be deserialized without any problems. However, if you now specify `serialVersionUID` to be `1`, then there will be clash and the deserialization won't work (the serialVersionUIDs don't match). Therefore, in these cases I think it is better to use the JVM computed value. In this case it is `-8207597677599370334L`.

Orviss: IDEA computes it for me if there is a `Serializable` class without `serialVersionUID`. I found an algorithm here (it is for Java 6 but it probably did not change much):
<https://docs.oracle.com/javase/6/docs/platform/serialization/spec/class.html#4100>

Now I noticed that it also takes interface names into account so only by adding `Serializable` it could have changed (not sure if `Serializable` is considered though since every class would have it :D ). It still depends on how the web application server does the serialization...

PR #2259, opened 2018-08-05
Enable checkstyle for web module

ahornace:
Hello,

I noticed that web module does not have checkstyle plugin.

Thanks :)

### DIFF
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -186,6 +186,11 @@ Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
                     </argLine>
                 </configuration>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


Orviss:
I noticed there is a problem. When I run `mvn clean compile` it fails with the following:
```bash
[ERROR] Failed to execute goal on project opengrok-web: Could not resolve dependencies for project org.opengrok:opengrok-web:war:1.1-rc35: Failure to find org.opengrok:opengrok:jar:tests:1.1-rc35 in http://repo.maven.apache.org/maven2/ was cached in the local repository, resolution will not be reattempted until the update interval of maven.org has elapsed or updates are forced
```

I don't see a reason why this happens. The `test` scope should be ignored when I just run `compile`. 

Works properly when `checkstyle` plugin is not specified...

vladak:
Similar failures have been hunting me for a long time (at least before your `mavenize` branch got merged in). Not sure why Maven starts searching in the remote repository.

vladak:
I decided to merge this in since I no longer run Maven build per partes :-)

PR #2262, opened 2018-08-06
make jrcs external dependency

vladak:
Replaces the `jrcs` module with dependencies. The annotation is performed by parsing the output from the `blame` command.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/RCSRepository.java
```diff
@@ -79,37 +86,23 @@ boolean fileHasAnnotation(File file) {
 
     @Override
     Annotation annotate(File file, String revision) throws IOException {
-        File rcsFile = getRCSFile(file);
-        return rcsFile == null ? null : annotate(file, revision, rcsFile);
-    }
+        ArrayList<String> argv = new ArrayList<>();
```
Orviss: Maybe use `List` interface?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RCSRepository.java
```diff
@@ -47,6 +44,16 @@
 
     private static final long serialVersionUID = 1L;
 
+    /**
+     * the property name used to obtain the command to get annotation for this repository
```
Orviss: Javadoc should have sentences - first sentence will be brief info. So I would start with upper case letter and end with `.`. CheckStyle has some checks for this. We can enable them later if the need arises.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RazorRepository.java
```diff
@@ -256,10 +256,6 @@ InputStream getHistoryGet(String parent, String basename, String rev) {
     Annotation annotate(File file, String revision)
             throws IOException {
         // @TODO : Rename & Delete Support
-        File rcsFile = getRazorArchiveRCSFileFor(file);
-        if (rcsFile != null && rcsFile.exists()) {
-            return RCSRepository.annotate(file, revision, rcsFile);
-        }
```
tulinkry: why is this no longer needed?

vladak: It should work as long as Razor automatically creates the `RCS/XYZ.txt,v` file.

PR #2264, opened 2018-08-07
properly create temporary directory

vladak:
fixes problem (race condition) reported by Sonar


opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
```diff
@@ -30,6 +30,8 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
```
Orviss: Is `Path` import needed?

PR #2265, opened 2018-08-07
Opt parser private

vladak:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -104,7 +104,11 @@ public final class Indexer {
     private static RuntimeEnvironment env = null;
     private static String host = null;
 
-    public static OptionParser openGrok = null;
+    public static OptionParser getOptParser() {
+        return optParser;
+    }
+
+    private static OptionParser optParser = null;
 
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -1147,4 +1147,8 @@ private static void pauseToAwaitProfiler() {
 
     private Indexer() {
     }
+
+    public static OptionParser getOptParser() {
```
Orviss: I wouldn't add getter just because test needs it. If test needs the access, then I would suggest using reflection.

tulinkry: +1

vladak: Any hints on how to do it ? I tried the method described in https://stackoverflow.com/questions/1196192/how-to-read-the-value-of-a-private-field-from-a-different-class-in-java however Indexer is static, its constructor is private and getClass() cannot be called from static context (at least IDEA says so).

Orviss: You do not have to use `getClass()` method. You can use class field -> `Indexer.class`.

PR #2267, opened 2018-08-07
parent member should not shadow parent class field

vladak:


opengrok-indexer/src/main/java/org/opengrok/indexer/history/AccuRevRepository.java
```diff
@@ -90,11 +90,11 @@
     private static final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
 
     private String depotName = null;
-    private String parent = null;
+    private String parentInfo = null;
```
Orviss: Why not use `parent` attribute from the `Repository`? It has `protected` access. Or does this attribute represent something different?

PR #2270, opened 2018-08-07
Sonar fixes for SCCS

vladak:


opengrok-indexer/src/main/java/org/opengrok/indexer/history/SCCSRepository.java
```diff
@@ -187,17 +190,17 @@ History getHistory(File file) throws HistoryException {
 
     @Override
     String determineParent(boolean interactive) throws IOException {
-        File parentFile = new File(getDirectoryName() + File.separator
-                + "Codemgr_wsdata" + File.separator + "parent");
+        File parentFile = Paths.get(getDirectoryName(),CODEMGR_WSDATA, "parent").toFile();
```
Orviss: Missing space before `CODEMGR_WSDATA`.

vladak: That's the result of IDEA annotating the arguments with names, it makes it harder to see the spaces.

Issue #2240, opened 2018-07-30
growing number of opengrok directories in Tomcat temporary area


vladak:
In the `/var/tomcat8/temp/` directory I see lots of empty directories with `opengrok` prefix, e.g.:

```
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok408091792513120463
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok5126593944445059938
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok1504167965668909705
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok6408331709879366844
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok8730009372603513500
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok3899762656954060544
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok2633702641013476723
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok5410711030578992205
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok4951472061130336907
drwx------   2 webservd webservd       2 Jul 30 08:05 opengrok4621036241443139481
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok9076596853661318875
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok8389696774557015597
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok688709581457593757
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok5227950753932224750
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok7727174665308681446
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok725675884455545561
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok7295946628718278133
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok5832858987571345938
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok3700590860536061517
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok1010355954905738740
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok8881200868190539397
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok4392975536131186015
drwx------   2 webservd webservd       2 Jul 30 08:06 opengrok2625372493496774060
```

Looks like multiple directories are created per minute.

[...]

PR #2280, opened 2018-08-09
delete temporary file on exception

vladak:
The easier part of #2240 to get rid of stale temporary files.

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
+++ b/suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
@@ -120,24 +120,29 @@ public class ChronicleMapAdapter implements PopularityMap {
 
         Path tempFile = Files.createTempFile("opengrok", "chronicle");
 
-        map.getAll(tempFile.toFile());
+        ChronicleMap<BytesRef, Integer> m;
 
-        String field = map.name();
+        try {
+            map.getAll(tempFile.toFile());
 
-        map.close();
+            String field = map.name();
 
-        Files.delete(f.toPath());
+            map.close();
 
-        ChronicleMap<BytesRef, Integer> m = ChronicleMap.of(BytesRef.class, Integer.class)
-                .name(field)
-                .averageKeySize(newMapAvgKey)
-                .entries(newMapSize)
-                .keyReaderAndDataAccess(BytesRefSizedReader.INSTANCE, new BytesRefDataAccess())
-                .createOrRecoverPersistedTo(f);
+            Files.delete(f.toPath());
 
-        m.putAll(tempFile.toFile());
-
-        Files.delete(tempFile);
+            m = ChronicleMap.of(BytesRef.class, Integer.class)
+                    .name(field)
+                    .averageKeySize(newMapAvgKey)
+                    .entries(newMapSize)
+                    .keyReaderAndDataAccess(BytesRefSizedReader.INSTANCE, new BytesRefDataAccess())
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
```diff
@@ -120,24 +120,29 @@ public void resize(final int newMapSize, final double newMapAvgKey) throws IOExc
...
+                    .keyReaderAndDataAccess(BytesRefSizedReader.INSTANCE, new BytesRefDataAccess())
+                    .createOrRecoverPersistedTo(f);
+            m.putAll(tempFile.toFile());
+        } catch (IOException e) {
+            throw e;
+        } finally {
+            Files.delete(tempFile);
+        }
 
         map = m;
```
Orviss: I would move this assignment inside the `try` block. It makes it clear that this will be only done if no exception is thrown.

suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
```diff
@@ -120,24 +120,29 @@ public void resize(final int newMapSize, final double newMapAvgKey) throws IOExc
 
         Path tempFile = Files.createTempFile("opengrok", "chronicle");
 
-        map.getAll(tempFile.toFile());
+        ChronicleMap<BytesRef, Integer> m;
```
Orviss: If you move assignment below inside the `try` block, then there will be no need for this declaration.

suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
```diff
@@ -120,24 +120,29 @@ public void resize(final int newMapSize, final double newMapAvgKey) throws IOExc
...
-
-        Files.delete(tempFile);
+            m = ChronicleMap.of(BytesRef.class, Integer.class)
+                    .name(field)
+                    .averageKeySize(newMapAvgKey)
+                    .entries(newMapSize)
+                    .keyReaderAndDataAccess(BytesRefSizedReader.INSTANCE, new BytesRefDataAccess())
+                    .createOrRecoverPersistedTo(f);
+            m.putAll(tempFile.toFile());
+        } catch (IOException e) {
```
Orviss: I think there is no need for this catch block. Or is it?

suggester/src/main/java/org/opengrok/suggest/popular/impl/chronicle/ChronicleMapAdapter.java
```diff
@@ -120,24 +120,29 @@ public void resize(final int newMapSize, final double newMapAvgKey) throws IOExc
...
-        Files.delete(f.toPath());
+            map.close();
 
-        ChronicleMap<BytesRef, Integer> m = ChronicleMap.of(BytesRef.class, Integer.class)
-                .name(field)
-                .averageKeySize(newMapAvgKey)
-                .entries(newMapSize)
-                .keyReaderAndDataAccess(BytesRefSizedReader.INSTANCE, new BytesRefDataAccess())
-                .createOrRecoverPersistedTo(f);
+            Files.delete(f.toPath());
```
tulinkry: Rename please `f` to something more descriptive as it is a member variable.

tulinkry: Also any chance this isn't executed when there's an exception in something above?

PR #2282, opened 2018-08-09
handle JSON decode error for older json module versions

vladak:
The tools scripts are written so they only work on Python 3. However, in Python 3.4 the `ValueError` is thrown instead of `json.decoder.JSONDecodeError` so this needs special handling.

### DIFF
--- a/tools/src/main/python/readconfig.py
+++ b/tools/src/main/python/readconfig.py
@@ -26,6 +26,7 @@ import json
 import yaml
 import sys
 
+# The following is to make the json parsing work on Python 3.4.
 try:
     from json.decoder import JSONDecodeError
 except ImportError:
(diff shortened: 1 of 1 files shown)


tools/src/main/python/readconfig.py
```diff
@@ -26,6 +26,11 @@
 import yaml
 import sys
 
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
```
tulinkry: I'd just add a note that this is because python 3.4 for anybody in the future looking into this

vladak: Maybe Python should introduce `twenty` package that would bridge 3.4 -> 3.5 (just like there is the `six` package for transition from Python 2 to 3)...

PR #2285, opened 2018-08-09
Fix stale temporary directories created by suggester

Orviss:
Hello,

fixes problem of stale directories that were created by a suggester. It seems to work fine -> temporary directories are deleted upon stopping the web application. 

Thanks :)


vladak:
Does the `close()` actually get called only when the webapp is stopped ?

Orviss:
> Does the close() actually get called only when the webapp is stopped ?

Yes. The same temporary directory will be reused also for rebuilds.

I think that it is not a problem. There is no way that the temp directory would stay there forever. If the app crashes then that very probably means that Tomcat crashed too. So the Tomcat's temporary directory will be cleared.

vladak:
> Yes. The same temporary directory will be reused also for rebuilds.

I see. This is fine then.

PR #2294, opened 2018-08-13
enforce Universal ctags

vladak:


dev/install-universal_ctags.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+git clone https://github.com/universal-ctags/ctags.git
+cd ctags
+./autogen.sh
+./configure
+make
+make install
```
tulinkry: `./configure && make && make install` to catch the failures?

dev/before_install
```diff
@@ -12,3 +12,5 @@ elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
 	brew upgrade python
 	pip3 install pep8 flake8
 fi
+
+sudo ./dev/install-universal_ctags.sh
```
tulinkry: On travis it's enough to have it in path? No need to set the `SYSTEM_CTAGS_PROPERTY` for indexer?
I mean if the location of the installation does or doesn't matter here.

vladak: It seems that in Travis environment it is sufficient to have the ctags binary in `PATH`. The test run reports
```
[WARNING] Tests run: 727, Failures: 0, Errors: 0, Skipped: 6
```
which is in line what is reported for `master`:
```

```

PR #2295, opened 2018-08-14
Improve suggester temporary files handling

Orviss:
Hello,

this way there is no temporary directory created for the whole time the application is running. `WFSTCompletionLookup` creates a temporary file in the directory `java.io.tmpdir` and removes it when it is not needed anymore.

Thanks :)

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
+++ b/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
@@ -46,8 +46,8 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
```diff
@@ -66,11 +66,13 @@
  */
 class SuggesterProjectData implements Closeable {
 
+    private static final String TMP_DIR_PROPERTY = "java.io.tmpdir";
```
tulinkry: I don't have much background on this.

1. Isn't there some builtin constant?
2. What if this property is not set at all? Is it possible to happen?

vladak: I think leaving the temporary files in Tomcat temporary directory makes more sense. There is usually too much clutter already in system temp dir.

Orviss: > Isn't there some builtin constant?

I haven't found one. I think this is similar to `line.separator` pre Java 8.

> What if this property is not set at all? Is it possible to happen?

I don't think so. Even `Files` (via `TempFileHelper`) in Java are using this property. So if it would not be set then I suppose all hell would break loose.

> I think leaving the temporary files in Tomcat temporary directory makes more sense. There is usually too much clutter already in system temp dir.

This actually is the Tomcat's temporary dir because Tomcat sets this property to custom value.

PR #2297, opened 2018-08-15
fix the logic and report ctags version

vladak:
The logic for ctags checking is wrong. It causes tests to fail if there is no valid ctags binary.


vladak:
This will fix bunch of failures for #2302 .

PR #2298, opened 2018-08-15
add Lucene as dependency to the web module

vladak:
IDEA was unhappy because it could not resolve dependency to compile `SuggesterQueryBuilder.java` due to the imports:
```
import org.apache.lucene.queryparser.classic.ParseException;
import org.apache.lucene.search.Query;
```

This change makes it happy again.

### DIFF
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -101,6 +101,11 @@
             <artifactId>org.suigeneris.jrcs.diff</artifactId>
             <version>0.4.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queryparser</artifactId>
+            <version>7.3.1</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -104,7 +104,7 @@
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-queryparser</artifactId>
-            <version>7.3.1</version>
+            <version>${lucene.version}</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


opengrok-web/pom.xml
```diff
@@ -101,6 +101,11 @@
             <artifactId>org.suigeneris.jrcs.diff</artifactId>
             <version>0.4.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queryparser</artifactId>
+            <version>7.3.1</version>
```
Orviss: Please use Lucene version variable.

Orviss:
Is this really necessary? It works for me without problems.

Orviss:
If you look at `indexer` or `suggester`, both of these have this dependency. And since `web` depends on them, then it has it transitively. What exact error does IDEA show?

vladak:
It underlines the imports with red.

Orviss:
> It underlines the imports with red.

This does not happen to me. I will try clean checkout later this evening.

Orviss:
Clean checkout. Imported from external model (Maven). IDEA 2018.1 Ultimate.

Imports are not underlined.

screen shot 2018-08-16 at 20 05 49

vladak:
Maybe this has to do with my use case - I am opening each Maven project in separate IDEA window. Will re-check.

tarzanek:
seen the same here when java EE module with tomcat support wasn't enabled
resp. when I imported project BEFORE I enabled the tomcat support in Idea Ultimate
I suppose Idea CE will have the same prob ... 
(and I am wondering about Netbeans, maybe they work out of box)

ahornace:
`lucene` should be added transitively. Please reopen if needed.

Issue #2302, opened 2018-08-15
fix test failures on Windows


vladak:
The AppVeyor builds currently exhibit number of test case failures. These should be investigated and fixed.

Feel free to submit pull requests that fix just one problem.

vladak:
Fixed by a series of PRs/commits.

PR #2308, opened 2018-08-16
normalize newlines

vladak:
should help with #2302

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/haskell/HaskellXrefTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/haskell/HaskellXrefTest.java
@@ -122,8 +122,10 @@ public class HaskellXrefTest {
             expectedOutputSteam.close();
         }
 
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/haskell/HaskellXrefTest.java
```diff
@@ -122,8 +122,10 @@ public void sampleTest() throws IOException {
             expectedOutputSteam.close();
         }
 
-        String actual[] = new String(sampleOutputStream.toByteArray(), "UTF-8").split("\n");
-        String expected[] = new String(expectedOutputSteam.toByteArray(), "UTF-8").split("\n");
+        String actual[] = new String(sampleOutputStream.toByteArray(), "UTF-8").
```
Orviss: Why not use `split("\\r?\\n")` as suggested in <https://stackoverflow.com/questions/454908/split-java-string-by-new-line> ?

Issue #2307, opened 2018-08-16
Open grok doesn't support markdown tables (for README.md files for instance)


lupuleasa-core:
Hi wonderful people,

I have worked on the docs for my enterprise codebase on github and I noticed that opengrok doesn't support markdown tables, and I was wondering if it is possible to add support for them, if needed I could contribute to it.

[Table Markdown](https://help.github.com/articles/organizing-information-with-tables/) of the form:

    | First Header  | Second Header |
    | ------------- | ------------- |
    | Content Cell  | Content Cell  |
    | Content Cell  | Content Cell  |

## Screenshots of issue:

- **How it looks on github:**

![image](https://user-images.githubusercontent.com/5671120/44205736-0fdeeb80-a160-11e8-9154-ab34d442299e.png)

- **How it looks on our opengrok:**

![image](https://i.imgur.com/I9BkTq0.png)

&nbsp;

I searched previously for "markdown" and "README.md" and I haven't seen this mentioned.

Thanks,

Lup.

vladak:
Feel free to submit a pull request.

lupuleasa-core:
I have looked a bit into the code, and this is what I found so far:

- From the ["markdown" search](https://github.com/oracle/opengrok/search?q=markdown&unscoped_q=markdown), it seems there is a `function init_markdown_converter() ` that utilizes a script called showdown `$.script.loadScript('js/showdown-1.4.2.min.js').done(function () {`

After googling, it seems to be related to [this](https://github.com/showdownjs/showdown) repository

The latest version is 1.8.2

Is it possible to look into upgrading this?

I can do it later on today.

lupuleasa-core:
Just tested it on [http://demo.showdownjs.com](http://demo.showdownjs.com)

It seems tables are supported in the latest version, and I quote:

## Tables

Tables aren't part of the core Markdown spec, but they are part of GFM and Showdown supports them by turning on the option `tables`.

Colons can be used to align columns.

In the new version, the outer pipes (`|`) are optional, matching GFM spec. 

You also don't need to make the raw Markdown line up prettily.

You can also use other markdown syntax inside them.

```md
| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| **col 3 is**  | right-aligned | $1600 |
| col 2 is      | *centered*    |   $12 |
| zebra stripes | ~~are neat~~  |    $1 |
```

lupuleasa-core:
Maybe the `tables` option is not enabled.

[...]

PR #2310, opened 2018-08-16
Showdown update for opengrok - Markdown extension support

lupuleasa-core:
Updated to fix issue #2307 (Open grok doesn't support markdown tables - for README.md files for instance).

The markdown engine support was 2 years old, now it is the latest version.

Lup.


vladak:
Thanks !

PR #2311, opened 2018-08-17
Markdown showdown upgrade - Enable support GitHub-like markdown options in opengrok README.md parsing

lupuleasa-core:
## Use this to test:

Add it to a README.md, then view on opengrok

```
- Table test

| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| **col 3 is**  | right-aligned | $1600 |
| col 2 is      | *centered*    |   $12 |
| zebra stripes | ~~are neat~~  |    $1 |

- Simple line break test

"If you DO want to add soft line breaks (which translate to `<br>` in HTML) to a paragraph, you can do so by adding 3 space ` ` characters to the end of the line ()."

This line should break   
because it has 3 spaces at the end   
even though no line break between
-while this one shouldn't because it has no spaces

- Strikethrough support

a ~~strikethrough~~ element

- Simplified auto link

www.google.com should appear as a link automatically

- parseImgDimension test (not available on gihub, but nice to have for opengrok)

![cat image](http://i.imgur.com/7g8uq7n.jpg=250x250 "Cute cat")

![cat ref][id]

[id]: http://i.imgur.com/7g8uq7n.jpg=250x250

- githubmentionslink

Hi, my name is @lupuleasa-core 

```

## On github it turns into:

- Table test

| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| **col 3 is**  | right-aligned | $1600 |
| col 2 is      | *centered*    |   $12 |
| zebra stripes | ~~are neat~~  |    $1 |

- Simple line break test

"If you DO want to add soft line breaks (which translate to `<br>` in HTML) to a paragraph, you can do so by adding 3 space ` ` characters to the end of the line ()."

This line should break   
because it has 3 spaces at the end   
even though no line break between
-while this one shouldn't because it has no spaces

- Strikethrough support

a ~~strikethrough~~ element

- Simplified auto link

www.google.com should appear as a link automatically

- parseImgDimension test (not available on gihub, but nice to have for opengrok)

cat image

![cat ref][id]

[id]: http://i.imgur.com/7g8uq7n.jpg=250x250

- githubmentionslink

Hi, my name is @lupuleasa-core


lupuleasa-core:
From issue #2307 

Previous fix in #2311

lupuleasa-core:
If the above is not working, maybe setting the option globally will do the trick

## Globally
Setting a "global" option affects all instances of showdown

`showdown.setOption('optionKey', 'value');`

lupuleasa-core:
CI seems to be failing because I missed a semicolon `;` at the following line

`converter.setOption('ghMentionsLink', true)`

lupuleasa-core:
Unsure why the CI is failing, any ideas?

vladak:
This is just the Windows build (see #2302). Otherwise looks good.

vladak:
Could you test the changes yourself ?

Dne pá 17. 8. 2018 14:19 uživatel Lup <notifications@github.com> napsal:

lupuleasa-core:
At the moment, I am busy with work and I won't be able to test it, since that requires either setting up OpenGrok on my machine, or contacting the devops at our company and manually patching it there.

vladak:
Setting up the build environment takes couple of minutes, see the developer
intro wiki.

Dne po 20. 8. 2018 9:04 uživatel Lup <notifications@github.com> napsal:

lupuleasa-core:
I doubt it's gonna take that long, since I never did it before.
I have to wrap up work since I will go on leave at the end of this week for 10 days, so I will probably get back in 20 days.

Lup.

tulinkry:
The ci fail doesn't seem to have anything in common with this change as this is a plain js change.

opengrok-web/src/main/webapp/js/utils-0.0.24.js
```diff
@@ -1498,6 +1498,13 @@ function init_markdown_converter() {
                     var $dataMarkdownDownloadEl = $(this)
                     if (converter === null) {
                         converter = new showdown.Converter();
+                        converter.setOption('tables', true);
+                        converter.setOption('simpleLineBreaks', true);
+                        converter.setOption('strikethrough', true);
+                        converter.setOption('tasklists', true);
+                        converter.setOption('simplifiedAutoLink', true);
+                        converter.setOption('parseImgDimension', true);
+                        converter.setOption('ghMentionsLink', true)
```
vladak: missing semi-colon ?

vladak:
I tested the changes and:
  - it causes badges to be displayed under each other (rather than being displayed along side)
  - table rendering works
  - line break works
  - autolink works
  - simplified link works
  - image insertion seems to work
  - strike-through works
  - github mentions link does not work (even though I fixed the missing semi-colon)

vladak:
The badges problem is caused by `simpleLineBreaks`.

PR #2312, opened 2018-08-17
Project delete fixes

vladak:
Just found out that deletion of a project that does not have xrefs generated and/or is lacking a repository fails. This change remedies that.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
@@ -169,7 +169,7 @@ public class ProjectsController {
         // Lastly, remove data associated with the project.
         logger.log(Level.INFO, "deleting data for project {0}", projectName);
         for (String dirName: new String[]{IndexDatabase.INDEX_DIR, IndexDatabase.XREF_DIR}) {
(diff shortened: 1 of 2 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -167,9 +169,12 @@ public void deleteProject(@PathParam("project") final String projectName)
         // Lastly, remove data associated with the project.
         logger.log(Level.INFO, "deleting data for project {0}", projectName);
         for (String dirName: new String[]{IndexDatabase.INDEX_DIR, IndexDatabase.XREF_DIR}) {
-
-            IOUtils.removeRecursive(Paths.get(env.getDataRootPath() +
-                    File.separator + dirName + File.separator + projectName));
+            java.nio.file.Path path = Paths.get(env.getDataRootPath(),dirName, projectName);
```
tulinkry: why the full path for `Path` ?
```
env.getDataRootPath(),di
                     ^^^ missing space
```

vladak: had to use full path, otherwise it will use `javax.ws.rs.Path`.

Issue #2030, opened 2018-03-05
'TokenStream contract violation: close() call missing' error for .jar files


ahornace:
Test output:
```
Mar 05, 2018 12:28:35 PM org.opensolaris.opengrok.index.Indexer prepareIndexer
INFO: Scanning for repositories...
Mar 05, 2018 12:28:35 PM org.opensolaris.opengrok.index.Indexer prepareIndexer
INFO: Done scanning for repositories (0s)
Mar 05, 2018 12:28:35 PM org.opensolaris.opengrok.index.Indexer doIndexerExecution
INFO: Starting indexing
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.analysis.Ctags$1 run
WARNING: Error from ctags: ctags: Warning: Language "pascal" already defined
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.analysis.Ctags$1 run
WARNING: Error from ctags: ctags: Warning: Language "pascal" already defined
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.analysis.Ctags$1 run
WARNING: Error from ctags: ctags: Warning: Language "pascal" already defined
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.analysis.Ctags$1 run
WARNING: Error from ctags: ctags: Warning: Language "pascal" already defined
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.analysis.Ctags$1 run
WARNING: Error from ctags: ctags: Warning: Language "pascal" already defined
Mar 05, 2018 12:28:36 PM org.opensolaris.opengrok.index.IndexDatabase lambda$indexParallel$1
WARNING: ERROR addFile(): /private/var/folders/b7/t9s65dpx1blcklnlfvy1__340000gn/T/source14347768915021023503opengrok/java/Main.class
java.lang.IllegalStateException: TokenStream contract violation: close() call missing
	at org.apache.lucene.analysis.Tokenizer.setReader(Tokenizer.java:90)
	at org.apache.lucene.analysis.Analyzer$TokenStreamComponents.setReader(Analyzer.java:412)
	at org.apache.lucene.analysis.Analyzer.tokenStream(Analyzer.java:202)
	at org.apache.lucene.document.Field.tokenStream(Field.java:505)
	at org.apache.lucene.index.DefaultIndexingChain$PerField.invert(DefaultIndexingChain.java:730)
[...]

PR #2317, opened 2018-08-27
Use more native path separators on Windows

vladak:
Primarily this should fix the recently reported brokeness on Windows where backward slashes ended up in the index.

While at it, I also replaced the leading project path slash with native path separator.


vladak:
Ideally, the former fix should be tested using `SearchEngine`'s `search()`, `scoreDocs()` and `doc()` functions to retrieve the documents and asserting that their `PATH` fields do not contains backward slashes. However, the `SearchEngineTest#testSearch()` is currently commented out due to #2030 so this cannot be reliably tested.

Issue #2316, opened 2018-08-24
indexer parameter "-e" (--economical) shows only blank pages when running indexer without -H


shyoshiz:
When I create index with -e (--economical) parameter, I can see file list correctly.  However, when I click the source file to see the contents, only blank page is displayed.

vladak:
Confirmed on the latest master (cset 31256c90e23def96ac9b26214b043921cf804c84)

shyoshiz:
I think this issue is not fixed yet on both 1.1-rc39/1.1-rc40.  When I click the source file to see the contents, it shows blank page.  No xref is generated.  Please let me know how can I help to get more detail info.  Here is my usage (note:  I don't use default source/data location.):

`  indexer.py -C -D \
      -j ${JAVA_HOME}/bin/java \
      -J=-Djava.util.logging.config.file=${OPENGROK_INSTANCE_BASE}/logging.properties \
      -J=-Djava.io.tmpdir=/opt/pisces/opengrok-test/tmp \
      -J=-Xmx40960m -J=-Xms20480m -J=-server -J=-d64 \
      -a ${OPENGROK_LIB}/opengrok.jar \
      -- -P \
      --verbose --depth 24 -r off -m 256 -e \
      -c ${OPENGROK_CTAGS} \
      -s ${OPENGROK_INSTANCE_BASE}/src \
      -d ${OPENGROK_INSTANCE_BASE}/data \
      -W ${OPENGROK_CONFIGURATION}/configuration.xml
  deploy.py -D \
      -c ${OPENGROK_CONFIGURATION}/configuration.xml \
         ${OPENGROK_LIB}/source.war \
         ${OPENGROK_TOMCAT_BASE}/webapps`

vladak:
Actually, you are indexing without creating history cache (there is no -H option used) so in `list.jsp` this will go to the else branch first in order to get the redirect (so that the latest revision will appear in the URL):

```java
345     } else {
346         // requesting cross referenced file
347         File xrefFile = null;
348 
349         // Get the latest revision and redirect so that the revision number
350         // appears in the URL.
351         String location = cfg.getLatestRevisionLocation();
352         if (location != null) {
353             response.sendRedirect(location);
354             return;
355         } else {
356             xrefFile = cfg.findDataFile();
357         }
358 
359         if (xrefFile != null) {
360 %>
361 <div id="src" data-navigate-window-enabled="<%= navigateWindowEnabled %>">
362     <pre><%
[...]

PR #2319, opened 2018-08-28
Re-gen xref if missing. Fix list.jsp w.r.t. isGenerateHtml().

vladak:
Replacement for #2180 so that it applies after the recent source code tree restructuring.


vladak:
Nothing significant changed from the original pull request, I've just cherry picked the changeset and fixed javadoc issue.

I verified this also fixes #2316.

PR #2324, opened 2018-08-29
Config merge switch over

vladak:
I forgot to update `projadm.py` to use the `config-merge.py`. Here are the changes that make it possible, e.g.:

```shell
/opengrok/dist/bin/projadm.py -b /opengrok -R /opengrok/etc/readonly_configuration.xml \
    -u -r --jar /opengrok/dist/lib/opengrok.jar
```

This will:
  - retrieve current config from webapp
  - run `config-merge.py` on `/opengrok/etc/readonly_configuration.xml` and the current config and stores the result in `/opengrok/etc/configuration.xml`
  - uploads `/opengrok/etc/configuration.xml` to the webapp

### DIFF
--- a/tools/src/main/python/projadm.py
+++ b/tools/src/main/python/projadm.py
@@ -103,7 +103,8 @@ def install_config(doit, src, dst):
         sys.exit(1)
 
 
-def config_refresh(doit, logger, basedir, uri, configmerge, jar_file, roconfig):
+def config_refresh(doit, logger, basedir, uri, configmerge, jar_file,
+                   roconfig):
     """
     Refresh current configuration file with configuration retrieved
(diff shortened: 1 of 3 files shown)


tools/src/main/python/projadm.py
```diff
@@ -281,6 +292,7 @@ def project_delete(doit, logger, project, uri):
                                basedir=args.base,
                                uri=uri,
                                configmerge=configmerge_file,
```
tulinkry: what if the configmerge_file is not defined here? can it happen?

vladak: If `roconfig` is not set, then it does not matter if `configmerge` is set. Otherwise, it will be checked in the block above.

PR #2325, opened 2018-08-29
add API endpoint for include files reload

vladak:
Tested with:

- create `footer_include` file
- deploy webapp
- check the footer is displayed
- change include file content
- reload:
```shell
curl -v -X PUT http://localhost:8080/source/api/v1/system/includes/reload
```
- check the changed footer is displayed

### DIFF
--- a/opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
+++ b/opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
@@ -42,7 +42,7 @@ public class SystemControllerTest extends JerseyTest {
         Path includeRootPath = Files.createTempDirectory("systemControllerTest");
         File includeRoot = includeRootPath.toFile();
         env.setIncludeRoot(includeRoot.getAbsolutePath());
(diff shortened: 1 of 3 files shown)


tulinkry:
update the wiki

vladak:
Sure, https://github.com/oracle/opengrok/wiki/Web-services will be updated as soon as this merges in.

vladak:
For the test I had to use the `RestApp` magic otherwise Grizzly fails when resolving the suggester dependency for `SystemController`.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        super.tearDown();
+    }
```
tulinkry: are these two needed?

vladak: Looks like not.

PR #2327, opened 2018-08-30
add API endpoint to refresh path descriptions

vladak:
Hides path description compilation into binary file behind RESTful API endpoint.

Tested with:

- create `/opengrok/etc/paths.tsv` with some sample path descriptions
- call the API to refresh the descriptions:
```
curl -i -X POST -H "Content-Type: text/plain" \
    --data-binary "@/opengrok/etc/paths.tsv" \
    http://localhost:8080/source/api/v1/system/pathdesc
```
- view projects with said path descriptions in the UI, make sure they have updated descriptions


vladak:
will update https://github.com/oracle/opengrok/wiki/Web-services once this is merged in.

vladak:
https://github.com/oracle/opengrok/wiki/Webapp-configuration#path-descriptions will need to be updated too

PR #2342, opened 2018-09-06
if economy mode is on and history off, generate xref on the fly

vladak:
This is a bit of a hack to make `list.jsp` jump into the code that generates xrefs on the fly if latest revision cannot be determined and economy mode is on.

Also, an error message was added to avoid empty page appearing on error.

### DIFF
--- a/opengrok-web/src/main/webapp/list.jsp
+++ b/opengrok-web/src/main/webapp/list.jsp
@@ -349,8 +349,7 @@ Click <a href="<%= rawPath %>">download <%= basename %></a><%
         // requesting cross referenced file
         File xrefFile = null;
 
-        // Get the latest revision and redirect so that  the revision number
-        // appears in the URL.
+        // Get the latest revision and redirect so that the revision number appears in the URL.
         String location = cfg.getRevisionLocation(cfg.getLatestRevision());
         if (location != null) {
             response.sendRedirect(location);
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/webapp/list.jsp
```diff
@@ -344,13 +349,21 @@ Click <a href="<%= rawPath %>">download <%= basename %></a><%
         // requesting cross referenced file
         File xrefFile = null;
 
-        // Get the latest revision and redirect so that the revision number
+        // Get the latest revision and redirect so that  the revision number
```
tulinkry: extra space?

opengrok-web/src/main/webapp/list.jsp
```diff
@@ -362,6 +375,10 @@ Click <a href="<%= rawPath %>">download <%= basename %></a><%
             Util.dumpXref(out, xrefFile, compressed, request.getContextPath());
     %></pre>
 </div><%
+        } else {
+            String error = "Failed to get xref file";
+%>
+<p class="error"><%= error %></p><%
```
tulinkry: you don't have to do the variable expansion, do you?

vladak: what do you mean ?

tulinkry: `<p class="error">Failed to get xref file</p>`

vladak: I see, yes. Originally I wanted to make the error to include the path to the file however this is not necessary.

PR #2346, opened 2018-09-07
get repositories for a project automatically

vladak:
This change makes it possible to run per-project indexing with just `-H projName` and the Indexer will get the list of repositories from the configuration.

The repository specification was separated into new `--repository` option.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -791,8 +791,7 @@ public final class Indexer {
 
     private static void checkConfiguration() {
         if (repositories.size() > 0 && !cfg.isHistoryEnabled()) {
-            System.out.println("Repositories were specified however history is off");
-            System.exit(1);
+            die("Repositories were specified however history is off");
         }
     }
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -397,7 +397,7 @@ public final class Indexer {
         }
 
         OptionParser configure = OptionParser.scan(parser -> {
-            parser.on("-R configPath").Do( cfgFile -> {
+            parser.on("-R configPath").Do(cfgFile -> {
                 try {
                     cfg = Configuration.read(new File((String)cfgFile));
                 } catch(IOException e) {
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -499,19 +504,9 @@ public static WebAddress parseWebAddress(String webAddr) {
...
-                Do( repo -> {
-                    String repository = (String) repo;
-                    if (repository.equals("")) {
-                        cfg.setHistoryEnabled(true);  // all repositories
-                    } else {
-                        repositories.add((String)repository);  // specific repository
-                    }
-                }
-            );
+            parser.on("-H", "--history", "Enable history.").Do( v -> {
```
tulinkry: is the space here what idea suggests to you?
```
Do( v -> {
   ^
```

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -788,6 +789,13 @@ public static WebAddress parseWebAddress(String webAddr) {
         return argv;
     }
 
+    private static void checkConfiguration() {
+        if (repositories.size() > 0 && !cfg.isHistoryEnabled()) {
+            System.out.println("Repositories were specified however history is off");
```
tulinkry: why not use `die()` ?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -98,7 +98,7 @@
     private static String configFilename = null;
     private static int status = 0;
 
-    private static final ArrayList<String> repositories = new ArrayList<>();
+    private static final Set<String> repositories = new HashSet<>();
```
tulinkry: the trick here just to do the dedup? or is there any other magic?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: no, I just followed the pre-existing code style in the method.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: yes, this is just to avoid duplicities.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
tulinkry: the style doesn't seem consistent in this parser options anyway

vladak: ok, fixed them all not to include the whitespace.

tulinkry: perfect

Issue #2318, opened 2018-08-27
supply Universal ctags for Appveyor build


vladak:
Currently the Appveyor builds lack Universal ctags so a number of tests is skipped there.

PR #2359, opened 2018-09-27
sanitize more paths on windows, fixes #2335

### LINKED ISSUE #2335: repository not displayed on the index page on Windows
As reported in #2315, after the changes in csets b2315481a20b1e8658027d7ded7e38e3b017fdce or 31256c90e23def96ac9b26214b043921cf804c84 the repositories stopped appearing on the index page on Windows.

tarzanek:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
@@ -844,7 +844,7 @@ public final class Util {
      * @return the mangled path.
      */
     public static String path2uid(String path, String date) {
-        return path.replace('/', '\u0000') + "\u0000" + date;
+        return path.replace(File.separatorChar, '\u0000') + "\u0000" + date;
     }
 
     /**
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
@@ -339,7 +339,7 @@ public class Project implements Comparable<Project>, Nameable, Serializable {
         // Try to match each project path as prefix of the given path.
         final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         if (env.hasProjects()) {
-            final String lpath = path.replace(File.separatorChar, '/');
+            final String lpath = path;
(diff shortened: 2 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -844,7 +844,7 @@ public static void readableLine(int num, Writer out, Annotation annotation,
      * @return the mangled path.
      */
     public static String path2uid(String path, String date) {
-        return path.replace('/', '\u0000') + "\u0000" + date;
+        return path.replace(File.separatorChar, '\u0000') + "\u0000" + date;
```
vladak: how is this going to interact with path sanitization done in populateDocument() ?

tarzanek: it makes that sanitization kinda useless, hmm, I think we need to define what to do and how
originally whole opengrok used "/" everywhere - newer versions of windows don't have a problem to recognize /var/opengrok and translate it automatically to C:\var\opengrok
so if we are going to skip File.separator and go back to "/" everywhere it should work - why was it fixed to File.separator? (looking at the changesets I couldn't find the original bugs, resp. the problem in them)

tarzanek: why was the path sanitized? did it conflict with lucene somehow, how? /me curious

vladak: Lucene does not like literal backslashes because it uses them for escaping.

vladak:
As for AppVeyor builds, see #2302. There was never full 100% PASS build and the coverage is less than optimal (see #2318). In this state (with tests failing) one can easily break something without even noticing (such as breaking already failing test even more). That's why I proposed to do the mini hackaton before 1.1 release to fix all failing Windows tests.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
vladak: As for the sanitization we need to decide where it will be done. If in populateDocument() or in path manipulating methods.

PR #2362, opened 2018-10-01
merge command options smarter

vladak:
Tested with (`hg` is present on the system):
```
/indexer.py -J=-Dorg.opengrok.indexer.history.Mercurial=/foo/bar -D -a /var/tmp/opengrok-1.1-rc41/lib/opengrok.jar -- -h
/indexer.py -D -a /var/tmp/opengrok-1.1-rc41/lib/opengrok.jar -- -h
```

### DIFF
--- a/tools/src/main/python/indexer.py
+++ b/tools/src/main/python/indexer.py
@@ -95,34 +95,18 @@ def merge_properties(base, extra):
     :return: merged list
     """
 
-    ret = set([])
-    for item in extra:
-        if item.find("=") != -1:
-            name, _ = item.split("=")
-            if get_property(name, base) is None:
-                ret.add(item)
-        else:
-            ret.add(item)
+    extra_prop_names = set(map(lambda x: x.split('=')[0], base))
 
+    ret = set([])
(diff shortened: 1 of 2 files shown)


tools/src/main/python/indexer.py
```diff
@@ -47,40 +47,82 @@ def __init__(self, command, logger=None, java=None, jar='opengrok.jar',
...
+    for cmd in SCM_COMMANDS.keys():
+        executable = get_command(logger, None, cmd, level=logging.INFO)
+        if executable:
+            properties.append("{}={}".
+                              format(SCM_COMMANDS[cmd], executable))
+
+    return properties
+
+
+def merge_properties(base, extra):
```
tehe: I'd introduce a 'name' cache - should probably be a set, see. https://stackoverflow.com/questions/2831212/python-sets-vs-lists - (i.e. the first element of the item.split("=")) and then I'd process 'base' first (populating the cache along the way) and then proceed with processing 'extra' depending on whether given name is in the cache or not.

Issue #2382, opened 2018-10-05
setting configuration property via API is not propagated to all threads


tulinkry:
Today I think I hit a race condition when doing the parallel indexing and using the sync.py script (mostly from the wiki).

My flow is following:

1. Delete all unwanted projects with projadm.py
2. Download all sources from the SCM
3. Add all projects with projadm.py (no index at this step)
4. Generate groups and publish them to the webapp
5. Index the sources using sync.py script with mirroring.

I'm using 10 repositories, some of them take longer time to index but most of them can finish in almost the same time. I'm using the number of workers 4.

When the sync.py script is finished, with no errors, the page itself displays only 9 (sometimes 8) projects most of the time, with one of them missing randomly. After looking in the webapp configuration the project itself is not marked as indexed (therefore not visible).

However, looking at the indexer logs, there is a 10 times a mention that the uri `projects/{p}/indexed` has been fired and return 204.

```bash
-bash-4.1$ cat /opengrok/log/**/opengrok0.0.log | grep "indexed HTTP/1.1" | wc -l
10
```

Also looking at the tcpdump communication (posted below), I can see 10 calls for the indexed url.

I couldn't dig any deeper at this point, but there's something wrong going on.

Sync-config.yml:
```yaml
commands:
- command:
  - http://localhost:8150/source/api/v1/messages
  - POST
  - cssClass: info
    duration: PT2H
    tags: ['%PROJECT%']
    text: resync + reindex in progress
- command: [ /opengrok/bin/mirror.py, -U, 'http://localhost:8150/source', -c, /opengrok/etc/mirror-config.yml ]
- command: [
    /opengrok/bin/reindex-project.py,
    -j,
    '/usr/java/jdk1.8.0_60/bin/java',
    -J=-d64,
    '-J=-XX:-UseGCOverheadLimit',
    -J=-Xmx12g,
    -J=-server,
    -J=-Dorg.opengrok.indexer.history.git=/opt/git/git283/bin/git,
    --jar,
    /opengrok/lib/opengrok.jar,
    -t,
    /opengrok/etc/logging.properties.template,
    -p,
    '%PROJ%',
    -d,
    /opengrok/log/%PROJECT%,
    -P,
    '%PROJECT%',
    --uri,
    'http://localhost:8150/source',
    --,
    -U,
    'http://localhost:8150/source',
    --renamedHistory,
    'off',
    -r,
    dirbased,
    --memory,
    '256',
    -c,
[...]

PR #2388, opened 2018-10-06
Hotfix for non-deterministic results for partial index

tulinkry:
Under some circumstances the configuration view per each thread
can become desynchronized when using partial index.

The issue itself should be tackled in more robust way.

Described properly in #2382.


tarzanek:
hmm, doing refresh of config inside auth filter ... not a great architecture design ... 
could it be moved to one of the jsp pages? or it's too late there?

tulinkry:
You can't do it in jsp as rest api doesn't use jsp. And the jsp have this call anyway I think.
Again, this is just a hotfix, proper fix should be addressed for this.

vladak:
This might actually make the memory leaks (as reported by Tomcat during redeploy) worse, so better to speed up the proper fix.

PR #2389, opened 2018-10-07
Packaging python scripts

### LINKED ISSUE #2251: consider shipping tools Python scripts as python package
As mentioned in #2245, the Python scripts in `tools`

> should be a python package and the distribution version of this package we should ship with each opengrok release.

tulinkry:
fixes #2251

This includes:
 - renaming tools to opengrok-tools
 - packaging the python scripts as opengrok-tools python package
 - running tests in maven for the package
 - using virtual env
 - including the python package in final distribution in tools/opengrok-tools.tar.gz


tulinkry:
There's more to go as I'm not really skilled with maven.

 - the python package should be generated to target directory (now is generated to dist and build directories)
 - maven clean should clean the generated python files
 - think about something what to do with the python package version (replace with the current opengrok version?)

tulinkry:
If anyone can examined these it'd be nice. I didn't write the original scripts so I'm not sure if that was just a copy paste mistake.
```
opengrok_tools/sync.py:61:5: F821 undefined name 'logger'
opengrok_tools/sync.py:69:5: F841 local variable 'output' is assigned to but never used
opengrok_tools/mirror.py:57:5: F841 local variable 'output' is assigned to but never used
opengrok_tools/mirror.py:58:5: F841 local variable 'dirs_to_process' is assigned to but never used
```

vladak:
> think about something what to do with the python package version (replace with the current opengrok version?)

yes please.

vladak:
The fixes to the lint errors look okay.

tulinkry:
```
Name: opengrok-tools
Version: 1.1rc48
Summary: Tools for managing OpenGrok instance
Home-page: https://github.com/OpenGrok/OpenGrok
Author: Oracle
Author-email: opengrok-users@yahoogroups.com
License: CDDL
Location: /Users/ktulinger/OpenGrok/opengrok-tools/env/lib/python3.6/site-packages/opengrok_tools-1.1rc48-py3.6.egg
Requires: jsonschema, pyyaml, requests, resource
Y
```

distribution/assembly.xml
```diff
@@ -19,6 +19,11 @@
         <outputDirectory>lib</outputDirectory>
         <destName>source.war</destName>
       </file>
+      <file>
+        <source>${project.basedir}/../opengrok-tools/dist/opengrok-tools-1.1rc48.tar.gz</source>
```
vladak: hard-coded version ?

opengrok-tools/setup.py
```diff
@@ -0,0 +1,58 @@
+import os
+import unittest
+from setuptools import setup
+
+
+def readme():
+    with open(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'src', 'main', 'python', 'README.md'), 'r') as readme:
```
vladak: this line looks a bit long, does the script pass `flake8` ?

opengrok-tools/setup.py
```diff
@@ -0,0 +1,58 @@
...
+    package_dir={
+        'opengrok_tools': 'src/main/python/opengrok_tools',
+        'all': 'src/main/python/opengrok_tools/all',
+        'all.scm': 'src/main/python/opengrok_tools/all/scm',
+        'all.utils': 'src/main/python/opengrok_tools/all/utils',
+    },
+    url='https://github.com/OpenGrok/OpenGrok',
+    license='CDDL',
+    author='Oracle',
+    author_email='opengrok-users@yahoogroups.com',
```
vladak: `opengrok-dev` would be more fitting I think

opengrok-tools/src/main/python/README.md
```diff
@@ -0,0 +1,105 @@
...
+or make a distribution tarball.
+
+```bash
+python setup.py sdist
+ls -l dist/
+```
+
+### Installation on the target system
+
+Use the distribution tarball and run pip.
```
vladak: `pip`

opengrok-tools/setup.py
tulinkry: Maybe it’s excluded from the check because the flake tests are part of Travis and it succeeded. Let me check

distribution/assembly.xml
vladak: Also, will this mean that user will have to extract the tools separately after extracting the distribution tarball ?

tulinkry: No, following the instalattion instruction, the tar.gz file is a python package

```bash
python3 -m pip install opengrok_tools-{version}.tar.gz
```

tulinkry: So extracting the distribution tarball and installing tools with python from the tools subdirectory.

opengrok-tools/setup.py
tulinkry: added

opengrok-tools/setup.py
tulinkry: done

opengrok-tools/src/main/python/README.md
tulinkry: done

distribution/assembly.xml
tulinkry: I removed the hardcoded version but it will still break in the future.
The problem here is that the python setup.py removes the dash from the version name 1.1-rc48 making it only 1.1rc48.

tulinkry: Or eventually if we have the domain opengrok.org we can have there self hosted pypi repository for the opengrok tools https://pypi.org/project/pypiserver/

opengrok-tools/setup.py
```diff
@@ -0,0 +1,63 @@
...
+def my_test_suite():
+    test_loader = unittest.TestLoader()
+    test_suite = test_loader.discover(
+        os.path.join(SCRIPT_DIR, 'src', 'test', 'python'), pattern='test_*.py')
+    return test_suite
+
+
+setup(
+    name='opengrok-tools',
+    version='1.1rc57',
```
vladak: is there a way how to avoid hardcoding the version ? Should we be using OpenGrok version at all here actually ?

tulinkry: We should and I think maven can do it, I hope someone how is with better experience with maven can handle this

tulinkry: > >    think about something what to do with the python package version (replace with the current opengrok version?)

> yes please. That said, each major script has its own version.

You already said that we should use the opengrok version, that's why I changed it.

vladak:
Pls file a new issue to get the version number resolved.

PR #2401, opened 2018-10-09
Print to the stream related to the exit code

### LINKED ISSUE #2379: ConfigMerge Java program does not print help
```bash
./bin/config-merge.py -j /usr/java/jdk1.8.0/bin/java --jar ./lib/opengrok.jar -- -h
```

prints just a blank line

tulinkry:
Python scripts can then gather the output according to the exit code.

fixes #2379

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigMerge.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigMerge.java
@@ -27,6 +27,7 @@ import java.beans.PropertyDescriptor;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.PrintStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigMerge.java
```diff
@@ -160,18 +161,18 @@ public static void main(String[] argv) {
...
-        System.err.println(name + " [-h] <config_file_base> <config_file_new>");
-        System.err.println();
-        System.err.println("OPTIONS:");
-        System.err.println("Help");
-        System.err.println("-?                   print this help message");
-        System.err.println("-h                   print this help message");
-        System.err.println();
+    private static final void a_usage(PrintStream out) {
+        out.println("Usage:");
+        out.println(name + " [-h] <config_file_base> <config_file_new>");
```
vladak: Please add some comment saying that base is usually read-only configuration etc.

PR #2402, opened 2018-10-09
add global tunable for navigate window

vladak:
Tested with this read-only configuration:
```XML
<java version="1.8.0_60" class="java.beans.XMLDecoder">
  <void property="navigateWindowEnabled">
     <boolean>true</boolean>
  </void>

  <void property="projects">                                                    
   <void method="put">                                                          
    <string>OpenGrok</string>                                                   
     <void property="navigateWindowEnabled">                                    
      <boolean>false</boolean>                                                  
     </void>                                                                    
    </object>                                                                   
   </void>                                                                      
  </void>   

 </object>
</java>
```
The feature is that the resulting configuration (written with `-W` indexer option) will have the flag set in each of the projects that does not override it, the result of the way serialization is done.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -274,6 +274,12 @@ public final class Configuration {
      */
     private boolean listDirsFirst = true;
 
+    /**
+     * A flag if the navigate window should be opened by default when browsing
+     * the source code of projects.
+     */
+    private boolean navigateWindowEnabled;
+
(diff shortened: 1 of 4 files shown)


tulinkry:
I don't like this as it is duplicating properties; on the code side lgtm.

vladak:
@tulinkry you mean that the variable name is the same in `Configuration` and `Project` classes ?

tulinkry:
Yes, not the name, but the design. We moved away from having tunable in configuration for all projects to have per-project configuration and now we're trying to have both.

vladak:
I think it makes sense - having a global property overridable on per project basis.

PR #2411, opened 2018-10-12
Mirror incoming check

vladak:
Tested with a sample Mercurial and Git repo.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/mirror.py
@@ -43,13 +43,15 @@ from all.utils.opengrok import get_repos, get_config_value, get_repo_type
 from all.utils.readconfig import read_config
 from all.utils.repofactory import get_repository
 from all.utils.utils import is_exe, check_create_dir, get_int, diff_list
+from all.scm.repository import RepositoryException
+
 
 major_version = sys.version_info[0]
-if (major_version < 3):
+if major_version < 3:
     print("Need Python 3, you are running {}".format(major_version))
     sys.exit(1)
 
-__version__ = "0.2"
+__version__ = "0.3"
 
 # "constants"
 HOOK_TIMEOUT_PROPERTY = 'hook_timeout'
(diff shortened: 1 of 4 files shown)


opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -43,30 +43,85 @@
...
+                              None,
+                              command_timeout)
+        if not repo:
+            logger.error("Cannot get repository for {}".
+                         format(repo_path))
+            ret = 1
+        else:
+            repos.append(repo)
+
+    return repos, ret
```
tulinkry: Consider using None for three-state return?

vladak:
Also, the reason I selected `git update --dry-run` is that I wanted to avoid parsing the output of `git status` (in the case of `git fetch && git status -uno`).

tulinkry:
I see `git pull --dry-run` not `update`

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -352,14 +345,21 @@ def main():
...
-                # The error was already logged in get_repos_for_project()
+            repos = []
+            try:
+                global_config = {'commands': config.get(COMMANDS_PROPERTY),
+                                 'proxy': proxy,
+                                 'command_timeout': command_timeout,
+                                 'source_root': source_root,
+                                 'uri': uri}
+                repos = get_repos_for_project(logger, args.project,
+                                              ignored_repos, **global_config)
```
tulinkry: I think you can also write it like this

```python
get_repos_for_project(logger, args.project, ignored_repos,  commands=config.get(..), proxy=proxy, command_timeout=command_timeout, source_root=source_root, uri=uri)
```

choose the way you prefer.

vladak:
Yep, it should've read `pull'.

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -43,30 +43,78 @@
...
+                              # of repo_path is absolute path.
+                              kwargs['source_root'] + repo_path,
+                              repo_type,
+                              project,
+                              kwargs['commands'],
+                              kwargs['proxy'],
+                              None,
+                              kwargs['command_timeout'])
+        if not repo:
+            raise RepositoryException("Cannot get repository for {}".
```
tulinkry: +1 nice

PR #2416, opened 2018-10-16
continuous output printing in indexer.py

vladak:
This change enables continuous printing of Indexer output to stdout. The output is still stored in pertaining buffers.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/all/utils/command.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/all/utils/command.py
@@ -40,7 +40,8 @@ class TimeoutException(Exception):
 class Command:
     """
     wrapper for synchronous execution of commands via subprocess.Popen()
-    and getting their output (stderr is redirected to stdout) and return value
+    and getting their output (stderr is redirected to stdout by default)
+    and exit value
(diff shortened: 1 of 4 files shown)


opengrok-tools/src/main/python/opengrok_tools/all/utils/command.py
```diff
@@ -169,6 +174,9 @@ def run(self):
 
                     self.out.append(line)
 
+                    if self.doprint:
+                        print(line.strip())
```
tulinkry: I'm little bit worried that while the intention is clear (to remove the trailing `\n`) this will destroy formatted messages with `tab` or `spaces` at the beginning of the line. I can imagine for example exception stacktraces which are indented in the log.

Could you please check it?

tulinkry:
And also the wrecker build hasn't passed.

vladak:
the Wercker build failed on its own.

opengrok-tools/src/main/python/opengrok_tools/all/utils/command.py
vladak: of course I wanted to use rstrip()

PR #2418, opened 2018-10-16
add JAX-B dependencies

vladak:
Should fix RESTful API problems with Java 9+.


vladak:
I wonder if suggester needs this (or will need) as well.

tarzanek:
let's merge, I hope the probs will go away

Issue #2161, opened 2018-06-15
Memory leaks on redeploy


Orviss:
Memory usage from the demo instance:
![usedmemory](https://user-images.githubusercontent.com/12401160/41471393-ce051c0c-70b3-11e8-8d23-b5bca5ad1903.png)

There was no Tomcat restart until the last week (only redeploys -> and quite a lot). So it seems that there might be some memory leaks. Maybe related to `ThreadLocal` usage? However, here <https://wiki.apache.org/tomcat/MemoryLeakProtection> they say it should already be fixed by Tomcat.

I've noticed that this might be the case when I was looking into generic Java memory leaks and someone mentioned Tomcat redeploys.

However, I'm not sure if this is considered a bug or not. I have next to no experience with this. Could someone please elaborate? :) Thanks.

vladak:
Some leaks were also reported in #1761

vladak:
Marking as a bug for now - if the demo instance (assuming it is running just OpenGrok) grows from 0.2 GB to 1.8 GB in a week then something is wrong.

vladak:
As for `ThreadLocal`, it is used for the `RuntimeEnvironment`:

```java
    /**
     * Creates a new instance of RuntimeEnvironment. Private to ensure a
     * singleton anti-pattern.
     */
    private RuntimeEnvironment() {
        configuration = new Configuration();
        threadConfig = new ThreadLocal<Configuration>() {
            @Override
            protected Configuration initialValue() {
                return configuration;
            }
        };
    }
```

Nurm78:
Since upgrading to opengrok 1.1-rc28, I also see an increased number of These Messages in the catalina log.
My System is a Windows 2008 R2 SP1 with 8 GB RAM and JDK 1.8.172 and Apache tomcat 8.5.23.

In the attached catalina log, many Analyser Reports a Memory leak. I do not get the same log pattern every day, most of the days, the configuration and the PlainAnalyzer are the only ones which logs memory leaks.
[catalina.2018-07-19.log](https://github.com/oracle/opengrok/files/2219179/catalina.2018-07-19.log)

vladak:
The problem of `ThreadLocal` causing memory leak in the presence of thread pools is described e.g. on:
  - https://www.quora.com/What-are-some-tips-tricks-and-gotchas-when-using-Java-ThreadLocal
[...]

PR #2423, opened 2018-10-18
Configuration rework

vladak:
Avoids storing `Configuration` in `ThreadLocal` variable.

The intention was to keep `Configuration` class as close to *POJO* as possible, i.e. the lock that protects it is in `RuntimeEnvironment`.


vladak:
Also, might help with #2161.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -246,28 +240,71 @@ private String getCanonicalPath(String s) {
...
+            readLock.lock();
+            return invokeGetter(configuration, fieldName);
+        } catch (IOException e) {
+            return null;
+        } finally {
+            if (readLock != null) {
+                readLock.unlock();
+            }
+        }
+    }
```
tulinkry: Shorter version, for all of the cases
```java
        configLock.readLock().lock();
        try {
            return invokeGetter(configuration, fieldName);
        } catch (IOException e) {
            return null;
        } finally {
            configLock.readLock().unlock();
        }
```

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -246,28 +240,71 @@ private String getCanonicalPath(String s) {
         }
     }
 
+    private Object getConfigurationValue(String fieldName) {
```
tulinkry: I'd vote for 
```java
private <T> T getConfigurationValue(String fieldName, T type) {
    return (T) invokeGetter( ... );
}

getConfigurationValue("sourceRoot", String.class);
```

opengrok-indexer/src/main/java/org/opengrok/indexer/util/ClassUtil.java
```diff
@@ -76,116 +76,140 @@ public static void remarkTransientFields(Class targetClass) {
...
+     *
+     * @param obj the object
+     * @param fieldName name of the field which will be changed
+     * @param value desired value
+     *
+     * @throws IOException if any error occurs (no suitable method, bad conversion, ...)
+     */
+    public static void invokeSetter(Object obj, String fieldName, String value) throws IOException {
+        Method setter = getSetter(obj, fieldName);
+        Class<?> c = setter.getParameterTypes()[0];
```
tulinkry: A note that the size is checked elsewhere would be nice

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1347,17 +1407,24 @@ public void writeConfiguration(File file) throws IOException {
...
+            readLock.lock();
+
+            Response r = ClientBuilder.newClient()
+                    .target(host)
+                    .path("api")
+                    .path("v1")
+                    .path("configuration")
+                    .queryParam("reindex", true)
+                    .request()
+                    .put(Entity.xml(configuration.getXMLRepresentationAsString()));
```
tulinkry: Is it neccessary to lock the whole api call which can take a non-trivial amount of time while the only thing which needs the lock is this `configuration.getXMLRepresentationAsString()`

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1510,7 +1581,8 @@ public void reloadIncludeFiles(Configuration configuration) {
     }
 
     public Configuration getConfiguration() {
-        return this.threadConfig.get();
+        // TODO lock/clone ?
```
tulinkry: or maybe completely delete this and use the appropriate setters/getters?

tulinkry:
While this is a best effort, it's not 100% thread safe. Even on the issue use case, when marking a project as indexed.

```java
202      public void markIndexed(@PathParam("project") final String projectName) throws Exception {
203  
204          Project project = env.getProjects().get(projectName);
205          if (project != null) {
                 // now you're mutating the state of the configuration and it is not locked or anything
                 // while this doesn't case much harm, adding project or other mutations can cause
                 // an exception
206              project.setIndexed(true);
207  
208              // Refresh current version of the project's repositories.
209              List<RepositoryInfo> riList = env.getProjectRepositoriesMap().get(project);
210              if (riList != null) {
211                  for (RepositoryInfo ri : riList) {
212                      Repository repo = getRepository(ri, false);
213  
214                      if (repo != null && repo.getCurrentVersion() != null && repo.getCurrentVersion().length() > 0) {
215                          // getRepository() always creates fresh instance
216                          // of the Repository object so there is no need
217                          // to call setCurrentVersion() on it.
218                          ri.setCurrentVersion(repo.determineCurrentVersion());
219                      }
220                  }
221              }
222              suggester.refresh(projectName);
223          } else {
224              logger.log(Level.WARNING, "cannot find project {0} to mark as indexed", projectName);
225          }
          ....
232      }
```

Because of mentioned problem I would add methods like this for read and write, if someone wants to run some code while being locked.

```java
    public void withReadLock(Consumer<Configuration> callback) {
        configLock.readLock().lock();
        try {
            callback.accept(this.configuration);
        } finally {
            configLock.readLock().unlock();
        }
        // usage
        withReadLock(config -> {
            config.setAllowLeadingWildcard(true);
        });
    }
```

This way you can use it for doing such operations while being locked.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -497,20 +535,8 @@ public boolean hasGroups() {
...
-     * @return this instance
-     */
-    public RuntimeEnvironment register() {
-        threadConfig.set(configuration);
-        return this;
+        populateGroups(groups, new TreeSet<>(getProjects().values()));
+        setConfigurationValue("groups", groups);
     }
 
     /**
```
tulinkry: repository_map doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -542,10 +568,9 @@ public String getUrlPrefix() {
...
-        return ctags != null ? ctags : (value =
-            threadConfig.get().getCtags()) != null ? value :
-            System.getProperty(SYSTEM_CTAGS_PROPERTY,
-            "ctags");
+        return ctags != null ? ctags :
+                (value = (String)getConfigurationValue("ctags")) != null ? value :
+                System.getProperty(SYSTEM_CTAGS_PROPERTY,"ctags");
     }
 
     /**
```
tulinkry: setting ctags doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -592,19 +617,19 @@ public void setMandoc(String value) {
     }
```
tulinkry: setting mandoc doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -775,15 +807,15 @@ public void setDefaultProjects(Set<Project> defaultProject) {
      * @return the default projects (may be null if not specified)
      */
     public Set<Project> getDefaultProjects() {
-        return threadConfig.get().getDefaultProjects();
+        return (Set<Project>)getConfigurationValue("defaultProjects");
```
tulinkry: if you modify the returned set from multiple threads you have a problem here

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1230,27 +1283,27 @@ public String getConfigURI() {
     }
```
tulinkry: config uri doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1261,7 +1314,7 @@ public void setListDirsFirst(boolean flag) {
      */
     public short getContextLimit() {
         return contextLimit != null ? contextLimit :
-            threadConfig.get().getContextLimit();
+                (short)getConfigurationValue("contextLimit");
     }
 
     /**
```
tulinkry: setting context limit doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1290,7 +1343,7 @@ public void setContextLimit(Short value)
      */
     public short getContextSurround() {
         return contextSurround != null ? contextSurround :
-            threadConfig.get().getContextSurround();
+                (short)getConfigurationValue("contextSurround");
     }
 
     /**
```
tulinkry: setting context surround doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1347,17 +1407,24 @@ public void writeConfiguration(File file) throws IOException {
...
+                    .request()
+                    .put(Entity.xml(configuration.getXMLRepresentationAsString()));
+
+            if (r.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
+                throw new IOException(r.toString());
+            }
+        } finally {
+            readLock.unlock();
         }
     }
```
tulinkry: generate project repository map doesn't need locking?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: shortened all 3 versions.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -110,6 +111,8 @@
 
     private Statistics statistics = new Statistics();
 
+    public IncludeFiles includeFiles;
```
tulinkry: public? please no

vladak: what would be the alternative ?

tulinkry: I know that exposing it with getter is in a sense the same, but keep it please consistent with the rest of the file. This class is the most used in OpenGrok so rather make it clean.

vladak: okay, used getter.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: I think this is not necessary. Firstly, it is not part of configuration, secondly it gets set only initially.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: right, the returned set needs to be immutable.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: I think that's fine, it only gets set once during the initial bootstrap.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: same case as ctags.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -468,9 +506,9 @@ public boolean hasProjects() {
      */
     public void setProjects(Map<String,Project> projects) {
         if (projects != null) {
-            populateGroups(getGroups(), new TreeSet<Project>(projects.values()));
+            populateGroups(getGroups(), new TreeSet<>(projects.values()));
```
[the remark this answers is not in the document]
vladak: wrapped in write lock.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: it initialized as `ConcurrentHashMap` so my guess is no. Also, it is defined within `RuntimeEnvironment`, not `Configuration`.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/ClassUtil.java
```diff
@@ -77,7 +77,7 @@ public static void remarkTransientFields(Class targetClass) {
         }
     }
 
-    private static Object getObjectValue(Class c, String value) throws IOException {
+    private static Object getObjectValue(String fieldName, Class c, String value) throws IOException {
```
tulinkry: What is `value` here? getting a value?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/ClassUtil.java
```diff
@@ -183,30 +189,29 @@ private static Method getSetter(Object obj, String fieldName) throws IOException
...
         // TODO this will get the setter again => refactor
         invokeSetter(obj, fieldName, objValue);
     }
 
+    /**
+     *
+     * @param obj the object
+     * @param fieldName name of the field which will be changed
+     * @param value desired value
+     * @throws IOException
```
tulinkry: maybe add a note that all exceptions from the reflection are propagated as ioexception

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1699,6 +1704,7 @@ public void applyConfig(String configuration, boolean reindex, boolean interacti
      * @param reindex is the message result of reindex
      * @param interactive true if in interactive mode
      *
+     * Assumes it is called with the {@code configLock} held in write mode.
```
tulinkry: What is the problem to wrap the inside of this body in write lock block?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -734,7 +734,12 @@ public void setRepositories(List<RepositoryInfo> repositories) {
     }
     
     public void removeRepositories() {
-        setConfigurationValue("repositories", null);
+        try {
+            configLock.writeLock().lock();
+            configuration.repositories =  null;
```
tulinkry: `setRepositories`?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1300,60 +1287,19 @@ public int getTabSize() {
...
-     * <p>
-     * N.b. the value is not mediated to {@link Configuration}.
-     * @param value a defined value or {@code null} to reset to use the
-     * {@link Configuration#getContextSurround()}
-     * @throws IllegalArgumentException if {@code value} is negative
-     */
-    public void setContextSurround(Short value)
-            throws IllegalArgumentException {
-        if (value < 0) {
-            throw new IllegalArgumentException("value is negative");
```
tulinkry: if you set it to negative in the read only config, will it throw an exception to signal an error now?

tulinkry: also the same for the other deduplicated value

vladak:
```
  <void property="contextLimit">                                                
    <short>-100</short>                                                         
  </void>                                                                       
  <void property="contextSurround">                                             
    <short>-100</short>                                                         
  </void> 
```
fail with:
```
$ java -cp 'distribution/target/dist/*' org.opengrok.indexer.index.Indexer -R /var/opengrok/etc/readonly_configuration.xml -s /var/opengrok/src -d /var/opengrok/data/
Oct 23, 2018 1:28:55 PM org.opengrok.indexer.configuration.Configuration read
INFO: Reading configuration from /var/opengrok/etc/readonly_configuration.xml
ERROR: java.lang.IllegalArgumentException: Invalid value for "contextSurround" - "-100". Expected value greater or equal than 0
$ java -cp 'distribution/target/dist/*' org.opengrok.indexer.index.Indexer -R /var/opengrok/etc/readonly_configuration.xml -s /var/opengrok/src -d /var/opengrok/data/
Oct 23, 2018 1:29:46 PM org.opengrok.indexer.configuration.Configuration read
INFO: Reading configuration from /var/opengrok/etc/readonly_configuration.xml
ERROR: java.lang.IllegalArgumentException: Invalid value for "contextLimit" - "-100". Expected value greater than 0
```

tulinkry:
I'm not ready. I again misclicked.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/StatUtil.java
```diff
@@ -0,0 +1,127 @@
+/*
```
tulinkry: I like StatisticsUtils more.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/StatUtil.java
```diff
@@ -0,0 +1,127 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+ /*
+  * Copyright (c) 2006, 2018, Oracle and/or its affiliates. All rights reserved.
```
tulinkry: no 2006

tulinkry:
Did you try the test case in the issue?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/StatUtil.java
vladak: I think it needs to be preserved, assuming the code comes from at least 2006. I am not a lawyer, though.

tulinkry: ok

vladak:
As for the solution to be 100% thread safe I am not sure if this is even desired. The code for setting the `indexed` property of a project looks like this:

```java
        if (project != null) {
            project.setIndexed(true);

            // Refresh current version of the project's repositories.
            List<RepositoryInfo> riList = env.getProjectRepositoriesMap().get(project);
            if (riList != null) {
                for (RepositoryInfo ri : riList) {
                    Repository repo = getRepository(ri, false);

                    if (repo != null && repo.getCurrentVersion() != null && repo.getCurrentVersion().length() > 0) {
                        // getRepository() always creates fresh instance
                        // of the Repository object so there is no need
                        // to call setCurrentVersion() on it.
                        ri.setCurrentVersion(repo.determineCurrentVersion());
                    }
                }
            }
            suggester.refresh(projectName);
        }
```

This could be wrapped around by write lock (e.g. using the `Consumer` interface) - and making other changes - e.g. `getProject()` grabs the read lock which might be ok given the caller would be holding the write lock however there could be a case where lock upgrade is necessary and there might be a risk of deadlocks.

In the code above the project or a repository can be removed/changed while the code is running. There does not seem to be a risk of corruption (e.g. project-repository map is concurrent). In the worst case an exception will be thrown I think.

vladak:
@tulinkry feel free to run your test case on the latest bits in this PR.

tulinkry:
I tried my setup as in #2382 and I can go even after the first index. Everything is fined until step 5. where in step 4 I pull the configuration from the webapp, merge it with groups, and push it back to the webapp.

```
 -- PULL the config
> GET /api/v1/configuration HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 
< Content-Type: application/xml
< Transfer-Encoding: chunked
< Date: Tue, 23 Oct 2018 14:33:42 GMT
< 
{ [16271 bytes data]
-- PUSH the config
* Connected to localhost (::1) port 8080 (#0)
> PUT /api/v1/configuration HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/xml
> Content-Length: 13485
> Expect: 100-continue
> 
< HTTP/1.1 100 
* We are completely uploaded and fine
```

After this point the webapplication is hung (when you pres refresh, it is still loading indefinitely).

The process list has only this related:
`69853 ttys010    0:00.02 curl -v -X PUT --data @/var/folders/w4/fr4pd7zn0x1f9lwc8hsmhbwh0000gp/T/tmp.p50oBBtY --header Content-Type: application/xml http://localhost:8080/api/v1/configuration` and this process stays there.

vladak:
Obviously, the idea of wrapping the contents of `applyConfig()` under write lock did not fare well. The reason is that the PUT configuration eventually triggers `HistoryGuru#invalidateRepositories()` which needs some configuration properties however cannot get them since the getters use read lock and the write lock is now held. The deadlocked thread stacks then look like this:
```
 ffff80ffbf39332a lwp_cond_wait (4194ee8, 4194ed0, 0, 0)
 ffff80ffbe9db1a0 __1cGParkerEpark6Mbl_v_ () + 2a8
 ffff80ffbec3ad17 Unsafe_Park () + 1ef
 ffff80ffad477eea * *sun/misc/Unsafe.park(ZJ)V [compiled] 
 ffff80ffad58deec * java/util/concurrent/locks/AbstractQueuedSynchronizer.doAcquireShared(I)V+83 (line 1950)
 ffff80ffad4543ac * org/opengrok/indexer/configuration/RuntimeEnvironment.getSourceRootPath()Ljava/lang/String;+3 (line 818)
 ffff80ffaca08040 * org/opengrok/indexer/history/RepositoryInfo.getDirectoryName()Ljava/lang/String;+3 (line 286)
 ffff80ffaca08040 * org/opengrok/indexer/history/RepositoryFactory.getRepository(Lorg/opengrok/indexer/history/RepositoryInfo;Z)Lorg/opengrok/indexer/history/Repository;+5 (line 374)
 ffff80ffaca08040 * org/opengrok/indexer/history/HistoryGuru$3.run()V+8 (line 1896)
 ffff80ffaca08302 * java/util/concurrent/Executors$RunnableAdapter.call()Ljava/lang/Object;+4 (line 1022)
 ffff80ffaca08114 * java/util/concurrent/FutureTask.run()V+42 (line 520)
```
while the thread processing the PUT that owns the write lock is waiting for these threads to finish:
```
 ffff80ffbf39332a lwp_cond_wait (24b68b8, 24b68a0, 0, 0)
 ffff80ffbe9db1a0 __1cGParkerEpark6Mbl_v_ () + 2a8
 ffff80ffbec3ad17 Unsafe_Park () + 1ef
 ffff80ffad477eea * *sun/misc/Unsafe.park(ZJ)V [compiled]
 ffff80ffad58deec * java/util/concurrent/locks/AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(I)V+72 (line 2013)
 ffff80ffaca082bd * java/util/concurrent/locks/AbstractQueuedSynchronizer.acquireSharedInterruptibly(I)V+24 (line 2605)
 ffff80ffaca082bd * java/util/concurrent/CountDownLatch.await()V+5 (line 462)
 ffff80ffaca082bd * org/opengrok/indexer/history/HistoryGuru.invalidateRepositories(Ljava/util/Collection;Z)V+169 (line 1871)
 ffff80ffaca082bd * org/opengrok/indexer/configuration/RuntimeEnvironment.setConfiguration(Lorg/opengrok/indexer/configuration/Configuration;Ljava/util/List;Z)V+119 (line 3044)
 ffff80ffaca082bd * org/opengrok/indexer/configuration/RuntimeEnvironment.setConfiguration(Lorg/opengrok/indexer/configuration/Configuration;Z)V+4 (line 2994)
 ffff80ffaca082bd * org/opengrok/indexer/configuration/RuntimeEnvironment.applyConfig(Lorg/opengrok/indexer/configuration/Configuration;ZZ)V+13 (line 3266)
 ffff80ffaca082bd * org/opengrok/indexer/configuration/RuntimeEnvironment.applyConfig(Ljava/lang/String;ZZ)V+41 (line 3228)
 ffff80ffaca082bd * org/opengrok/web/api/v1/controller/ConfigurationController.set(Ljava/lang/String;Z)V+15
 ffff80ffaca007a7 * org/opengrok/web/api/v1/controller/ConfigurationController.set(Ljava/lang/String;Z)V+-7880
```

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1625,44 +1620,34 @@ public void applyConfig(String configuration, boolean reindex, boolean interacti
...
+        }
 
-            // start/stop the watchdog if necessary
-            if (isAuthorizationWatchdog() && config.getPluginDirectory() != null) {
-                watchDog.start(new File(config.getPluginDirectory()));
-            } else {
-                watchDog.stop();
-            }
+        // start/stop the watchdog if necessary
+        if (isAuthorizationWatchdog() && config.getPluginDirectory() != null) {
```
tulinkry: `config.getPluginDirectory()` this is ok without locking?

tulinkry: also there are more calls like this here (getPluginStack, messageLimit)

tulinkry:
I run my setup now for a 400 times and the only issue is the one described in #2445 which can be a part of a next PR.

So once you include at least a simple test case as in #2382 we're ready to go.

vladak:
You mean to write a unit test ?

Dne čt 25. 10. 2018 13:29 uživatel Kryštof Tulinger <
notifications@github.com> napsal:

tulinkry:
Yes. That one which discovered the misbehaviour.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConfigurationControllerTest.java
```diff
@@ -236,4 +241,50 @@ public void testSuggesterServiceNotifiedOnConfigurationChange() {
...
+        executor.shutdown();
+
+        // Check thread's view of the variable.
+        for (int j = 0; j < values.length; j++) {
+            assertEquals(newValue, values[j]);
+        }
+
+        // Revert back to default.
+        env.setHitsPerPage(origValue);
+    }
```
tulinkry: If you try it without the fix, does this test usually fail or?

vladak: I think it would need to call `register()` so that each thread has its own distinct configuration snapshot.

vladak: Using `PageConfig` should do the trick, let me try.

tulinkry: (it fails if you remove the hotfix for it)

vladak: changed the test case so that it is simpler and actually fails without the fix in place.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: this is actually #2445, lets make it in another pr

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: nice

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: answer this and i think we're ready to go

vladak: the configuration is already set so using locked versions is fine and actually necessary.

PR #2426, opened 2018-10-18
use /bin/sh instead of /bin/false to make sure exit value is 1

vladak:
As @gco mentioned, `/bin/false` returns 255 on Solaris (unlike 1 on Linux etc.) which is perfectly fine w.r.t. POSIX which says: "The false utility shall always exit with a value other than zero." So the Python tests that rely on particular return value need to use something else.

### DIFF
--- a/opengrok-tools/src/test/python/test_commands.py
+++ b/opengrok-tools/src/test/python/test_commands.py
@@ -41,45 +41,45 @@ class TestApp(unittest.TestCase):
                         [{"command": ['foo']}, {"command": ["bar"]}]))
         self.assertEqual("opengrok-master", str(cmds))
 
-    @unittest.skipUnless(os.path.exists('/bin/true') and os.path.exists('/bin/false'), "requires Unix")
+    @unittest.skipUnless(os.path.exists('/bin/true') and os.path.exists('/bin/cp'), "requires Unix")
     def test_run_retcodes(self):
         cmds = Commands(CommandsBase("opengrok-master",
                         [{"command": ["/bin/echo"]},
                          {"command": ["/bin/true"]},
-                         {"command": ["/bin/false"]}]))
+                         {"command": ["/bin/cp"]}]))
         cmds.run()
         # print(p.retcodes)
         self.assertEqual({'/bin/echo opengrok-master': 0,
                           '/bin/true opengrok-master': 0,
-                          '/bin/false opengrok-master': 1}, cmds.retcodes)
+                          '/bin/cp opengrok-master': 1}, cmds.retcodes)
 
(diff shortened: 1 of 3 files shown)


opengrok-tools/src/test/python/test_commands.py
```diff
@@ -41,45 +41,45 @@ def test_str(self):
                         [{"command": ['foo']}, {"command": ["bar"]}]))
         self.assertEqual("opengrok-master", str(cmds))
 
-    @unittest.skipUnless(os.path.exists('/bin/true') and os.path.exists('/bin/false'), "requires Unix")
+    @unittest.skipUnless(os.path.exists('/bin/true') and os.path.exists('/bin/cp'), "requires Unix")
```
tulinkry: cp is not used here

opengrok-tools/src/test/python/test_commands.py
```diff
@@ -41,45 +41,45 @@ def test_str(self):
...
+                         {"command": ["/bin/ls"]}]))
         cmds.run()
         # print(p.retcodes)
         self.assertEqual({'/bin/echo opengrok-master': 0,
                           '/bin/true opengrok-master': 0,
-                          '/bin/false opengrok-master': 1}, cmds.retcodes)
+                          '/bin/ls opengrok-master': 2}, cmds.retcodes)
 
-    @unittest.skipUnless(os.path.exists('/usr/bin/true') and os.path.exists('/usr/bin/false'), "requires Unix")
+    @unittest.skipUnless(os.path.exists('/bin/true') and os.path.exists('/bin/ls') and os.path.exists('/bin/echo'), "requires Unix")
```
tulinkry: it had a reason there was a /usr. this way it worked for me on mac.

tulinkry: if you remove it then it runs just twice the same test

vladak: on highSierra I see:
```
[
bash
cat
chmod
cp
csh
date
dd
df
domainname
echo
ed
expr
hostname
kill
ksh
launchctl
link
ln
ls
mkdir
mv
pax
ps
pwd
rm
rmdir
sh
sleep
stty
sync
tcsh
test
unlink
wait4path
zsh
```

tulinkry: yes exactly, there's no `false` nor `true` in it. I wanted to run the test also on my machine and they're in `usr/bin`

vladak: The tests need to use common denominator, duplication is wrong.

vladak:
Also, I had to remove `opengrok_tools` from the `sys.path.insert` calls in the test scripts so that `mvn verify` when run from the `opengrol-tools` directory works again.

PR #2429, opened 2018-10-18
Opengrok tools mvn

tulinkry:
Generating the output under the `target` directory. That also makes it clean when running `mvn clean`.

NOTE: when running `python3 setup.py sdist` the output won't go to `target`. This is just for maven builds.

### DIFF
--- a/distribution/assembly.xml
+++ b/distribution/assembly.xml
@@ -20,7 +20,7 @@
         <destName>source.war</destName>
       </file>
       <file>
-        <source>${project.basedir}/../opengrok-tools/dist/opengrok-tools-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}${parsedVersion.qualifier}.tar.gz</source>
+        <source>${project.basedir}/../opengrok-tools/target/dist/opengrok-tools-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}${parsedVersion.qualifier}.tar.gz</source>
         <outputDirectory>tools</outputDirectory>
         <destName>opengrok-tools.tar.gz</destName>
       </file>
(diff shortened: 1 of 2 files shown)


opengrok-tools/pom.xml
```diff
@@ -86,54 +130,37 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
...
                         <configuration>
                             <executable>env/bin/python</executable>
-                            <workingDirectory>${project.build.directory}/../</workingDirectory>
+                            <workingDirectory>${project.build.directory}</workingDirectory>
                             <arguments>
                                 <argument>setup.py</argument>
+                                <argument>install</argument>
                                 <argument>test</argument>
                             </arguments>
+                            <skip>${skipTests}</skip>
```
vladak: you will need to merge with yourself here I believe.

distribution/assembly.xml
```diff
@@ -20,7 +20,7 @@
         <destName>source.war</destName>
       </file>
       <file>
-        <source>${project.basedir}/../opengrok-tools/dist/opengrok-tools-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}${parsedVersion.qualifier}.tar.gz</source>
+        <source>${project.basedir}/../opengrok-tools/target/dist/opengrok-tools-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}${parsedVersion.qualifier}.tar.gz</source>
```
vladak: is there perhaps a Maven define that would expand to `target` ?

tulinkry: `${project.build.directory}` I think

PR #2431, opened 2018-10-19
Reading the python version from pom.xml

### LINKED ISSUE #2393: opengrok-tools version in setup.py needs to be updated automatically or have its own versioning scheme
opengrok-tools python package contains a hard-coded version

```python
setup(
    name='opengrok-tools',
    version='1.1rc58',
    packages=[ ...
)
```

We should either
1. run some maven execution to synchronize it with the current version in maven
2. introduce new versioning scheme (for example starting from 0.0.1) and increasing always when there is a change in the package

Let's do some discussion in this thread

tulinkry:
Use maven native plugins to read and transform
the module version and place it into the setup.py.

fixes #2393

When releasing 1.1 directly, I'd still verify it by hand if everything goes correctly, but there should not be a problem with the transition to the direct release.

### DIFF
--- a/distribution/assembly.xml
+++ b/distribution/assembly.xml
@@ -20,9 +20,9 @@
         <destName>source.war</destName>
       </file>
       <file>
-        <source>${project.basedir}/../opengrok-tools/${project.build.directory}/dist/opengrok-tools-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}${parsedVersion.qualifier}.tar.gz</source>
+        <source>${project.basedir}/../opengrok-tools/${project.build.directory}/dist/opengrok-tools-${project.python.package.version}.tar.gz</source>
         <outputDirectory>tools</outputDirectory>
-        <destName>opengrok-tools.tar.gz</destName>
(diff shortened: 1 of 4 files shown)


tulinkry:
Also in the distribution the package is now in `tools/opengrok-tools-<version>.tar.gz`. If this isn't appropriate (we want to skip the version), let me know.

vladak:
This will actually make the release process more complicated since it would require to run `mvn package`. Is it somehow possible to trigger the `setup.py` substitution via separate goal ?

tulinkry:
mvn package is run at the travis, isn't it? So this becomes part of the release package on the travis side

tulinkry:
I understood that you bumb the version with maven locally, then make a tag, upload it (with new versions) and then travis calls mvn package which builds the correct package.

vladak:
That's right. This is good enough then.

tulinkry:
Let's try it with the next rc release

tulinkry:
>  Is it somehow possible to trigger the setup.py substitution via separate goal or ideally make it dependent on versions:set

This is actually what this change does. It takes it from pom.xml where the version is bumped with versions:set.

PR #2434, opened 2018-10-20
deploy.py: repackage WAR files using standard zipfile module

edigaryev:
Remove dependency on jar/zip/unzip executables and simplify code a bit.


edigaryev:
>I don't have any background on this, no install require dependency in setup.py needed?

Yes, it's in the Python's standard library:

https://docs.python.org/3/library/zipfile.html

vladak:
Nice, thanks !

PR #2435, opened 2018-10-22
Using weak tag instead of just last modification date

### LINKED ISSUE #2405: opengrok main page becomes outdated because of caching
I figured out today that the main page (or other pages also maybe) send `Cache-Control: max-age=1800` which can make it

1. outdated for some time after the reindex
2. break the cookies (user search for projects, clicks on home, on homepage there is the cached set of cookie projects)
3. display messages that are no longer present in the system

only hitting full refresh (ctrl+shift+R) can fetch the updated version.

This must have been introduced to opengrok lately.

tulinkry:
This enables us to reflect all other factors
   which contribute to the resource freshness
 - the last file timestamp
 - messages on the project
 - the timestamp of last index
 - the opengrok version
We assemble all these to the weak tag,
   and the resource is reloaded only in any of these change

OpenGrok version is added to the weak tag attributes because of utils.js. Until now event when we increase the version of utils.js, the cached xref still pointed to the old version causing some drama with javascript. This way with every new OpenGrok version, each xref will be reloaded (no need for ctrl+shift+R).

fixes #2405

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
@@ -41,6 +41,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
@@ -51,8 +52,8 @@ import java.util.stream.Collectors;
 import javax.servlet.ServletRequest;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
-import org.suigeneris.jrcs.diff.Diff;
-import org.suigeneris.jrcs.diff.DifferentiationFailedException;
+import javax.servlet.http.HttpServletResponse;
+import org.opengrok.indexer.Info;
 import org.opengrok.indexer.analysis.AnalyzerGuru;
 import org.opengrok.indexer.analysis.ExpandTabsReader;
 import org.opengrok.indexer.analysis.FileAnalyzer.Genre;
@@ -70,6 +71,8 @@ import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.search.QueryBuilder;
 import org.opengrok.indexer.util.IOUtils;
 import org.opengrok.indexer.web.messages.MessagesContainer.AcceptedMessage;
+import org.suigeneris.jrcs.diff.Diff;
+import org.suigeneris.jrcs.diff.DifferentiationFailedException;
 
 /**
  * A simple container to lazy initialize common vars wrt. a single request. It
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -1619,4 +1622,73 @@ public void checkSourceRootExistence() throws IOException {
...
+     * <p>
+     * If the resource was modified, appropriate headers in the response are filled.
+     *
+     * @param response the http response for setting the headers
+     * @return true if resource was not modified; false otherwise
+     */
+    public boolean isNotModified(HttpServletRequest request, HttpServletResponse response) {
+        String currentEtag = String.format("W/\"%s\"",
+                Objects.hash(
+                        // last modified time as utc timestamp in millis
```
vladak: UTC

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -1619,4 +1622,73 @@ public void checkSourceRootExistence() throws IOException {
...
+
+    /**
+     * Get all project related messages. These include
+     * <ol>
+     * <li>Main messages</li>
+     * <li>Messages with tag = project name</li>
+     * <li>Messages with tag = project's groups names</li>
+     * </ol>
+     *
+     * @return the sorted set of messages according the accept time
```
vladak: according to

opengrok-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -56,11 +56,8 @@
     <filter-mapping>
         <filter-name>ExpiresHalfHourFilter</filter-name>
         <url-pattern>/history/*</url-pattern>
```
vladak: why not history ?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -1619,4 +1622,73 @@ public void checkSourceRootExistence() throws IOException {
...
+     * </ul>
+     * <p>
+     * <p>
+     * If the resource was modified, appropriate headers in the response are filled.
+     *
+     * @param response the http response for setting the headers
+     * @return true if resource was not modified; false otherwise
+     */
+    public boolean isNotModified(HttpServletRequest request, HttpServletResponse response) {
+        String currentEtag = String.format("W/\"%s\"",
```
vladak: is the value specified by any standard ?

opengrok-web/src/main/webapp/WEB-INF/web.xml
tulinkry: history does not include mast.jsp so this does not affect it anyway

vladak: Does this mean that history page will not display any messages ?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
tulinkry: yes
https://tools.ietf.org/html/rfc7232#section-2.3
https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html

the computed value itself of course not

vladak: It would be good to reference the standards in the comments.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -1619,4 +1622,73 @@ public void checkSourceRootExistence() throws IOException {
...
+                        // all project related messages which changes the view
+                        getProjectMessages(),
+                        // last timestamp value
+                        getEnv().getDateForLastIndexRun().getTime(),
+                        // OpenGrok version has changed since the last time
+                        Info.getVersion()
+                )
+        );
+
+        String headerEtag = request.getHeader("If-None-Match");
```
vladak: Is the literal value of the header name perhaps available in a pre-existing class ?

tulinkry: yes

opengrok-web/src/main/webapp/WEB-INF/web.xml
tulinkry: No. It means it will work like now, with sometimes inconsistent view as mentioned in #2405 because of the 30 minutes caching interval. This is also a subject to change if desired, but in different PR.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -1619,4 +1623,75 @@ public void checkSourceRootExistence() throws IOException {
...
+        String headerEtag = request.getHeader(HttpHeaders.IF_NONE_MATCH);
+
+        if (headerEtag != null && headerEtag.equals(currentEtag)) {
+            // weak ETag has not changed, return 304 NOT MODIFIED
+            response.setStatus(HttpServletResponse.SC_NOT_MODIFIED);
+            return true;
+        }
+
+        // return 200 OK
+        response.setHeader("ETag", currentEtag);
```
vladak: ditto

PR #2441, opened 2018-10-24
enable Pylint

vladak:


vladak:
will try.

dev/install-python-packages.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pip3 install --upgrade pip
+pip3 install -r opengrok-tools/src/main/python/requirements.txt
```
tulinkry: this should not be needed, if you install opengrok-tools with pip the requirements are donwloaded automatically

vladak: the opengrok-tools package is not installed during build I think

tulinkry: It is installed when tests phase is on. So why do you need to install its requirements?

vladak: because pylint needs the modules and cannot find them

tulinkry: aha, ok then

vladak: Also, I'd like to be able to check the sources without installing the package. After all, the checks precede the install in the pom file.

vladak:
This will have to be fixed for Windows environment too I believe, will be subject of separate PR.

vladak:
The pylint check for the re-throw action had to be disabled. (https://stackoverflow.com/questions/2228790/why-does-pylint-give-error-e0702-raising-nonetype-on-this-raise-statement)

dev/install-python-packages.sh
vladak: Actually, running pylint from venv is reasonable because it does not clash with system installed site packages.

PR #2463, opened 2018-10-29
modify the opengrok-tools env paths on windows

tulinkry:
Refactor of the property naming, verifying that the build + tests pass correctly on windows build. Ensure python3.4 in appveyor.

When user wants to use his custom version of python, he can use
```bash
mvn -Dpython.system.executable=/opt/python50/bin/python test
```
it will create the virtul env with this base python.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ install:
         [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\uctags-bin.zip", "C:\uctags")
       }
 build_script:
-    - mvn -B -V -Dorg.opengrok.indexer.analysis.Ctags=c:\uctags\ctags.exe clean verify
+    - mvn -B -V -Dorg.opengrok.indexer.analysis.Ctags=c:\uctags\ctags.exe -DskipTests clean verify
 cache:
   - C:\maven\ -> appveyor.yml
   - C:\Users\appveyor\.m2\ -> pom.xml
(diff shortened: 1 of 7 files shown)


appveyor.yml
```diff
@@ -27,7 +27,7 @@ install:
         [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\uctags-bin.zip", "C:\uctags")
       }
 build_script:
-    - mvn -B -V -Dorg.opengrok.indexer.analysis.Ctags=c:\uctags\ctags.exe clean verify
+    - mvn -B -V -Dorg.opengrok.indexer.analysis.Ctags=c:\uctags\ctags.exe -DskipTests clean verify
```
vladak: We need to run the tests although number of them currently fail.

vladak:
How to solve this problem w.r.t. distribution ?

tulinkry:
This is not ready. I’m just using the windows build to see if it pass.

29. 10. 2018 v 9:37, Vladimir Kotal <notifications@github.com>:

tulinkry:
And I don’t think there’s a problem with distribution. If you install the tools tarball with python 3, it should all work out of the box.



29. 10. 2018 v 9:37, Vladimir Kotal <notifications@github.com>:

appveyor.yml
tulinkry: sure

dev/install-python-packages.sh
```diff
@@ -2,7 +2,7 @@
 
 python3 -m pip install --upgrade pip
 
-python3 -m pip install pep8 flake8 virtualenv
+python3 -m pip install pep8 virtualenv
 if [[ $? != 0 ]]; then
```
tulinkry: it is installed in opengrok-tools pom.xml

tulinkry:
Can you review now? I did some updates. Also the description is updated.

vladak:
Nice work !

PR #2465, opened 2018-10-29
swallow FileAlreadyExistsException in doLink()

vladak:
The thread that lost the race needs to move on.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -26,6 +26,7 @@ package org.opengrok.indexer.index;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.DirectoryStream;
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
```diff
@@ -372,11 +373,16 @@ private void doLink(PendingSymlinkageExec lnk) throws IOException {
                     sourceParentFile.exists()) {
                 Files.createSymbolicLink(sourcePath, Paths.get(lnk.targetRel));
             }
+        } catch (FileAlreadyExistsException e) {
```
tulinkry: is the double check for `exists` neccessary now?

vladak: yes, that's for the parent directory :-)

tulinkry: aha, ok.

PR #2467, opened 2018-10-29
add better console logger

vladak:
With the changes:
```
$ opengrok-projadm -a foo
Adding project foo
could not add project foo for web application on http://localhost:8080/source
could not get configuration from web application on http://localhost:8080/source
$ opengrok-projadm -a foo >/dev/null
could not add project foo for web application on http://localhost:8080/source
could not get configuration from web application on http://localhost:8080/source
$ opengrok-projadm -a foo 2>/dev/null
Adding project foo
```

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/log.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/log.py
@@ -28,7 +28,7 @@ import sys
 def get_console_logger(name, level=logging.INFO, format='%(message)s'):
     """
     Get logger that logs logging.ERROR and higher to stderr, the rest
-    to stdout.
+    to stdout. For logging.DEBUG level more verbose format is used.
 
     :param name: name of the logger
     :param level: base logging level
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/log.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/log.py
@@ -49,15 +49,20 @@ class MaxLogLevelFilter(logging.Filter):
         return False
 
 
-def get_console_logger(name, level=logging.INFO):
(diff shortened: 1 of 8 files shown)


opengrok-tools/src/main/python/opengrok_tools/utils/log.py
```diff
@@ -0,0 +1,79 @@
...
+
+def get_console_logger(name, level=logging.INFO):
+    """
+    Get logger that logs logging.ERROR and higher to stderr, the rest
+    to stdout.
+    :param level: base logging level
+    :param name: name of the logger
+    :return: logger
+    """
+    formatter = logging.Formatter('%(message)s')
```
tulinkry: this will change also all other log messages to be in this format (in debug mode - for urllib and so on), is it intended?

vladak: With `propagate = False` it should not affect any other logger, including the root one.

opengrok-tools/src/main/python/opengrok_tools/utils/log.py
```diff
@@ -0,0 +1,85 @@
...
+
+#
+# Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+#
+
+import logging
+import sys
+
+
+class MinLogLevelFilter(logging.Filter):
```
tehe: Please see https://docs.python.org/3/library/logging.html#filter-objects:

> Changed in version 3.2: You don’t need to create specialized Filter classes, or use other classes with a filter
> method: you can use a function (or other callable) as a filter. 

So maybe you don't need these classes at all?

vladak: good point. works fine with lambdas.

PR #2469, opened 2018-10-30
upgrade requests to 2.20.0

vladak:
looks like this is needed to get the latest secure version.

### DIFF
--- a/opengrok-tools/requirements.txt
+++ b/opengrok-tools/requirements.txt
@@ -6,7 +6,7 @@ jsonschema==2.6.0
 JsonSir==0.0.2
 Python-EasyConfig==0.1.7
 PyYAML==3.13
-requests~=2.20.0
+requests==2.20.0
 Resource==0.2.1
 six==1.11.0
 urllib3==1.23
--- a/opengrok-tools/requirements.txt
+++ b/opengrok-tools/requirements.txt
@@ -6,7 +6,7 @@ jsonschema==2.6.0
 JsonSir==0.0.2
 Python-EasyConfig==0.1.7
 PyYAML==3.13
-requests==2.19.1
+requests~=2.20.0
 Resource==0.2.1
 six==1.11.0
 urllib3==1.23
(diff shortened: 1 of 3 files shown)


opengrok-tools/requirements.txt
```diff
@@ -6,7 +6,7 @@ jsonschema==2.6.0
 JsonSir==0.0.2
 Python-EasyConfig==0.1.7
 PyYAML==3.13
-requests==2.19.1
+requests~=2.20.0
```
tulinkry: if you need to force some version for the package, it must be in setup.py

this file is only a snapshot of the exact dev environment where this package was developed and not a file with requirements to install.

opengrok-tools/src/main/python/opengrok_tools/sync.py
```diff
@@ -95,7 +95,7 @@ def main():
                         help='URI of the webapp with context path')
     args = parser.parse_args()
 
-    if args.debug:
+    git if args.debug:
```
tulinkry: i guess windows has not been switched when you started typing ;-)

opengrok-tools/requirements.txt
vladak: Thing is the Github security vulnerability scanner interprets the `requirements.txt` file when emitting warnings.

opengrok-tools/src/main/python/opengrok_tools/sync.py
vladak: d'oh

opengrok-tools/requirements.txt
tulinkry: Ah. But then you shoud make it available also for users in setup.py, that's how you enforce a specific version range in the installation phase.

vladak: okay. It's strange to have this information stored in two places.

tulinkry: This requirements.txt is a generated file and the only purpose is to for some other developer to get the exact copy of your dev environment.

The source of truth is the setup.py.

vladak:
Also, had to fix the local version because of:
```
$ python3 setup.py install
/home/vkotal/opengrok-git-vladak/opengrok-tools/env/lib/python3.6/site-packages/setuptools/dist.py:406: UserWarning: The version specified ('0.0.1local') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
```

PR #2478, opened 2018-10-31
Actually skip history cache creation when history is disabled for project

edigaryev:
This typo/bug was introduced in https://github.com/oracle/opengrok/commit/2b892162bcf709338fb6776ed184615f9b9890c5.


vladak:
d'oh! clearly better unit test is needed.

Issue #2488, opened 2018-11-06
allow to specify log level on command line for tools


vladak:
It should be possible for the tools commands to specify log level on the command line. This would obsolete the `-D` option.

PR #2489, opened 2018-11-06
Tools logging redux

vladak:
After the changes in cset 846db8c5 to `CommandSequence` which replaced `logging.basicConfig` with `get_console_logger()` log records emitted in any worker were multiplied by the number of workers. This happened because `__init__()` therein was called for each worker in `sync.py` and each added new logging handlers (of the same types). The way logging in Python works is that each handler is called, thus leading to the excessive messaging. The first change address the problem by getting back to `getLogger()` (without `basicConfig()`).

However, this lead to the creation of new logger based on root logger, rather than child logger because the `__name__` in `CommandSequence` is `opengrok_tools.utils.commandsequence` which does not match the `__name__` of `sync.py` which is `opengrok_tools.sync`. As a consequence, the format used for log records would not match the one set by the `get_console_logger()`. The second change addresses this problem by basic the logger on `opengrok_sync`. Now, this still produces sort of unexpected results, e.g. `filelock.py` is 3rd party module that correctly calls `getLogger(__name__)` however also emits log records with `INFO` level so this will end up in the output which can be annoying especially when running from `cron`. The cure for now is to redirect `stdout` to `/dev/null`. The better solution is tracked by #2488.


opengrok-tools/src/main/python/opengrok_tools/utils/log.py
```diff
@@ -25,6 +25,10 @@
 import sys
 
 
+def get_class_basename():
+    return __name__.split('.')[0]
```
tulinkry: What if the name does not have a suffix?
Does it work with for example opengrok-indexer?

vladak: Then it will take the basename which is fine. `opengrok-indexer` will create new `opengrok_tools` logger just like `opengrok-sync` et al.

vladak: Actually all of the CLI tools have a suffix, e.g. if I insert `print(__name__)` in `indexer.py#main()`, it will report `opengrok_tools.indexer`.

tulinkry: aha so this is to get the `opengrok_tools`

PR #2495, opened 2018-11-08
Fix C++ parsing of hex/binary literals with digit separators

tjwalton:
We noticed that hexadecimal and binary literals in C++ with digit separators are parsed incorrectly in OpenGrok, so this patch fixes them by using the same idiom as decimal and octal integers.

I accept the Oracle Contributor Agreement for this patch.


vladak:
Thanks !

PR #2498, opened 2018-11-10
adding a test case for opengrok python scripts

tulinkry:
Quick sanity check before deployment.

### DIFF
--- a/opengrok-tools/pom.xml
+++ b/opengrok-tools/pom.xml
@@ -297,7 +297,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                                 <argument>setup.py</argument>
                                 <argument>test</argument>
                             </arguments>
-                            <skip>false</skip>
+                            <skip>${skipTests}</skip>
                         </configuration>
                         <phase>test</phase>
                         <goals>


opengrok-tools/src/test/python/test_binaries.py
```diff
@@ -0,0 +1,61 @@
...
+import unittest
+
+from parameterized import parameterized
+
+from opengrok_tools.utils.command import Command
+
+
+class TestApp(unittest.TestCase):
+    @parameterized.expand((
+            ('opengrok'),
```
vladak: ?

opengrok-tools/src/test/python/test_binaries.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_opengrok_binary(self, command):
+        """
+        Test that installed command is able to run
+        :param command: the command name
+        :return:
+        """
+        pass
+        cmd = Command([command, '--help'])
+        cmd.execute()
+        self.assertEqual(0, cmd.getretcode())
```
vladak: what happens if the program ends with exception ?

tulinkry: When I fix the PATH issue the test fails

```
  File "/Users/ktulinger/OpenGrok/opengrok-tools/target/.eggs/parameterized-0.6.1-py3.6.egg/parameterized/parameterized.py", line 392, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/Users/ktulinger/OpenGrok/opengrok-tools/target/src/test/python/test_binaries.py", line 55, in test_opengrok_binary
    self.assertEqual(0, cmd.getretcode())
AssertionError: 0 != 1
```

tulinkry:
Again it’s visible in few builds ago that the appveyor is successful

Issue #2085, opened 2018-04-26
paths.tsv data included inconsistently.


blu28:
using OpenGrok  1.1-rc25

When the paths.tsv data is included at the top of the browse repo page, following the word "xref:" it is inserted as is. But when it is inserted on the search results page, at the beginning of each repo header, it is processed to convert special HTML characters to their equivalent HTML entities. The processing should be the same in both places. I am trying to insert HTML code via the paths. tsv file, so I would prefer that the substitution not be done since that would prevent me for achieving my goal. The admin can always do the substitution themselves if it is a problem, but there is no way to insert HTML if the substitution is done automatically.

vladak:
I think using HTML for path descriptions was not part of the original idea. The display should be consistent, though.

blu28:
I am sure it wasn't in the original design, but why disallow it?

vladak:
Letting people breaking HTML in their instances by themselves is fine I guess :-)

PR #2500, opened 2018-11-10
Fix inconsistency in handling inclusion of paths.tsv data [issue 2085]

blu28:


tulinkry:
I’m out to try it - what kind of misbehavior does it fix?

vladak:
looks like a fix for #2085

PR #2501, opened 2018-11-10
Project repository map testing PR

tulinkry:


tulinkry:
@vladak let's have a look

tulinkry:
I wasn't able to synchronize it better then whole setConfiguration (the invokations depend on each other so multiple threads spoil the call)

wercker.yml
```diff
@@ -24,6 +24,7 @@ build:
       name: Install Python packages
       code: ./dev/install-python-packages.sh
     - wercker/maven@1.3.8:
+      debug: true
```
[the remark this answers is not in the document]
tulinkry: sure

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+        /*
+         * Now run setting a value in parallel which triggers configuration reload.
+         */
+        for (int i = 0; i < TASK_COUNT; i++) {
+            futures.add(threadPool.submit(() -> {
+                Response put = target("configuration")
+                        .path("projectsEnabled")
+                        .request()
+                        .put(Entity.text("true"));
```
vladak: maybe it should set the true/false property value randomly - in case someone implements smarter property setting (that would become no-op if the value is already set)

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
```diff
@@ -0,0 +1,188 @@
...
+        for (int i = 0; i < TASK_COUNT; i++) {
+            futures.add(threadPool.submit(() -> {
+                env.applyConfig(false, false);
+
+                assertTestedProjects();
+            }));
+        }
+
+        threadPool.shutdown();
+        threadPool.awaitTermination(100, TimeUnit.MINUTES);
```
vladak: that's too long I think. Travis will terminate the build run if there is no output in 30 minutes IARC.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
```diff
@@ -0,0 +1,188 @@
...
+    public void testConcurrentCInvalidateRepositories() throws InterruptedException, ExecutionException {
+        final ExecutorService threadPool = Executors.newFixedThreadPool(THREAD_COUNT);
+        final List<RepositoryInfo> repositoryInfos = env.getRepositories();
+        List<Future<?>> futures = new LinkedList<>();
+
+        /*
+         * Now run apply config in parallel
+         */
+        for (int i = 0; i < TASK_COUNT; i++) {
+            futures.add(threadPool.submit(() -> {
```
vladak: why do you actually need the futures ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -977,7 +977,7 @@ public void run() {
         repositories.clear();
         newrepos.forEach((_key, repo) -> { putRepository(repo); });
 
-        elapsed.report(LOGGER, "done invalidating repositories");
+        elapsed.report(LOGGER, String.format("done invalidating repositories (%d)", newrepos.size()));
```
vladak: make it evident that the count is number of repos

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1426,30 +1426,28 @@ public void signalTorefreshSearcherManagers(List<String> subFiles, String host)
...
-                Project proj;
-                String repoPath;
-                try {
-                    repoPath = PathUtils.getPathRelativeToSourceRoot(
-                            new File(r.getDirectoryName()), 0);
-                } catch (ForbiddenSymlinkException e) {
-                    LOGGER.log(Level.FINER, e.getMessage());
-                    continue;
-                }
+    public void generateProjectRepositoriesMap() throws IOException {
```
vladak: does this really need to be public ? (esp. now that the `repository_map` locking is removed)

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
```diff
@@ -0,0 +1,188 @@
...
+        repository.destroy();
+
+        env.setProjects(origProjects);
+        env.setRepositories(origRepositories);
+        env.setSourceRoot(origSourceRootPath);
+        env.setDataRoot(origDataRootPath);
+    }
+
+    @Test
+    @ConditionalRun(RepositoryInstalled.GitInstalled.class)
```
vladak: can you annotate the whole class as being Git dependent ? The `setUp()` copies the repositories (rather than cloning them which would be arguably better) however this would be of no use if the tests cannot be run.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
tulinkry: because when you trigger `get` it will rethrow all exceptions from the thread (i. e. the asserts)

vladak: I see. Can you add a comment there ?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
tulinkry: ok, but this should never happen as the calls terminates really fast

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
tulinkry: then there should be a different test, this one would become needleslly complex

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
tulinkry: ok

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
tulinkry: ok

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
tulinkry: yes

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: no, it does not

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ConcurrentConfigurationControllerTest.java
vladak: the assertions do not care about the property value.

vladak: perhaps use something like:
```java
Random rand = new Random(System.currentTimeMillis());
for (...) {
    ...put(rand.nextInt(2) == 1 ? Entity.text("true") : Entity.text("false")
}
```

tulinkry: Assertions do care, if you set enabledProjects = false, then you can't expect 20 projects in the next assertion.

vladak: I see. Let's leave it then.

PR #2504, opened 2018-11-11
Opengrok tools use pytest instead of unittest

tulinkry:
Consider using pytest as a default test library for our python tools.


opengrok-tools/pom.xml
```diff
@@ -309,7 +311,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                                 <argument>setup.py</argument>
                                 <argument>test</argument>
                             </arguments>
-                            <skip>${skipTests}</skip>
+                            <skip>false</skip>
```
vladak: why this change ?

tulinkry: ah, this is leftover

vladak:
Could you explain the rationale a bit ? The tests were duplicated because of a change to `setuptools` however this change requires certain version of that **and** switches to `pytest`. So, why do the latter ?

tulinkry:
The problem with duplicated test is fixed by either of it. It’s still necessary to have higher version of setuptools for pytest 3+.

Choosing pytest over embedded unittest is a matter of personal choice supported by wide usage of pytest in other python project where I inspired by myself. Although you can achieve basically the same thing - testing the code, pytest offers less verbose code and is prepared for interaction with mocking libraries like flexmock or betamax which we don’t use (yet).

To sum it up, it doesn’t bring more than unittest at the moment, except for colours.

tulinkry:
You can get more information about the differences for example here https://www.slant.co/versus/9148/9149/~unittest_vs_pytest

PR #2505, opened 2018-11-11
Fix casing issues

### LINKED ISSUE #1768: Turkish character search problem regarding to specific character I
OpenGrok search is case insensitive search which is a great feature. Due to one simple exception in Turkish characters upper lowercase rule searching can be painful.

When I want to search "date", I can search with both "date" and "DATE", and both match files containing "date" in any mixed case of letters. 

But,
When I want to search "time" or "TIME", these are not same in Turkish upper lower case rules.  So resultsets are completely different. which first matches *i** second matches *I**.

In Turkish 'I' -> 'ı' and 'İ' -> 'i'.  I am running in turkish locale to display turkish characters of comments properly.  In order to match both codes containing "time" word I have to search with both TIME and time and this problem leads to degradation of user experience.

Actually this is not a bug because system is running as expected. Is it possible to do some improvement on this specific issue?

idodeclare:
Hello,

Please consider for integration this patch to fix casing issues:

- Fix "README" matching for mixed-case. E.g. "Readme" files will now be formatted in OpenGrok.
- Review `to[Lower,Upper]Case` w.r.t. `Locale`. All reviewed cases except `Locale.English` in
Repository were not dependent on either US, English, or Default locale; and would be properly
specified as `ROOT`. Fixes #1768 for Turkish.
- Use `String` `toLowerCase()` and not `Character` `toLowerCase()` in `PathTokenizer` for context-sensitive and 1:M character mappings.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -866,8 +866,7 @@ public class AnalyzerGuru {
 
             // Try matching the prefix.
             if (dotpos > 0) {
-                factory
-                        = pre.get(path.substring(0, dotpos).toUpperCase(Locale.getDefault()));
+                factory = pre.get(path.substring(0, dotpos).toUpperCase(Locale.ROOT));
                 if (factory != null) {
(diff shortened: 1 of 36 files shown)


vladak:
Looks reasonable, thanks !

PR #2509, opened 2018-11-14
making protected constant again public

tulinkry:


vladak:
what is the reason for this change ?

tulinkry:
check #dev channel on slack

tulinkry:
the reason is that in https://github.com/oracle/opengrok/pull/2507/commits/76e97724efa6dc7474d43bc22d46854818f706a3 I accidentally changed the cookie name constant visibility

vladak:
I see, fails to compile the JSP now.

st 14. 11. 2018 9:07 odesílatel Kryštof Tulinger <notifications@github.com>
napsal:

PR #2511, opened 2018-11-14
add driveon option to sync.py

vladak:
This change makes possible running the whole sequence of commands even though some of them request the break using the exit code of 2.


opengrok-tools/src/test/python/test_commandSequence.py
```diff
@@ -77,6 +77,20 @@ def test_exit_2_handling():
     assert not cmds.failed
```
tulinkry: please use snake case for the filename as that is best practice for python

opengrok-tools/src/test/python/test_commandSequence.py
```diff
@@ -77,6 +77,20 @@ def test_exit_2_handling():
...
 
 
+@pytest.mark.skipif(not os.path.exists('/bin/sh') or not os.path.exists('/bin/echo'), reason="requires Unix")
+def test_driveon_flag():
+    cmds = CommandSequence(CommandSequenceBase("opengrok-master",
+                                               [{"command": ["/bin/sh", "-c",
+                                                             "echo " + CommandSequence.PROJECT_SUBST + "; exit 2"]},
+                                                {"command": ["/bin/echo"]}], driveon=True))
```
tulinkry: I'd vote for `skip_failed` or similar

opengrok-tools/src/test/python/test_commandSequence.py
```diff
@@ -77,6 +77,20 @@ def test_exit_2_handling():
     assert not cmds.failed
 
 
+@pytest.mark.skipif(not os.path.exists('/bin/sh') or not os.path.exists('/bin/echo'), reason="requires Unix")
+def test_driveon_flag():
+    cmds = CommandSequence(CommandSequenceBase("opengrok-master",
+                                               [{"command": ["/bin/sh", "-c",
+                                                             "echo " + CommandSequence.PROJECT_SUBST + "; exit 2"]},
```
tulinkry: is this 80 columns?

opengrok-tools/src/test/python/test_commandSequence.py
vladak: you mean rename `driveon` ? The skip failed would not be precise because in fact the cycle is terminated for failed commands (i.e. those that return non-zero but also non-2) if this option is True.

opengrok-tools/src/test/python/test_commandSequence.py
vladak: i.e. `test_command_sequence.py` ?

opengrok-tools/src/test/python/test_commandSequence.py
vladak: most likely not. the test files are still exempted from style checks I think.

vladak: The reason was originally path manipulation however this is gone so probably we can remove the excludes.

tulinkry:
Yes

16. 11. 2018 v 22:35, Vladimir Kotal <notifications@github.com>:

opengrok-tools/src/test/python/test_commandSequence.py
tulinkry: yes

opengrok-tools/src/test/python/test_commandSequence.py
tulinkry: yes

opengrok-tools/src/test/python/test_commandSequence.py
tulinkry: I don't see this in the code, I see that every error is ignored and the cycle is never terminated when this option = True that's why I'd be for the other name.

vladak: Ah, logic error. Should be fixed now.

tulinkry: ok

PR #2512, opened 2018-11-15
fix ClearCase history support for UNIX/Linux snapshot views

unhipzippo:
Simple tweak to ClearCase repository code to detect snapshot views on UNIX/Linux, in addition to its existing detection of snapshot views on Windows.


vladak:
Thanks !

Issue #2513, opened 2018-11-15
JDK9+ support


vladak:
This will need some workaround for the `coveralls-maven-plugin` as described in #2255.

vladak:
Once solved, this will make it possible to broaden the build matrix in Travis (`jdk: oraclejdk9` for Linux builds and bumping `xcode` version for macOS builds)

tulinkry:
How about to skip the coverall for jdk9?

vladak:
Well, I'm worried that it will be forgotten forever then. Maybe there is a
way how to compile patched version during the build of the plugin and use
that.

pá 16. 11. 2018 17:53 odesílatel Kryštof Tulinger <notifications@github.com>
napsal:

vladak:
Maybe you're right. It's not our problem that external dependency is broken. Maybe the Coveralls guys will realize that this is a problem as people will stop using the service.

tulinkry:
Personally, I'd drop it for newer jdk. It's not like our life depend on it so there is not a problem I think.

PR #2515, opened 2018-11-15
expand Travis jdk matrix with Oracle JDK 9 for Linux

vladak:
approaches #2513

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
 jobs:
   include:
     - stage: test
+      name: Linux + Oracle JDK 8
       os: linux
       dist: trusty
       sudo: required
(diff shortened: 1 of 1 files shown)


vladak:
Tried to enable Oracle JDK 11 however this failed miserably - https://travis-ci.org/vladak/OpenGrok/jobs/455513037

vladak:
Normally, one would specify the JDK versions using `jdk:` list so that the build matrix is expanded automatically however this does not work with build stages (limitation of Travis), hence the copy-n-paste-n-modify approach.

tulinkry:
> Tried to enable Oracle JDK 11 however this failed miserably - https://travis-ci.org/vladak/OpenGrok/jobs/455513037 <https://travis-ci.org/vladak/OpenGrok/jobs/455513037>

Isn’t is a problem of surefire? Maybe newer version? Just guessing because it doesn’t look like the problem is on our side.

vladak:
Actually, I am going to withdraw this because the JDK 9 Linux build failed due to Coveralls Maven plugin. It was passing in my fork build which was certainly strange.

PR #2517, opened 2018-11-16
distribute the dist readme

tulinkry:
distribute the distribution readme not the dev readme


vladak:
Thanks for fixing my change !

PR #2518, opened 2018-11-18
using group param in search url

tulinkry:
Using the "group" parameter in search url. It basically works that when user submits the form, the checkbox status is rewritten so that for a full group a "group" checkbox is checked and the projects in that group is "unchecked".


vladak:
Also, what happens if user manually selects all projects in a group ?

PR #2520, opened 2018-11-19
enable JDK 9

vladak:
try to enable JDK 9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,14 @@ jobs:
       script: dev/main
       before_install: dev/before_install
       before_script: dev/before
+    - stage: test
+      name: macOS + Oracle JDK 10
+      os: osx
+      osx_image: xcode9.4
+      install: true
+      script: dev/main
+      before_install: dev/before_install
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,8 +33,27 @@ jobs:
...
+    - stage: test
+      name: macOS + Oracle JDK 8
+      os: osx
+      osx_image: xcode9.3
+      install: true
+      script: dev/main
+      before_install: dev/before_install
+      before_script: dev/before
+    - stage: test
+      name: macOS + Oracle JDK 10
```
tulinkry: why not linux and jdk 10?

vladak: see https://docs.travis-ci.com/user/reference/trusty/#jvm-clojure-groovy-java-scala-images

vladak: for macOS, I wanted to see if JDK 10 works at all, it does not have to be there.

tulinkry: aha, ok not supported.

PR #2525, opened 2018-11-20
JDK 11

vladak:
The major part of the change is passing `--illegal-access=permit` to `java` for JDK 9+. The only way I could make this to work is to also introduce JDK 8 profile - having maven surefire plugin to be present in the `<plugins>` section and then also in a JDK9+ profile (with the argument line containing the option) does not work.

I had to bump the chronicle-map to the latest release otherwise suggester tests were failing in `init()`.

The macOS builds currently fail with:
```
[INFO] --- exec-maven-plugin:1.6.0:exec (Install python package for tests/pylint) @ opengrok-tools ---

Download error on https://pypi.org/simple/pytest-runner/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1051) -- Some packages may not be found!

Couldn't find index page for 'pytest-runner' (maybe misspelled?)
```
however given I did not modify the `opengrok-tools` module this should not be related to my changes and could well be transient (`master` builds do not currently fail).


dev/install-python-packages.sh
```diff
@@ -2,7 +2,7 @@
 
 python3 -m pip install --upgrade pip
 
-python3 -m pip install pep8 virtualenv
+python3 -m pip install pep8 virtualenv certifi
```
tulinkry: this is not needed if you install it in tools/pom.xml

opengrok-indexer/pom.xml
```diff
@@ -254,6 +254,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                         -->
                         <id>replace-in-jflex-sources</id>
                         <phase>generate-sources</phase>
+
```
tulinkry: extra newline?

opengrok-indexer/pom.xml
```diff
@@ -367,23 +409,60 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
...
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <excludes>
-                        <!-- Test helper class with name that confuses surefire -->
-                        <exclude>**/TestRepository.java</exclude>
-                    </excludes>
-                    <argLine>${surefireArgLine}</argLine>
-                </configuration>
-            </plugin>
-
```
tulinkry: you can leave this here and update just the argline in the profile

opengrok-web/pom.xml
```diff
@@ -136,18 +136,28 @@
...
                 </configuration>
             </plugin>
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <argLine>${surefireArgLine}</argLine>
-                </configuration>
-            </plugin>
```
tulinkry: again, this can stay here and you can update just the argline in the profile

plugins/pom.xml
```diff
@@ -73,18 +73,54 @@ Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
...
-          <plugin>
-              <groupId>org.apache.maven.plugins</groupId>
-              <artifactId>maven-surefire-plugin</artifactId>
-              <configuration>
-                  <systemPropertyVariables>
-                      <junit-force-all>true</junit-force-all>
-                  </systemPropertyVariables>
-                  <argLine>${surefireArgLine}</argLine>
-              </configuration>
-          </plugin>
```
tulinkry: the same as above

suggester/pom.xml
```diff
@@ -93,18 +93,28 @@
...
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <argLine>${surefireArgLine}</argLine>
-                </configuration>
-            </plugin>
```
tulinkry: the same as above

dev/install-python-packages.sh
vladak: Neither actually works. I will have to find out why `setup.py` fails to find the CA certificates. Maybe I will split the macOS changes into separate PR.

opengrok-indexer/pom.xml
vladak: yep

dev/install-python-packages.sh
vladak: actually, Travis now says the macOS builds pass which is weird since I saw them failing.

vladak: They definitely fail. Also, the macOS builds are broken since the recent JDK 9 push. Looks like this changed with more recent Python as we switched from xcode8.3 to 9.3.

opengrok-web/pom.xml
vladak: what do you mean exactly ?

vladak:
Also, jacoco has to be upgraded: https://github.com/jacoco/jacoco/issues/663

vladak:
Another gotcha is that JDK versions identified in the Maven profiles' `<activation>` elements - e.g. JDK 8 is `1.8`, JDK 11 is `18.9`.

vladak:
Also, `jacoco:report` has to be present on the Maven command line before the coveralls goal is invoked, otherwise the build fails with strange errors in Coveralls plugin processing - it will find that Jacoco reports line counts higher than the current length of some files (I think it was happening in the JavaScript generated lexer).

vladak:
A link a found handy: https://winterbe.com/posts/2018/08/29/migrate-maven-projects-to-java-11-jigsaw/

PR #2543, opened 2018-11-26
disable symlink test for Windows

vladak:


opengrok-indexer/src/test/java/org/opengrok/indexer/util/PathUtilsTest.java
```diff
@@ -77,6 +79,7 @@ public void shouldHandleEffectivelySameInputs() throws IOException {
     }
 
     @Test
+    @ConditionalRun(UnixPresent.class)
```
tulinkry: shouldn't there be also @Rule fiild in this class?

PR #2549, opened 2018-11-26
Close streams while extracting archive

Orviss:
While looking at the failing Windows tests @tarzanek found out that these streams are not closed properly.

Note: `assertNotNull(out);` was redundant since constructor will never return `null`

Thanks :)

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/util/FileUtilities.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/util/FileUtilities.java
@@ -53,11 +53,12 @@ public class FileUtilities {
             if (ze.isDirectory()) {
                 file.mkdirs();
             } else {
-                InputStream in = zipfile.getInputStream(ze);
-                assertNotNull(in);
-                FileOutputStream out = new FileOutputStream(file);
-                assertNotNull(out);
-                copyFile(in, out);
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/util/FileUtilities.java
```diff
@@ -53,11 +53,10 @@ public static void extractArchive(File sourceBundle, File root) throws IOExcepti
             if (ze.isDirectory()) {
                 file.mkdirs();
             } else {
-                InputStream in = zipfile.getInputStream(ze);
-                assertNotNull(in);
-                FileOutputStream out = new FileOutputStream(file);
-                assertNotNull(out);
-                copyFile(in, out);
+                try (InputStream in = zipfile.getInputStream(ze); FileOutputStream out = new FileOutputStream(file)) {
+                    assertNotNull(in);
```
tulinkry: Can’t we do better than assert? IOexception?

Orviss: We could. But I told myself that it's just for tests so why bother :D

Orviss: But honestly I did not like the assert as well :) I can change it when I get home.

PR #2561, opened 2018-11-29
Retry on ctags tagLine == null. Adjust ctags logging.

idodeclare:
Hello,

Please consider for integration this patch to eject a problematic ctags instance from the pool and to retry for the affected file.

Log messages are adjusted slightly for clarity.

Thank you.


vladak:
Looks good, thanks !

PR #2569, opened 2018-12-01
Use jetty-jspc. Fix opensearch.jsp exposed error.

idodeclare:
Hello,

Please consider for integration this patch to add a jetty-jspc step to affirm that the JSP files compile without error. One JSP, `opensearch.jsp`, had an error and is fixed.

* Set `mergeFragment=false` so pre-compiled JSP classes will not be included in the WAR.
* Also, use `skipTests` not `maven.test.skip` w.r.t. the short-circuiting the setting up of testdata.

Thank you.

### DIFF
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -400,7 +400,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                             <goal>run</goal>
                         </goals>
                         <configuration>
-                            <target unless="maven.test.skip">
+                            <target unless="skipTests">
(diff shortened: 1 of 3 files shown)


vladak:
Nice ! I always wanted to have this capability.

tulinkry:
Isn't there an issue about some compile jsp things?

idodeclare:
@tulinkry, none that I’m aware of.

I was longing for the compilation check that NetBeans did when you built the web project, and jetty-jspc seems the modern way to do it with maven.

tulinkry:
I think I meant this one https://github.com/oracle/opengrok/issues/1322 but that's sort of a different issue.

Issue #2404, opened 2018-10-10
ArrayIndexOutOfBoundsException in determineCurrentVersion for git


tulinkry:
I recently received an exception, which eventually hides one of my projects (because the repository information is missing)

```
09-Oct-2018 12:17:59.478 WARNING [invalidate-repos-18] org.opengrok.indexer.history.HistoryGuru$3.run Could not create git for '/var/opengrok/src/project
 java.lang.ArrayIndexOutOfBoundsException: 32
	at sun.util.calendar.BaseCalendar.getCalendarDateFromFixedDate(BaseCalendar.java:453)
	at java.util.GregorianCalendar.computeFields(GregorianCalendar.java:2397)
	at java.util.GregorianCalendar.computeFields(GregorianCalendar.java:2312)
	at java.util.Calendar.setTimeInMillis(Calendar.java:1804)
	at java.util.Calendar.setTime(Calendar.java:1770)
	at java.text.SimpleDateFormat.format(SimpleDateFormat.java:943)
	at java.text.SimpleDateFormat.format(SimpleDateFormat.java:936)
	at java.text.DateFormat.format(DateFormat.java:345)
	at org.opengrok.indexer.history.GitRepository.determineCurrentVersion(GitRepository.java:725)
	at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:150)
	at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:187)
	at org.opengrok.indexer.history.HistoryGuru$3.run(HistoryGuru.java:948)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

Last 10 git log records look like this:
```log
2018-08-31T15:47:34-04:00#202af7b Name Surname BUGFIX: Lorem Ipsum
2018-08-27T10:32:42-04:00#317495a Name Surname FEATURE: Lorem Ipsum
2018-08-20T09:54:42-04:00#ddf6817 Name Surname Merge "FEATURE: Lorem Ipsum"
[...]

PR #2570, opened 2018-12-02
refactoring date formats to be synchronized per repository

### LINKED ISSUE #2534: Repositories are missing on the main page when using parallel reindex
Continuing on #2445, the repository state is not consistent still.

The is some issue with the project repository map when it comes to concurrent index. My setup is following, 260 git repositories, doing parallel reindex with `opengrok-sync` and 8 workers.

All projects are marked as indexed (in the current webapp configuration), so this part looks ok. However, I still miss some of them on the main page, and the problem is somewhere around the project repository map. Please see comments https://github.com/oracle/opengrok/issues/2445#issuecomment-441071246 and https://github.com/oracle/opengrok/issues/2445#issuecomment-441074965 about the issue.

tulinkry:
fixes #2534


tulinkry:
Also fixes #2404

PR #2572, opened 2018-12-02
allowing wildcards in mirror-config

### LINKED ISSUE #2568: Allow wildcards in mirror-config.yml
Support wildcards in mirror-config.yml. I would like to write the config somewhat like this

```yaml
#
# Per project configuration.
#
projects:
  group-*:
    ignored_repos:
      - /group-*/testdata/*
```

or even better would be (if the repository root wouldn't be necessary):
```yaml
#
# Per project configuration.
#
projects:
  group-*:
    ignored_repos:
      - testdata/*
```

This is actually the opengrok structure and my system doesn't have all the repository commands, so I would like to ignore the test data for my runs.

tulinkry:
fixes #2568

This supports writing wildcards in mirror-config.yml and also uses the relative path to the matched repository:
```yml
#
# Per project configuration.
#
projects:
  group-*:
    ignored_repos:
      - testdata/*
```

=> meaning `group-*/testdata/*` repositories.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/mirror.py
@@ -30,6 +30,7 @@
 """
 
 import argparse
+import fnmatch
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


vladak:
Pls update the wiki.

tulinkry:
updated

PR #2573, opened 2018-12-03
temporarily disable testGetSuggestionsMultipleProjects

vladak:
The failure in `testGetSuggestionsMultipleProjects()` occurs from time to time, esp. when it is least convenient such as when creating a release. Thus, this change will disable the test until #2030 is solved.


opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SuggesterControllerTest.java
```diff
@@ -237,6 +237,8 @@ public void testGetSuggestionUnknownField() {
         assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), r.getStatus());
     }
 
+    // temporarily disabled, see https://github.com/oracle/opengrok/issues/2030
+    /*
```
tulinkry: Could you use @Ignore or other means provided be junit to skip the test rather then commenting it out?

PR #2575, opened 2018-12-03
fixes OpenGrok-2574 revisit maven patch after jflex update

tarzanek:

### DIFF
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -177,6 +177,25 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
 
     <build>
         <plugins>
+<!-- jflex:generate plugin HAS to be the first one which runs in generate-sources phase!!! -->
+            <plugin>
+                <groupId>de.jflex</groupId>
+                <artifactId>jflex-maven-plugin</artifactId>
+                <version>1.7.0</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>generate</goal>
+                        </goals>
+                        <configuration>
+                            <lexDefinitions>
+                                <lexDefinition>${basedir}/target/jflex-sources</lexDefinition>
+                            </lexDefinitions>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
             <plugin>
                 <groupId>pl.project13.maven</groupId>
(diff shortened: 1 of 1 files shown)


opengrok-indexer/pom.xml
```diff
@@ -259,7 +278,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                         </goals>
                         <configuration>
                           <filesToInclude>
-                              ${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/java/JavaSymbolTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/javascript/JavaScriptSymbolTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/plain/PlainFullTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/plain/PlainSymbolTokenizer.java
+                              ${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/java/JavaSymbolTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/javascript/JavaScriptSymbolTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/json/JsonSymbolTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/plain/PlainFullTokenizer.java,${basedir}/target/generated-sources/jflex/org/opengrok/indexer/analysis/plain/PlainSymbolTokenizer.java
```
vladak: does it perhaps accept the `**` notation ?

vladak:
what about the test case ?

opengrok-indexer/pom.xml
tarzanek: it could, I only added json so far, without more tests I hesitate to add more

tarzanek:
I will open a test case bug/PR soon, this could be merged

tarzanek:
the deal is that there is a test case, but it didn't work for negative test, I will fix it in next PR

opengrok-indexer/pom.xml
vladak: Fair enough.

vladak:
Done deal, will wait for the promised negative test case :-)

PR #2576, opened 2018-12-03
add opengrok version to the main page

tulinkry:
Show the version when one hovers over the logo at the bottom of the web page.

### DIFF
--- a/opengrok-web/src/main/webapp/foot.jspf
+++ b/opengrok-web/src/main/webapp/foot.jspf
@@ -23,7 +23,6 @@ Portions Copyright 2011 Jens Elkner.
 java.util.Date,
 java.util.EnumSet,
 
-org.apache.commons.lang3.StringUtils,
 org.opengrok.indexer.Info,
 org.opengrok.indexer.web.PageConfig,
 org.opengrok.indexer.web.Prefix"
--- a/opengrok-web/src/main/webapp/foot.jspf
+++ b/opengrok-web/src/main/webapp/foot.jspf
@@ -44,7 +44,7 @@ org.opengrok.indexer.web.Prefix"
         %>
     <div id="footer">
 <p><a href="http://opengrok.github.com/OpenGrok/"
- title="Served by OpenGrok (<%= Info.getVersion() %> - <%= StringUtils.left(Info.getRevision(), 7) %>)"><span id="fti"></span></a></p>
+ title="Served by OpenGrok (<%= Info.getVersion() %> - <%= Info.getShortRevision() %>)"><span id="fti"></span></a></p>
 <% if(dateForLastIndexRun != null) { %>
 <p>Last Index update <%= dateForLastIndexRun %></p>
 <%}%>
--- a/opengrok-web/src/main/webapp/foot.jspf
+++ b/opengrok-web/src/main/webapp/foot.jspf
@@ -20,12 +20,15 @@ Copyright (c) 2007, 2018, Oracle and/or its affiliates. All rights reserved.
 
 Portions Copyright 2011 Jens Elkner.
 --%><%@page import="
-java.util.EnumSet,
 java.util.Date,
+java.util.EnumSet,
 
+org.apache.commons.lang3.StringUtils,
+org.opengrok.indexer.Info,
 org.opengrok.indexer.web.PageConfig,
 org.opengrok.indexer.web.Prefix"
-%><%!
+%>
+<%!
 /* ---------------------- foot.jspf members start --------------------- */
     private static EnumSet<Prefix> needAddDiv = EnumSet.of(
         Prefix.HIST_L, Prefix.XREF_P, Prefix.DIFF_P, Prefix.MORE_P,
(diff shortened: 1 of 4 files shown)


opengrok-web/src/main/webapp/foot.jspf
```diff
@@ -41,7 +44,7 @@ org.opengrok.indexer.web.Prefix"
         %>
     <div id="footer">
 <p><a href="http://opengrok.github.com/OpenGrok/"
- title="Served by OpenGrok"><span id="fti"></span></a></p>
+ title="Served by OpenGrok (<%= Info.getVersion() %> - <%= StringUtils.left(Info.getRevision(), 7) %>)"><span id="fti"></span></a></p>
```
vladak: why is it necessary to hard code the 7 ?

tulinkry: Because the change set is too long so it is wrapped with a new line `3728a7212725abe56cfdf6fa94bdb20fa453c0e0` so I made it shorter to 7 chars (that is what github uses).

tulinkry: I didn't want to make this a constant as this information is not directly visible and I think not really worth it unless it is used on more places in the application.

vladak: Could you do something like this and call `getShortRevision()` instead ?

```java
diff --git a/opengrok-indexer/pom.xml b/opengrok-indexer/pom.xml
index 4e7705129..fc01de5bc 100644
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -183,7 +183,7 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                 <version>2.2.4</version>
                 <configuration>
                     <includeOnlyProperties>
-                        <includeOnlyProperty>^git.commit.id$</includeOnlyProperty>
+                        <includeOnlyProperty>^git.commit.id</includeOnlyProperty>
                     </includeOnlyProperties>
                 </configuration>
                 <executions>
@@ -238,6 +238,10 @@ Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
                                 ${project.build.outputDirectory}/org/opengrok/indexer/info.properties
                             </filesToInclude>
                             <replacements>
+                                <replacement>
+                                    <token>git.commit.id.abbrev</token>
+                                    <value>changeset_short</value>
+                                </replacement>
                                 <replacement>
                                     <token>git.commit.id</token>
                                     <value>changeset</value>
diff --git a/opengrok-indexer/src/main/java/org/opengrok/indexer/Info.java b/opengrok-indexer/src/main/java/org/opengrok/indexer/Info.java
index 3b3b22cc2..561616823 100644
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Info.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Info.java
@@ -37,6 +37,7 @@ public final class Info {
 
     private static final String VERSION;
     private static final String REVISION;
+    private static final String REVISION_SHORT;
 
     static {
         try (InputStream in = Info.class.getResourceAsStream("info.properties")) {
@@ -45,6 +46,7 @@ public final class Info {
             }
             VERSION = properties.getProperty("version", "unknown");
             REVISION = properties.getProperty("changeset", "unknown");
+            REVISION_SHORT = properties.getProperty("changeset_short", "unknown");
         } catch (IOException ioe) {
             throw new RuntimeException(ioe);
         }
@@ -59,13 +61,15 @@ public final class Info {
     }
 
     /**
-     * get full version (product vMajor revMinor)
+     * get full version (product vMajor revMinor and changeset ID)
      * @return full version
      */
     public static String getFullVersion() {
         return "OpenGrok v" + VERSION + " rev " + REVISION;
     }
 
+    public static String getShortRevision() { return REVISION_SHORT; }
+
     /**
      * get minor version
      * @return minor version
```

vladak: I think it is definitely more standard solution that requires no prior knowledge about the hash.

vladak: I mean if the SCM offers this then it makes sense to use it and avoids people wondering in the future why-the-h*** is the 7 doing there.

tulinkry: oukie doukie

tulinkry: My intention was to put at least something there before the release and then think about a better aproach how to display it later (probably in full version).

opengrok-web/src/main/webapp/foot.jspf
```diff
@@ -20,12 +20,15 @@ Copyright (c) 2007, 2018, Oracle and/or its affiliates. All rights reserved.
 
 Portions Copyright 2011 Jens Elkner.
 --%><%@page import="
-java.util.EnumSet,
 java.util.Date,
+java.util.EnumSet,
 
+org.apache.commons.lang3.StringUtils,
```
vladak: is this still needed ?

tulinkry: heh, no

PR #2582, opened 2018-12-08
Support ctags for historical revisions

idodeclare:
Hello,

Please consider for integration this patch to support running ctags for historical revisions.

- Add `webappCtags` configuration flag, with `--webappCtags` switch, to indicate if the webapp is eligible to run ctags.
- Add `Repository.getHistoryGet()` override to allow sub-classes to override to avoid a full in-memory version; and override for Git and Mercurial.
- Revise `BoundedBlockingObjectPool` as LIFO-to-FIFO so the webapp does not start extraneous
  instances; Indexer switches to FIFO performance when the queue pool is emptied.
- make `IndexerParallelizer` a lazy property of `RuntimeEnvironment`, and make the executors of `IndexerParallelizer` also lazy properties. Move the history-related executors into `IndexerParallelizer` so the lifecycle of all indexing/history executors are controlled in the same class.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -24,14 +24,10 @@
 package org.opengrok.indexer.history;
 
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.io.OutputStream;
 import java.io.Reader;
 import java.nio.file.Paths;
 import java.text.ParseException;
@@ -203,7 +199,7 @@ public class GitRepository extends Repository {
             Executor executor = new Executor(Arrays.asList(argv), directory,
                     RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
             int status = executor.exec();
-            result.iterations = transferExecutorOutput(sink, executor);
+            result.iterations = copyBytes(sink, executor.getOutputStream());
 
             /*
              * If exit value of the process was not 0 then the file did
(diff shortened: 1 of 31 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -171,14 +174,18 @@ Executor getRenamedFilesExecutor(final File file, String sinceRevision) throws I
...
+     * made to {@code sink}
      * @param fullpath full pathname of the file
      * @param rev revision
-     * @return contents of the file in revision rev
+     * @return {@code true} if any contents were found
      */
-    private InputStream getHistoryRev(String fullpath, String rev) {
-        InputStream ret = null;
-        File directory = new File(getDirectoryName());
+    private boolean getHistoryRev(BufferSink sink, int[] outIterations,
```
tulinkry: please dont use array for out param, think about something better, object reference or returning a pair or smth else

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexerParallelizer.java
```diff
@@ -49,64 +50,182 @@
...
+     * thread might call those methods simultaneously with this method.
+     * <p>
+     * The JVM will await any instantiated thread pools until they are
+     * explicitly shut down. A principle intention of this method is to
+     * facilitate OpenGrok test classes that run serially. The non-test
+     * processes using {@link IndexerParallelizer} -- i.e. {@code opengrok.jar}
+     * indexer or opengrok-web -- would only need a one-way shutdown; but they
+     * call this method satisfactorily too.
+     */
+    public void bounce() {
```
tulinkry: I still don't get why this is called bounce. I'd prefer clear or close.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
```diff
@@ -579,6 +580,31 @@ protected String getRepoRelativePath(final File file)
...
+                if (len > 0) {
+                    ++iterations;
+                    sink.write(buffer, 0, len);
+                }
+            }
+            return iterations;
+        }
+    }
+
+    class HistoryRevResult {
```
tulinkry: does it need to be non-static? It doesn't need any state from the enclosing class

idodeclare: Hmmm what does it matter for a class with no methods or initializers?

tulinkry: It contains unnecessary reference to the enclosing class

vladak:
Looks like after the recent commits this is good to go.

tarzanek:
so let's merge

PR #2584, opened 2018-12-10
add serialVersionUID to silence lint

vladak:
The serial version UID checks resurfaces in IDEA build so it is necessary to address them.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -57,7 +57,7 @@ public class GitRepository extends Repository {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GitRepository.class);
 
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -57,7 +57,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GitRepository.class);
 
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = -6126297612958508386L;
```
tulinkry: why is such an ugly number necessary?

vladak: This is what IDEA generated (which I think is based on the default Java algorithm).

tulinkry: ok

PR #2600, opened 2018-12-27
--noindex should suppress history cache generation

vladak:
This change modifies the behavior of `--noIndex` to encompass also the history cache generation. This way it is possible to change the index page before the initial reindex has complete not to display error about the configuration.


tarzanek:
I'd say merge, lgtm.

Fwiw - all those ifs need to be made simpler, hence my arguing about kotlin and getting to functional thinking and programming

PR #2605, opened 2018-12-31
Fix #2604, add TandemFilename and TandemPath

### LINKED ISSUE #2604: long filenames cause dirty index due to pending file completer terminating the indexing with failure
Same repository for 1.0 and 1.1
Problem with long file names:
```
2018-12-29 09:54:18.458+0300 WARNING t2876 PendingFileCompleter.doRename: Failed to move file: /og1.1/data/xref/MAIN/Test/Sunkey/Src/Autotests.ABC_DEF/Stage2/US3-US4/ABC_DEF_MART_На_ПО_нет_ABC_номера_Добавление_доп_номеров_Федеральные_Бесплатная_категория_номера_Печать_документов_Повторная_печать_1316603.cs.gz.org_opengrok -> /og1.1/data/xref/MAIN/Test/Sunkey/Src/Autotests.ABC_DEF/Stage2/US3-US4/ABC_DEF_MART_На_ПО_нет_ABC_номера_Добавление_доп_номеров_Федеральные_Бесплатная_категория_номера_Печать_документов_Повторная_печать_1316603.cs.gz
2018-12-29 09:54:19.399+0300 WARNING t2750 IndexDatabase.finishWriting: An error occurred while finishing writer and completer
java.io.IOException: 34 failures (0.0%) while renaming pending files
```

idodeclare:
Hello,

Please consider for integration this patch to ensure that OpenGrok tandem filenames (i.e. filenames that operate in tandem with source files; e.g. `.gz` xref files and `.org_opengrok` interim files) do not exceed 255 byte filenames.

This is backward-compatible with existing 1.1 `DATA_ROOT` outputs* because the new algorithm is only really effective for files that could not be created due to exceeding 255 byte names.

Also:
- Do case-insensitive comparison for `.gz` in `GZIPAnalyzer`.
- Add a `serialVersionUID` to silence lint.

Thank you.

*for filesystems with 255 byte name limits, such as Linux, Windows, and macOS.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
@@ -1,0 +1,173 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
+ */
+
+package org.opengrok.indexer.util;
+
+import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * Represents a utility class for creating a filename to operate in tandem with
+ * an original filename by adding a new file extension but limiting the length
+ * of the new filename to 255 UTF-8 encoded bytes if necessary by truncating
+ * and packing in a Base64-encoded SHA-256 hash of the original file name.
+ */
+public class TandemFilename {
+
+    private static final int MAX_BYTES = 255;
+
+    /**
+     * One fewer than {@link #MAX_BYTES} as a cap for simple concatenation to
+     * avoid the possibility of easily fabricating a collision against this
+     * algorithm. I.e., a 255 byte tandem filename will always include a
+     * computed hash and not just be the concatenation of original filename
+     * plus new extension.
+     */
+    private static final int MAX_CAT_BYTES = MAX_BYTES - 1;
+
+    /**
+     * "Instances of Base64.Encoder class are safe for use by multiple
+     * concurrent threads." --Oracle.
+     */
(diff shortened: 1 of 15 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
```diff
@@ -0,0 +1,173 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * Represents a utility class for creating a filename to operate in tandem with
+ * an original filename by adding a new file extension but limiting the length
+ * of the new filename to 255 UTF-8 encoded bytes if necessary by truncating
```
vladak: why is it necessary to leave the part of the filename in ? (as opposed to using just the digest)

opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
```diff
@@ -0,0 +1,173 @@
...
+ * and packing in a Base64-encoded SHA-256 hash of the original file name.
+ */
+public class TandemFilename {
+
+    private static final int MAX_BYTES = 255;
+
+    /**
+     * One fewer than {@link #MAX_BYTES} as a cap for simple concatenation to
```
vladak: could you expand on the collision ?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
idodeclare: I prefer to be able to correlate files in `xref/` using just my human brain, so I chose to stick with a straight-forward concatenation for filenames that encode as anything up to 254 bytes and for anything longer to pack the digest into the tail of the name and limit the total to 255 bytes.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
idodeclare: Trivial collision is prevented by disallowing simple concatenation to `MAX_BYTES`.

E.g.  in a simplified example, if `MAX_BYTES` were 16 and a base64 digest were five bytes long, then for a short name, `filename1`, a tandem filename for xref would be `filename1.gz`. For a long name, `filename0123456789`, a tandem filename for xref might be `filename5cdb0.gz`

If simple concatenation were allowed to 16 bytes, then one could drop in a file named `filename5cdb0`, and OpenGrok would concatenate for a tandem filename for xref as `filename5cdb0.gz` and collide with the long file above.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
vladak: Initially I thought you do not trust SHA-256 to produce unique digests :-) It certainly helps in tracking down which file is which.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/TandemFilename.java
vladak: I see, that makes sense; I like the robustness.

vladak:
Thanks, nice !

PR #2614, opened 2019-01-05
Bugfix/eiffel

idodeclare:
Hello,

Please consider for integration this patch with a fix for Eiffel to support its Special Characters and to limit its Character Constant.

Thank you.


vladak:
Thanks !

Issue #2046, opened 2018-03-20
get rid of json-simple


vladak:
json-simple is more of a proof of concept rather than serious library. I took a random look inside its code and discovered a comment saying that its parser is not thread safe. This sounds like a time bomb. Replacing it with gson would be better.

vladak:
Currently json-simple is used for:
  - `Statistics`
  - `JSONSearchServlet`
  - `egrok` Eclipse plugin
  - passing data from `Messages` to UI

vladak:
Also, this is the only jar file that is present in binary form in the OpenGrok repository, all the other dependencies are automatically downloaded.

vladak:
Given that Jersey uses json-jackson, it makes sense to use it for other JSON code as well.

vladak:
The binary is no longer part of the repo, was replaced by Maven dependency.

PR #2620, opened 2019-01-12
convert Statistics to use Jackson for JSON processing

vladak:
This change gets rid of the JSON simple library in favor of Jackson.

There were couple of notable decisions I had to do:
  - empty/initial Statistics will be encoded into full JSON (avoiding the special case)
  - the averages are recomputed with each request so that getters/setters can be clean
  - no JSON object passing around anymore - just strings
  - StatsController converts the Statistics to JSON implicitly

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
@@ -22,14 +22,12 @@
  */
 package org.opengrok.indexer.web;
 
+import com.fasterxml.jackson.core.*;
+import org.junit.*;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.util.function.Function;
-import org.json.simple.JSONObject;
-import org.junit.After;
-import org.junit.AfterClass;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.Test;
 
 /**
  *
@@ -203,29 +201,42 @@ public class StatisticsTest {
      * Test of toJson method, of class Statistics.
      */
     @Test
-    public void testToJsonOnInstance() {
-        checkToJson(new Function<Statistics, JSONObject>() {
+    public void testToJsonOnInstance() throws IOException {
+        checkToJson(new Function<Statistics, String>() {
             @Override
-            public JSONObject apply(Statistics t) {
-                return t.toJson();
+            public String apply(Statistics t) {
+                try {
+                    return t.toJson();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+                return null;
             }
         });
     }
 
     /**
-     * Test of from method, of class Statistics.
+     * Test of fromJson method, of class Statistics.
      */
     @Test
     @SuppressWarnings("unchecked")
-    public void testFrom() {
-        JSONObject input = new JSONObject();
-        input.put(Statistics.STATISTIC_MINUTES, 10L);
-        input.put(Statistics.STATISTIC_REQUESTS, 100L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE, 543L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_AVG, 54312.0);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MIN, 49L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MAX, 4753L);
-        Statistics stat = Statistics.from(input);
+    public void testFrom() throws IOException {
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        JsonFactory jfactory = new JsonFactory();
+        JsonGenerator jGenerator = jfactory
+                .createGenerator(stream, JsonEncoding.UTF8);
+
+        jGenerator.writeStartObject();
+        jGenerator.writeNumberField(Statistics.STATISTIC_MINUTES, 10L);
+        jGenerator.writeNumberField(Statistics.STATISTIC_REQUESTS, 100L);
+        jGenerator.writeNumberField(Statistics.STATISTIC_REQUESTS_PER_MINUTE, 543L);
+        jGenerator.writeNumberField(Statistics.STATISTIC_REQUESTS_PER_MINUTE_AVG, 54312.0);
+        jGenerator.writeNumberField(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MIN, 49L);
+        jGenerator.writeNumberField(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MAX, 4753L);
+        jGenerator.writeEndObject();
+        jGenerator.close();
+        String jsonString = new String(stream.toByteArray(), "UTF-8");
+        Statistics stat = Statistics.fromJson(jsonString);
 
         Assert.assertNotNull(stat);
         Assert.assertEquals(10L, stat.getMinutes());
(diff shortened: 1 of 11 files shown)


vladak:
This is another step towards completion of #2046.

pom.xml
```diff
@@ -66,6 +66,8 @@ Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
         <compileTarget>1.8</compileTarget>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jersey.version>2.27</jersey.version>
+        <!-- Jackson version needs to match the version of Jackson Jersey depends on otherwise weird things will happen. -->
+        <jackson.version>2.8.10</jackson.version>
```
tulinkry: maybe you can use exclude for deps like this to be sure whats happening.

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -240,34 +251,68 @@ public void testFrom() {
...
+    public void testToJsonOnClass() throws IOException {
+        checkToJson(new Function<Statistics, String>() {
             @Override
-            public JSONObject apply(Statistics t) {
-                return Statistics.toJson(t);
+            public String apply(Statistics t) {
+                try {
+                    return Statistics.toJson(t);
+                } catch (JsonProcessingException e) {
+                    e.printStackTrace();
```
tulinkry: assert fail?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -203,35 +201,48 @@ public void testAddRequestMultipleTimeSum() {
...
+    public void testToJsonOnInstance() throws IOException {
+        checkToJson(new Function<Statistics, String>() {
             @Override
-            public JSONObject apply(Statistics t) {
-                return t.toJson();
+            public String apply(Statistics t) {
+                try {
+                    return t.toJson();
+                } catch (IOException e) {
+                    e.printStackTrace();
```
tulinkry: assert fail?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -203,35 +201,48 @@ public void testAddRequestMultipleTimeSum() {
...
+                }
+                return null;
```
tulinkry: assert fail?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -240,34 +251,68 @@ public void testFrom() {
...
             @Override
-            public JSONObject apply(Statistics t) {
-                return Statistics.toJson(t);
+            public String apply(Statistics t) {
+                try {
+                    return Statistics.toJson(t);
+                } catch (JsonProcessingException e) {
+                    e.printStackTrace();
+                }
+                return null;
```
tulinkry: assert fail?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -203,35 +201,48 @@ public void testAddRequestMultipleTimeSum() {
...
-        input.put(Statistics.STATISTIC_REQUESTS, 100L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE, 543L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_AVG, 54312.0);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MIN, 49L);
-        input.put(Statistics.STATISTIC_REQUESTS_PER_MINUTE_MAX, 4753L);
-        Statistics stat = Statistics.from(input);
+    public void testFrom() throws IOException {
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        JsonFactory jfactory = new JsonFactory();
+        JsonGenerator jGenerator = jfactory
```
tulinkry: what's the point of using generator when you could just write simple json string as a string?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
```diff
@@ -240,34 +251,68 @@ public void testFrom() {
...
+        Assert.assertNotNull(getJsonLongField(result, Statistics.STATISTIC_MINUTES));
+        Assert.assertEquals(-325L, (long) getJsonLongField(result, Statistics.STATISTIC_MINUTES));
+        Assert.assertNotNull(getJsonLongField(result, Statistics.STATISTIC_REQUESTS));
+        Assert.assertEquals(145L, (long) getJsonLongField(result, Statistics.STATISTIC_REQUESTS));
+        Assert.assertNotNull(getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE));
+        Assert.assertEquals(1000L, (long) getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE));
+        Assert.assertNotNull(getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE_MIN));
+        Assert.assertEquals(107L, (long) getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE_MIN));
+        Assert.assertNotNull(getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE_MAX));
+        Assert.assertEquals(106L, (long) getJsonLongField(result, Statistics.STATISTIC_REQUESTS_PER_MINUTE_MAX));
```
tulinkry: is there any benefit in doing this complex stuff with getjsonLongField instead of using Statistics.fromJson and then comparing the values with expected values.

pom.xml
vladak: Not sure if you can use dynamic dependencies (version wildcard, in this case something like 2.9.*) in Maven. It would be better to enforce this positively - somehow extract the Jackson version Jersey depends on and make it explicit dependency.

I did not really want to have a implicit dependency however maybe that is more robust than this - the risk of Jersey to be removed is fairly low (and the `Statistics` tests would break anyway).

tulinkry: Yes ok. Maybe you could use some maven plugin to extract the version but it leads to complex solution to not-really-a-problem thing. So I would stick with this.

vladak: The same question was asked on https://stackoverflow.com/questions/23891067/maven-access-version-of-transitive-dependency-via-property however received no answer so far (4 years).

vladak: https://howtodoinjava.com/maven/maven-dependency-management/ spells out this problem nicely:
> Transitivity brings a very serious problem when different versions of the same artifacts are included by different dependencies. It may cause version mismatch issue in runtime.

It is a bit surprising that Maven (or rather the standard plugins) does not have a solution built in.

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
vladak: The advantage is that the other fields do not have to be specified - this also tests the parsing of not fully formed Statistics JSON string.

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
vladak: This sort of follows the idea done previously. I realized it can serve also as a test that will make sure the JSON serialization has a certain form.

vladak: I concur, this is more about structure, not whitespaces. That's why I think using generator is better than the test in `StatisticsUtilTest#testLoadStatistics()`.

vladak: On the other hand, using concrete JSON representation is closer to whitebox testing - e.g. testing idempotence can pass without making sure that any JSON processing actually happened which is arguably not a problem when calling a generator from JSON library :-)

opengrok-indexer/src/test/java/org/opengrok/indexer/web/StatisticsTest.java
vladak: I like to have to construct the JSON string programmatically, it is less error prone.

PR #2623, opened 2019-01-14
lucene 7.6

tarzanek:


vladak:
This will mean reindexing from scratch ?

tarzanek:
not really, it seems the format stayed the same
even when I did luke rebuild it worked, some features we don't use changed, otherwise it's pretty much a bug fix

tarzanek:
I will merge it later today/tomorrow since this is low risk

PR #2624, opened 2019-01-16
Fix #1999 : REFS aren't indexed if DEFS is empty

### LINKED ISSUE #1999: REFS aren't indexed if DEFS is empty
For e.g. a shell script with no aliases, functions, scripts, or heredocs — and thus no tags from ctags (I'm using Universal) — then REFS are not indexed either. This seems odd, but I am not sure if there is a historical motivation for it.

For the Sh example above, it is also counterintuitive because it left out expected results in a Symbol search while showing other files.

idodeclare:
Hello,

Please consider for integration this patch to index REFS independently from DEFS and to bump the affected analyzers' specialized version numbers.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/fortran/FortranAnalyzer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/fortran/FortranAnalyzer.java
@@ -40,7 +40,18 @@ public class FortranAnalyzer extends AbstractSourceCodeAnalyzer {
         super(factory, new JFlexTokenizer(new FortranSymbolTokenizer(
             FileAnalyzer.dummyReader)));
     }
-    
+
+    /**
+     * Gets a version number to be used to tag processed documents so that
+     * re-analysis can be re-done later if a stored version number is different
+     * from the current implementation.
+     * @return 20180208_00
+     */
+    @Override
+    protected int getSpecializedVersionNo() {
+        return 20180208_00; // Edit comment above too!
+    }
+
     /**
      * Creates a wrapped {@link FortranXref} instance.
      * @param reader the data to produce xref for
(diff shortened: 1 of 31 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/fortran/FortranAnalyzer.java
```diff
@@ -40,7 +40,18 @@
...
+     * Gets a version number to be used to tag processed documents so that
+     * re-analysis can be re-done later if a stored version number is different
+     * from the current implementation.
+     * @return 20180208_00
+     */
+    @Override
+    protected int getSpecializedVersionNo() {
+        return 20180208_00; // Edit comment above too!
+    }
+
```
tulinkry: 👍

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/plain/PlainAnalyzer.java
```diff
@@ -108,15 +108,18 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut)
...
                 byte[] tags = defs.serialize();
                 doc.add(new StoredField(QueryBuilder.TAGS, tags));                
             }
         }
+        /*
+         * This is to explicitly use appropriate analyzer's token stream to
+         * work around #1376: symbols search works like full text search.
+         */
+        OGKTextField ref = new OGKTextField(QueryBuilder.REFS,
+                this.symbolTokenizer);
```
tulinkry: the new line break is not necessary here I think but that's not important

tulinkry:
Looks ok.

idodeclare:
Thank you for the review, @tulinkry

tarzanek:
this looks OK, however I don't know if REFS have any value without DEFS in the file (yes, you should get existing symbols in index at least, if local file has none detected)

tarzanek:
so the question is whether this won't waste more cpu cycles than add value

tarzanek:
that said I am OK with merge

vladak:
Merged based on Slack #dev discussion.

PR #2627, opened 2019-01-17
fix some Javadoc warnings

vladak:
Travis JDK 11 build seems to be failing due to Javadoc warnings. This change fixes some of them.


vladak:
Had to do some workarounds due to https://github.com/jflex-de/jflex/issues/542

PR #2632, opened 2019-01-18
Replace RAMDirectory with ByteBuffersDirectory in tests

### LINKED ISSUE #2454: org.apache.lucene.store.RAMDirectory deprecated
Another issue uncovered by the changes for #2246:
```
suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java:[34,31] org.apache.lucene.store.RAMDirectory in org.apache.lucene.store has been deprecated
...
```
Also, https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/store/RAMDirectory.html says:

> Warning: This class is not intended to work with huge indexes. Everything beyond several hundred megabytes will waste resources (GC cycles), because it uses an internal buffer size of 1024 bytes, producing millions of byte[1024] arrays. This class is optimized for small memory-resident indexes. It also has bad concurrency on multithreaded environments.
>
> It is recommended to materialize large indexes on disk and use MMapDirectory, which is a high-performance directory implementation working directly on the file system cache of the operating system, so copying data to Java heap space is not useful.

ahornace:
fixes #2454

From https://lucene.apache.org/core/7_6_0/core/org/apache/lucene/store/ByteBuffersDirectory.html :
> A heap-based directory like this one can have the advantage in case of ephemeral, small, short-lived indexes when disk syncs provide an additional overhead.

which I think is exactly what we need for tests.

Thanks :)

### DIFF
--- a/suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java
+++ b/suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
  */
 package org.opengrok.suggest;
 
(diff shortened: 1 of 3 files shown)


vladak:
Cool !

PR #2639, opened 2019-01-23
use Jackson for JSON serialization of Messages

vladak:
This change converts Message handling to use Jackson for JSON handling and finally shoves JSON-simple out of the door. The JSON format of the Message changed a bit to accommodate easier serialization.

Tested with:
  1. `curl -X POST --data '{ "tags": [ "foo" ], "cssClass": "class", "text": "test message", "duration": "PT10M" }' -H 'Content-Type:application/json' http://localhost:8080/source/api/v1/messages`
  1. `curl -X GET http://localhost:8080/source/api/v1/messages`
  1. check the UI

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -75,7 +75,7 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
 
         {
           "tags": ["main"],
-          "cssClass": "cssClass",
+          "cssClass": "class",
           "text":"test message",
           "duration":"PT10M"
         }
@@ -98,18 +98,17 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
 
         [
           {
-            "acceptedTime": "2018-06-28T17:49:01.793Z",
-            "message": {
-              "tags": ["main"],
-              "cssClass": "cssClass",
-              "text": "test message",
-              "duration": "PT10M"
-            },
-            "expirationTime": "2018-06-28T17:59:01.793Z",
-            "expired":false
+            "expired": false,
+            "created": "2019-01-23 20:39:31 CET",
+            "tags": [
+              "main"
+            ],
+            "expiration": "2019-01-23 20:49:31 CET",
+            "cssClass": "class",
+            "text": "test&nbsp;message"
           }
         ]
-        
+
 ## Projects [/projects]
 
 ### returns a list of all projects [GET]
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
@@ -42,9 +42,10 @@ public interface JSONable {
         try {
             return mapper.writeValueAsString(this);
         } catch (JsonProcessingException e) {
-            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING, "failed to convert object of class {0} to JSON: {1}",
-                    new Object[]{this.getClass().toString(), this});
-            return "";
+            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING,
+                    String.format("failed to convert object of class {0} to JSON: {1}",
+                    this.getClass().toString(), this), e);
+            return EMPTY;
         }
     }
 }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
@@ -31,7 +31,7 @@ import java.util.logging.Level;
 
 public interface JSONable {
 
-    public static final String EMPTY = "";
+    String EMPTY = "";
 
     /**
      * convert object to JSON
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
@@ -31,6 +31,8 @@ import java.util.logging.Level;
 
 public interface JSONable {
 
+    public static final String EMPTY = "";
+
     /**
      * convert object to JSON
      * @return JSON string or empty string on error
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
@@ -1,0 +1,48 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ */
+
+package org.opengrok.indexer.web.messages;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.opengrok.indexer.logger.LoggerFactory;
+
+import java.util.logging.Level;
+
+public interface JSONable {
+
+    /**
+     * convert object to JSON
+     * @return JSON string or empty string on error
+     */
+    default String toJSON() {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            return mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING, "failed to convert object of class {0} to JSON: {1}",
+                    new Object[]{this.getClass().toString(), this});
+            return "";
+        }
+    }
+}
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
@@ -43,7 +43,7 @@ public interface JSONable {
             return mapper.writeValueAsString(this);
         } catch (JsonProcessingException e) {
             LoggerFactory.getLogger(JSONable.class).log(Level.WARNING,
-                    String.format("failed to convert object of class {0} to JSON: {1}",
+                    String.format("failed to convert object of class %s to JSON: %s",
                     this.getClass().toString(), this), e);
             return EMPTY;
         }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
@@ -32,6 +32,7 @@ import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 import org.hibernate.validator.constraints.NotBlank;
 import org.hibernate.validator.constraints.NotEmpty;
+import org.opengrok.indexer.web.Util;
 import org.opengrok.indexer.web.api.constraints.PositiveDuration;
 
 import java.io.IOException;
@@ -44,7 +45,7 @@ import java.util.TreeSet;
 
 import static org.opengrok.indexer.web.messages.MessagesContainer.MESSAGES_MAIN_PAGE_TAG;
 
-public class Message implements Comparable<Message> {
+public class Message implements Comparable<Message>, JSONable {
 
     @NotEmpty(message = "tags cannot be empty")
     private Set<String> tags = Collections.singleton(MESSAGES_MAIN_PAGE_TAG);
@@ -52,6 +53,7 @@ public class Message implements Comparable<Message> {
     private String cssClass = "info";
 
     @NotBlank(message = "text cannot be empty")
+    @JsonSerialize(using = HTMLSerializer.class)
     private String text;
 
     @PositiveDuration(message = "duration must be positive")
@@ -190,4 +192,29 @@ public class Message implements Comparable<Message> {
         }
     }
 
+    protected static class HTMLSerializer extends StdSerializer<String> {
+
+        private static final long serialVersionUID = -2843900664165513923L;
+
+        HTMLSerializer() {
+            this(null);
+        }
+
+        HTMLSerializer(final Class<String> cl) {
+            super(cl);
+        }
+
+        @Override
+        public void serialize(
+                final String string,
+                final JsonGenerator jsonGenerator,
+                final SerializerProvider serializerProvider
+        ) throws IOException {
+            if (string != null) {
+                jsonGenerator.writeString(Util.encode(string));
+            } else {
+                jsonGenerator.writeNull();
+            }
+        }
+    }
 }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
@@ -22,16 +22,21 @@
  */
 package org.opengrok.indexer.web.messages;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import org.hibernate.validator.constraints.NotBlank;
 
 import java.io.IOException;
+import java.text.DateFormat;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -223,19 +228,39 @@ public class MessagesContainer {
         return toRet == null ? new TreeSet<>() : toRet;
     }
 
-    public static class AcceptedMessage implements Comparable<AcceptedMessage> {
+    public static class AcceptedMessage implements Comparable<AcceptedMessage>, JSONable {
 
         private final Instant acceptedTime = Instant.now();
 
+        // The message member is ignored so that it can be flattened using the getters specified below.
+        @JsonIgnore
         private final Message message;
+
+        @JsonProperty("text")
+        @NotBlank(message = "text cannot be empty")
+        @JsonSerialize(using = Message.HTMLSerializer.class)
+        public String getText() {
+            return message.getText();
+        }
+
+        @JsonProperty("cssClass")
+        public String getCssClass() {
+            return message.getCssClass();
+        }
 
         private AcceptedMessage(final Message message) {
             this.message = message;
         }
 
+        @JsonProperty("created")
         @JsonSerialize(using = InstantSerializer.class)
         public Instant getAcceptedTime() {
             return acceptedTime;
+        }
+
+        @JsonProperty("tags")
+        public Set<String> getTags() {
+            return message.getTags();
         }
 
         public Message getMessage() {
@@ -246,6 +271,7 @@ public class MessagesContainer {
             return getExpirationTime().isBefore(Instant.now());
         }
 
+        @JsonProperty("expiration")
         @JsonSerialize(using = InstantSerializer.class)
         public Instant getExpirationTime() {
             return acceptedTime.plus(message.getDuration());
@@ -297,7 +323,8 @@ public class MessagesContainer {
                     final SerializerProvider serializerProvider
             ) throws IOException {
                 if (instant != null) {
-                    jsonGenerator.writeString(instant.toString());
+                    DateFormat formatter = DateFormat.getDateTimeInstance(DateFormat.LONG, DateFormat.LONG, Locale.ROOT);
+                    jsonGenerator.writeString(formatter.format(Date.from(instant)));
                 } else {
                     jsonGenerator.writeNull();
                 }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
@@ -1,0 +1,235 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ */
+
+package org.opengrok.indexer.web.messages;
+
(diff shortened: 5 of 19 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(MessagesUtils.class);
+
+    private void MessageUtils() {
+        // private to ensure static
+    }
+
+    private static final String emptyJSON;
+
+    static {
+        emptyJSON = "";
```
tulinkry: this can be set up in the previous statement

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+
+    static class TaggedMessagesContainer {
+        private String tag;
```
tulinkry: can be final

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+        private SortedSet<MessagesContainer.AcceptedMessage> messages;
```
tulinkry: can be final

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+            return emptyJSON;
+        }
+
+        ObjectMapper mapper = new ObjectMapper();
+
+        try {
+            return mapper.writeValueAsString(messages);
+        } catch (JsonProcessingException e) {
+            LOGGER.log(Level.WARNING, "failed to encode {} to JSON", messages);
+            return null;
```
tulinkry: this will cause the ui to blow on npe when doing String.length()

opengrok-indexer/src/test/java/org/opengrok/indexer/web/messages/JSONUtils.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import org.junit.Assert;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+public class JSONUtils {
```
tulinkry: can be final

opengrok-indexer/src/test/java/org/opengrok/indexer/web/messages/JSONUtils.java
```diff
@@ -0,0 +1,41 @@
...
+    private JSONUtils() {
+        // private to ensure statis
+    }
+
+    protected static Set<String> getNumberOfJSONFields(String jsonString) throws IOException {
```
tulinkry: this isn't returning the number but the fields

opengrok-indexer/src/test/java/org/opengrok/indexer/web/messages/MessageTest.java
```diff
@@ -59,4 +65,11 @@ public void createBadMessageTest6() {
         new Message("test", Collections.emptySet(), null, Duration.ofMinutes(1));
     }
 
+    @Test
+    public void message2JSON() throws IOException {
```
tulinkry: I'd be for messageToJson please

opengrok-web/src/main/webapp/js/utils-0.0.26.js
```diff
@@ -1117,7 +1117,7 @@
                             $ul.append(
                                     $('<li>')
                                     .addClass('message-group-item')
-                                    .addClass(tag.messages[j].class)
+                                    .addClass(tag.messages[j].cssclass)
```
tulinkry: bump a new version then

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -183,11 +193,46 @@ public void serialize(
...
+            }
+        }
+    }
+
+    public String toJSON() {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            return mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            return null;
```
tulinkry: I'm not sure if this directly would cause the ui to blow on npe when accessing isEmpty or length of the string.

tulinkry: However definitely a log statement is necessary

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -246,11 +264,21 @@ public boolean isExpired() {
...
         public Instant getExpirationTime() {
             return acceptedTime.plus(message.getDuration());
         }
 
+        public String toJSON() {
+            ObjectMapper mapper = new ObjectMapper();
+            try {
+                return mapper.writeValueAsString(this);
+            } catch (JsonProcessingException e) {
+                return null;
```
tulinkry: some log statement would help the later investigtion

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -183,11 +193,46 @@ public void serialize(
...
             try {
+                // TODO Util.decode()
                 return Duration.parse(context.readValue(parser, String.class));
             } catch (DateTimeParseException e) {
                 throw new IOException(e);
             }
         }
     }
 
+    private static class HTMLSerializer extends StdSerializer<String> {
```
tulinkry: why we need a html serializer? why is necessary to return encoded html markup in the message?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -161,7 +171,7 @@ public void serialize(
                 final SerializerProvider serializerProvider
         ) throws IOException {
             if (duration != null) {
-                jsonGenerator.writeString(duration.toString());
+                jsonGenerator.writeString(Util.encode(duration.toString()));
```
tulinkry: is it necessary to encode duration iso standard?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -46,14 +50,20 @@
 
 public class Message implements Comparable<Message> {
 
+    @JsonProperty("tags")
     @NotEmpty(message = "tags cannot be empty")
     private Set<String> tags = Collections.singleton(MESSAGES_MAIN_PAGE_TAG);
 
+    @JsonProperty("cssclass")
+    @JsonSerialize(using = HTMLSerializer.class)
```
tulinkry: why css class would need encoding? I'm not even sure if that is a legal class name with html encoded entities

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -297,7 +325,8 @@ public void serialize(
                     final SerializerProvider serializerProvider
             ) throws IOException {
                 if (instant != null) {
-                    jsonGenerator.writeString(instant.toString());
+                    SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z");
```
tulinkry: can't we use some standard format constant instead of this adhoc format?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+
+    /**
+     * Print messages for given tags into JSON array
+     *
+     * @param tags list of tags
+     * @return JSON array of the messages (the same as the parameter)
+     */
+    @SuppressWarnings("unchecked")
```
tulinkry: does it have an unchecked warning?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Print set of tagged messages into JSON object.
+     * TOOD should this be TaggedMessagesContainer.toJSON() ?
+     *
+     * @return JSON string or null
+     */
+    @SuppressWarnings("unchecked")
+    private static String taggedMessagesToJson(Set<TaggedMessagesContainer> messages) {
```
tulinkry: does it have an unchecked warning?

opengrok-indexer/src/test/java/org/opengrok/indexer/web/messages/MessagesContainerTest.java
```diff
@@ -144,4 +148,16 @@ public void getMessagesNullTest() {
         container.getMessages(null);
     }
 
+    // tests for serialization of MessagesContainer.AcceptedMessage
```
tulinkry: javadoc comment instead of inline comment

apiary.apib
```diff
@@ -98,18 +98,17 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
...
-              "cssClass": "cssClass",
-              "text": "test message",
-              "duration": "PT10M"
-            },
-            "expirationTime": "2018-06-28T17:59:01.793Z",
-            "expired":false
+            "expired": false,
+            "created": "2019-01-23 20:39:31 CET",
+            "tags": [
+              "foo"
```
tulinkry: the tag should be main, shouldn't it?

apiary.apib
```diff
@@ -98,18 +98,17 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
...
+            ],
+            "expiration": "2019-01-23 20:49:31 CET",
+            "cssclass": "class",
+            "text": "test message"
```
tulinkry: this is wrong (if you use the encode html entities)

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -246,11 +264,21 @@ public boolean isExpired() {
             return getExpirationTime().isBefore(Instant.now());
         }
 
+        @JsonProperty("expiration")
         @JsonSerialize(using = InstantSerializer.class)
         public Instant getExpirationTime() {
             return acceptedTime.plus(message.getDuration());
         }
 
+        public String toJSON() {
```
tulinkry: maybe it'd be worth it to extract this to an interface Jsonable with default method implementation to
1. not to duplicate the code
2. mark classes as jsonable

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -227,17 +232,30 @@ public void stopExpirationTimer() {
 
         private final Instant acceptedTime = Instant.now();
 
+        // The message member is ignored so that it can be flattened using the getters
+        // specified below.
```
tulinkry: I generally don't see a reason for this. Now it just duplicates/delegates all stuff from/to Message.

Now via the api you're setting different properties than you can see in the output as the input entity and output entity does not match.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
tulinkry: also this is not a valid json, valid empty json is {} or [], this is just empty string.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -46,14 +50,20 @@
 
 public class Message implements Comparable<Message> {
 
+    @JsonProperty("tags")
```
ahornace: Are not these annotations redundant? The variable has a proper name so no need for "renaming". Applies similarly below. https://stackoverflow.com/questions/12583638/when-is-the-jsonproperty-property-used-and-what-is-it-used-for

opengrok-indexer/src/test/java/org/opengrok/indexer/web/messages/JSONUtils.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.core.JsonToken;
+import org.junit.Assert;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+public class JSONUtils {
+    private JSONUtils() {
+        // private to ensure statis
```
ahornace:
```suggestion
        // private to ensure static
```

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    public static String messagesToJson(Project project, String... additionalTags) {
+        if (project == null) {
+            return emptyJSON;
+        }
+
+        List<String> tags = new ArrayList<>();
+        tags.addAll(Arrays.asList(additionalTags));
+        tags.add(project.getName());
+        project.getGroups().stream().forEach((Group t) -> {
```
ahornace: Maybe rewrite to
```java
project.getGroups().stream().map(Group::getName).forEach(tags::add);
```
?

Hopefully I wrote it correctly (no IDE suggestions :D ).

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Print messages for given tags into JSON array
+     *
+     * @param tags list of tags
+     * @return json array of the messages
+     * @see #messagesToJson(String...)
+     */
+    private static String messagesToJson(List<String> tags) {
+        String[] array = new String[tags.size()];
```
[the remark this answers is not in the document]
tulinkry: @ahornace This isn’t a new code however improvements are welcome

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -227,17 +232,30 @@ public void stopExpirationTimer() {
...
+        @JsonIgnore
         private final Message message;
 
+        @JsonProperty("text")
+        public String getText() { return message.getText(); };
```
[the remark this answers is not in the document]
vladak: I blame IDEA for this one. In fact in IDEA you never know how the code appears on disk, it seems to be dynamically modifying the appearance.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
vladak: this used to be there originally - see http://demo.opengrok.org/xref/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java?r=81b586e6#1319

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
vladak: probably not. Does not hurt to be extra careful, though. Again, this is in line with original encoding code.

apiary.apib
vladak: Well, the tag can be arbitrary. That said it makes the documentation a bit friendlier to make it appear like it was produced by the POST message above.

tulinkry: ah I see this "foo" is not connected to the "main" above.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -133,7 +133,7 @@ private static String taggedMessagesToJson(Set<TaggedMessagesContainer> messages
             return mapper.writeValueAsString(messages);
         } catch (JsonProcessingException e) {
             LOGGER.log(Level.WARNING, "failed to encode {} to JSON", messages);
-            return null;
+            return empty;
```
tulinkry: I think constant should be always named with capital letters. Our checkstyle does not enforce it?

vladak: seems not

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
vladak: Yeah, it's strange combination of 2 styles. In fact, IDEA offers to replace this with loop:
```java
        for (Group t : project.getGroups()) {
            tags.add(t.getName());
        }
```
but I guess these days stream API one liner is more fashionable.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
vladak: The `private static String taggedMessagesToJson(Set<TaggedMessagesContainer> messages)` was left as a copy of `toJSON()`. Ideas on how to make the `Set` `JSONable` are welcome.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
tulinkry: then go ahead, I wrote it when I was new in the streaming api

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
tulinkry: I guess only by introducing another named class (like taggedmessagecontainer) extending Set and implementing Jsonable

But I don't have a full context on the usage so maybe one helper function is the way to go .

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -183,11 +193,46 @@ public void serialize(
         public Duration deserialize(final JsonParser parser, final DeserializationContext context)
                 throws IOException {
             try {
+                // TODO Util.decode()
```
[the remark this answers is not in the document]
vladak: The problem on the surface is that there is no `Util.decode()`. This could be resolved by replacing the encoding/decoding with `StringEscapeUtils` methods from Apache commons.

However, there is another problem that is related - due to the flattening in `AcceptedMessage` there is no deserialization done. The question how to work with user supplied content - should he get what he sent or just get the internal representation as it is currently.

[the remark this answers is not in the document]
tulinkry: I think for Duration it's not a big deal.

I think we can do the internal representation as one day we might replace the static content in jsp with some javascript api calls to the messages endopint.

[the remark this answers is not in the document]
vladak: That's what I had in mind when I was looking at the JS code.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -297,7 +323,8 @@ public void serialize(
                     final SerializerProvider serializerProvider
             ) throws IOException {
                 if (instant != null) {
-                    jsonGenerator.writeString(instant.toString());
+                    DateFormat formatter = DateFormat.getDateTimeInstance(DateFormat.LONG, DateFormat.LONG, Locale.ROOT);
+                    jsonGenerator.writeString(formatter.format(Date.from(instant)));
```
tulinkry: nb i found `new ISO8601DateFormat()` from jackson databind

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
```diff
@@ -0,0 +1,48 @@
...
+     * convert object to JSON
+     * @return JSON string or empty string on error
+     */
+    default String toJSON() {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            return mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING, "failed to convert object of class {0} to JSON: {1}",
+                    new Object[]{this.getClass().toString(), this});
```
tulinkry: you're not logging the exception

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,235 @@
...
+
+public final class MessagesUtils {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MessagesUtils.class);
+
+    private void MessageUtils() {
+        // private to ensure static
+    }
+
+    private static final String EMPTY = "";
```
tulinkry: this can be part of jsonable interface
and also (I don't know if we use it) but there's String(s)Utils.EMPTY in apache libraries so maybe it is included

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -0,0 +1,235 @@
...
+        if (messages.isEmpty()) {
+            return EMPTY;
+        }
+
+        ObjectMapper mapper = new ObjectMapper();
+
+        try {
+            return mapper.writeValueAsString(messages);
+        } catch (JsonProcessingException e) {
+            LOGGER.log(Level.WARNING, "failed to encode {0} to JSON", messages);
```
tulinkry: you're not logging the exception

opengrok-web/src/main/webapp/js/utils-0.0.27.js
```diff
@@ -1117,7 +1117,7 @@
                             $ul.append(
                                     $('<li>')
                                     .addClass('message-group-item')
-                                    .addClass(tag.messages[j].class)
+                                    .addClass(tag.messages[j].cssclass)
```
tulinkry: I'm little bit lost now with the renaming, but isn't this currently "cssClass" again?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/MessagesControllerTest.java
```diff
@@ -58,10 +59,12 @@
     private final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
 
     private static class AcceptedMessageModel {
```
tulinkry: why not use `AcceptedMessage` from message container?

vladak: It may have to do with the `JerseyTest`. Anyhow, another advantage of this is that it verifies the structure of the JSON data without parsing.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
vladak: It's deprecated: https://github.com/FasterXML/jackson-databind/issues/1786

vladak: Also, I find ISO 8601 date format not very readable for humans.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/JSONable.java
```diff
@@ -42,9 +42,10 @@ default String toJSON() {
         try {
             return mapper.writeValueAsString(this);
         } catch (JsonProcessingException e) {
-            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING, "failed to convert object of class {0} to JSON: {1}",
-                    new Object[]{this.getClass().toString(), this});
-            return "";
+            LoggerFactory.getLogger(JSONable.class).log(Level.WARNING,
+                    String.format("failed to convert object of class {0} to JSON: {1}",
```
tulinkry: %s

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
tulinkry: ok

PR #2642, opened 2019-01-25
deal with null entities in AuthorizationFramework

vladak:
This provides basic protection for `null` entities.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
@@ -763,6 +763,12 @@ public final class AuthorizationFramework {
             return true;
         }
 
+        if (entity == null) {
+            LOGGER.log(Level.WARNING, "entity was null for request with parameters: {}",
+                    request.getParameterMap());
+            return false;
+        }
+
         Statistics stats = RuntimeEnvironment.getInstance().getStatistics();
 
         Boolean val;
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
```diff
@@ -763,6 +763,11 @@ private boolean checkAll(HttpServletRequest request, String cache, Nameable enti
             return true;
         }
 
+        if (entity == null) {
+            LOGGER.log(Level.WARNING, "entity was null for request {}", request);
```
tulinkry: maybe an entity class (getSimpleName) would also be a benefit for debuging

tulinkry: and did you try what request.toString returns? I guess it's not any valueable info just the reference number.

vladak: will use `request.getRequestURI()`

vladak: or maybe `request.getParameterMap()` would be more useful.

tulinkry: may be. a little bit too long eventually for searches

vladak: yes, however might be crucial for figuring out what went wrong.

Issue #2647, opened 2019-01-31
indexer of AOSP fails with OutOfMemoryError and other exceptions


brightchuh:
### Versions
- Java 
>openjdk version "1.8.0_191"
OpenJDK Runtime Environment (build 1.8.0_191-8u191-b12-0ubuntu0.16.04.1-b12)
OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)
- OpenGrok
>OpenGrok v1.1.2 rev 506fa6be49f1bd4dee6b6a60e16f7e7227407d0d

### Steps
- Install OpenGrok as https://github.com/oracle/opengrok/wiki/How-to-setup-OpenGrok
- Run first indexing
- Open localhost:8080/opengrok in firefox
### Problem
Some projects' files can not be opened in opengrok, showing "Failed to get xref file", as 
![2019-01-31 09-07-03](https://user-images.githubusercontent.com/10250143/52025797-af424000-2540-11e9-9207-21255828f108.png)

, while other projects are OK!
### Log
- index log
[index.zip](https://github.com/oracle/opengrok/files/2815326/index.zip)

vladak:
I see quite a few exceptions in the (huge) log, e.g.:
```
java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError
        at java.util.concurrent.ForkJoinTask.get(ForkJoinTask.java:1006)
        at org.opengrok.indexer.index.IndexDatabase.indexParallel(IndexDatabase.java:1227)
        at org.opengrok.indexer.index.IndexDatabase.update(IndexDatabase.java:509)
        at org.opengrok.indexer.index.IndexDatabase$1.run(IndexDatabase.java:227)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.OutOfMemoryError
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[...]

PR #2648, opened 2019-01-31
print index directory on exception

vladak:
As noticed in the log files supplied in #2647, the exception logging is missing index directory name so it is not clear which project's index was incomplete.


vladak:
Yep, the Suggester test failed in Appveyor due to inability to bind a network port.

PR #2650, opened 2019-01-31
Fix suggestions for groups selections

### LINKED ISSUE #2640: Suggester needs to use PageConfig.getRequestedProjects() to get list of projects
I am using version 1.1.2 on tomcat
Steps to re-produce
- Select multiple projects by clicking "Select All" button or by clicking on the group name containing multiple projects
- Start typing in any search box. We will see the error exclamation against Full Search text box and web server log will have NullPointerException 

Observation
If we select all those projects by clicking each one then there is no issue. 
Also if we refresh page after selecting the projects (by clicking select all or group name), suggester starts working.

ahornace:
fixes #2640

Hi,

the problem was that `$("#project").searchableOptionList().getSelection()` can return group elements.

Thanks :)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
  */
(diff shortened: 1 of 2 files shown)


vladak:
So I guess the idea in https://github.com/oracle/opengrok/issues/2640#issuecomment-457537498 is not needed ?

Still, it would be handy to fortify the `AuthorizationFilter` code to prevent null Projects from being passed on.

tulinkry:
> So I guess the idea in #2640 (comment) is not needed ?

I don't think so.

ahornace:
> So I guess the idea in #2640 (comment) is not needed ?

I don't think so either. The `AuthorizationFilter` is tied to suggestions, expects `projects[]` query param whereas `PageConfig` is tied to web frontend and requires `project` query param.

PR #2652, opened 2019-02-01
Feature/retire lex helpers

idodeclare:
Hello,

Please consider for integration this refactoring I finally got around to do after having unified the class hierarchy for xrefers and symbol tokenizers a while ago in 80ef15e08.

The delegates I previously used in order to have common JFlex specifications for xref/symtok for Perl, Ruby, Ada, and Eiffel can now be subsumed as language-specific, common xref/symtok base classes (which was the intention of 80ef15e08).

This way any interested parties who want to write analyzers don't have to concern themselves with the delegate pattern previously required when a common JFlex spec was desired (and which should be desired for analyzers going forward).

Additionally I caught a few places where single-char identifiers were not revised to be actively linked as was done for all other languages a year ago.

Thank you.


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/perl/PerlLexer.java
```diff
@@ -682,4 +671,58 @@ public HereDocSettings(String terminator, int state) {
 
         public int getState() { return state; }
     }
+
+    class PerlLexerData {
```
tulinkry: can it be static?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/ruby/RubyLexer.java
```diff
@@ -18,128 +18,53 @@
...
-     */
-    private char nestqchar;
-
-    /**
-     * When matching a quoting construct like %[], %w(), etc., the
-     * number of remaining end separators is stored. It starts at 1, and
-     * any nesting increases the value.
-     */
-    private int nendqchar;
+    private RubyLexerData d;
```
tulinkry: can this have some descriptive name?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/perl/PerlLexer.java
```diff
@@ -18,146 +18,53 @@
...
-    private Queue<HereDocSettings> hereSettings;
-
-    /**
-     * When matching a quoting construct like qq[], q(), m//, s```, etc., the
-     * operator name (e.g., "m" or "tr") is stored. Unlike {@code endqchar} it
-     * is not unset when the quote ends, because it is useful to indicate if
-     * quote modifier characters are expected.
-     */
-    private String qopname;
+    private PerlLexerData d;
```
tulinkry: can this have some descriptive name?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/ruby/RubyLexer.java
```diff
@@ -550,39 +476,116 @@ public int nameLength(String capture) {
...
+     */
+    abstract int SCOMMENT();
+
+    /**
+     * Subclasses must override to get the constant value created by JFlex to
+     * represent POD.
+     */
+    abstract int POD();
+
     private class HereDocSettings {
```
tulinkry: this class is the same like in PerlAnalyser, can you extract it so the code is not duplicated?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/perl/PerlLexer.java
idodeclare: Excised Perl stacking, which was accidentally ported from Ruby

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/ruby/RubyLexer.java
idodeclare: This is a parallel-evolved feature in the languages but not always the same; and cross-language, inter-analyzer dependencies are hard to reason about for the limited value of re-use.

Also the similar features have to be allowed to change behavior independently.

E.g. support for indented here-docs (partially but not 100% supported by OpenGrok for Ruby or Perl) arrived in Ruby years before Perl's, so `HereDocSettings` would have evolved differently from Perl's.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/ruby/RubyLexer.java
tulinkry: With variables like c, i, rem, o a simple `d` can become a candidate for a local variable, and one has to actually look for the definition and that it actually shared variable. This doesn't seem obvious for me. If that's a head of a stack, why not name it like that? `d` doesn't mean anything.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/ruby/RubyLexer.java
tulinkry: Ok, I agree for separating the analyzers logic.

It's still violating the YAGNI principle but for this instance it makes more sense to keep it separately.

opengrok-indexer/src/main/resources/analysis/perl/PerlXref.lex
```diff
@@ -122,6 +103,66 @@ import org.opengrok.indexer.web.HtmlConsts;
     }
 
     protected void skipLink(String s, Pattern p) { /* noop */ }
+
+    /**
```
tarzanek: where are below functions used?

idodeclare: They’re called by the abstract `PerlLexer` extended by the Perl JFlex xref and symtok classes. Formerly the values were injected through the helper delegate constructor but now can be mediated through method overrides.

tarzanek:
ok, let's merge, thank you!

tarzanek:
hopefully this didn't conflicted with https://github.com/oracle/opengrok/pull/2673 (resp. I hope vertilog didn't use these helpers ;-) ) @idodeclare Chris?

idodeclare:
No — all good, @tarzanek. Thank you, Lubos.

Issue #2643, opened 2019-01-28
Add additional CheckStyle rules


ahornace:
List of checks: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/sun_checks.xml

I'm not saying that we need to use all but some of them are quite useful.

vladak:
what else is needed after #2653 ?

ahornace:
> what else is needed after #2653 ?

Well, that depends on what we want. I would add at least whitespace checks, javadoc style checks, line length check and a few others. But there was quite a lot of errors and I was lazy while doing #2653 :D

tulinkry:
I'd be for some standard for javadoc definitelly. Line length is not a bad idea but keep it relaxed pls

vladak:
Okay, sounds good to me. Adding checks (of any type) as an afterthought is
often painful.

čt 7. 2. 2019 15:11 odesílatel Adam Hornáček <notifications@github.com>
napsal:

ahornace:
Fixed by #2896

PR #2653, opened 2019-02-01
Add more CheckStyle rules

ahornace:
Just a few additional rules that were easy to fix.

Thanks :)

### DIFF
--- a/dev/checkstyle/style.xml
+++ b/dev/checkstyle/style.xml
@@ -32,5 +32,14 @@
             <property name="offCommentFormat" value="&lt;editor-fold defaultstate=&quot;collapsed&quot; desc=&quot;Generated Code&quot;&gt;"/>
             <property name="onCommentFormat" value="&lt;/editor-fold&gt;"/>
         </module>
+
+        <module name="ModifierOrder"/>
+        <module name="RedundantModifier"/>
+
+        <module name="SimplifyBooleanExpression"/>
+        <module name="SimplifyBooleanReturn"/>
+
+        <module name="ArrayTypeStyle"/>
+        <module name="UpperEll"/>
     </module>
 </module>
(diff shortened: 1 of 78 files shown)


dev/checkstyle/style.xml
```diff
@@ -32,5 +32,14 @@
...
         </module>
+
+        <module name="ModifierOrder"/>
+        <module name="RedundantModifier"/>
+
+        <module name="SimplifyBooleanExpression"/>
+        <module name="SimplifyBooleanReturn"/>
+
+        <module name="ArrayTypeStyle"/>
+        <module name="UpperEll"/>
```
tulinkry: I'm little lazy...what is this one?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/EftarFile.java
```diff
@@ -87,56 +81,6 @@ public Node get(long hash) {
         }
     }
 
-    class FNode {
```
tulinkry: ?

dev/checkstyle/style.xml
ahornace: Incorrect: `1l`. Correct: `1L`. Rationale: `l` is similar to `1` in some fonts.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/EftarFile.java
ahornace: It was not used at all -> YAGNI. If we need it is needed again we can add it back.

tulinkry: ok

dev/checkstyle/style.xml
tulinkry: thank you

tulinkry:
Approaches #2643

PR #2654, opened 2019-02-01
add API endpoint to delete project index data

vladak:
This should be handy for cases like #2647.


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -135,35 +135,55 @@ public Response addProject(final String projectName) {
...
+            throws HistoryException {
 
-        Project proj = env.getProjects().get(projectName);
-        if (proj == null) {
+        Project project = env.getProjects().get(projectName);
+        if (project == null) {
             throw new IllegalStateException("cannot get project \"" + projectName + "\"");
         }
 
+        project.setIndexed(false);
```
tulinkry: this is not necessary if you consider you're removing it

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -189,6 +211,7 @@ public void deleteProject(@PathParam("project") final String projectName)
                     }
                 }).collect(Collectors.toSet()));
```
tulinkry: could you use filter here to ensure that "" wont go further? it's just asking for troubles in future

apiary.apib
```diff
@@ -145,6 +145,17 @@ and suggester data.
...
 + Response 204
 
+## Project data management [/projects/{project}/data]
+
+### delete project index data [DELETE]
+
+This will delete project index data but keep the project in web application configuration so that
+can be indexed from scratch.
+The project data (whatever is the project's directory under source root) will be left intact.
+This will also mark the project as not indexed.
```
tulinkry: what is the benefit of keeping it in the configuration when you need to reindex it from scratch?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
vladak: This is to avoid exceptions when performing e.g. all project search.

apiary.apib
vladak: I needed this multiple times, e.g. when indexer failed for some reason or when the repository was force pushed.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
vladak: Good idea.

apiary.apib
tulinkry: But can't you just delete the whole project and not just the data?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
tulinkry: You mean at the point between this line and the actual removal from the project map?

vladak: yep

apiary.apib
vladak: Sure, this is just a little bit less invasive.

vladak:
Actually just needed this today as https://github.com/NetBSD/src has rewritten its history.

vladak:
I just like to have the granularity, that's all.

vladak:
As part of the changes, I eradicated the code from Indexer that can be replaced by RESTful API calls.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ProjectsControllerTest.java
```diff
@@ -539,4 +538,8 @@ private void setHandleRenamedFilesToFalse(final String project) {
                 .put(Entity.text(Boolean.FALSE.toString()));
     }
 
+    @Test
+    public void testListFiles() {
+
+    }
```
tulinkry: :-D very robust

tulinkry:
> I just like to have the granularity, that's all.
Then ok

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/ProjectsControllerTest.java
vladak: :-)
oops, did not intend to commit this one.

apiary.apib
```diff
@@ -153,6 +172,12 @@ This entry point is used by the Indexer once it finishes indexing given project.
 
 + Response 204
 
+## Project index files [/projects/{project}/files]
+
+### Get a list of files tracked by the index database for given project [GET]
+
```
tulinkry: Is this a top level, or all files?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1371,7 +1370,13 @@ public void removeIndexChangedListener(IndexChangedListener listener) {
...
             }
             while (iter != null && iter.term() != null) {
-                files.add(Util.uid2url(iter.term().utf8ToString()));
+                String value = iter.term().utf8ToString();
+                if (value.isEmpty()) {
+                    iter.next();
+                    continue;
+                }
+
+                files.add(Util.uid2url(value));
```
tulinkry: why is this necessary? it wasn't used before?

vladak: The first term is always empty, it never seem to have worked correctly.

tulinkry: Aha, so I guess nobody tried this code for a long time.

apiary.apib
vladak: all files

tulinkry: ok

tulinkry:
Looks good

Issue #2588, opened 2018-12-12
Analyzer plugin framework


vladak:
As discussed on Slack #dev channel, some (namely @tarzanek) would like to add some new analyzers (e.g. PDF analyzer - #492) written in Scala/Kotlin/Closure/you-name-it. Given that core OpenGrok should stay pure Java, we need a plugin framework and interface for the analyzers to facilitate this.

The repositories for analyzers can be then hosted anywhere (our own would live under https://github.com/OpenGrok/) and the jars published to Maven central. This should also allow for 3rd party analyzers - just shove them under analyzer directory (configurable) and OpenGrok will load them automatically.

tulinkry:
We can somehow reuse the existing framework for authorization plugins.

vladak:
Yeah, that would be nice.

vladak:
Some things for consideration:
  - analyzer should be able to add items to the 'File type' drop down menu in the UI
  - should be possible to add options to ctags

PR #2658, opened 2019-02-02
Extracting base classes for analyzer factory and analyzers

tulinkry:
Some base class is needed as a reference for the framework

This does not add any functionality. I verified both unit tests and index size does not change.

approaches #2588


tarzanek:
lgtm, merging

PR #2659, opened 2019-02-02
Separate plugin framework

tulinkry:
Separating the idea with plugin framework to a dedicated class and making the authorization framework a subclass of it.

depends on #2658 **(do not review until this one is merged)**
approaches #2588


tarzanek:
lgtm

PR #2660, opened 2019-02-02
Analyzer plugin framework

tulinkry:
Introducing a analyzer plugin framework

- loading analyzers implementing an interface from a directory
- augmenting analyzer guru to use the loaded plugins
- making analyzer guru a property of RuntimeEnvironment
- loading the default analyzers as it used to be

Things to finish:
- it is poorly tested (no tests so far)
- I haven't tried to load an actual analyzer from a file (I just got a working state with the defaults analyzers)

depends on #2659 (do not review until this one is merged)
approaches #2588


opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginClassLoader.java
```diff
@@ -257,6 +255,8 @@ public Class loadClass(String name, boolean resolveIt) throws ClassNotFoundExcep
             }
         }
 
+        checkClassname(name);
```
tulinkry: this does not prevent loading classes which are already loaded by the parent loader

tulinkry:
@vladak for review

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -795,50 +326,55 @@ public static AnalyzerFactory findFactory(String factoryClassName)
...
      *
      * @param factoryClass the factory class
      * @return a file analyzer factory
-     *
-     * @throws ClassCastException if the class is not a subclass of {@code
-     * FileAnalyzerFactory}
-     * @throws IllegalAccessException if the constructor cannot be accessed
-     * @throws InstantiationException if the class cannot be instantiated
-     * @throws NoSuchMethodException if no-argument constructor could not be found
+     * @throws ClassCastException        if the class is not a subclass of {@code
```
ahornace: Why are you aligning the description? I personally don't like it because if I were to add  `VeryLongLongLongLongException` then I would have to modify other places that have nothing to do with my changes which just makes the diff hard to read. That's the reason we don't do following:
```
int abc = 1;
int a   = 2;
int ab  = 3;
```

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzersInfo.java
```diff
@@ -0,0 +1,180 @@
...
+            Comparator.comparingInt(String::length).thenComparing(String::toString);
+
+    /**
+     * Modified when
+     * {@link AnalyzerFramework#addExtension(String, AnalyzerFactory)}
+     * or
+     * {@link AnalyzerFramework#addPrefix(String, AnalyzerFactory)}
+     * are called to augment the value in {@link AnalyzerGuru#getVersionNo()}.
+     */
+    public final SortedSet<String> customizations;
```
ahornace: Could you please add getters? It just does't feel right this way.

opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/AnalyzerFrameworkTest.java
```diff
@@ -0,0 +1,12 @@
+package org.opengrok.indexer.analysis;
```
ahornace: I don't think it's a good idea to keep this here. Either add some tests or remove it -> YAGNI.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzersInfo.java
tulinkry: I intentionally don’t want to use getters and setters as this class does no provide any functionality

ahornace: Even simple DTO classes are always written with getters/setters.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
tulinkry: Idea does that for all of javadoc stuff.

Descriptions and the actual code are two different things.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzersInfo.java
tulinkry: I’ve seen codes which don’t follow that. If this is code style for opengrok I can change it of course.

opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/AnalyzerFrameworkTest.java
tulinkry: I was about to fill it but it looks like it won’t happen anywhen soon so I’ll delete for now

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/IAnalyzerPlugin.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.indexer.analysis;
+
+public interface IAnalyzerPlugin {
```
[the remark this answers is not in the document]
tulinkry: No doubt for that. The trouble here is that we’ve already exposed one interface for plugins with I so this is consistent with the current code.

We should make a change in a future to transfer both cases to the correct name but it’d be a breaking change and I think it shouldn’t be part of this pr.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,400 @@
...
+        reload();
+    }
+
+    /**
+     * Register a {@code FileAnalyzerFactory} instance into the given {@code analyzersInfo}.
+     */
+    private static void registerAnalyzer(AnalyzersInfo analyzersInfo, AnalyzerFactory factory) {
+        for (String name : factory.getFileNames()) {
+            AnalyzerFactory old = analyzersInfo.fileNames.put(name, factory);
+            assert old == null :
```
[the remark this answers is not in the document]
tulinkry: This is a very old code and should be refactored. No production code should ever rely on assert statements.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,400 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AnalyzerFramework.class);
+
+    /**
+     * The default {@code FileAnalyzerFactory} instance.
+     */
+    public static final AnalyzerFactory DEFAULT_ANALYZER_FACTORY = new FileAnalyzerFactory();
+
+    private static final IAnalyzerPlugin[] DEFAULT_PLUGINS = {
+            () -> DEFAULT_ANALYZER_FACTORY,
+            () -> new IgnorantAnalyzerFactory(),
```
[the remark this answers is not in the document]
tulinkry: Sure

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,400 @@
...
+
+    /**
+     * Free resources associated with all registered analyzers.
+     */
+    public void returnAnalyzers() {
+        getAnalyzersInfo().factories.forEach(AnalyzerFactory::returnAnalyzer);
+    }
+
+    /**
+     * Decode about two factory instances if they are different.
```
vladak: about ?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,400 @@
...
+     *
+     * @param a the first instance
+     * @param b the second instance
+     * @return true if we consider them different, false otherwise
+     */
+    private static boolean factoriesDifferent(AnalyzerFactory a, AnalyzerFactory b) {
```
vladak: should this class be `Comparable` then ?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1591,6 +1597,18 @@ public synchronized AuthorizationFramework getAuthorizationFramework() {
...
     }
 
+    /**
+     * Return the analyzer guru.
+     *
+     * @return the analyzer guru instance, never {@code null}.
+     */
+    public synchronized AnalyzerGuru getAnalyzerGuru() {
+        if (analyzerGuru == null) {
+            analyzerGuru = new AnalyzerGuru(getPluginDirectory());
```
vladak: I wonder if we should just split these 2 plugin directories into analyzers and authorization plugins. This way it is kind of confusing.

vladak:
I wonder how does this interacts with analysis performed from within the webapp.

vladak:
It would be nice if you could produce dummy analyzer plugin, ideally using language other than Java. That would also serve as a basis for documentation.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,400 @@
...
+public class AnalyzerFramework extends PluginFramework<IAnalyzerPlugin, AnalyzersInfo> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AnalyzerFramework.class);
+
+    /**
+     * The default {@code FileAnalyzerFactory} instance.
+     */
+    public static final AnalyzerFactory DEFAULT_ANALYZER_FACTORY = new FileAnalyzerFactory();
+
+    private static final IAnalyzerPlugin[] DEFAULT_PLUGINS = {
```
vladak: I assume this is a temporary measure until the framework is fully implemented and in the final state it will dynamically load all the plugins from a directory - no hard-coding needed.

tulinkry:
> It would be nice if you could produce dummy analyzer plugin, ideally using language other than Java. That would also serve as a basis for documentation.

This is beyond my knowledge, maybe @tarzanek can produce the pdf analyzer to test this

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
tulinkry: Yes this is for seamless transition from the old approach to the new approach.

That is a good point though. I'm affraid that with that approach we would need to distribute two jar files instead of one and more to force people to actually put the second one somewhere (plugins directory) and point the configuration directory there and I think this is a little bit more than what I'd expect to get opengrok working.

tulinkry:
> I wonder how does this interacts with analysis performed from within the webapp.

What do you mean?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: It's not necessary as it's called plugins directory and you can put all your plugins there be it analyzer or authorization plugin. The application frameworks will load only the plugins they support.

However, if it's desired I can split these.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
tulinkry: This is an old code, I can do this in another PR so I don't mix refactoring with actual changes.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
tulinkry: will fix

tulinkry:
Rebased and updated. Make it a priority for next comments as any change in analyzer guru makes this hard to rebase correctly.

tulinkry:
> > It would be nice if you could produce dummy analyzer plugin, ideally using language other than Java. That would also serve as a basis for documentation.

> This is beyond my knowledge, maybe @tarzanek can produce the pdf analyzer to test this.

I did a unit test with a dummy analyzer to test the basic functionality.

opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/AnalyzerFrameworkTest.java
tulinkry: I filled the test with actual code

idodeclare:
Please hold on merging. I have some questions and issues and will write them up from work later today.

idodeclare:
I do like this patch's refactoring the static `AnalyzerGuru` -- though in some files you nicely added a local variable for accessing the environment's instance while in others you repeat over-and-over syntax like `cfg.getEnv().getAnalyzerGuru()` which to me is not as nice.

---

The approach of this patch treats OpenGrok's supported analyzers as merely a collection of fully independent `AnalyzerFactory`-derived classes and thus misses accommodating all the ways OpenGrok enlists analyzers.

While OpenGrok can match filename extensions or prefixes, it also selects analyzers in order-dependent non-heuristic and heuristic stages, relying on `Matcher` implementations running in the correct sequence. This patch does not accommodate these latter stages in the new model -- i.e. there is no capability for an `AnalyzerFactory` plug-in to register itself at a desired place in sequence. Alternatively there is no revision in this patch to implement e.g. an order-independent scoring model rather than a sequential model of `Matcher` instances.

You write that your approach provides "seamless transition from the old approach to the new approach" but really the new approach would not support properly registering the existing set of analyzers; the application does not break only because you copy the static ordering for `DEFAULT_PLUGINS`.

For supporting @tarzanek's desire for a PDF plug-in, only a plug-in analyzer that keys off of file extension would succeed reliably. Any implemented `Matcher` of the PDF magic number would be appended as the last of the sequence and never match PDFs because of the order-precedence of `PlainAnalyzerFactory`.

---

For another example of how misinterpreting OpenGrok's supported analyzers as fully independent does not accomodate extensibility as more analyzers are incorporated: to fully support Objective-C, it is not enough to add an Objective-C analyzer and factory. `CAnalyzerFactory` must be revised not to dominate for every `.h` file and to cooperate to pick the better analyzer for an `.h` file.

---

This patch is lacking any approach to modify `AnalyzerGuru` `getVersionNo()` to reflect plugged-in analyzers, and so it breaks the existing support for avoiding the need for full re-indexing when analyzer selection logic is altered by the user or when the user deploys a new version.

---

@vladak in #2588 directs that with an analyzer plug-in framework it "should be possible to add options to Ctags," which I agree is a necessary part of a pull request named "Analyzer plug-in framework" that is not just a trivial class loader.

---

Please may I head off any suggestion that this patch is but an introductory stage of an analyzer framework by remarking that this patch in its form is just a trivial jar/class loader [unlike your nice, stack-preserving `AuthorizationFramework`] and does not tackle any API revision needed for more complete support of an analyzer plug-in (e.g. the aforementioned Ctags options or `Matcher` order or versioning).

Perhaps we should work to revise the order-dependent `Matcher` stages instead to be numerically scoring so that they can be executed in any order and then ranked.

Overall I think some effort should be made to tackle some of the API improvements before adding jar/class loading. Otherwise we may find ourselves in a difficult architectural position with permanent, limited support of meaningful plug-ins.

---

```
any change in analyzer guru makes this hard to rebase correctly
```
You bring up the difficulty of rebasing large branches, which vexes me too. May I ask why you casually reordered a number of methods in `AnalyzerGuru` making for harder review of this patch and also for difficult merges for others? It's inconsiderate in my opinion.

---

BTW there are a couple of regressions and a log message error in this patch which I did not line-comment because of high-level concerns which I wanted to bring up first.

---

In my opinion that IDEA style of spaced-out JavaDoc should not be accepted by us.

---

Lastly I am reluctantly forced to name you as a serial offender in lifting code to different files without complying with CDDL's requirements for maintaining in the new files the copyright and descriptive text giving attribution for the original source. `AnalyzerGuru`'s code is moved without attribution to `AnalyzersInfo` and `AnalyzerFramework`.

tulinkry:
> Please may I head off any suggestion that this patch is but an introductory stage of an analyzer framework by remarking that this patch in its form is just a trivial jar/class loader [unlike your nice, stack-preserving `AuthorizationFramework`] and does not tackle any API revision needed for more complete support of an analyzer plug-in (e.g. the aforementioned Ctags options or `Matcher` order or versioning).

> Perhaps we should work to revise the order-dependent `Matcher` stages instead to be numerically scoring so that they can be executed in any order and then ranked.

> Overall I think some effort should be made to tackle some of the API improvements before adding jar/class loading. Otherwise we may find ourselves in a difficult architectural position with permanent, limited support of meaningful plug-ins.

This is exactly what this is meant to be. Transition from the old state to the new state with as minimal number of changes so it can be reviewed easily rather than fat PR with mixed refactoring and features.

idodeclare:
Yes I thought you might discount everything I brought up in my review by claiming a jar/class loader as a meaningful initial stage when the forgone API changes are more significant in my opinion.

As I wrote, "I think some effort should be made to tackle some of the API improvements before adding jar/class loading."

tulinkry:
> I do like this patch's refactoring the static `AnalyzerGuru` -- though in some files you nicely added a local variable for accessing the environment's instance while in others you repeat over-and-over syntax like `cfg.getEnv().getAnalyzerGuru()` which to me is not as nice.

Will change, why not use an inline comment?

> This patch is lacking any approach to modify `AnalyzerGuru` `getVersionNo()` to reflect plugged-in analyzers, and so it breaks the existing support for avoiding the need for full re-indexing when analyzer selection logic is altered by the user or when the user deploys a new version.

No problem I'll change it, why not use an inline comment?

> @vladak in #2588 directs that with an analyzer plug-in framework it "should be possible to add options to Ctags," which I agree is a necessary part of a pull request named "Analyzer plug-in framework" that is not just a trivial class loader.

As it's in the description it only approaches the plugin framework - only the loading part. This is not to be meant as a complete solution rather a first step forward.

> You bring up the difficulty of rebasing large branches, which vexes me too. May I ask why you casually reordered a number of methods in `AnalyzerGuru` making for harder review of this patch and also for difficult merges for others? It's inconsiderate in my opinion.

IDEA -> refactor –> move. I can walk though it again by myself.

> BTW there are a couple of regressions and a log message error in this patch which I did not line-comment because of high-level concerns which I wanted to bring up first.

I'd like to know which ones as this does not really bring much new functionality nor changes any order for the analyzers.

> In my opinion that IDEA style of spaced-out JavaDoc should not be accepted by us.

In my opinion it does not matter. If this is a consensus I can conform to it.

> Lastly I am reluctantly forced to name you as a serial offender in lifting code to different files without complying with CDDL's requirements for maintaining in the new files the copyright and descriptive text giving attribution for the original source. `AnalyzerGuru`'s code is moved without attribution to `AnalyzersInfo` and `AnalyzerFramework`.

That's why we have reviews. Why not using inline comments for affected files?

tulinkry:
> For another example of how misinterpreting OpenGrok's supported analyzers as fully independent does not accomodate extensibility as more analyzers are incorporated: to fully support Objective-C, it is not enough to add an Objective-C analyzer and factory. `CAnalyzerFactory` must be revised not to dominate for every `.h` file and to cooperate to pick the better analyzer for an `.h` file.

This is not supported now and should be tackled when there is a need for it in my opinion.

idodeclare:
```
"why not use an inline comment"
```

When it's a broad problem, I put it in a top-level review which I don't think should affect your ability as the patch owner to correlate to the source code.

```
"This is not supported now and should be tackled when there is a need for it in my opinion"
```
I have raised pull requests related to this and have existing customizations for the same. If we're supporting a plug-in model where folks can customize OpenGrok without sharing their modifications, then my existing use cases are valid too.

tulinkry:
So as I understood the main concern is that we go from order-dependent approach to order-indepenent approach and the new loaded analyzers wouldn't quite work with that. I'll look into that.

tulinkry:
> If we're supporting a plug-in model where folks can customize OpenGrok without sharing their modifications, then my existing use cases are valid too.

If you or folks already customize their OpenGrok analyzers then I don't think a plug-in model should affect your or folks's ability to customize it in any way further.

idodeclare:
@vladak asked, "I wonder how does this interacts with analysis performed from within the webapp."

The webapp can use `AnalyzerGuru` to re-analyze e.g. an historical version so the question is whether all customizations that happens while indexing are (properly) applicable when run from the webapp (including your changes).

tulinkry:
> The webapp can use `AnalyzerGuru` to re-analyze e.g. an historical version so the question is whether all customizations that happens while indexing are (properly) applicable when run from the webapp (including your changes).

Will verify.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerFramework.java
```diff
@@ -0,0 +1,402 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
```
idodeclare: This seems derived from `AnalyzerGuru` so missing attribution.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -328,19 +148,21 @@
...
+    public long getVersionNo() {
         final int ver32 = 20190211_00; // Edit comment above too!
         long ver = ver32;
-        if (customizationHashCode != 0) {
-            ver |= (long)customizationHashCode << 32;
+        final int customizations = Objects.hash(this.pluginFramework.getAnalyzersInfo().customizations.toArray());
+        if (customizations != 0) {
+            ver |= (long) customizations << 32;
         }
         return ver;
```
idodeclare: I think this value should be plug-in-aware

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     * @param file    The file to index
+     * @param path    Where the file is located (from source root)
+     * @param fa      The analyzer to use on the file
+     * @param xrefOut Where to write the xref (possibly {@code null})
+     * @throws IOException               If an exception occurs while collecting the data
+     * @throws InterruptedException      if a timeout occurs
+     * @throws ForbiddenSymlinkException if symbolic-link checking encounters
+     *                                   an ineligible link
+     */
+    public void populateDocument(Document doc, File file, String path,
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     * Get an analyzer suited to analyze a file. This function will reuse
+     * analyzers since they are costly.
+     *
+     * @param in   Input stream containing data to be analyzed
+     * @param file Name of the file to be analyzed
+     * @return An analyzer suited for that file content
+     * @throws java.io.IOException If an error occurs while accessing the data
+     *                             in the input stream.
+     */
+    public AbstractAnalyzer getAnalyzer(InputStream in, String file) throws IOException {
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     *
+     * @param in   The input stream we want to get the content type for (if we
+     *             cannot determine the content type by the filename)
+     * @param file The name of the file
+     * @return The contentType suitable for printing to
+     * response.setContentType() or null if the factory was not found
+     * @throws java.io.IOException If an error occurs while accessing the input
+     *                             stream.
+     */
+    public String getContentType(InputStream in, String file) throws IOException {
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     *
+     * @param factory    The analyzer factory for this file type
+     * @param in         The input stream containing the data
+     * @param out        Where to write the result
+     * @param defs       definitions for the source file, if available
+     * @param annotation Annotation information for the file
+     * @param project    Project the file belongs to
+     * @throws java.io.IOException If an error occurs while creating the output
+     */
+    public static void writeXref(AnalyzerFactory factory, Reader in,
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     *                    {@link Util#dumpXref(java.io.Writer, java.io.Reader, java.lang.String)}
+     * @param factory     the analyzer factory for this file type
+     * @param in          the input stream containing the data
+     * @param out         a defined instance to write
+     * @param defs        definitions for the source file, if available
+     * @param annotation  annotation information for the file
+     * @param project     project the file belongs to
+     * @throws java.io.IOException if an error occurs while creating the output
+     */
+    public static void writeDumpedXref(String contextPath,
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+    }
+
+    /**
+     * Register the given prefix for the analyzer factory.
+     *
+     * @param prefix  the prefix
+     * @param factory the factory
+     * @see AnalyzerFramework#addPrefix(String, AnalyzerFactory)
+     */
+    public void addPrefix(String prefix, AnalyzerFactory factory) {
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1139,31 +673,244 @@ private static String readOpening(InputStream in, byte[] sig)
...
+    }
+
+    /**
+     * Register the given extension for the analyzer factory.
+     *
+     * @param extension the extension
+     * @param factory   the factory
+     * @see AnalyzerFramework#addExtension(String, AnalyzerFactory) (String, AnalyzerFactory)
+     */
+    public void addExtension(String extension, AnalyzerFactory factory) {
```
idodeclare: Why relocate?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzersInfo.java
```diff
@@ -0,0 +1,180 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
```
idodeclare: Missing attributions?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -149,8 +149,8 @@ public static void main(String[] argv) {
                 PrintStream helpStream = status != 0 ? System.err : System.out;
                 helpStream.println(helpUsage);
                 if (helpDetailed) {
-                    helpStream.println(AnalyzerGuruHelp.getUsage());
-                    helpStream.println(ConfigurationHelp.getSamples());
+                    System.err.println(AnalyzerGuruHelp.getUsage(RuntimeEnvironment.getInstance().getAnalyzerGuru()));
```
idodeclare: regression

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -149,8 +149,8 @@ public static void main(String[] argv) {
...
+                    System.err.println(ConfigurationHelp.getSamples());
```
idodeclare: regression

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -216,7 +216,8 @@ private PluginType loadClass(String classname) throws ClassNotFoundException,
 
         // check for implemented interfaces or extended superclasses
         for (Class intf1 : getSuperclassesAndInterfaces(c)) {
-            if (intf1.getCanonicalName().equals(classType.getCanonicalName())
+            if (intf1.getCanonicalName() != null && // anonymous classes
+                    intf1.getCanonicalName().equals(classType.getCanonicalName())
```
idodeclare: See line 226 below which logs re `IAuthorizationPlugin` that is no longer applicable for all `PluginFramework` subclasses

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
tulinkry: right

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
tulinkry: right

tarzanek:
ordering of analyzers is important to fix (note to this issue from Krystof)

tulinkry:
I've thought about possible solutions:
 - topological sort based on hardcoded analyzer dependencies (other analyzers full name) - robust, but can require more configuration
 - inserting new analyzers to a specified index (kinda error prone when the array gets bigger)
 - scoring from the matcher for all analyzers, and then sorting based on score (I wasn't going deep in this)

vladak:
Looks like this needs some work (and rebase), converting to draft.

PR #2673, opened 2019-02-09
Feature/verilog

idodeclare:
Hello,

Please consider for integration this patch to add a Verilog analyzer requested eagerly by a couple of folks a few months ago.

Also, I patched some additional nits found while testing:

* Do not throw if source file unavailable for `Context`.
* Output `--help` to stdout not stderr, with EXIT STATUS=0.
* Show full path when logging missing xref.
* Short-circuit option `-R` if `--help|-h|-?` is used.

Thank you.


tarzanek:
lgtm, merging

PR #2681, opened 2019-02-17
check git version in GitRepository#isWorking()

### LINKED ISSUE #2664: GitRepository isWorking should check git version
As we are using features from newer git version (date format), the repository should report if it's working via the `isWorking()` method. Currently just return code is checked.

```
    @Override
    public boolean isWorking() {
        if (working == null) {
            ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);
            working = checkCmd(RepoCommand, "--help");
        }
        return working;
    }
```

tulinkry:
fixes #2664

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
@@ -41,7 +41,7 @@ import java.util.stream.Stream;
  * @see <a href="https://semver.org/">https://semver.org/</a>
  */
 public class Version implements Comparable<Version> {
-    private final Integer versions[];
+    private final Integer[] versions;
 
     /**
      * Construct the version from integer parts.
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
@@ -1,0 +1,109 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
```diff
@@ -0,0 +1,121 @@
...
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
```
vladak: I think you should have @tulinkry there

opengrok-indexer/src/test/java/org/opengrok/indexer/util/VersionTest.java
```diff
@@ -0,0 +1,66 @@
+package org.opengrok.indexer.util;
```
vladak: missing license

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Version.java
```diff
@@ -0,0 +1,123 @@
...
```
vladak: I mean you do not have to include Oracle copyright at all if all of the code comes from elsewhere.

tulinkry: I don't mind about putting oracle license there.

PR #2684, opened 2019-02-17
escaping extra html in annotate tooltip

### LINKED ISSUE #2671: Angle brackets in change set comments not handled in popup
With Annotate on, hovering over a changeset number shows a popup description of the change.  In the summary section, if the Mercurial comment includes `<foo>` it's dropped from the display.  It's in the HTML `<a title="...">` using `&#60;` and `&#62;` so something is trying, but apparently it's getting reinterpreted later.

This issue is less visibly but more frequently present in the user section of the popup, where "`Joe Developer <joe.developer@example.com>`" has the e-mail address dropped.

tulinkry:
fixes #2671


tulinkry:
screenshot 2019-02-17 at 21 54 24

Issue #1576, opened 2017-05-04
selecting single project in two groups adds it twice to the selected projects


tulinkry:
Using single project in multiple groups; when adding this projects in the project picker from both groups then this project is listed twice on the right side next to the project picker.

Note: this is just the ui bug

vladak:
I spent some non-trivial amount of time going through, debugging and modifying the JavaScript code in `searchable-option-list` (SOL) only to find out that the problem actually comes into existence already in the JSP code in `menu.jspf`:
```java
    <select tabindex="8" class="q" id="project"
        name="project" multiple="multiple" size="<%=
        Math.min(15, projectsSize) %>"><%
        SortedSet<String> pRequested = cfg.getRequestedProjects();
        for (Group group : ph.getGroups()) {
            Set<Project> groupProjects = ph.getAllGrouped(group);
            if (groupProjects.size() > 0) {
                %><optgroup label="<%= group.getName() %>"><%
                for (Project p : groupProjects) {
                    if (!p.isIndexed()) {
                        continue;
                    }

                // TODO below "selected" has no effect if one refreshes the page
                // with F5

                %><option value="<%= p.getName() %>"<%
                    if (pRequested.contains(p.getName())) {
                        %> selected="selected"<%
                    }
```

This will result in multiple `options` tags with `selected` attribute if a project is in multiple groups. The "SOL" builds the `sol-option` stuff based on the select form contents in `_detectDataFromOriginalElement()`.

vladak:
The bottom of the problem is that there is 1-1 relationship between the project items in the pull down menu (`sol-option`) and the rectangle objects (`sol-selected-display-item`). In addition to that, clicking the small `x` in the rectangle (`sol-quick-delete`) to remove the project triggers the form submission (`$('#sbox').submit()`) which initiates a search.

If the above JSP code is modified to remove a project from the `pRequested` set after its `<option>` element was generated (so that at most one group will have the option `selected`), this works reasonably well. There is little quirk, though: once all projects are selected, the rectangles will contain the project twice (for simplicity I am assuming a setup where one project is listed in 2 groups). Once one of them is deselected, the submit is triggered and now only one of them will remain thanks to the modified JSP code. Deselecting the last one will remove it and trigger the search again and they are both gone for good.

[...]

PR #2686, opened 2019-02-18
avoid selecting duplicate projects multiple times

vladak:
alleviates the problem of #1576


opengrok-web/src/main/webapp/menu.jspf
```diff
@@ -102,10 +102,11 @@ document.domReady.push(function() { domReadyMenu(); });
                 %><option value="<%= p.getName() %>"<%
                     if (pRequested.contains(p.getName())) {
                         %> selected="selected"<%
+                        pRequested.remove(p.getName());
```
tulinkry: there is the same block of code few lines below which may need this too

PR #2692, opened 2019-02-19
Vary Level for printProgress(). No redundant getFileCount().

idodeclare:
Hello,

Please consider for integration this patch to remove the redundant directory traversal while indexing and to adjust logging levels, so that `INFO` is less verbose and `FINEST` fills out.

As an example with the OpenGrok source, previously the indexing would produce logs with the following aggregate number of messages per log level:

```
FINER      3291
INFO       2828
FINEST     1499
FINE       107
WARNING    18
CONFIG     14
```

`FINER` had the most messages, with `INFO` second, and `FINE` with barely any content.

With this patch, a run will produce the aggregate levels below, with `INFO` nicely terse, `FINEST` having the largest number of messages, and `FINE` and `FINER` having about an equal number of messages.

```
FINEST     4668
FINER      1528
FINE       1468
INFO       48
WARNING    18
CONFIG     14
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/DefaultIndexChangedListener.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/DefaultIndexChangedListener.java
@@ -41,21 +41,25 @@ public class DefaultIndexChangedListener implements IndexChangedListener {
 
     @Override
     public void fileAdd(String path, String analyzer) {
-        LOGGER.log(Level.INFO, "Add: {0} ({1})", new Object[]{path, analyzer});
+        if (LOGGER.isLoggable(Level.FINE)) {
+            LOGGER.log(Level.FINE, "Add: {0} ({1})", new Object[]{path, analyzer});
+        }
     }
 
     @Override
     public void fileRemove(String path) {
-        LOGGER.log(Level.INFO, "Remove file:{0}", path);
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/DefaultIndexChangedListener.java
```diff
@@ -41,21 +41,25 @@
 
     @Override
     public void fileAdd(String path, String analyzer) {
-        LOGGER.log(Level.INFO, "Add: {0} ({1})", new Object[]{path, analyzer});
+        if (LOGGER.isLoggable(Level.FINE)) {
```
tulinkry: why do you use if condition here and not in fileRemove or fileUpdate?

idodeclare: To short-circuit `Object` creation happening before the `log()` call, which the others are not doing

tulinkry: ah, ok.

tulinkry:
Could you rebase on current master and push so the tests are restared? The failure in wercker does not seem to be related.

idodeclare:
OK

tulinkry:
@vladak review & merge

vladak:
Looks good to me.

PR #2695, opened 2019-02-19
ie10-11 js fix

### LINKED ISSUE #2694: IE10 js error
opengrok 1.2.2
1. Object doesn't support property or method 'includes' -  utils-0.0.29.js:1787
2. Object doesn't support property or method "startsWith" -  utils-0.0.29.js:1868
3. Object doesn't support property or method "isInteger" - utils-0.0.29.js:1772

Teriand:
fixes #2694


vladak:
Pls do the move of the file to the new version.

opengrok-web/src/main/webapp/js/utils-0.0.29.js
```diff
@@ -2390,3 +2390,23 @@ function getCookie(cname) {
...
+    {
+        return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
+    };
+}
+
+if (!String.prototype.startsWith) {
+  String.prototype.startsWith = function(searchString, position) {
+    position = position || 0;
+    return this.indexOf(searchString, position) === position;
+  };
```
tulinkry: there's only one match, isn't it better to replace it inline?

opengrok-web/src/main/webapp/js/utils-0.0.29.js
```diff
@@ -2390,3 +2390,23 @@ function getCookie(cname) {
...
+//polyfills for IE10-11
+// References:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger
+
+
+if (!Number.isInteger)
+{
+    Number.isInteger = function numberIsInteger(value)
+    {
+        return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
```
tulinkry: there is only one match, isn't it better to do it inline or use some kind of alternative with parseInt() === NaN ?

vladak:
Also, use `fixes #2694` in the commit message for autoclose on merge.

opengrok-web/src/main/webapp/js/utils-0.0.29.js
```diff
@@ -1754,7 +1754,7 @@ function initAutocomplete(config, minisearch) {
 }
 
 function initMinisearchAutocomplete(config) {
-    if (config.allowedFields && !config.allowedFields.includes('full')) {
+    if (config.allowedFields && config.allowedFields.indexOf('full')<0) {
```
tulinkry: there should be a space `ull')<0)` -> `ull') < 0)`

opengrok-web/src/main/webapp/js/utils-0.0.29.js
```diff
@@ -1784,7 +1784,7 @@ function initMinisearchAutocomplete(config) {
 }
 
 function initAutocompleteForField(inputId, field, config, dataFunction, errorElemId) {
-    if (config.allowedFields && !config.allowedFields.includes(field)) {
+    if (config.allowedFields && config.allowedFields.indexOf(field)<0) {
```
tulinkry: there should be a space `field')<0)` -> `field') < 0)`

Teriand: thx, fixed

opengrok-web/src/main/webapp/js/utils-0.0.29.js
Teriand: thx, fixed

opengrok-web/src/main/webapp/js/utils-0.0.29.js
Teriand: i can change to simple
```
if (event.originalEvent.originalEvent.type.indexOf('key') == 0) {
```

opengrok-web/src/main/webapp/js/utils-0.0.29.js
Teriand: if caretPos need only simple number check,
i can change to 
```
typeof caretPos === 'number'
```

tulinkry:
Looks ok, can we merge?

PR #2697, opened 2019-02-21
specify required python version

vladak:
The minimum version is set according to our internal deployment, i.e. the version it is most tested with.


tulinkry:
The description does not match the change.

vladak:
Oops, forgot to commit.

PR #2701, opened 2019-02-26
avoid returning null objects from getIndexSearcher()

vladak:
This change corrects the semantics of the outcome of getIndexSearcher() - it should either raise exception or return null but now both.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
@@ -469,8 +469,7 @@ public class SearchHelper {
             Suggestion s = new Suggestion(proj);
             try {
                 if (!closeOnDestroy) {
-                    SuperIndexSearcher searcher;
-                    searcher = RuntimeEnvironment.getInstance().getIndexSearcher(proj);
+                    SuperIndexSearcher searcher = RuntimeEnvironment.getInstance().getIndexSearcher(proj);
                     searcherList.add(searcher);
                     ir = searcher.getIndexReader();
                 } else {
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
```diff
@@ -469,7 +469,8 @@ private void getSuggestion(Term term, IndexReader ir,
             Suggestion s = new Suggestion(proj);
             try {
                 if (!closeOnDestroy) {
-                    SuperIndexSearcher searcher = RuntimeEnvironment.getInstance().getIndexSearcher(proj);
+                    SuperIndexSearcher searcher;
+                    searcher = RuntimeEnvironment.getInstance().getIndexSearcher(proj);
```
tulinkry: what is the benefit in this change?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
```diff
@@ -503,7 +504,7 @@ private void getSuggestion(Term term, IndexReader ir,
                 }
             } catch (IOException e) {
                 LOGGER.log(Level.WARNING, "Got exception while getting "
-                        + "spelling suggestions: ", e);
+                        + "spelling suggestions for project " + proj + " :", e);
```
tulinkry: this is not important but I think there's an extra " " space before the ":"

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
vladak: ah, that's a leftover from earlier modification.

PR #2703, opened 2019-02-27
copy projects before modifying them

vladak:
No exceptions seen any more when performing searches with no match.

### DIFF
--- a/opengrok-web/src/main/webapp/menu.jspf
+++ b/opengrok-web/src/main/webapp/menu.jspf
@@ -86,8 +86,7 @@ document.domReady.push(function() { domReadyMenu(); });
     <select tabindex="8" class="q" id="project"
         name="project" multiple="multiple" size="<%=
         Math.min(15, projectsSize) %>"><%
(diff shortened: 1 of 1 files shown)


opengrok-web/src/main/webapp/menu.jspf
```diff
@@ -86,7 +86,8 @@ document.domReady.push(function() { domReadyMenu(); });
     <select tabindex="8" class="q" id="project"
         name="project" multiple="multiple" size="<%=
         Math.min(15, projectsSize) %>"><%
-        SortedSet<String> pRequested = cfg.getRequestedProjects();
+        SortedSet<String> pRequested = new TreeSet<>();
+        pRequested.addAll(cfg.getRequestedProjects());
```
tulinkry: how about `SortedSet<String> pRequested = new TreeSet<>(cfg.get...)`

vladak:
My only worry is that the proper solution will make the project picker yet more slower.

PR #2709, opened 2019-03-08
fail mirroring if incoming check raises exception

vladak:
One of our Mercurial projects failed to sync and this was not noticed in the logs because the `opengrok-mirror` script was run with `-I`. One problem is that incoming check failures are treated as soft, second problem is that Mercurial exit codes are not interpreted correctly - in this case `hg incoming` exits with 255 due to authentication failure.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/scm/mercurial.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/scm/mercurial.py
@@ -109,7 +109,7 @@ class MercurialRepository(Repository):
         cmd.execute()
         self.logger.debug(cmd.getoutputstr())
         retcode = cmd.getretcode()
-        if cmd.getstate() != Command.FINISHED or retcode not in [0,1]:
+        if cmd.getstate() != Command.FINISHED or retcode not in [0, 1]:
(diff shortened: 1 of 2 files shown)


opengrok-tools/src/main/python/opengrok_tools/scm/mercurial.py
```diff
@@ -108,12 +108,13 @@ def incoming(self):
                               env_vars=self.env, logger=self.logger)
         cmd.execute()
         self.logger.debug(cmd.getoutputstr())
-        if cmd.getstate() != Command.FINISHED:
+        retcode = cmd.getretcode()
+        if cmd.getstate() != Command.FINISHED or retcode not in [0,1]:
```
tulinkry: I don't get this, why `1` is special?

vladak: This is what Mercurial man page says for the `incoming` command:
```
       Returns 0 if there are incoming changes, 1 otherwise.
```

tulinkry: aha, perfect

Issue #2710, opened 2019-03-11
Enable Cross Origin Requests for OpenGrok


wy193777:
OpenGrok already has REST API provided for users. So cross origin requests should be enabled by default otherwise other services will not be able to take advantage of REST API.

vladak:
I guess this is about adding a CORS filter to `web.xml` + make it configurable.

tulinkry:
This should be part of the api only, so some filter for the REST API.
Currently public API is only:
 - SearchController.PATH
 - SuggesterController.PATH
 - SuggesterController.PATH + "/config"

other paths are available only for localhost so CORS filter does not make sense I think.

vladak:
Good point.

Dne út 12. 3. 2019 13:04 uživatel Kryštof Tulinger <notifications@github.com>
napsal:

wy193777:
Can we include this change to newest release? It's really hard to config the opengrok. I just want to run it with docker container.

vladak:
Yep, will make a release soonish.

Dne st 20. 3. 2019 20:44 uživatel Shenghan Gao <notifications@github.com>
napsal:

vladak:
Also, I wonder what makes it so hard for you to configure it ?

Dne st 20. 3. 2019 21:07 uživatel Vladimír Kotal <vlada@devnull.cz> napsal:

wy193777:
1. To let OpenGrok run properly, first need to install several dependencies, such as ctags, tomcat. I already had them.
2. After installed those dependencies, I need to unzip the build result, copy paste the source.war to Tomcat path.
3. After the application is up and running, I also have to manually edit `web.xml` and a configuration file somewhere. 
4. I also have to manually run indexer to index code. 

None of them is hard, but all steps combined take at least one or two hours to setup. Also sometimes there are conditions on local machine, then the configuration might took someone forever(my experience with tomcat). As a developer spoiled by `npm`, `python` and other tools, I would expect apps can run with copy paste a few commands.

wy193777:
Maybe we can add a Dockerfile for people to easily run the application from source code.

vladak:
That will probably happen in some time by folding the OpenGrok/docker repo
onto the main OpenGrok repo.

Dne st 20. 3. 2019 22:21 uživatel Shenghan Gao <notifications@github.com>
napsal:

PR #2711, opened 2019-03-13
Enable CORS for RESTful API

wy193777:
Trying to solve #2710 . Cannot run the app with `../mvnw tomcat7:redeploy` to test because always error. Already got tomcat7 up and running.
```
[ERROR] Failed to execute goal org.apache.tomcat.maven:tomcat7-maven-plugin:2.2:redeploy (default-cli) on project opengrok-web: Cannot invoke Tomcat manager: Connection reset -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```


opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    /**
+     * A preflight request is an OPTIONS request
+     * with an Origin header.
+     */
+    private static boolean isPreflightRequest(ContainerRequestContext request) {
+        return request.getHeaderString("Origin") != null
+                && request.getMethod().equalsIgnoreCase("OPTIONS");
```
tulinkry: prefight request per spec is not only OPTIONS with origin but also with special headers

> It is an OPTIONS request, using three HTTP request headers: Access-Control-Request-Method, Access-Control-Request-Headers, and the Origin header. (https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)

tulinkry: And in our special case we only support GET for the selected paths so essentially the prefight request would not be needed at all.

opengrok-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -47,6 +47,16 @@
...
     </filter>
+    <filter>
+        <filter-name>CorsFilter</filter-name>
+        <filter-class>org.opengrok.web.CorsFilter</filter-class>
+    </filter>
+    <filter-mapping>
+        <filter-name>CorsFilter</filter-name>
+        <url-pattern>/search</url-pattern>
+        <url-pattern>/suggest</url-pattern>
+        <url-pattern>/suggest/config</url-pattern>
```
tulinkry: it shouldn't be here (see my general comment on this PR) and the paths are `/api/v1/search` and so on.

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        if (isPreflightRequest(request)) {
+            response.getHeaders().add("Access-Control-Allow-Credentials", "true");
+            response.getHeaders().add("Access-Control-Allow-Methods",
+                    "GET, POST, PUT, DELETE, OPTIONS, HEAD");
+            response.getHeaders().add("Access-Control-Allow-Headers",
+                    // Whatever other non-standard/safe headers (see list above)
+                    // you want the client to be able to send to the server,
+                    // put it in this list. And remove the ones you don't want.
+                    "X-Requested-With, Authorization, " +
+                            "Accept-Version, Content-MD5, CSRF-Token");
```
tulinkry: can you extract this list as a private static set/list/array and use the value here?

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        if (request.getHeaderString("Origin") == null) {
+            return;
+        }
+
+        // If it is a preflight request, then we add all
+        // the CORS headers here.
+        if (isPreflightRequest(request)) {
+            response.getHeaders().add("Access-Control-Allow-Credentials", "true");
+            response.getHeaders().add("Access-Control-Allow-Methods",
+                    "GET, POST, PUT, DELETE, OPTIONS, HEAD");
```
tulinkry: for the mentioned paths we support only `GET`, although I'm not against leaving this here in the full power

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+                    // put it in this list. And remove the ones you don't want.
+                    "X-Requested-With, Authorization, " +
+                            "Accept-Version, Content-MD5, CSRF-Token");
+        }
+
+        // Cross origin requests can be either simple requests
+        // or preflight request. We need to add this header
+        // to both type of requests. Only preflight requests
+        // need the previously added headers.
+        response.getHeaders().add("Access-Control-Allow-Origin", "*");
```
tulinkry: also quoting from the mozzila spec

> **Credentialed requests and wildcards**

> When responding to a credentialed request, the server must specify an origin in the value of the Access-Control-Allow-Origin header, instead of specifying the "*" wildcard.

> Because the request headers in the above example include a Cookie header, the request would fail if the value of the Access-Control-Allow-Origin header were "*". But it does not fail: Because the value of the Access-Control-Allow-Origin header is "http://foo.example" (an actual origin) rather than the "*" wildcard, the credential-cognizant content is returned to the invoking web content

so if you are to support crendentialized request, you need to add some additional logic here

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        // If it is a preflight request, then we add all
+        // the CORS headers here.
+        if (isPreflightRequest(request)) {
+            response.getHeaders().add("Access-Control-Allow-Credentials", "true");
+            response.getHeaders().add("Access-Control-Allow-Methods",
+                    "GET, POST, PUT, DELETE, OPTIONS, HEAD");
+            response.getHeaders().add("Access-Control-Allow-Headers",
+                    // Whatever other non-standard/safe headers (see list above)
```
tulinkry: which list?

vladak:
Is it possible to add tests for this change ? (Mockito ?)

wy193777:
> for

Need to learn how to use it? Is OpenGrok code base already included a test about HTTP response so I can write a similar one.

vladak:
bunch of the tests already use Mockito like `WebappListenerTest`, `LocalhostFilterTest.java` etc.

I think this should be just a matter of supplying the headers via `when().thenReturn()` like suggested in https://stackoverflow.com/questions/37314469/how-to-mock-httpservletrequest-with-headers

wy193777:
The test command listed on [Wiki](https://github.com/oracle/opengrok/wiki/Developer-intro) is not working at all. Actually I tried run `./mvnw test` with some definitely failed tests and nothing failed.

wy193777:
Strange, why I cannot get test failed locally...

wy193777:
Tried my best to run it on my machine but failed. The information provided on [the wiki page](https://github.com/oracle/opengrok/wiki/Developer-intro) isn't so helpful for me. Always get errors when try to run `tomcat7:run` or `tomcat7:redeploy`.

vladak:
The Maven Tomcat plugin setup needs revisiting. I was not pressed to make it working because IDEA Ultimate edition (as opposed to the Community edition) does the redeploy for me.

vladak:
I verified that the Maven Tomcat plugin steps actually work  - I was able to successfully redeploy the app:
```
opengrok-web (master *)$ ../mvnw -DskipTests=true tomcat7:redeploy
[INFO] Scanning for projects...
[INFO] 
[INFO] ---------------------< org.opengrok:opengrok-web >----------------------
[INFO] Building OpenGrok Web 1.2.4
[INFO] --------------------------------[ war ]---------------------------------

...

[INFO] --- tomcat7-maven-plugin:2.2:redeploy (default-cli) @ opengrok-web ---
[INFO] Deploying war to http://localhost:8080/source  
Uploading: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true
Uploaded: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true (25200 KB at 12456.4 KB/sec)

[INFO] tomcatManager status code:200, ReasonPhrase:OK
[INFO] OK - Deployed application at context path /source
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  26.820 s
[INFO] Finished at: 2019-03-15T14:46:09+01:00
[INFO] ------------------------------------------------------------------------
```

how exactly it is failing on your end ?

opengrok-web/src/test/java/org/opengrok/web/CorsFilterTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    public void CorsTest() {
+        testBoth("https://example.org", Arrays.asList("*"));
+    }
+
+    private void testBoth(String origin, List<Object> headerValue) {
+        CorsFilter filter = new CorsFilter();
+        ContainerRequestContext request = mock(ContainerRequestContext.class);
+        when(request.getHeaderString("Origin")).thenReturn(origin);
```
vladak: use `CORS_REQUEST_HEADER`

vladak:
Also, you will need to get an OCA (https://www.oracle.com/technetwork/community/oca-486395.html). For starters, just state you are okay with the terms and initiate the process please.

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
vladak: I don't see pre-flight requests being handled by the current implementation.

vladak:
Also, please add `fixes #2710` into commit comments.

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
tulinkry: Cause it's not needed.

wy193777:
Here is my output when run `../mvnw -DskipTests=true tomcat7:redeploy`. I installed tomcat7 from apt-get.
```
[INFO] --- tomcat7-maven-plugin:2.2:redeploy (default-cli) @ opengrok-web ---
[INFO] Deploying war to http://localhost:8080/source  
Uploading: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true
2056/34190 KB   
[INFO] I/O exception (java.net.SocketException) caught when processing request: Connection reset
[INFO] Retrying request
Uploading: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true
2056/34190 KB   
[INFO] I/O exception (java.net.SocketException) caught when processing request: Connection reset
[INFO] Retrying request
Uploading: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true
2056/34190 KB   
[INFO] I/O exception (java.net.SocketException) caught when processing request: Connection reset
[INFO] Retrying request
Uploading: http://localhost:8080/manager/text/deploy?path=%2Fsource&update=true
2054/34190 KB   
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.100 s
[INFO] Finished at: 2019-03-15T11:57:50-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.tomcat.maven:tomcat7-maven-plugin:2.2:redeploy (default-cli) on project opengrok-web: Cannot invoke Tomcat manager: Connection reset -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```
My `/var/lib/tomcat7/conf/tomcat-users.xml`:
```xml
<?xml version='1.0' encoding='utf-8'?>
  <role rolename="admin"/>
  <role rolename="manager-script"/>
  <user password="admin1" roles="manager-script,admin" username="admin1"/>
</tomcat-users>
```

My `~/.m2/settings.xml`:
```xml
<settings>
  <servers>
    <server>
      <id>OpenGrok</id>
      <username>admin1</username> <!-- you can use different name -->
      <password>admin1</password> <!-- you can use different password -->
    </server>
  </servers>
</settings>
```
I think it is related to how I install tomcat or something.

wy193777:
Here is tomcat localhost access log:
```
127.0.0.1 - - [15/Mar/2019:11:57:50 -0700] "PUT /manager/text/deploy?path=%2Fsource&update=true HTTP/1.1" 403 979
127.0.0.1 - - [15/Mar/2019:11:57:50 -0700] "PUT /manager/text/deploy?path=%2Fsource&update=true HTTP/1.1" 403 979
127.0.0.1 - - [15/Mar/2019:11:57:50 -0700] "PUT /manager/text/deploy?path=%2Fsource&update=true HTTP/1.1" 403 979
127.0.0.1 - - [15/Mar/2019:11:57:50 -0700] "PUT /manager/text/deploy?path=%2Fsource&update=true HTTP/1.1" 403 979
```

wy193777:
Also, can we move wiki to source code so contributors can improve wiki by create pull requests.

wy193777:
I'm okay with the terms and already submitted the signed agreement.

wy193777:
Got some jsp error after manually deploy source.war to tomcat.

```Mar 15, 2019 10:55:40 AM org.apache.catalina.core.StandardContext loadOnStartup
SEVERE: Servlet [lister] in web application [/source] threw load() exception
org.apache.jasper.JasperException: Unable to compile class for JSP: 

An error occurred at line: 189 in the jsp file: /list.jsp
Type mismatch: cannot convert from FileAnalyzerFactory to AnalyzerFactory
186:                 BufferedInputStream bin =
187:                     new BufferedInputStream(new FileInputStream(resourceFile));
188:                 try {
189:                     AnalyzerFactory a = AnalyzerGuru.find(basename);
190:                     AbstractAnalyzer.Genre g = AnalyzerGuru.getGenre(a);
191:                     if (g == null) {
192:                         a = AnalyzerGuru.find(bin);


An error occurred at line: 190 in the jsp file: /list.jsp
The method getGenre(String) in the type AnalyzerGuru is not applicable for the arguments (AnalyzerFactory)
187:                     new BufferedInputStream(new FileInputStream(resourceFile));
188:                 try {
189:                     AnalyzerFactory a = AnalyzerGuru.find(basename);
190:                     AbstractAnalyzer.Genre g = AnalyzerGuru.getGenre(a);
191:                     if (g == null) {
192:                         a = AnalyzerGuru.find(bin);
193:                         g = AnalyzerGuru.getGenre(a);


An error occurred at line: 192 in the jsp file: /list.jsp
Type mismatch: cannot convert from FileAnalyzerFactory to AnalyzerFactory
189:                     AnalyzerFactory a = AnalyzerGuru.find(basename);
190:                     AbstractAnalyzer.Genre g = AnalyzerGuru.getGenre(a);
191:                     if (g == null) {
192:                         a = AnalyzerGuru.find(bin);
193:                         g = AnalyzerGuru.getGenre(a);
194:                     }
195:                     if (g == AbstractAnalyzer.Genre.IMAGE) {


An error occurred at line: 193 in the jsp file: /list.jsp
The method getGenre(String) in the type AnalyzerGuru is not applicable for the arguments (AnalyzerFactory)
190:                     AbstractAnalyzer.Genre g = AnalyzerGuru.getGenre(a);
191:                     if (g == null) {
192:                         a = AnalyzerGuru.find(bin);
193:                         g = AnalyzerGuru.getGenre(a);
194:                     }
195:                     if (g == AbstractAnalyzer.Genre.IMAGE) {
196: %>


An error occurred at line: 219 in the jsp file: /list.jsp
The method writeDumpedXref(String, FileAnalyzerFactory, Reader, Writer, Definitions, Annotation, Project) in the type AnalyzerGuru is not applicable for the arguments (String, AnalyzerFactory, Reader, JspWriter,
 Definitions, Annotation, Project)
216:                         // SRCROOT is read with UTF-8 as a default.
217:                         r = IOUtils.createBOMStrippedReader(bin,
218:                             StandardCharsets.UTF_8.name());
219:                         AnalyzerGuru.writeDumpedXref(request.getContextPath(), a,
220:                                 r, out, defs, annotation, project);
221:     %></pre>
222: </div><%


An error occurred at line: 250 in the jsp file: /list.jsp
Type mismatch: cannot convert from FileAnalyzerFactory to AnalyzerFactory
247:             }
248:         } else {
249:             // requesting a previous revision or needed to generate xref on the fly (economy mode).
250:             AnalyzerFactory a = AnalyzerGuru.find(basename);
```

wy193777:
@vladak I just added needed tests and also tested the app locally.

opengrok-web/src/main/java/org/opengrok/web/CorsEnable.java
```diff
@@ -0,0 +1,12 @@
...
+import javax.ws.rs.NameBinding;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@NameBinding
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface CorsEnable { }
```
tulinkry: could you document it with a javadoc?

opengrok-web/src/main/java/org/opengrok/web/api/v1/RestApp.java
```diff
@@ -33,7 +34,9 @@
     public static final String API_PATH = "/api/v1";
     
     public RestApp() {
+        System.setProperty("sun.net.http.allowRestrictedHeaders", "true"); // necessary to test CORS from controllers
         register(new SuggesterAppBinder());
+        register(CorsFilter.class);
```
tulinkry: if you move it to `org.opengrok.web.api.v1` then this is no longer necessary

opengrok-web/src/main/java/org/opengrok/web/api/v1/RestApp.java
```diff
@@ -33,7 +34,9 @@
     public static final String API_PATH = "/api/v1";
     
     public RestApp() {
+        System.setProperty("sun.net.http.allowRestrictedHeaders", "true"); // necessary to test CORS from controllers
```
tulinkry: can't you set it somewhere it tests? so the production code is not poluted with this

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+        repository = new TestRepository();
+
+        repository.create(SearchControllerTest.class.getResourceAsStream("/org/opengrok/indexer/index/source.zip"));
+
+        env.setHistoryEnabled(false);
+        env.setProjectsEnabled(true);
+        Indexer.getInstance().prepareIndexer(env, true, true,
+                false, null, null);
+        env.setDefaultProjectsFromNames(Collections.singleton("__all__"));
+        Indexer.getInstance().doIndexerExecution(true, null, null);
```
tulinkry: is it necessary to run the indexer to test simple cors header in the response?

opengrok-web/src/main/java/org/opengrok/web/api/v1/RestApp.java
tulinkry: or like this https://stackoverflow.com/questions/32137121/system-setpropertysun-net-http-allowrestrictedheaders-true-is-not-working you can set it for the surefire argline.

opengrok-web/src/main/java/org/opengrok/web/CorsFilter.java
vladak: how so ?

tulinkry: Based on the doc I sent in other responses here https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests. The cors has two request categories

1. simple requests
2. prefighted requests

Quoting the mozilla spec: **simple request meets all the following conditions**:


    1. The only allowed methods are:
        GET
        HEAD
        POST
    2 .Apart from the headers set automatically by the user agent (for example, Connection, User-Agent, or any of the other headers with names defined in the Fetch spec as a “forbidden header name”), the only headers which are allowed to be manually set are those which the Fetch spec defines as being a “CORS-safelisted request-header”, which are:
        Accept
        Accept-Language
        Content-Language
        Content-Type (but note the additional requirements below)
        DPR
        Downlink
        Save-Data
        Viewport-Width
        Width
    3. The only allowed values for the Content-Type header are:
        application/x-www-form-urlencoded
        multipart/form-data
        text/plain
    4. No event listeners are registered on any XMLHttpRequestUpload object used in the request; these are accessed using the XMLHttpRequest.upload property.
    5. No ReadableStream object is used in the request.


For all paths which should be covered with CORS on our side, all uses GET method (1), don't have body (so Content-Type header is needless) (2+3), don't upload files (4), don't need to use streaming api for receiving simple html page (5). Therefore prefight request aren't necessary.

vladak:
As for the wiki, if you scroll down to the bottom of the wiki pages, there is a link to clone the repository (https://github.com/oracle/opengrok.wiki.git). Not sure how that works w.r.t. pull requests so give it a try please.

vladak:
This looks good. If anybody needs the CORS value to be configurable, please raise another PR.

tulinkry:
There's more complex solution if the need arises https://static.javadoc.io/com.thetransactioncompany/cors-filter/2.6/com/thetransactioncompany/cors/CORSFilter.html.

PR #2712, opened 2019-03-13
annotate output from commands

vladak:
This change provides annotation of the commands run by the `mirror.py` script.


opengrok-tools/src/main/python/opengrok_tools/scm/mercurial.py
```diff
@@ -45,6 +45,7 @@ def get_branch(self):
         cmd = self.getCommand(hg_command, work_dir=self.path,
                               env_vars=self.env, logger=self.logger)
         cmd.execute()
+        self.logger.info("output of " + cmd)
```
tulinkry: you missed the colon here

PR #2720, opened 2019-03-17
adding python tool specific version

tulinkry:


opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -49,6 +49,9 @@
 from .utils.repofactory import get_repository
 from .utils.utils import is_exe, check_create_dir, get_int, diff_list, \
     is_web_uri
+
+# do not reorder this import, it must be imported after utils.
+# (for me) idea reorders the import to the top, causing an import error
```
tulinkry: see #2721

PR #2721, opened 2019-03-17
Moving repofactory to scm module

tulinkry:
depends on #2720

Removing the cross module dependency forcing the particular order of imports.


vladak:
merge pls

tulinkry:
rebased

PR #2722, opened 2019-03-17
renaming modules to align with the rest of the codebase

tulinkry:
Renaming modules to align with the rest of the code base
 
- renaming plugins to opengrok-plugins
 - renaming suggester to opengrok-suggester

We can close this if you think it's not relevant.


ahornace:
I personally would do it the other way around. `opengrok` is implicit in this case.

vladak:
I feel the opengrok prefix is redundant too. It certainly does not help with command line completion.

tulinkry:
I don’t mind either way but I just want it consistent. Let me close it and make it he other way around. That will possibly make it hard for folks to rebase.

PR #2725, opened 2019-03-19
Suppressing the double render problem on error

tulinkry:
Using set instead of lists does not add the same record twice.

The records were doubled because (assume we get an exception on index.jsp page):
1. we partially render the index.jsp
2. it includes httpheader.jspf which registers the scripts into PageConfig
3. exception occurs, the error page is being rendered instead
4. error page calls httpheader.jspf again and it adds the script again to the same PageConfig (because the request has not changed)
5. error page renders the scripts twice

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
@@ -23,13 +23,13 @@
  */
 package org.opengrok.indexer.web;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
 import java.util.TreeMap;
+import java.util.TreeSet;
 
 /**
  * A list-like container for javascripts in the page.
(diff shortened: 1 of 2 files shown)


vladak:
Nice rootcausing !

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
```diff
@@ -111,11 +111,14 @@ public String toHtml() {
...
+            .comparingInt(Script::getPriority)
+            .thenComparing(Script::getScriptData);
+
     /**
-     * Scripts which will be written to the page. We assume that the length
-     * could be the same as for {@link #SCRIPTS}.
+     * Scripts which will be written to the page.
      */
-    private final List<Script> outputScripts = new ArrayList<>(SCRIPTS.size());
+    private final NavigableSet<Script> outputScripts = new TreeSet<>(SCRIPTS_COMPARATOR);
```
vladak: why exactly do you need NavigableSet ?

vladak:
also, add your copyright.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
tulinkry: to have it sorted based on the priority, because some scripts are dependent on each other

tulinkry:
These are my files :-D. I'm there as an author already.

PR #2726, opened 2019-03-20
Fixing git viewing renamed revision

tulinkry:
I noticed that viewing revision of renamed files was broken. The getHistory expects a full path file while the find renamed name returns the name relative to the repository directory. This leads that for renamed files the revision was never found as the filename after renaming was invalid.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -215,10 +215,14 @@ public class GitRepository extends Repository {
              * not exist or internal git error occured.
              */
             result.success = (status == 0);
-        } catch (Exception exp) {
+        } catch (Exception exception) {
             LOGGER.log(Level.SEVERE,
-                    "Failed to get history for file {0} in revision {1}: ",
-                        new Object[]{fullpath, rev, exp.getClass().toString(), exp});
+                    String.format(
+                            "Failed to get history for file %s in revision %s:",
+                            fullpath, rev
+                    ),
+                    exception
+            );
         }
         return result;
     }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -196,8 +196,9 @@ public class GitRepository extends Repository {
              * Be careful, git uses only forward slashes in its command and output (not in file path).
              * Using backslashes together with git show will get empty output and 0 status code.
              */
-            String filename = fullpath.substring(getDirectoryName().length() + 1)
-                                      .replace(File.separatorChar, '/');
+            String filename = Paths.get(getDirectoryName()).relativize(Paths.get(fullpath))
+                                   .toString()
+                                   .replace(File.separatorChar, '/');
             ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);
             String[] argv = {
                 RepoCommand,
@@ -263,8 +264,12 @@ public class GitRepository extends Repository {
                 });
                 return false;
             }
-            if (origpath != null && !origpath.equals(fullpath)) {
-                result = getHistoryRev(sink, origpath, rev);
+
+            if (origpath != null) {
+                final String fullRenamedPath = Paths.get(getDirectoryName(), origpath).toString();
+                if (!fullRenamedPath.equals(fullpath)) {
+                    result = getHistoryRev(sink, fullRenamedPath, rev);
+                }
             }
         }
 
@@ -325,7 +330,7 @@ public class GitRepository extends Repository {
             ABBREV_LOG,
             "--abbrev-commit",
             "--name-status",
-            "--pretty=format:commit %h%b",
+            "--pretty=format:commit %h%n%d",
             "--",
             fileInRepo
         };
@@ -367,7 +372,7 @@ public class GitRepository extends Repository {
                     new Object[]{fullpath, String.valueOf(status), changeset});
             return null;
         }
-        
+
         return originalFile;
     }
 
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -321,7 +327,7 @@ protected String findOriginalName(String fullpath, String changeset)
             ABBREV_LOG,
             "--abbrev-commit",
             "--name-status",
-            "--pretty=format:commit %h%b",
+            "--pretty=format:commit %h%n%d",
```
tulinkry: %h - commit hash
%n - new line
%d - file references (renames, add, remove)

vladak:
Would be nice if this could be covered by tests like Mercurial.

tulinkry:
Renamed files are covered with tests, but I don't get why this wasn't caught.

vladak:
It would be good to know. Otherwise the tests cannot be trusted.

tulinkry:
The test are not actually testing the behaviour of findRenamedFile. Let me fix it.

tulinkry:
Rewrote the tests so they're meaning full.

tulinkry:
This effectivelly swallowed the exception so noone could really say what was wrong.
https://github.com/oracle/opengrok/blob/f98c36aeb22b23e626acb37c4c02db1c74b485d9/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java#L218-L222

For this particular cases it mostly way IndexOutOBException for the substring operation.

tulinkry:
I don't know why javadoc hates me suddenly :D

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -215,10 +217,14 @@ private HistoryRevResult getHistoryRev(
              * not exist or internal git error occured.
              */
             result.success = (status == 0);
-        } catch (Exception exp) {
+        } catch (Exception exception) {
```
vladak: you mentioned that the out-of-bounds exception was swalloed by this catch. Is using `Exception` maybe too broad then ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: so this in fact does not change the output, or does it ?

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
```diff
@@ -350,47 +337,44 @@ public void testGetHistoryForTwiceRenamed() throws Exception {
...
-                + "}\n";
+        runRenamedTest(Paths.get("moved", "renamed2.c").toString(), "67dfbe26", new_content);
+        runRenamedTest(Paths.get("moved", "renamed2.c").toString(), "1086eaf5", new_content);
+
+        runRenamedTest(Paths.get("moved", "renamed.c").toString(), "1086eaf5", new_content);
 
-        runRenamedTest(Paths.get("moved", "renamed.c").toString(), "b641394", exp_str);
-        runRenamedTest(Paths.get("renamed.c").toString(), "ce4c98e", exp_str);
+        // old content (before revision b6413947 inclusively)
+        runRenamedTest(Paths.get("moved2", "renamed2.c").toString(), "b6413947", old_content);
```
vladak: these look like a candidate for array + forEach() ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: or move the `filename` assignment outside of the `try` block ?

tulinkry: No to the problem was it wasn't even logged with the bad logger parameters.

tulinkry: I just changed it so it's being logged

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
tulinkry: Not really.
%b is too verbose - body + files
%d is just the files

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
tulinkry: Is it more readable though? I would leave it this way so a test failure can point to a particular line and also it's not that much lines.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: Okay. Still large try/catch blocks with generic catch make me nervous.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
vladak: If you complement the assert with a message it should be enough.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: I see.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
tulinkry: done

PR #2730, opened 2019-03-21
first stab on lucene 8.0.0

tarzanek:
lucene 8.0 by default fetches only max first 1000 hits
also scoring was changed

This request is missing:
1) do we want to detect after search if we have more than 1000 hits and requery? (I assume yes, will add it)
2) what score should be in suggesters?  getMaxScore (especially in the sloppy one) ? @ahornace ?

Docs: http://lucene.apache.org/core/8_0_0/changes/Changes.html#v8.0.0 (and https://github.com/apache/lucene-solr/blob/branch_8_0/lucene/MIGRATE.txt and patches that are inside issues on changes page, some of them miss chains though)

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.opengrok</groupId>
     <artifactId>opengrok-top</artifactId>
-    <version>1.2.8</version>
+    <version>1.2.9</version>
   </parent>
 
   <artifactId>opengrok-dist</artifactId>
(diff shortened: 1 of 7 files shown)


tarzanek:
and the test fails - org.apache.lucene.index.IndexFormatTooOldException: Format version is not supported (resource BufferedChecksumIndexInput(MMapIndexInput(path="/private/var/folders/6r/6kynxrpn1dzdjzgr6t_s7kx80000gn/T/data5012722428846211249/index/segments_3"))): 6 (needs to be between 7 and 9). This version of Lucene only supports indexes created with release 6.0 and later.

so oldIndex.zip we bundle has 5.0 indexes, so we will need to bump it (or bundle https://lucene.apache.org/core/8_0_0/backward-codecs/overview-summary.html ?)

PR #2731, opened 2019-03-22
make it possible to mirror more projects at once

vladak:
`mirror.py` was always good candidate for refactoring. I used the opportunity to add the capability of mirroring multiple projects in one run. The intention is to use it in the Docker script (https://github.com/OpenGrok/docker/issues/27).

This change also adds `--all` switch mirrors all indexed projects. This is mutually exclusive with project arguments of the script.


vladak:
Another hopefully positive side effect is that configuration is now checked before the run.

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -0,0 +1,400 @@
...
+        if project_config.get(DISABLED_PROPERTY):
+            logger.info("Project {} disabled, exiting".
+                        format(project_name))
+            return CONTINUE_EXITVAL
+
+    #
+    # Cache the repositories first. This way it will be known that
+    # something is not right, avoiding any needless pre-hook run.
+    #
+    ret = 0
```
tulinkry: you start to use symbolic values here, can you convert 0 and 1 to some names as well?

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -0,0 +1,400 @@
...
+
+    # Check if any of the repositories contains incoming changes.
+    if check_incoming:
+        got_incoming = False
+        for repo in repos:
+            try:
+                if repo.incoming():
+                    logger.debug('Repository {} has incoming changes'.
+                                 format(repo))
+                    got_incoming = True
```
tulinkry: I wonder if it makes sense to reindex only the repository which actually changed (if it's supported at this point by the indexer)

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -0,0 +1,400 @@
...
+    # Quick sanity check.
+    known_project_tunables = [DISABLED_PROPERTY, CMD_TIMEOUT_PROPERTY,
+                              HOOK_TIMEOUT_PROPERTY, PROXY_PROPERTY,
+                              IGNORED_REPOS_PROPERTY, HOOKS_PROPERTY]
+
+    if not multiple_project_config:
+        return True
+
+    for project_name in multiple_project_config.keys():
+        project_config = multiple_project_config.get(project_name)
```
tulinkry: you can iterate per both key and value using `items()`

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -0,0 +1,400 @@
...
+        diff = diff_list(project_config.keys(),
+                         known_project_tunables)
```
tulinkry: python set supports native difference,
also I could expect different results from diff_list than set difference so the function name is not the best choice here

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
 
     logger = get_console_logger(get_class_basename(), args.loglevel)
 
+    if len(args.project) > 0 and args.all:
+        logger.fatal("Cannot use both project list and -a/--all")
+        sys.exit(1)
```
tulinkry: why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
...
+
+    if not args.all and len(args.project) == 0:
+        logger.fatal("Need at least one project or --all")
+        sys.exit(1)
```
tulinkry: why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
...
+
     if args.config:
         config = read_config(logger, args.config)
         if config is None:
-            logger.error("Cannot read config file from {}".format(args.config))
+            logger.fatal("Cannot read config file from {}".format(args.config))
```
tulinkry: why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
...
-
-    # Make sure the log directory exists.
-    logdir = config.get(LOGDIR_PROPERTY)
-    if logdir:
-        check_create_dir(logger, logdir)
-
     uri = args.uri
     if not is_web_uri(uri):
-        logger.error("Not a URI: {}".format(uri))
+        logger.fatal("Not a URI: {}".format(uri))
```
tulinkry: why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
...
         sys.exit(1)
     logger.debug("web application URI = {}".format(uri))
 
+    if not check_configuration(config):
+        sys.exit(1)
```
tulinkry: no log? why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -145,60 +84,38 @@ def main():
...
+
+    # Save the source root to avoid querying the web application.
     source_root = get_config_value(logger, 'sourceRoot', uri)
     if not source_root:
-        sys.exit(1)
+        return False
```
tulinkry: why?
why not use print_exc_exit ?

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -213,203 +130,38 @@ def main():
...
-    if project_config:
-        if project_config.get(DISABLED_PROPERTY):
-            logger.info("Project {} disabled, exiting".
-                        format(args.project))
-            sys.exit(CONTINUE_EXITVAL)
-
-    lock = FileLock(os.path.join(tempfile.gettempdir(),
-                                 args.project + "-mirror.lock"))
+    projects = args.project
+    if len(args.project) == 1:
```
tulinkry: `len(projects)` and

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -213,203 +130,38 @@ def main():
...
+        lockfile = args.project[0] + "-mirror"
```
tulinkry: `projects[0]`

tulinkry:
How exactly you mirror multiple projects at one? I still see a sequential loop in there.

vladak:
the main loop is in `main()`:
```python
       for project_name in projects:
                if args.batch:
                    get_batch_logger(logdir, project_name,
                                     args.loglevel,
                                     args.backupcount,
                                     get_class_basename())

                project_result = mirror_project(config, project_name,
                                                args.incoming,
                                                args.uri, source_root)
                if project_result == 1:
                    ret = 1
```

it should be possible to take this one step further and make it parallelized.

opengrok-tools/src/main/python/opengrok_tools/mirror.py
vladak: the error is logged already within the function. `print_exc_exit()` was originally for exceptions only.

opengrok-tools/src/main/python/opengrok_tools/mirror.py
vladak: I prefer logging over printing directly to `stderr`.

opengrok-tools/src/main/python/opengrok_tools/scm/teamware.py
```diff
@@ -41,21 +41,18 @@ def __init__(self, logger, path, project, command, env, hooks, timeout):
...
-                self.logger.error("Cannot get PATH env var")
-                raise OSError
+                raise OSError("Cannot get PATH env var")
 
             path += ":" + command
             self.env['PATH'] = path
         else:
-            self.logger.error("Cannot get path to Teamware commands")
-            raise OSError
+            raise RepositoryException("Cannot get path to Teamware commands")
```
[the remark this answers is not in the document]
vladak: good point. I think `OSError` should be thrown only in case of problem not related to repository imlementation/environment.

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.mark.skipif(not os.name.startswith("posix"), reason="requires posix")
+def test_valid_project_config_hook():
+    tmpdir = tempfile.TemporaryDirectory()
+    with open(os.path.join(tmpdir.name, "foo.sh"), 'w+') as tmpfile:
+        tmpfile.write("foo\n")
+        st = os.stat(tmpfile.name)
+        os.chmod(tmpfile.name, st.st_mode | stat.S_IEXEC)
+        config = {"foo": {HOOKS_PROPERTY: {"pre": "foo.sh"}}}
+        assert check_project_configuration(config, hookdir=tmpdir.name)
+        shutil.rmtree(tmpdir.name)
```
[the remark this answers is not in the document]
vladak: ah yes, thanks.

opengrok-tools/src/main/python/opengrok_tools/utils/log.py
```diff
@@ -122,7 +125,40 @@ def get_console_logger(name, level=logging.INFO, format='%(message)s'):
     logger = logging.getLogger(name)
     logger.setLevel(level)
     logger.propagate = False
+    logger.handlers = []
```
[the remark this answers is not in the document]
vladak: The problem is that each project needs to log to its own log file. This is done via the `RotatingFileHandler`.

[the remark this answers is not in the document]
tulinkry: aha, so you get the same instance of loggers but attache different files, seems fine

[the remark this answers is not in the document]
vladak: I was thinking about making the logger tree wide so that each project would get its own logger that would be directly under the root logger, i.e. creating a logger hierarchy.

The logger used by a sub-module would look like  `opengrok_tools.<PROJECT_NAME>.foo.bar`. The `get_batch_logger()` is called with base class name, i.e. `opengrok_tools` and is non-propagating. Hence, log message emitted by `bar` would propagate up to the batch logger and stop there.

However, I am not sure what trick to use to inject the project part so that `logging.getLogger(__name__)` works seamlessly in sub-modules.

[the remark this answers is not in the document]
vladak: A colleague pointed out that the `handlers` member is not starting with underscore so it can be probably be modified without too big risk.

tulinkry:
There were 11 unresolved comments and you merged it?

vladak:
I believe all of them were resolved.

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -0,0 +1,400 @@
...
+
+def get_repos_for_project(project_name, ignored_repos, **kwargs):
+    """
+    :param project_name: project name
+    :param ignored_repos: list of ignored repositories
+    :param kwargs: argument dictionary
+    :return: list of Repository objects
+    """
+
+    logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
tulinkry: I mean it should be, as it's a logger. Like in https://github.com/oracle/opengrok/pull/2731#discussion_r272487322

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -0,0 +1,115 @@
...
+    PROJECTS_PROPERTY
+
+
+def test_empty_project_configuration():
+    config = {}
+    assert check_project_configuration(config)
+
+
+def test_none_project_configuration():
+    assert check_project_configuration(None)
```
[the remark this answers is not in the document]
tulinkry: I think that's a matter of opinion but in that case I'd expect empty object rather than None.

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
tulinkry: No, I meant a general idea for an enhancement. As I see it, now when we get incoming=True from **any** of the repositories (there might by several in one project), we continue:

1) Synchronize all repositories in the project (which is not necessary for these which got incoming=False
2) Run the indexer for the whole project (again not necessary as other repositories have not changed).

PR #2733, opened 2019-03-28
improving the color palette for annotations

### LINKED ISSUE #1361: color scheme used in annotations for changesets is hard to perceive
It seems people have general problem with understanding the colors in the Annotate view:

> 
> It seemed obvious to me the colors indicated how recent the change was made.
> I thought at first the scale was:
>
> yellow= recent < orange = more recent < dark orange = most recent
>
> But it is actually
>
> dark orange = recent < orange = more recent < yellow = most recent 

Something more obvious is needed. Perhaps the thermo analogy ? (the whole gradient from blue to red)

tulinkry:
fixes #1361
This is heavily inspired by annotate in idea so we get some better feel out of it.

- colours on all revisions
- most recent is yellow (see the screenshots)
- most recent is also marked with "*"

Screenshot 2019-03-28 at 16 49 44
Screenshot 2019-03-28 at 16 49 56
Screenshot 2019-03-28 at 16 50 02

For two adjacent revisions there is a slightly different shade of the color, still not easy to perceive it very quickly, but definitely faster then as it used to be (5a and 98 are one revision apart).

Screenshot 2019-03-28 at 17 31 14


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -743,16 +744,8 @@ public static void readableLine(int num, Writer out, Annotation annotation,
...
-                        version number, 1 is the most recent
-                        generates css classes version_color_n
-                     */
-                    int versionNumber = Math.max(1,
-                            annotation.getFileVersionsCount()
-                            - annotation.getFileVersion(r) + 1);
-                    out.write(" version_color_" + versionNumber);
-                }
+                out.write("\" style=\"background-color: ");
+                out.write(annotation.getColors().getOrDefault(r, "inherit"));
```
vladak: can you avoid calling `getColors()` for every line ?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -770,6 +763,9 @@ public static void readableLine(int num, Writer out, Annotation annotation,
             }
             StringBuilder buf = new StringBuilder();
             htmlize(r, buf);
+            if (annotation.getFileVersion(r) == annotation.getRevisions().size()) {
```
vladak: add comment

vladak: Also, I wonder if this widening could cause some UI problems ? Is there perhaps better way to mark the latest revision ? (bold font, different font color, etc.)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -190,6 +191,53 @@ public int getFileVersionsCount() {
...
+         * We use the basic of hsl(60, 90%, 80%) and rotate the HSL circle with step equal 360/revisions_count.
+         *
+         * For two adjacent revisions, we adjust the saturation and lightness a bit so we get a smooth visible
+         * transition.
+         *
+         * The idea of the computation is taken from
+         * https://stackoverflow.com/questions/470690/how-to-automatically-generate-n-distinct-colors
+         */
+        for (String revision : revisions) {
+            colors.put(revision, String.format("hsl(%d, %d%%, %d%%)", hue, 90 - (even ? 0 : 10), 80 - (even ? 0 : 10)));
```
vladak: move `(even ? 0 : 10)` to a var (`adjust` ?) for better readability ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -190,6 +191,53 @@ public int getFileVersionsCount() {
...
+     *
+     * @return map of (revision, css color string) for each revision in {@code getRevisions()}
+     * @see #getRevisions()
+     */
+    public Map<String, String> getColors() {
+        Map<String, String> colors = new HashMap<>();
+        boolean even = true;
+        int hue = 60;
+        final List<String> revisions =
+                getRevisions()
```
vladak: What if you display annotation for older revision of the file ? Then the set of revisions would be smaller.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -190,6 +191,53 @@ public int getFileVersionsCount() {
...
+         * For two adjacent revisions, we adjust the saturation and lightness a bit so we get a smooth visible
+         * transition.
+         *
+         * The idea of the computation is taken from
+         * https://stackoverflow.com/questions/470690/how-to-automatically-generate-n-distinct-colors
+         */
+        for (String revision : revisions) {
+            colors.put(revision, String.format("hsl(%d, %d%%, %d%%)", hue, 90 - (even ? 0 : 10), 80 - (even ? 0 : 10)));
+            hue += 360 / getRevisions().size();
+            hue %= 360;
```
vladak: I wonder if it makes sense to avoid coloring the revisions beyond certain threshold. Usually the really old revisions are not that important.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -190,6 +191,53 @@ public int getFileVersionsCount() {
...
+     * and compute their color. Secondly, use all other revisions in order
+     * which is undefined and generate the rest of the colors for them.
+     *
+     * @return map of (revision, css color string) for each revision in {@code getRevisions()}
+     * @see #getRevisions()
+     */
+    public Map<String, String> getColors() {
+        Map<String, String> colors = new HashMap<>();
+        boolean even = true;
+        int hue = 60;
```
vladak: I'd assume the most recent version would start with red and not yellow.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: lol, yes

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: IDEA uses that together with bold font.  I can add a +1 to widestRevision to avoid any UI problems. Adding asterisk or any other mark seems fine to me to mark the revision.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: It used to be yellow so I kept it consistent.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: I really disliked it as it used to be with just few revisions, reasons:

1) it feels like some thing went wrong when some revisions have colour and some don't
2) colours on all revisions separate them visually (not just emphasizing the more recent revisions)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: Yes exactly.

tulinkry: Colors are not preserved for the revision, the palette will be different for older revision, starting from yellow with bigger steps. I don't see it as a problem.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: used lazy instantiation

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: Maybe it's time to change it to red :-) The old and new schemes are so different it does not matter with what color to start. Or maybe better yet: make it consistent with IDEA behavior.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: Maybe it's time to change it to red :-) The old and new schemes are so different it does not matter with what color to start. Or maybe better yet: make it consistent with IDEA behavior.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: Okay, as long as getRevisions() does the right thing..

Would be nice to have this covered by tests.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
vladak: I'd prefer the bold font too.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: My worry is that once you wrap around the H spectrum the colors might start repeating.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: ok

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: The colors never start repeating because I only use the full circle. The step is proportional to revisions.size.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: After brief overview IDEA uses fixed linear gradient for 5 stop colors (roughly green, gray, blue, yellow, red) with interleaving 4 steps. Generating total 21 color and distribute the revisions among this pallete. I can prepare an example for that although I find this hsl approach nicer.

tulinkry: Screenshot 2019-04-06 at 11 24 50

tulinkry:
The idea uses this rainbow on the annotations:
Screenshot 2019-04-06 at 11 24 50

I tried the approach and it looks like this:

Screenshot 2019-04-06 at 11 23 02
Screenshot 2019-04-06 at 11 23 08

Not bad. I'd still prefer the overall color palette generated with hsl, but this isn't bad. You decide. My only worry is about copyrights as this is reversed engineered from IDEA. Is that ok?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: Still, I guess there is a threshold when the number of revisions stop making sense.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: I like the HSL approach better, partly also because the saturation is higher. On the other hand, lots of folks are used to IDEA these days and could be confused with our custom scheme.

What happens in IDEA if there are more than 21 revisions ? Do these work like hash buckets ?

vladak:
Foremost, I am not a lawyer. It seems to me that as long as you take just the idea it should be fine (as long as it is not patented :-D). Could you try cranking the saturation a bit ? The colors look a tad more faded than in the original rainbow design.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
tulinkry: I know the picture here looks little bit awful but in reality it looks quite ok on the opengrok page.

For 22+ revisions this is like a hash buckets yes. The interval of revisions is mapped to the single target color.

tulinkry:
This is back to the `rgb` color space, so no saturation here. I'd need to choose different stop colours for the rainbow (by hand). These exactly match the IDEA based on color picker. (Repeating:) The image here looks little faded but when you see it on the page it is similar to what you see in IDEA and it's not bad.

vladak:
Ok, go ahead then. I think it makes sense in this case to stick with de-facto standard color scheme rather than pushing/inventing our own.

tulinkry:
Time to revive this

vladak:
Will take a look. Are you done making changes ?

Dne čt 20. 6. 2019 18:01 uživatel Kryštof Tulinger <notifications@github.com>
napsal:

tulinkry:
Ill fix the tests but then it’s ready

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: And this hasn’t been fully addressed

tulinkry:
Seems like it's passing

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
tulinkry: done

tulinkry:
ready

vladak:
Okay, let's merge this and see what the feedback will be.

PR #2736, opened 2019-03-29
LDAP plugin changes

vladak:
The hardcoded set of LDAP attributes in `LdapUser/LdapUserAttrPlugin` should go away in favor of configuration. The attribute names in the configuration will be comma separated.

Sample config looks like this:
```xml
            <void method="add">
                    <void property="name">
                        <string>opengrok.auth.plugin.LdapUserPlugin</string>
                </void>
                <void property="flag">
                    <string>REQUISITE</string>
                </void>
                <void property="setup">
                    <void method="put">
                         <string>objectclass</string>
                         <string>person</string>
                    </void>
                    <void method="put">
                         <string>attributes</string>
                         <string>mail,uid</string>
                    </void>
                </void>
                </object>
            </void>
```

Also, the attribute presence check was softened (made best effort) - if given LDAP attribute cannot be found in `LdapUserPlugin#fillSession()` it will just continue.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationException.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationException.java
@@ -18,26 +18,24 @@
  */
 
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
  */
 
 package org.opengrok.indexer.authorization;
 
-/**
- * Unchecked exception to be thrown when LDAP server pool is down.
- */
-public class LdapException extends RuntimeException {
-    public static final long serialVersionUID = -1;
+public class AuthorizationException extends RuntimeException {
 
-    public LdapException() {
-        super();
+    private static final long serialVersionUID = -1;
+
+    public AuthorizationException(Throwable ex) {
+        super(ex);
     }
 
-    public LdapException(String str) {
-        super(str);
+    AuthorizationException(String msg) {
+        super(msg);
     }
 
-    public LdapException(String str, Throwable ex) {
-        super(str, ex);
+    AuthorizationException(String msg, Throwable ex) {
+        super(msg, ex);
     }
 }
--- a/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -28,6 +28,8 @@ import java.nio.file.Paths;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Stream;
 import javax.servlet.http.HttpServletRequest;
 import opengrok.auth.entity.LdapUser;
@@ -41,6 +43,9 @@ import org.opengrok.indexer.configuration.Project;
  * @author Krystof Tulinger
  */
 public class LdapAttrPlugin extends AbstractLdapPlugin {
+
+    private static final Logger LOGGER = Logger.getLogger(LdapAttrPlugin.class.getName());
+
 
     protected static final String ATTR_PARAM = "attribute"; // LDAP attribute name to check
     protected static final String FILE_PARAM = "file";
@@ -92,7 +97,7 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
         updateSession(req, sessionAllowed);
 
         if ((ldapUser = (LdapUser) req.getSession().getAttribute(LdapUserPlugin.SESSION_ATTR)) == null) {
-            // TODO log
+            LOGGER.log(Level.FINE, "cannot get {0} attribute", LdapUserPlugin.SESSION_ATTR);
             return;
         }
 
--- a/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -42,7 +42,7 @@ import org.opengrok.indexer.configuration.Project;
  */
 public class LdapAttrPlugin extends AbstractLdapPlugin {
 
-    protected static final String ATTR_PARAM = "attribute";
+    protected static final String ATTR_PARAM = "attribute"; // LDAP attribute name to check
     protected static final String FILE_PARAM = "file";
 
     private static final String SESSION_ALLOWED_PREFIX = "opengrok-attr-plugin-allowed";
@@ -87,33 +87,31 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
         Boolean sessionAllowed = false;
         LdapUser ldapUser;
         Map<String, Set<String>> records;
-        Set<String> attributes;
+        Set<String> attributeValues;
 
         updateSession(req, sessionAllowed);
 
         if ((ldapUser = (LdapUser) req.getSession().getAttribute(LdapUserPlugin.SESSION_ATTR)) == null) {
+            // TODO log
             return;
         }
 
-        if (ldapAttr.equals("mail")) {
-            sessionAllowed = whitelist.contains(ldapUser.getMail());
-        } else if (ldapAttr.equals("uid")) {
-            sessionAllowed = whitelist.contains(ldapUser.getUid());
-        } else if (ldapAttr.equals("ou")) {
-            sessionAllowed = ldapUser.getOu().stream().anyMatch((t) -> whitelist.contains(t));
-        } else if ((attributes = (Set<String>) ldapUser.getAttribute(ldapAttr)) != null) {
-            sessionAllowed = attributes.stream().anyMatch((t) -> whitelist.contains(t));
+        // Check attributes cached in LDAP user object first, then query LDAP server
+        // (and if found, cache the result in the LDAP user object).
+        attributeValues = ldapUser.getAttribute(ldapAttr);
+        if (attributeValues != null) {
+            sessionAllowed = attributeValues.stream().anyMatch((t) -> whitelist.contains(t));
         } else {
             if ((records = getLdapProvider().lookupLdapContent(user, new String[]{ldapAttr})) == null) {
                 return;
             }
 
-            if (records.isEmpty() || (attributes = records.get(ldapAttr)) == null) {
+            if (records.isEmpty() || (attributeValues = records.get(ldapAttr)) == null) {
                 return;
             }
 
-            ldapUser.setAttribute(ldapAttr, attributes);
-            sessionAllowed = attributes.stream().anyMatch((t) -> whitelist.contains(t));
+            ldapUser.setAttribute(ldapAttr, attributeValues);
+            sessionAllowed = attributeValues.stream().anyMatch((t) -> whitelist.contains(t));
         }
 
         updateSession(req, sessionAllowed);
--- a/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -99,7 +99,7 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
         updateSession(req, sessionAllowed);
 
         if ((ldapUser = (LdapUser) req.getSession().getAttribute(LdapUserPlugin.SESSION_ATTR)) == null) {
-            LOGGER.log(Level.FINE, "cannot get {0} attribute", LdapUserPlugin.SESSION_ATTR);
+            LOGGER.log(Level.WARNING, "cannot get {0} attribute", LdapUserPlugin.SESSION_ATTR);
             return;
         }
 
@@ -111,6 +111,8 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
         } else {
             try {
                 if ((records = getLdapProvider().lookupLdapContent(user, new String[]{ldapAttr})) == null) {
+                    LOGGER.log(Level.WARNING, "cannot lookup attributes {0} for user {1}",
+                        new Object[]{ldapAttr, user});
                     return;
                 }
             } catch (LdapException ex) {
@@ -118,6 +120,8 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
             }
 
             if (records.isEmpty() || (attributeValues = records.get(ldapAttr)) == null) {
+                LOGGER.log(Level.WARNING, "empty records or attribute values {0} for user {1}",
+                        new Object[]{ldapAttr, user});
                 return;
             }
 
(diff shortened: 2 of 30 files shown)


vladak:
The other change makes LDAP server pools failures visible in the UI. New `LdapError` unchecked exception is thrown from `LdapFacade` on failure so that `error.jsp` can be used. I had to remove the page header from `error.jsp` to prevent nested exception that happens when project picker is being rendered.

Also, I added RESTful API hooks for better observability of LDAP server pool failures/recovery. The whole config can look like this:

```xml
<java version="1.8.0_65" class="java.beans.XMLDecoder">
        <object class="opengrok.auth.plugin.configuration.Configuration">
                <void property="interval">
                        <int>900000</int>
                </void>
                <void property="searchBase">
                        <string>dc=foo,dc=com</string>
                </void>
                <void property="hooks">
                 <object class="opengrok.auth.plugin.util.Hooks">
                  <void property="fail">
                   <object class="opengrok.auth.plugin.util.Hook">
                    <void property="URI">
                     <string>http://localhost:8080/source/api/v1/messages</string>
                    </void>
                    <void property="content">                                                 
                     <string>{ "tags": [ "main" ], "cssClass": "class", "text": "test message", "duration": "PT10M" }</string>
                    </void>
                   </object>
                  </void>
                 </object>
                </void>
                <void property="servers">
                        <void method="add">
                                <object class="opengrok.auth.plugin.ldap.LdapServer">
                                        <void property="name">
                                                <string>ldap://ldap.foo.com</string>
                                        </void>
                                        <void property="timeout">
                                                <int>1000</int>
                                        </void>
                                </object>
                        </void>
                </void>
        </object>
</java>
```

except the message would not be displayed since the error page takes over  - this is just for demonstration only and can be used to trigger/silence warnings elsewhere.

vladak:
The next change makes the LDAP related tunables more flexible, namely:
  - it is now possible to configure search timeout and the max. number of results
  - LDAP connect timeout is now configurable on server pool level

vladak:
The decoder changes accept full class names only in the configuration, e.g.:
```xml
           <void method="add">
                    <void property="name">
                        <string>opengrok.auth.plugin.UserPlugin</string>
                    </void>
                    <void property="flag">
                        <string>REQUISITE</string>
                    </void>
                    <void property="setup">
                        <void method="put">
                             <string>decoder</string>
                             <string>opengrok.auth.plugin.decoders.OSSOHeaderDecoder</string>
                        </void>
                    </void>
                </object>
            </void>
```

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/LdapError.java
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ */
+
+package org.opengrok.indexer.authorization;
+
+/**
+ * Unchecked exception to be thrown when LDAP server pool is down.
+ */
+public class LdapError extends RuntimeException {
```
tulinkry: why call it error when it's an exception?

plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -87,33 +87,31 @@ public void fillSession(HttpServletRequest req, User user) {
         Boolean sessionAllowed = false;
         LdapUser ldapUser;
         Map<String, Set<String>> records;
-        Set<String> attributes;
+        Set<String> attributeValues;
 
         updateSession(req, sessionAllowed);
 
         if ((ldapUser = (LdapUser) req.getSession().getAttribute(LdapUserPlugin.SESSION_ATTR)) == null) {
+            // TODO log
```
tulinkry: todo?

plugins/src/opengrok/auth/plugin/UserPlugin.java
```diff
@@ -39,43 +41,70 @@
...
     public static final String REQUEST_ATTR = "opengrok-user-plugin-user";
 
-    private IUserDecoder decoder = new OSSOHeaderDecoder();
+    private IUserDecoder decoder;
+
+    public UserPlugin() {
+    }
+
+    // for testing
+    public UserPlugin(IUserDecoder decoder) {
```
tulinkry: protected? package private for tests only?

plugins/src/opengrok/auth/plugin/util/Hooks.java
```diff
@@ -0,0 +1,41 @@
...
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ */
+
+package opengrok.auth.plugin.util;
+
+public class Hooks {
```
tulinkry: serializable

plugins/src/opengrok/auth/plugin/util/Hook.java
```diff
@@ -0,0 +1,35 @@
...
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ */
+
+package opengrok.auth.plugin.util;
+
+public class Hook {
```
tulinkry: serializable

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -89,6 +96,11 @@
      */
     private List<LdapServer> servers = new ArrayList<>();
 
+    /**
+     * server hooks
+     */
+    Hooks hooks;
```
tulinkry: i know it doesnt make any real difference but all other members have their 'private' specified

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -292,22 +320,26 @@ private SearchControls prepareSearchControls() {
         if (errorTimestamp > 0 && errorTimestamp + interval > System.currentTimeMillis()) {
             if (!reported) {
                 reported = true;
-                LOGGER.log(Level.SEVERE, "Server pool is still broken");
+                LOGGER.log(Level.SEVERE, "LDAP server pool is still broken");
             }
-            return null;
+            throw new LdapError("LDAP server pool is still broken");
```
tulinkry: why don't you throw a checked exception and convert it to unchecked in the plugin itself?

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -317,11 +349,18 @@ private SearchControls prepareSearchControls() {
...
+                    if ((hook = hooks.getRecover()) != null) {
+                        RestfulClient.postIt(hook.getURI(), hook.getContent());
+                    }
+                }
                 return processResult(sr, mapper);
             }
         } catch (NameNotFoundException ex) {
             LOGGER.log(Level.SEVERE, "The LDAP name was not found.", ex);
-            return null;
+            throw new LdapError("The LDAP name was not found.");
```
tulinkry: how about to signal the parent exception?

tulinkry: is the log now really needed?

tulinkry:
Can you post some screenshot of the UI when the LdapError is thrown?

plugins/src/opengrok/auth/plugin/UserPlugin.java
```diff
@@ -39,43 +41,70 @@
...
     public void unload() {
     }
 
-    @Override
-    public boolean isAllowed(HttpServletRequest request, Project project) {
+    private User getUser(HttpServletRequest request) {
         User user;
-        if ((user = (User) request.getAttribute(REQUEST_ATTR)) == null) {
+
+        if ((user =(User)request.getAttribute(REQUEST_ATTR))==null) {
```
tulinkry: missing space?

tulinkry:
> Also, the attribute presence check was softened (made best effort) - if given LDAP attribute cannot be found in LdapUserPlugin#fillSession() it will just continue.

isn't it a security risk?

plugins/src/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -125,22 +146,19 @@ public void fillSession(HttpServletRequest req, User user) {
...
+            }
         }
 
-        if (!records.containsKey("mail") || records.get("mail").isEmpty()) {
-            LOGGER.log(Level.WARNING, "mail record for user {0} is not present or empty",
-                    user);
-            return;
+        Map<String, Set<String>> attrSet = new HashMap<>();
+        for (String attrName : attributes) {
+            attrSet.put(attrName, records.get(attrName));
```
tulinkry: you now copy the map to a new map, right? isn't copy constructor better solution?

plugins/src/opengrok/auth/plugin/util/Hook.java
```diff
@@ -0,0 +1,35 @@
+/*
```
tulinkry: I'd be for name `webhook` or similar to clearly state this is only http related

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/LdapError.java
vladak: true, I wanted to rename but then forgot.

plugins/src/opengrok/auth/plugin/LdapUserPlugin.java
vladak: `attributes` is array

tulinkry: Right. I didn't see that.

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -317,11 +349,17 @@ private SearchControls prepareSearchControls() {
             for (namingEnum = server.search(dn, filter, controls); namingEnum.hasMore();) {
                 SearchResult sr = namingEnum.next();
                 reported = false;
+                if (errorTimestamp > 0) {
+                    errorTimestamp = 0;
+                    WebHook hook;
+                    if ((hook = webHooks.getRecover()) != null) {
+                        RestfulClient.postIt(hook.getURI(), hook.getContent());
```
tulinkry: consider using `CompletableFuture.supplyAsync( ... );` so you don't slow down the request?

vladak:
This is how the error page will look like in case of LDAP problem:
ldapException

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -292,22 +320,26 @@ private SearchControls prepareSearchControls() {
...
+                LOGGER.log(Level.SEVERE, "LDAP server pool is still broken");
             }
-            return null;
+            throw new LdapException("LDAP server pool is still broken");
         }
 
         if (fail > servers.size() - 1) {
             // did the whole rotation
-            LOGGER.log(Level.SEVERE, "Tried all servers in a pool but no server works");
+            LOGGER.log(Level.SEVERE, "Tried all LDAP servers in a pool but no server works");
```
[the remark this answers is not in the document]
vladak: actually yes. Otherwise the catalina log will not contain records for post mortems.

vladak:
As for the attributes, if the attribute is not found by the `LdapUserPlugin` this does not still constitute an error, it is up to the subsequent modules (`LdapAttrPlugin`) to actually perform the check. As far as I understand it, missing user will behave differently.

plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -107,8 +109,12 @@ public void fillSession(HttpServletRequest req, User user) {
         if (attributeValues != null) {
             sessionAllowed = attributeValues.stream().anyMatch((t) -> whitelist.contains(t));
         } else {
-            if ((records = getLdapProvider().lookupLdapContent(user, new String[]{ldapAttr})) == null) {
-                return;
+            try {
+                if ((records = getLdapProvider().lookupLdapContent(user, new String[]{ldapAttr})) == null) {
```
tulinkry: can it return null now with your changes?

plugins/src/opengrok/auth/plugin/LdapFilterPlugin.java
```diff
@@ -85,9 +87,13 @@ public void fillSession(HttpServletRequest req, User user) {
         String expandedFilter = expandFilter(ldapFilter, ldapUser, user);
         LOGGER.log(Level.FINER, "expanded filter for user {0} into ''{1}''",
                 new Object[]{user, expandedFilter});
-        if ((records = getLdapProvider().lookupLdapContent(null, expandedFilter, dn)) == null) {
-            LOGGER.log(Level.FINER, "failed to get content for user from LDAP server");
-            return;
+        try {
+            if ((records = getLdapProvider().lookupLdapContent(null, expandedFilter, dn)) == null) {
```
tulinkry: can it return null with your changes?

plugins/src/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -133,11 +135,15 @@ public void fillSession(HttpServletRequest req, User user) {
         }
 
         String filter = getFilter(user);
-        if ((records = getLdapProvider().lookupLdapContent(null, filter, attributes)) == null) {
-            LOGGER.log(Level.WARNING, "failed to get LDAP attributes ''{3}'' for user ''{0}'' " +
-                            "with filter ''{1}''",
-                    new Object[]{user, filter, String.join(", ", attributes)});
-            return;
+        try {
+            if ((records = getLdapProvider().lookupLdapContent(null, filter, attributes)) == null) {
```
tulinkry: can it return null with your changes?

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationException.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    AuthorizationException(String msg) {
+        super(msg);
+    }
+
+    AuthorizationException(String msg, Throwable ex) {
+        super(msg, ex);
+    }
+}
```
tulinkry: nice

plugins/src/opengrok/auth/plugin/ldap/LdapFacade.java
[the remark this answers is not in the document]
tulinkry: ok

plugins/src/opengrok/auth/plugin/LdapAttrPlugin.java
vladak: which changes do you mean ?

tulinkry: specifically the changes that you throw an exception instead of returning null

vladak: I think it can return null if `processResult()` returns `null`. The `SearchResult` class is not really specific in when it can return null; it contains some curious comments such as:
```java

    /**
     * Sets the attributes of this search result to <code>attrs</code>.
     * @param attrs The non-null attributes to use. Can be empty.
     * @see #getAttributes
     */
    public void setAttributes(Attributes attrs) {
        this.attrs = attrs;
        // ??? check for null?
    }
```

so my guess is that `null` still might be returned.

tulinkry: ok

PR #2744, opened 2019-04-08
parallelize repository mirroring

vladak:
will be handy for https://github.com/OpenGrok/docker/pull/32

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/mirror.py
@@ -56,34 +56,19 @@ if major_version < 3:
 __version__ = "0.8"
 
 
-class workerConfig():
-    """
-    class to wrap parameters of the worker process
-    """
-    def __init__(self, project_name, logdir, loglevel, backupcount, config,
-                 incoming, uri, source_root, batch):
+def worker(args):
+    project_name, logdir, loglevel, backupcount, config, incoming, uri, \
+        source_root, batch = args
 
-        self.project_name = project_name
-        self.logdir = logdir
-        self.loglevel = loglevel
-        self.backupcount = backupcount
-        self.config = config
-        self.incoming = incoming
-        self.uri = uri
-        self.source_root = source_root
-        self.batch = batch
-
-
-def worker(cfg):
-    if cfg.batch:
-        get_batch_logger(cfg.logdir, cfg.project_name,
-                         cfg.loglevel,
-                         cfg.backupcount,
+    if batch:
+        get_batch_logger(logdir, project_name,
+                         loglevel,
+                         backupcount,
                          get_class_basename())
 
-    return mirror_project(cfg.config, cfg.project_name,
-                          cfg.incoming,
-                          cfg.uri, cfg.source_root)
+    return mirror_project(config, project_name,
+                          incoming,
+                          uri, source_root)
 
 
 def main():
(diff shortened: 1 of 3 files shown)


opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -52,7 +53,37 @@
...
+def worker(cfg):
+    if cfg.batch:
+        get_batch_logger(cfg.logdir, cfg.project_name,
+                         cfg.loglevel,
+                         cfg.backupcount,
+                         get_class_basename())
+
+    return mirror_project(cfg.config, cfg.project_name,
+                          cfg.incoming,
+                          cfg.uri, cfg.source_root)
```
tulinkry: I personally developed a decorator for pool map function, you can use it if you feel it's better

```
def pool_callback(callback):
    """
    Decorator for auto expanding the thread pool arguments from a single
    variable list to a function arguments.
    :param callback: the actual executing method
    :return: a function calling the callback with proper argument list
    """
    return lambda args: callback(*args)
```

Usage
```
@pool_callback
def worker(project, logdir, loglevel, ...):
    pass
```

opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -34,6 +34,7 @@
 import os
 import sys
 import tempfile
+from multiprocessing import Pool, cpu_count
```
tulinkry: is process pool better than thread pool?

vladak: I think in this case it does not matter as calling `sync()` from `scm` modules will result in process spawning.

opengrok-tools/src/main/python/opengrok_tools/mirror.py
vladak: So the 2nd arg to `pool.map` is then list of lists ?

tulinkry: Yes

tulinkry:
```
@pool_callback
def callback(arg1, arg2, arg3):
    return arg1, arg2, arg3

ThreadPool(1).map(callback, [[1, 2, 3]])
```

vladak: This does not work with regular `Pool`:
```python
with Pool(processes=int(1)) as pool:                                            
    ret = pool.map(callback, [[1, 2, 3], [4, 5, 6]], 1)
```
which fails with:
```
  File "./pool_callb.py", line 25, in <module>
    ret = pool.map(callback, [[1, 2, 3], [4, 5, 6]], 1)
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 288, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 670, in get
    raise self._value
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 450, in _handle_tasks
    put(task)
  File "/usr/lib/python3.6/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object 'pool_callback.<locals>.<lambda>'
```
I think this is due to the passing/serialization of worker arguments to worker processes. Looks like it wants to send the lambda across and use it only when it reaches the worker.

vladak: In the end I unpacked the list manually.

tulinkry: ah so, the interprocess communication is little bit different, ok.

PR #2750, opened 2019-04-12
add Mellon header decoder

vladak:
This is a simple decoder for the https://github.com/UNINETT/mod_auth_mellon SAML implementation.

Needs `mod_headers` to convert the email Apache environment variable from `mod_auth_mellon` to HTTP header. In Apache config it looks like this:

```
        RequestHeader unset MELLON_email
        RequestHeader set MELLON_email "%{MELLON_email}e" env=MELLON_email
```

Also, makes some of the error messages more informative.

### DIFF
--- a/plugins/src/opengrok/auth/plugin/decoders/MellonHeaderDecoder.java
+++ b/plugins/src/opengrok/auth/plugin/decoders/MellonHeaderDecoder.java
@@ -1,0 +1,62 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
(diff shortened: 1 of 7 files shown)


plugins/src/opengrok/auth/plugin/decoders/MellonHeaderDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+ * and that the {@code mod_auth_mellon} is setup to create Apache environment variable
+ * containing the e-mail address and the {@code mod_headers} Apache module is set to
+ * pass the value of this variable in HTTP header called {@code MELLON_email}, i.e.:
+ * {@code RequestHeader set email "%{MELLON_email}e" env=MELLON_email}
+ */
+public class MellonHeaderDecoder implements IUserDecoder {
+
+    private static final Logger LOGGER = Logger.getLogger(MellonHeaderDecoder.class.getName());
+
+    static String MELLON_EMAIL_HEADER = "MELLON_email";
```
tulinkry: can be final also

PR #2757, opened 2019-04-24
Add Dockerfile to the Repo

wy193777:
Added Dockerfile and related scripts from [OpenGrok/Docker](https://github.com/OpenGrok/docker) repo. So devs can run their code directly in docker if they have difficulties setup environment.

Also make indexer flags configurable from passed environment variables when run the image.


vladak:
Besides non-optimal structuring of the changes (the docker related files should be put under single directory plus the `.gitignore` changes do not seem to be related), the main problem is the coexistence with the [OpenGrok/docker](https://github.com/OpenGrok/docker) repo. The main purpose of the repo is to automatically create and upload new image to Docker hub whenever there is a OpenGrok release. I'd like to avoid having duplicate code (and I mean the dockerfile too) in both locations. Now, having docker stuff in the main repository would strip away some of the complexity - currently release build in [oracle/OpenGrok](https://github.com/oracle/OpenGrok) kicks off a build in [OpenGrok/docker](https://github.com/OpenGrok/docker) which then downloads the release tarball, creates docker image and uploads it to docker hub. The image build cannot be done in Docker hub itself because of Github request throttling. The centralization would lead to needing just one secret token (for Docker hub) and the Docker image creation and upload can be done in distinct Travis build stage.

vladak:
So, for starters, restructure the files so that they reside in single directory + think of way how to parametrize the dockerfile so that it can be used for both developer and release builds.

wy193777:
Already made the change, can we move this PR forward?

vladak:
This still needs a way how to resolve the developer/release Docker builds, i.e. moving the files from one repo to another is easy, the more difficult part will be on me then. Let me think about how to complete the transition.

vladak:
Merged this with bunch of changes to make the build usable for both developer and release builds.

Issue #2746, opened 2019-04-09
Perforce wildcards need to be protected


cross:
Hello.  I am trying to set up OpenGrok 1.2.6 on an ubuntu system with multiple repositories.  Some of my repositories are perforce, and among a few problems I'm facing, I'm seeing the following:

`00:18:57 WARNING: Non-zero exit status 1 from command [/usr/bin/p4, files, class-map#O.ini] in directory /data/opengrok/src/myproj/dir: Invalid revision number 'O.ini'.
`

As you would guess, `/data/opengrok/src/myproj/dir/class-map#O.ini` is in fact in my sources.

Looking at https://www.perforce.com/manuals/v17.1/cmdref/Content/CmdRef/filespecs.synopsis.limitations.html I found that using ~~HTML-escapes~~ URI-escapes work for these characters.  Testing:

`p4 files ./myproj/dir/class-map%23O.ini` from within my source tree succeeds and gives me the information on that `class-map#O.ini` file.

Please adjust your perforce repository module to understand these translations in files with reserved characters in them.

cross:
Apologies, my original comment was misleading.  HTML-escapes would be wrong, they need to be URI-escapes.  Sorry for that.  Your latest comment is correct in describing the translations needed.

Also, while changing `src/main/java/org/opengrok/indexer/history/PerforceRepository.java#isInP4Depot()` would avoid the issue I reported, I fear that something further down into handling perforce repository actions may also be needed.  Assumedly after `isInP4Depot()` returns that it is (in a p4 depot), other operations would also attempt to operate on the filename and need the same change.

vladak:
Definitely, had the same thought when traversing the source code, i.e. `PerforceHistoryParser.java` will probably need some changes as well.

cross:
Thanks.  I forked and downloaded, and made changes to `PerforceHistoryParser.java` and `PerforceRepository.java` that compile.  But, unfortunately, I know enough java to work basic code, and don't know maven very well.

I was able to compile it, but `./mvnw package -DskipTests` (Ubuntu 18) reports an error related to opengrok-tools. This I'm sure is an environment problem on my system unrelated to this issue, but if you could help me figure out how to get past this, I can test my changes and work towards a pull request if/when they work.

```
[INFO] OpenGrok Web ....................................... SUCCESS [  3.148 s]
[INFO] OpenGrok tools ..................................... FAILURE [  0.851 s]
[INFO] Distribution ....................................... SKIPPED
[...]

PR #2762, opened 2019-04-30
Perforce filenames 2746

cross:
This avoids the errors from perforce when using filenames containing reserved characters, and will fix issue #2746 .

I am not a java developer so much, so I'm happy to adjust the code to match core opinions of how it should be implemented.  Let me know.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
@@ -87,7 +87,7 @@ public class PerforceHistoryParser {
         cmd.add("p4");
         cmd.add("filelog");
         cmd.add("-lti");
-        cmd.add(file.getName() + PerforceRepository.getRevisionCmd(rev));
+        cmd.add(PerforceRepository.protectPerforceFilename(file.getName()) + PerforceRepository.getRevisionCmd(rev));
         Executor executor = new Executor(cmd, file.getCanonicalFile().getParentFile());
         executor.exec();
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
@@ -19,6 +19,7 @@
 
 /*
  * Copyright (c) 2008, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Portions Copyright (c) 2019, Chris Ross <cross@distal.com>.
  */
 
 package org.opengrok.indexer.history;
@@ -35,6 +36,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.opengrok.indexer.util.Executor;
+import static org.opengrok.indexer.history.PerforceRepository.protectPerforceFilename;
 
 /**
  * Parse source history for a Perforce Repository
@@ -87,7 +89,7 @@ public class PerforceHistoryParser {
         cmd.add("p4");
         cmd.add("filelog");
         cmd.add("-lti");
-        cmd.add(PerforceRepository.protectPerforceFilename(file.getName()) + PerforceRepository.getRevisionCmd(rev));
+        cmd.add(protectPerforceFilename(file.getName()) + PerforceRepository.getRevisionCmd(rev));
         Executor executor = new Executor(cmd, file.getCanonicalFile().getParentFile());
         executor.exec();
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.AbstractMap.SimpleImmutableEntry;
 
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.logger.LoggerFactory;
@@ -61,6 +62,34 @@ public class PerforceRepository extends Repository {
         ignoredFiles.add(".p4config");
     }
 
+    public static String protectPerforceFilename(String name) {
+        ArrayList<SimpleImmutableEntry<String,String>>   transmap = new ArrayList<SimpleImmutableEntry<String,String>>() {
+            {
+                add(new SimpleImmutableEntry<>("#", "%23"));
+                add(new SimpleImmutableEntry<>("%", "%25"));
+                add(new SimpleImmutableEntry<>("\\*", "%2A"));
+                add(new SimpleImmutableEntry<>("@", "%40"));
+            }
+        };
+        String t = name;   /* XXX remove? */
+        for (SimpleImmutableEntry<String,String> ent : transmap) {
+            t = t.replaceAll(ent.getKey(), ent.getValue());
+        }
+        /* Instead of iterating over a list, just force each? */
+/*
+        String t = name.replaceAll("#", "%23");
+        t = t.replaceAll("%", "%25");
+        t = t.replaceAll("\\*", "%2A");
+        t = t.replaceAll("@", "%40");
+*/
+        if (name != t) {
+            LOGGER.log(Level.INFO,
+                       "protectPerforceFilename: replaced ''{0}'' with ''{1}''",
+                       new Object[]{name, t});
+        }
+        return t;
+    }
+
     @Override
     public Annotation annotate(File file, String rev) throws IOException {
         ArrayList<String> cmd = new ArrayList<>();
@@ -68,7 +97,7 @@ public class PerforceRepository extends Repository {
         cmd.add(RepoCommand);
         cmd.add("annotate");
         cmd.add("-qci");
-        cmd.add(file.getPath() + getRevisionCmd(rev));
+        cmd.add(protectPerforceFilename(file.getPath()) + getRevisionCmd(rev));
 
         Executor executor = new Executor(cmd, file.getParentFile(),
                 RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
@@ -87,7 +116,7 @@ public class PerforceRepository extends Repository {
         cmd.add(RepoCommand);
         cmd.add("print");
         cmd.add("-q");
-        cmd.add(basename + getRevisionCmd(rev));
+        cmd.add(protectPerforceFilename(basename) + getRevisionCmd(rev));
         Executor executor = new Executor(cmd, new File(parent));
         // TODO: properly evaluate Perforce return code
         executor.exec();
@@ -140,7 +169,7 @@ public class PerforceRepository extends Repository {
         boolean status = false;
         if (testRepo.isWorking()) {
             ArrayList<String> cmd = new ArrayList<>();
-            String name = file.getName();
+            String name = protectPerforceFilename(file.getName());
             File dir = file.getParentFile();
             if (file.isDirectory()) {
                 dir = file;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
@@ -65,8 +65,10 @@ public class PerforceRepository extends Repository {
     public static String protectPerforceFilename(String name) {
         ArrayList<SimpleImmutableEntry<String,String>>   transmap = new ArrayList<SimpleImmutableEntry<String,String>>() {
             {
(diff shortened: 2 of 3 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
```diff
@@ -35,8 +35,9 @@
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.AbstractMap.SimpleImmutableEntry;
 
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
```
vladak: pls avoid wildcard imports

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
```diff
@@ -61,14 +62,44 @@ public PerforceRepository() {
...
+        }
+        /* Instead of iterating over a list, just force each? */
+/*
+        String t = name.replaceAll("%", "%25");
+        t = t.replaceAll("#", "%23");
+        t = t.replaceAll("\\*", "%2A");
+        t = t.replaceAll("@", "%40");
+*/
+        if (name != t) {
+            LOGGER.log(Level.INFO,
```
vladak: this should probably be `Level.FINEST`

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
```diff
@@ -61,14 +62,44 @@ public PerforceRepository() {
...
+                add(new SimpleImmutableEntry<>("#", "%23"));
+                add(new SimpleImmutableEntry<>("\\*", "%2A"));
+                add(new SimpleImmutableEntry<>("@", "%40"));
+            }
+        };
+        String t = name;   /* XXX remove? */
+        for (SimpleImmutableEntry<String,String> ent : transmap) {
+            t = t.replaceAll(ent.getKey(), ent.getValue());
+        }
+        /* Instead of iterating over a list, just force each? */
```
vladak: not a bad idea at all. keep it simple.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
```diff
@@ -61,14 +62,44 @@ public PerforceRepository() {
         ignoredFiles.add(".p4config");
     }
 
+    public static String protectPerforceFilename(String name) {
```
vladak: this method could be protected perhaps or package private

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
```diff
@@ -105,4 +106,25 @@ public void testHistoryAndAnnotations() throws Exception {
...
+                add(new SimpleImmutableEntry<>("@seventeen", "%40seventeen"));
+                add(new SimpleImmutableEntry<>("upNdown(", "upNdown("));
+                add(new SimpleImmutableEntry<>("tst#99", "tst%2399"));
+            }
+        };
+        for (SimpleImmutableEntry<String,String> ent : testmap) {
+            String prot = instance.protectPerforceFilename(ent.getKey());
+            assertEquals("Improper protected filename, "+prot+" != "+ent.getValue(), ent.getValue(), prot);
+        }
+//        instance.update();
```
vladak: ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
cross: So you'd prefer I just call replaceAll four times?  Easy enough, that's why I left the comment in.  I'll use that.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
```diff
@@ -105,4 +106,25 @@ public void testHistoryAndAnnotations() throws Exception {
             }
         }
     }
+
+    @Test
+    public void testProtectFilename() throws Exception {
+        PerforceRepository instance = new PerforceRepository();
+        ArrayList<SimpleImmutableEntry<String,String>>   testmap = new ArrayList<SimpleImmutableEntry<String,String>>() {
```
vladak: I am pretty sure the diamond operator does not have to be fully specified on both sides.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
```diff
@@ -105,4 +106,25 @@ public void testHistoryAndAnnotations() throws Exception {
...
+            {
+                add(new SimpleImmutableEntry<>("Testfile 34", "Testfile 34"));
+                add(new SimpleImmutableEntry<>("Test%52", "Test%2552"));
```
vladak: maybe add another test case with multiple occurrences of chars to be replaced (tests `relplaceAll()`)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
```diff
@@ -87,7 +87,7 @@ public static History getRevisions(File file, String rev) throws IOException {
         cmd.add("p4");
         cmd.add("filelog");
         cmd.add("-lti");
-        cmd.add(file.getName() + PerforceRepository.getRevisionCmd(rev));
+        cmd.add(PerforceRepository.protectPerforceFilename(file.getName()) + PerforceRepository.getRevisionCmd(rev));
```
vladak: import and use the method directly

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
cross: As it's called from PerforceHistoryParser, I assumed it needed to be public.  Is there a relationship I'm missing?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
cross: I'm sorry, I don't understand what you're asking for here.  This is modeled after the other method calls to methods of PerforceRepository, including PerforceRepository.getRevisionCmd in this same line.  Can you clarify what you'd like to see?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
cross: Yeah, that was originally for my debugging and I thought it might come out.  But leaving it in at FINE or FINEST makes sense, so I'll change it as you suggest.

cross:
I see that coveralls notes 0 of 17 new lines covered, but the unit test I added should cause coverage of the new method I added.  Do I have to do something to get that new method "covered"?

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
cross: I could not get it to compile without the inner types, I tried numerous omissions.  I am not very good at java, so may be wrong, but I didn't see how to make it work.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
cross: removed.  I think I didn't know why it was there.  And, I've since wondered if I should bother having an instance in this test, since it's testing a static method anyway.  Left a comment to that affect in the updated commit.

cross:
@vladak Can you look over my most recent commit which fixes some things you mentioned?  There are still other questions I have open in comments here, if I could get feedback we can close any open issues.  Thanks.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
vladak: If you add
```java
import static org.opengrok.indexer.history.PerforceRepository.protectPerforceFilename;
```
then you can replace the line with just:
```java
            String prot = protectPerforceFilename(ent.getKey());
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
vladak: The method can be package private (i.e. omit all the specifiers so the method signature becomes just `static String protectPerforceFilename(String name)`) because it is used by the class itself and also by the test case which resides in the same package.

If you use any decent IDE for editing the sources, it will suggest the change in fact.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
vladak: same thing as with the test method - import the static method and use it directly without having to specify the class.

vladak:
One more note - you might want to add your copyright to the modified files.

vladak:
Also, if you add the text `fixes #2746` as a commit message, it will automatically close the related issue upon merging the pull request.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
vladak: Oh, this cannot be done because you are using anonymous inner class for initializing the array list.

The alternative I had in mind:
```java
        ArrayList<SimpleImmutableEntry<String,String>> testmap = new ArrayList<>();
        testmap.add(new SimpleImmutableEntry<>("Testfile 34", "Testfile 34"));
        testmap.add(new SimpleImmutableEntry<>("Test%52", "Test%2552"));
...
```

vladak:
I think the lack of coverage is caused by not running any Perforce test as part of running the build within Travis (which happens to be the environment used to submit the results to Coveralls). This is because the test class is annotated with `@ConditionalRun(RepositoryInstalled.PerforceInstalled.class)` so if Perforce runtime is not installed (which is the case with Travis) then all the tests in the test class will be skipped.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
cross: Ahh, okay.  I tried to be fancy by initializing in the declaration.  But, I see the value in the simpler approach you have above.  I'll do that, thanks.

cross: Related question.  I have an instance of the class instantiated here, but am only testing a static method.  Would it be preferable just to call the static method, or instantiate a class object in this test?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
cross: Okay.  Can do.  I was just following the example of the calls to `getRevisionCmd`, but happy to do it this way instead.  Thanks for explaining how to accomplish it.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
cross: Well, I'm old school, my IDE is vi.  :-)  Thanks, change made.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
cross: Actually, before I push up my last commit.  @vladak - How would you prefer I engage this method in this test?  From an instanced instance of PerforceRepository, or just as a static method in PerforceRepository?  Neither is more correct, I don't think, so I leave it to preference.  Let me know, so I can take out the question comment here on that subject.

vladak: Just use the static method directly.

cross:
@vladak Do you have a suggestion as to whether I call the static method directly from the test function, or within an instance of PerforceRepository?  If you don't care, I'll leave it as it, but will remove the questioning comment before asking the PR be merged.

vladak:
Just do the static import and call it directly.

Dne pá 10. 5. 2019 21:03 uživatel Chris Ross <notifications@github.com>
napsal:

cross:
Done and done.  Thanks @vladak .  Ready for merge if you agree.

vladak:
Looks fine, thanks !

PR #2769, opened 2019-05-15
Perforce incremental history

cross:
This implements PerforceRepository.getHistory(File,String) to support incremental history.


vladak:
Looks sane, thanks !

PR #2779, opened 2019-05-24
ignore eclipse and vscode ide files

wy193777:
So people use eclipse and vscode will feel a little bit happier.


vladak:
Thanks !

PR #2719, opened 2019-03-16
Templating projects and repositories on the main page

tulinkry:
depends on #2718

Using jstl custom tags to implement reusable code.

PR #2796, opened 2019-06-07
escape string literal using single quotes

vladak:
The advice from https://stackoverflow.com/questions/8898815/how-to-use-both-single-and-double-quotes-inside-jstl-el-expression seems to fix the problem of attribute value evaluation.


vladak:
This is follow-up to #2719.

vladak:
I verified in a custom repo that using double quotes and passing the set by value is enough.

Issue #2798, opened 2019-06-11
how to have multiple Perforce repositories as projects


cross:
This is a question.  I have a source tree, /disk/src, and under that are top-level directories for multiple (currently 12) projects from multiple source repositories.  I am running opengrok indexer with `-s /disk/src -d /disk/data -H -P -S -G`.  What the output from the console shows, and the log file, includes:

```
11:08:10 INFO: Scanning for repositories...
11:08:11 INFO: Done scanning for repositories, found 1 repositories (took 825 ms)
11:08:11 INFO: Generating history cache for all repositories ...
11:08:11 INFO: Creating historycache for 1 repositories
11:08:11 INFO: Creating historycache for /disk/src (PerforceRepository) without renamed file handling
11:41:42 INFO: Done historycache for /data/opengrok/src (took 0:33:31)
11:41:42 INFO: Done historycache for all repositories (took 0:33:31)
11:41:42 INFO: Done...
11:41:42 INFO: Starting indexing
11:41:43 INFO: Waiting for the executors to finish
11:41:43 INFO: Creating historycache for /disk/src (PerforceRepository) without renamed file handling
11:41:43 INFO: Creating historycache for /disk/src (PerforceRepository) without renamed file handling
11:41:43 INFO: Creating historycache for /disk/src (PerforceRepository) without renamed file handling
11:41:43 INFO: Creating historycache for /disk/src (PerforceRepository) without renamed file handling
```

It seems to be creating projects properly, when I view results in the U/I later, but.  Are the log messages just misleading, or am I doing something wrong?

(side note, only some of the projects are Perforce, some are git, it seems to always say `/disk/src (PerforceRepository)` which is in a way the most concerning part of the incorrect log messages...)

vladak:
Those identical paths do not look right to me. Also, the indexer found just one repository. If you run the indexer with -W, what is inside the XML configuration in terms of projects/repositories ?

vladak:
As for repository detection, for some repository types it is quite simple, e.g. for Git it is sufficient to have `.git` directory: https://github.com/oracle/opengrok/blob/f23ad94568a09e115e8d48c641573dcf2ef7408a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java#L528-L534

That said, it muse be within the search depth which is 3 by default, see the comment:
[...]

PR #2801, opened 2019-06-12
disallow repository path equal to source root

vladak:
The detection of most repository types relies on the existence of special directory/file (e.g. the `.git` directory for Git). Perforce runs external command to detect whether given directory belongs to a repository. When misconfigured this could lead to strange indexing behavior.

This should help to detect+prevent cases as described in #2798.


vladak:
Another place to fix that would be `RepositoryInfo#setDirectory()` however that would rely on classes that override this method to call `super`.

tulinkry:
Did you setup a test case for it or the existing test cases cover it?

vladak:
I am yet to think of a test case. It will have to be a bit artificial.

PR #2803, opened 2019-06-13
add missing brackets for method call

vladak:
This should fix the missing context path in project links on index page.


vladak:
Turns out JSTL checking is not that strong. Also, IDEA Ultimate does not follow through the method calls/references for `request` in tag libraries, maybe there is some import missing ?

PR #2809, opened 2019-06-14
Add support for getting CurrentVersion of PerforceRepository

cross:
Implement determineCurrentVersion() for PerforceRepository

note, I tested this and it works.  However, I've found that after loading the root in the Web U/I, then updating one of the perforce repositories, the web U/I doesn't refresh.  It is still showing the version test of that repository when I first loaded the page.  Shift-refresh doesn't help, and waiting an hour didn't help.

Have I missed something to make this know it needs to recheck via determineCurrentVersion()?

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
@@ -272,6 +272,24 @@ public class PerforceRepository extends Repository {
 
     @Override
     String determineCurrentVersion(boolean interactive) throws IOException {
-        return null;
+        File directory = new File(getDirectoryName());
+        List<String> cmd = new ArrayList<>();
+
+        ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);
+        cmd.add(RepoCommand);
+        cmd.add("changes");
+        cmd.add("-t");
+        cmd.add("-m");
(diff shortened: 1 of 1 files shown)


vladak:
The refresh of repository information happens only after reindex that is finished with sending the new configuration to the web app (with the -U option).

cross:
Screen Shot 2019-06-14 at 16 03 56

And, a few minutes after initially restarting tomcat, I updated that source tree.  The relevant perforce command (that's now used by determineCurrentVersion()) in the tree shows:

`Change 1450791 on 2019/06/14 11:43:10 by otheruser@otherclient 'other message'`

I presume if I restart tomcat I'll see that in the U/I, but I also presume there should be some way the web interface knows to recheck...

cross:
Ahh.  Okay.  Thanks vladak...  Got it. Should be good to merge this then when tests complete.

vladak:
If you restart Tomcat (or redeploy the web app) then the refreshed information will show up because the web app will load the fresh configuration from disk (assuming it was written using the -W option of the indexer). As I mentioned above, the configuration (which contains the repository information with current version) will be refreshed also by sending the new one at the end of the indexing with the -U option.

vladak:
Yep, I will just wait for Wercker build to complete as it is faster than Travis (which expands into build matrix) just to see if checkstyle checks are clear.

vladak:
Thanks !

PR #2814, opened 2019-06-18
JSTL vs Paths

vladak:
correctly displays subrepository path on the index page

### DIFF
--- a/opengrok-web/src/main/webapp/WEB-INF/tags/repository.tag
+++ b/opengrok-web/src/main/webapp/WEB-INF/tags/repository.tag
@@ -42,7 +42,7 @@ Portions Copyright (c) 2019, Krystof Tulinger <k.tulinger@seznam.cz>.
 
 <c:if test="${isSubrepository}">
     <c:set var="name"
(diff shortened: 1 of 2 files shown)


opengrok-web/src/main/webapp/WEB-INF/tags/repository.tag
```diff
@@ -42,7 +41,7 @@ Portions Copyright (c) 2019, Krystof Tulinger <k.tulinger@seznam.cz>.
 
 <c:if test="${isSubrepository}">
     <c:set var="name"
-           value="${Paths.get(pageConfig.sourceRootPath).relativize(Paths.get(repositoryInfo.directoryName))}"/>
+           value="${pageConfig.getRelativePath(pageConfig.sourceRootPath, repositoryInfo.directoryName)}"/>
```
tulinkry: you're not using the static call but of course it does not matter

tulinkry:
Good job tackling the issue!

opengrok-web/src/main/webapp/WEB-INF/tags/repository.tag
vladak: Good point.

PR #2817, opened 2019-06-19
add sub-repositories of the same type as parent

vladak:
This change restricts the discovery of sub-repositories to the type of repository already found.


vladak:
This is done more for efficiency (esp. those repository types which do not just look for special files/directories but execute a command to detect a repository). Also, mixing repository types is uncommon, I think.

tulinkry:
I agree that this is an edge case. Let's merge

PR #2846, opened 2019-07-02
checking return values for incoming check

### LINKED ISSUE #2843: python tools - --incoming check does not return the correct exit value
```
ktulinger:opengrok ktulinger$ opengrok-mirror --loglevel INFO -c /var/opengrok/etc/mirror-config-generated.yml --incoming -U http://localhost:8080 project-x
output of ['/usr/local/bin/git', 'pull', '--dry-run']:

No incoming changes for repositories in project project-x
ktulinger:opengrok ktulinger$ echo $?
0
```

As I understood the return value should be 2, so the sync chain is skipped however this is never met.

tulinkry:
- refactoring the code to use the constants

fixes #2843

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/mirror.py
@@ -38,6 +38,11 @@ from multiprocessing import Pool, cpu_count
 
 from filelock import Timeout, FileLock
 
+from .utils.exitvals import (
+    FAILURE_EXITVAL,
+    CONTINUE_EXITVAL,
+    SUCCESS_EXITVAL
+)
 from .utils.log import get_console_logger, get_class_basename, \
     fatal, get_batch_logger
 from .utils.opengrok import get_config_value, list_indexed_projects
@@ -71,7 +76,7 @@ def worker(args):
 
 
 def main():
-    ret = 0
+    ret = SUCCESS_EXITVAL
 
     parser = argparse.ArgumentParser(description='project mirroring',
                                      parents=[get_baseparser(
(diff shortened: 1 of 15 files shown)


opengrok-tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -177,11 +182,13 @@ def main():
                 except KeyboardInterrupt:
                     sys.exit(1)
                 else:
-                    if any([True for x in project_results if x == 1]):
-                        ret = 1
+                    if any([True for x in project_results if x == FAILURE_EXITVAL]):
+                        ret = FAILURE_EXITVAL
+                    if all([True for x in project_results if x == CONTINUE_EXITVAL]):
```
vladak: I think you need to check indeed for all. Something like:
```python
(len([x for x in project_results if x == 2]) == len(project_results))
```

vladak: or something like:
```python
 all(list(map(lambda x: True if x == 1 else False, project_results)))
```

vladak:
There is the logic problem + style issues, otherwise this is good to go.

opengrok-tools/src/main/python/opengrok_tools/mirror.py
tulinkry: yes I discovered the problem also

tulinkry: I went for `all([x == CONTINUE_EXITVAL for x in project_results])`

tulinkry: which is essentially your last suggestion

tulinkry:
How about now?

PR #2847, opened 2019-07-02
Python tools improved incoming check

### LINKED ISSUE #2844: python tools --incoming check should check if the project has no index yet
I'm not sure if this is implemented or not right now but I'm adding an issue to find this out.

When using the incoming check for mirror.py + sync.py combo (after #2843 is fixed), I suspect that for the first time the sync.py is run (projects are freshly clonned to src_root, no index exists for them), the incomming check would skip the indexing for all of the projects (=no changes as expected because it has just been clonned). So the check could be more robust and check the index presence and also the scm changes presence.

tulinkry:
Improving --incoming check to check also the index presence
for the particular project.
    
After that, the name "incoming" could be a little misleading
so I refactored the name to "check changes".
    
fixes #2844

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -26,6 +26,7 @@ import re
 import os
 import fnmatch
 import logging
+import urllib
 
 from .exitvals import (
     FAILURE_EXITVAL,
@@ -33,7 +34,8 @@ from .exitvals import (
     SUCCESS_EXITVAL
 )
 from .utils import is_exe, check_create_dir, get_int
-from .opengrok import get_repos, get_repo_type
+from .webutil import get
+from .opengrok import get_repos, get_repo_type, get_uri
 from .hook import run_hook
 
 from ..scm.repofactory import get_repository
@@ -196,13 +198,13 @@ def get_project_properties(project_config, project_name, hookdir):
         use_proxy, ignored_repos
 
 
-def mirror_project(config, project_name, check_incoming, uri,
+def mirror_project(config, project_name, check_changes, uri,
                    source_root):
     """
     Mirror the repositories of single project.
     :param config global configuration dictionary
     :param project_name: name of the project
-    :param check_incoming:
+    :param check_changes:
     :param uri
     :param source_root
     :return exit code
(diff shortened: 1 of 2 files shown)


opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -244,22 +246,44 @@ def mirror_project(config, project_name, check_incoming, uri,
...
+            r.raise_for_status()
+            if not bool(r.json()):
+                changes_detected = True
+                logger.debug('Project {} has not been indexed yet'
+                             .format(project_name))
+        except ValueError as e:
+            logger.error('Unable to parse project \'{}\' indexed flag: {}'
+                         .format(project_name, e))
+            return FAILURE_EXITVAL
+        except Exception as e:
```
vladak: maybe too broad ?

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -244,22 +246,44 @@ def mirror_project(config, project_name, check_incoming, uri,
...
+            logger.error('Unable to determine project \'{}\' indexed flag: {}'
+                         .format(project_name, e))
+            return FAILURE_EXITVAL
 
-        if not got_incoming:
+        # check if the project has any new changes in the scm
```
vladak: SCM

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -196,13 +198,13 @@ def get_project_properties(project_config, project_name, hookdir):
         use_proxy, ignored_repos
 
 
-def mirror_project(config, project_name, check_incoming, uri,
+def mirror_project(config, project_name, check_changes, uri,
```
vladak: the function is getting too long. refactoring needed.

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -196,13 +198,13 @@ def get_project_properties(project_config, project_name, hookdir):
...
                    source_root):
     """
     Mirror the repositories of single project.
     :param config global configuration dictionary
     :param project_name: name of the project
-    :param check_incoming:
+    :param check_changes:
```
vladak: add some description pls

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
tulinkry: changed to HTTPError from requests

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
tulinkry: added

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
tulinkry: agreed

vladak: pls file a new issue then or maybe even do it now

tulinkry: https://github.com/oracle/opengrok/issues/2849

PR #2851, opened 2019-07-03
Python mirror tool refactor and test

vladak:
To prevent further regressions of #2843, I added a simple unit test. While at it, I refactored the meat of the mirroring - the `mirror_project()` function as suggested in the recent code review.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -339,7 +339,7 @@ def mirror_project(config, project_name, check_changes, uri,
 
     #
     # If one of the repositories fails to sync, the whole project sync
-    # is treated as failed, i.e. the program will return 1.
+    # is treated as failed, i.e. the program will return FAILURE_EXITVAL.
     #
     for repo in repos:
         logger.info("Synchronizing repository {}".
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -244,7 +244,7 @@ def process_changes(repos, project_name, uri):
         if not bool(r.json()):
             changes_detected = True
             logger.info('Project {} has not been indexed yet'
-                         .format(project_name))
+                        .format(project_name))
     except ValueError as e:
         logger.error('Unable to parse project \'{}\' indexed flag: {}'
                      .format(project_name, e))
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -229,7 +229,7 @@ def process_changes(repos, project_name, uri):
     """
     :param repos: repository list
     :param project_name: project name
-    :return: exit code or None
+    :return: exit code
     """
     logger = logging.getLogger(__name__)
 
@@ -274,7 +274,7 @@ def process_changes(repos, project_name, uri):
                     format(project_name))
         return CONTINUE_EXITVAL
 
-    return None
+    return SUCCESS_EXITVAL
 
 
 def mirror_project(config, project_name, check_changes, uri,
@@ -330,7 +330,7 @@ def mirror_project(config, project_name, check_changes, uri,
     # Check if the project or any of its repositories have changed.
     if check_changes:
         r = process_changes(repos, project_name, uri)
-        if r is not None:
+        if r != SUCCESS_EXITVAL:
             return r
 
     if not process_hook("pre", prehook, source_root, project_name, proxy,
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -243,7 +243,7 @@ def process_changes(repos, project_name, uri):
         r.raise_for_status()
         if not bool(r.json()):
             changes_detected = True
-            logger.debug('Project {} has not been indexed yet'
+            logger.info('Project {} has not been indexed yet'
                          .format(project_name))
     except ValueError as e:
         logger.error('Unable to parse project \'{}\' indexed flag: {}'
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -200,6 +200,83 @@ def get_project_properties(project_config, project_name, hookdir):
         use_proxy, ignored_repos
 
 
+def process_hook(hook_ident, hook, source_root, project_name, proxy,
+                 hook_timeout):
+    """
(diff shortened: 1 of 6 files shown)


opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -200,6 +200,83 @@ def get_project_properties(project_config, project_name, hookdir):
...
+
+    # check if the project is a new project - full index is necessary
+    try:
+        r = get(logger, get_uri(uri, 'api', 'v1', 'projects',
+                                urllib.parse.quote_plus(project_name),
+                                'property', 'indexed'))
+        r.raise_for_status()
+        if not bool(r.json()):
+            changes_detected = True
+            logger.debug('Project {} has not been indexed yet'
```
tulinkry: I know this is my line but it may be info as this is for the whole project, to match the `logger.info('No incoming changes ... `

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -109,3 +114,59 @@ def test_valid_config():
...
+    Test that the special CONTINUE_EXITVAL value bubbles all the way up to
+    the mirror.py return value.
+    """
+
+    class MockResponse:
+
+        # mock json() method always returns a specific testing dictionary
+        @staticmethod
+        def json():
+            return {"indexed": "true"}
```
tulinkry: be careful here, because the call to the api does not return a dict, it returns the boolean literal

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -109,3 +114,59 @@ def test_valid_config():
...
+        repo_name = "parent_repo"
+        repo_path = os.path.join(source_root, repo_name)
+        cloned_repo_name = "cloned_repo"
+        cloned_repo_path = os.path.join(source_root, cloned_repo_name)
+        project_name = "foo"    # does not matter for this test
+
+        os.mkdir(repo_path)
+
+        def mock_get_repos(*args, **kwargs):
+            from opengrok_tools.scm.repofactory import get_repository
```
tulinkry: any special reason why the import is here?

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -109,3 +114,59 @@ def test_valid_config():
...
+
+            return [get_repository(cloned_repo_path,
+                                   "git", project_name,
+                                   None, None, None, None)]
+
+        def mock_get(*args, **kwargs):
+            return MockResponse()
+
+        # Clone a Git repository so that it can pull.
+        import pygit2
```
tulinkry: any special reason why the import is here?

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -109,3 +114,59 @@ def test_valid_config():
...
+        pygit2.init_repository(repo_path, True)
+        pygit2.clone_repository(repo_path, cloned_repo_path)
```
tulinkry: I'm not that convinced that it is worth bringing this dependency for simple `git init` and `git clone` commands call.

opengrok-tools/src/test/python/test_mirror.py
```diff
@@ -109,3 +114,59 @@ def test_valid_config():
     assert check_configuration({PROJECTS_PROPERTY:
                                 {"foo": {PROXY_PROPERTY: True}},
                                 PROXY_PROPERTY: "proxy"})
+
+
+@pytest.mark.skipif(not os.name.startswith("posix"), reason="requires posix")
```
tulinkry: any reason why this should work only on posix?

opengrok-tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -200,6 +200,83 @@ def get_project_properties(project_config, project_name, hookdir):
...
+                             'repository {}'.format(repo))
+                return FAILURE_EXITVAL
+
+    if not changes_detected:
+        logger.info('No incoming changes for repositories in '
+                    'project {}'.
+                    format(project_name))
+        return CONTINUE_EXITVAL
+
+    return None
```
tulinkry: I don't like mixing ints and None in return values, maybe SUCCESS_EXITVAL would be better?

opengrok-tools/src/test/python/test_mirror.py
vladak: It is used just for this one function.

opengrok-tools/src/test/python/test_mirror.py
vladak: Recently I realized that imports can be anywhere so I am a bit on the path to use it as much as I can (a little bit like the joy of variable declaration of C99 compared to C89). Will use it more sparingly once I get enough :-)

opengrok-tools/src/test/python/test_mirror.py
vladak: I think both approaches have their pros and cons. I used pygit2 mainly because it is an API (and APIs are out there for consumption), and widely used one. If I were to use command execution, I'd have to deal with return codes, error strings and rely on the command line interface to stay stable which is arguably less likely than the programmatic API. Also, this should encourage anyone to add more tests to the mirroring - which is minimally covered - seeing how easy it is.

opengrok-tools/src/test/python/test_mirror.py
vladak: should work on Windows too, let me try.

opengrok-tools/src/test/python/test_mirror.py
tulinkry: however this function is executed always when this file is executed so it does not bring anything extra

opengrok-tools/src/test/python/test_mirror.py
tulinkry: One thing is that we use it consistently I think in these tools at the top of the file.
Second thing is that it gives benefit when the function with the import is not executed (so the import does not trigger) but in this case it is executed each time

opengrok-tools/src/test/python/test_mirror.py
tulinkry: fair enough

opengrok-tools/src/test/python/test_mirror.py
tulinkry: more it hides the actual runtime dependency for each run

opengrok-tools/src/test/python/test_mirror.py
vladak: hangs in the newly added test for some reason, will keep posix for now.

opengrok-tools/src/test/python/test_mirror.py
vladak: okay, will group them in the beginning

opengrok-tools/src/test/python/test_mirror.py
tulinkry: ok

PR #2854, opened 2019-07-04
These tests were written using Diffblue Cover.

ThomasPerkins1123:
Hi,

I've analysed your code base and noticed that `org.opengrok.indexer.util.BooleanUtil` in the `opengrok` module is not fully tested.

These tests have been created by [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.


vladak:
This is fairly trivial test. Any reasonable extension of test coverage is welcome. Thanks !

PR #2868, opened 2019-07-18
refresh repository README on Docker hub after image push

vladak:
The README file on Docker hub is not automatically refreshed. This is because the Docker image builds are performed in Travis and not on Docker hub.

Using https://stackoverflow.com/a/55082686/11582827 and https://success.docker.com/article/how-do-i-authenticate-with-the-v2-api I made this work without enabling automated builds in Docker hub. The API does not seem to be publicly documented but works.

### DIFF
--- a/dev/before_install
+++ b/dev/before_install
@@ -20,7 +20,8 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
 	    python3 \
 	    python3-venv \
 	    python3-pip \
-	    nodejs
+	    nodejs \
+	    jq
 	if [[ $? != 0 ]]; then
 		echo "cannot install extra packages"
 		exit 1
--- a/dev/before_install
+++ b/dev/before_install
@@ -38,7 +38,7 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
 elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
 	brew update
 
-	brew install cvs libgit2
+	brew install cvs libgit2 jq
 	if [[ $? != 0 ]]; then
 		echo "cannot install extra packages"
 		exit 1
--- a/dev/docker.sh
+++ b/dev/docker.sh
@@ -106,7 +106,7 @@ TOKEN=$(curl -s -H "Content-Type: application/json" -X POST \
     -d '{"username": "'${DOCKER_USERNAME}'", "password": "'${DOCKER_PASSWORD}'"}' \
     https://hub.docker.com/v2/users/login/ | jq -r .token)
 if [[ -z $TOKEN ]]; then
-	echo "Cannot get auth token"
+	echo "Cannot get auth token to publish the README file"
 	exit 1
 fi
 
--- a/dev/docker.sh
+++ b/dev/docker.sh
@@ -48,13 +48,13 @@ docker ps -a
 
 # Travis can only work on master since it needs encrypted variables.
 if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then
-	echo "Not pushing docker image for pull requests"
+	echo "Not pushing Docker image for pull requests"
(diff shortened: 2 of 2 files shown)


dev/docker.sh
```diff
@@ -13,50 +13,15 @@
...
+IMAGE="opengrok/docker"
 
 # Travis can only work on master since it needs encrypted variables.
 if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then
-	echo "Not publishing docker image for pull requests"
+	echo "Not building docker image for pull requests"
 	exit 0
 fi
 
-# The push only works on the main repository.
```
tulinkry: why is this no longer needed?

dev/docker.sh
```diff
@@ -72,14 +37,77 @@ if [[ -z $DOCKER_PASSWORD ]]; then
...
+		echo "Successfully pushed README to Docker Hub"
+	else
+		printf "Unable to push README to Docker Hub, response code: %s\n" "${code}"
+		exit 1
+	fi
+}
+
+TOKEN=$(curl -s -H "Content-Type: application/json" -X POST \
+    -d '{"username": "'${DOCKER_USERNAME}'", "password": "'${DOCKER_PASSWORD}'"}' \
+    https://hub.docker.com/v2/users/login/ | jq -r .token)
```
tulinkry: do you ensure jq is installed?

dev/docker.sh
```diff
@@ -72,14 +37,77 @@ if [[ -z $DOCKER_PASSWORD ]]; then
...
+if [[ -z $TOKEN ]]; then
+	echo "Cannot get auth token"
```
tulinkry: Cannot get auth token to publish the readme (so as you know in the log why it needs the token)

dev/docker.sh
```diff
@@ -72,14 +37,77 @@ if [[ -z $DOCKER_PASSWORD ]]; then
 	exit 1
 fi
 
+VERSION="$TRAVIS_TAG"
+VERSION_SHORT=$( echo $VERSION | cut -d. -f1,2 )
```
tulinkry: how about default "latest" ? when you build the image in your local dev env

dev/docker.sh
```diff
@@ -72,14 +37,77 @@ if [[ -z $DOCKER_PASSWORD ]]; then
...
+		exit 1
+	fi
+
+	local code=$(jq -n --arg msg "$(<${input_file})" \
+	    '{"registry":"registry-1.docker.io","full_description": $msg }' | \
+	        curl -s -o /dev/null  -L -w "%{http_code}" \
+	           https://cloud.docker.com/v2/repositories/"${image}"/ \
+	           -d @- -X PATCH \
+	           -H "Content-Type: application/json" \
+	           -H "Authorization: JWT ${token}")
```
tulinkry: just to understand it...this reads the input file (=README) and sends it through the api to docker, right?

tulinkry: looks like nice oneliner but little hard to follow at first sight

vladak: yes. I copied the whole thing from elsewhere.

dev/docker.sh
vladak: mismerge, should be fixed now.

dev/docker.sh
vladak: fixed the mismerge

dev/before_install
```diff
@@ -20,7 +20,8 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
 	    python3 \
 	    python3-venv \
 	    python3-pip \
-	    nodejs
+	    nodejs \
+	    jq
```
tulinkry: just curious...why not in the osx branch?

vladak: just forget, adding.

Issue #2872, opened 2019-07-23
Tomcat security and API


unhipzippo:
(Testing on OpenGrok 1.2.24, AdoptOpenJDK 11.0.3, Tomcat 9.0.21 on Oracle Linux 7.6)

If OpenGrok is running inside a Tomcat instance which uses Tomcat security to protect the entire OpenGrok webapp (rather than using the OpenGrok-specific authorization framework), the Tomcat security defaults to extending to the OpenGrok API as well. Since the OpenGrok indexer's call to the configuration API endpoint doesn't expect/include authorization, this results in indexer completion never informing the webapp of the new configuration.

Somewhat related issues:

https://github.com/oracle/opengrok/issues/2352
https://github.com/oracle/opengrok/issues/2635

In order to permit the indexer to update the webapp with new configuration, https://github.com/oracle/opengrok/wiki/Authorization recommends exempting API calls from security entirely:

```
   <security-constraint>
       <web-resource-collection>                                               
           <web-resource-name>API endpoints are checked separately by the web app</web-resource-name>
           <url-pattern>/api/*</url-pattern>                                   
       </web-resource-collection>                                              
   </security-constraint>
```

However, since the API **search** endpoint can return data which may be considered privileged (e.g. lines of source code), permitting completely public access to the **search** / **suggest** endpoints might not be ideal. (@tulinkry alludes to this in https://github.com/oracle/opengrok/issues/2635#issuecomment-474227453)

Is there a recommended alternative configuration if we want to apply security to API calls?

unhipzippo:
One possibility we're experimenting with is some tweaks to both Tomcat and to our preexisting front-end proxy (nginx):

- Proxy (nginx) on port :80, OpenGrok listening on localhost only, e.g. localhost:8080.
  e.g. Tomcat server.xml:

```
    <Connector address="127.0.0.1" port="8080" protocol="HTTP/1.1"
               connectionTimeout="20000"
               maxHttpHeaderSize="16384"
               redirectPort="8443" />
```

- Add front-end proxy rules to explicitly block all paths under /<webapp_context/api/v1/ EXCEPT for search, suggest and suggest/config endpoints.
  e.g. nginx .../conf/vhosts/opengrok.conf:

```
upstream opengrok-main-http-backend {
    server 127.0.0.1:8080;
    keepalive 10;
    keepalive_timeout 15s;
}

server {
    server_name opengrok.company.com;

    listen *:80;

    location /source/ {

[...]

PR #2875, opened 2019-07-23
add decoder for HTTP Basic auth

vladak:
As discussed in #2872, the current documentation on https://github.com/oracle/opengrok/wiki/Authorization gives dangerous advice in terms of enabling HTTP Basic authentication for the web application.

This decoder addition forms a basis for subsequent work (e.g. adding an authorization plugin that checks the user against a whitelist, similarly to what the pre-existing Ldap plugins do).

I will update the above mentioned documentation once this is merged in.

### DIFF
--- a/plugins/test/opengrok/auth/plugin/decoders/UserPrincipalDecoderTest.java
+++ b/plugins/test/opengrok/auth/plugin/decoders/UserPrincipalDecoderTest.java
@@ -30,16 +30,15 @@ import org.junit.Before;
 import org.junit.Test;
 
 import static org.junit.Assert.assertNull;
-import static opengrok.auth.plugin.decoders.HttpBasicAuthHeaderDecoder.AUTHORIZATION_HEADER;
 
-public class HttpBasicAuthDecoderTest {
+public class UserPrincipalDecoderTest {
     DummyHttpServletRequestUser dummyRequest;
-    HttpBasicAuthHeaderDecoder decoder = new HttpBasicAuthHeaderDecoder();
+    UserPrincipalDecoder decoder = new UserPrincipalDecoder();
 
     @Before
     public void setUp() {
         dummyRequest = new DummyHttpServletRequestUser();
-        dummyRequest.setHeader(AUTHORIZATION_HEADER, "Basic Zm9vOmJhcg==");
+        dummyRequest.setHeader("authorization", "Basic Zm9vOmJhcg==");
     }
 
     @Test
--- a/plugins/test/opengrok/auth/plugin/decoders/UserPrincipalDecoderTest.java
+++ b/plugins/test/opengrok/auth/plugin/decoders/UserPrincipalDecoderTest.java
@@ -38,11 +38,12 @@ public class UserPrincipalDecoderTest {
     @Before
     public void setUp() {
         dummyRequest = new DummyHttpServletRequestUser();
-        dummyRequest.setHeader("authorization", "Basic Zm9vOmJhcg==");
     }
 
     @Test
-    public void testAll() {
+    public void testHttpBasicDecoding() {
+        dummyRequest.setHeader("authorization", "Basic Zm9vOmJhcg==");
+
         User result = decoder.fromRequest(dummyRequest);
 
         Assert.assertNotNull(result);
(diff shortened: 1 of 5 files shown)


vladak:
Tested by adding this to `WEB-INF/web.xml`:
```xml
    <security-constraint>
        <web-resource-collection>
            <web-resource-name>API endpoints are checked separately by the web app</web-resource-name>
            <url-pattern>/api/*</url-pattern>
        </web-resource-collection>
    </security-constraint>

    <security-constraint>
        <web-resource-collection>
            <web-resource-name>In general everything needs to be authenticated</web-resource-name>
            <url-pattern>/*</url-pattern> <!-- protect the whole application -->
        </web-resource-collection>

        <auth-constraint>
            <role-name>*</role-name>
        </auth-constraint>

        <user-data-constraint>
            <transport-guarantee>NONE</transport-guarantee>
        </user-data-constraint>
    </security-constraint>

    <security-role>
        <role-name>*</role-name>
    </security-role>

    <login-config>
        <auth-method>BASIC</auth-method>
    </login-config>
```
and creating `conf/tomcat-users.xml`:
```xml
<?xml version='1.0' encoding='utf-8'?>
<tomcat-users xmlns="http://tomcat.apache.org/xml"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="http://tomcat.apache.org/xml tomcat-users.xsd"
              version="1.0">

  <user username="foo" password="bar" roles="tomcat,manager-script"/>
</tomcat-users>
```
and adding this to read-only configuration:
```xml
  <void property="pluginStack">
        <void property="stack">
            <!-- get user cred from HTTP headers -->
            <void method="add">
                    <void property="name">
                        <string>opengrok.auth.plugin.UserPlugin</string>
                    </void>
                    <void property="flag">
                        <string>REQUISITE</string>
                    </void>
                    <void property="setup">
                        <void method="put">
                             <string>decoder</string>
                             <string>opengrok.auth.plugin.decoders.HttpBasicAuthHeaderDecoder</string>
                        </void>
                    </void>
                </object>
            </void>
        </void>
  </void>
```
Running the indexer with:
```
-s
/var/opengrok/src
-d
/var/opengrok/data
-P
-H
-S
-G
-R
/var/opengrok/etc/readonly_configuration-BasicHttpAuth.xml
-W
-U
http://localhost:8080/source
```
works fine.

vladak:
To make sure this covers `/search` API endpoint checks I added project/group matching so that the stack config looks like this:
```xml
  <void property="pluginStack">                                                 
        <void property="stack">                                                 
            <!-- get user cred from HTTP headers -->                            
            <void method="add">                                                 
                    <void property="forGroups">                                 
                            <void method="add">                                 
                                <string>Solaris</string>                        
                            </void>                                             
                    </void>                                                     
                    <void property="forProjects">                               
                            <void method="add">                                 
                                <string>solaris-userland</string>               
                            </void>                                             
                    </void>                                                     
                    <void property="name">                                      
                        <string>opengrok.auth.plugin.UserPlugin</string>        
                    </void>                                                     
                    <void property="flag">                                      
                        <string>REQUISITE</string>                              
                    </void>                                                     
                    <void property="setup">                                     
                        <void method="put">                                     
                             <string>decoder</string>                           
                             <string>opengrok.auth.plugin.decoders.HttpBasicAuthHeaderDecoder</string>
                        </void>                                                 
                    </void>                                                     
                </object>                                                       
            </void>                                                             
        </void>                                                                 
  </void>          
```
and did a non-localhost request (`http_proxy= curl -s 'http://10.x.y.z:8080/source/api/v1/search?full=hack&projects=solaris-userland&projects=jna' | jq`) and compared that to the results received from before the change - the `solaris-userland` was no longer included in the results.

Also, to add some detail about the testing, the read-only configuration had this:
```xml
  <void property="pluginDirectory">                                             
   <string>/var/opengrok/share/auth/plugins</string>                            
   <!-- <string>/usr/opengrok/share/plugins</string> -->                        
  </void>
```
The directory contained `plugins.jar` from the original install and I copied the new decoder binary to `/var/opengrok/share/auth/plugins/opengrok/auth/plugin/decoders/HttpBasicAuthHeaderDecoder.class`.

plugins/test/opengrok/auth/plugin/decoders/UserPrincipalDecoderTest.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.junit.Assert.assertNull;
+
+public class UserPrincipalDecoderTest {
+    DummyHttpServletRequestUser dummyRequest;
+    UserPrincipalDecoder decoder = new UserPrincipalDecoder();
+
+    @Before
+    public void setUp() {
+        dummyRequest = new DummyHttpServletRequestUser();
+        dummyRequest.setHeader("authorization", "Basic Zm9vOmJhcg==");
```
tulinkry: no blocking; this is just for one test so it could be only in the related method.

PR #2878, opened 2019-07-25
make cleanup a list in CommandSequence

vladak:
Allows to have multiple cleanup "commands" in `opengrok-sync` configuration:

Sample config:
```yml
commands:
  # returns 1
  - command: [cat, '/foo']
cleanup:
  - command: [ls, '/etc']
  - command: [echo, 'cleanup']
```
Even if any of the cleanup commands fail, the execution continues, unlike what is done for the main sequence of commands.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
@@ -48,6 +48,9 @@ class CommandSequenceBase:
         self.failed = False
         self.retcodes = {}
         self.outputs = {}
+        if cleanup and not isinstance(cleanup, list):
+            raise Exception("cleanup is not a list of commands")
+
         self.cleanup = cleanup
         self.loglevel = loglevel
         self.driveon = driveon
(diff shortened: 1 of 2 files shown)


opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
```diff
@@ -173,13 +176,17 @@ def run(self):
...
+
+        for cleanup_cmd in self.cleanup:
+            if is_web_uri(cleanup_cmd.get("command")[0]):
+                self.call_rest_api(cleanup_cmd)
             else:
-                command_args = self.cleanup.get("command")
+                command_args = cleanup_cmd.get("command")
                 self.logger.debug("Running cleanup command '{}'".
                                   format(command_args))
                 cmd = Command(command_args,
```
tulinkry: it looks like it works but just to verify - when one of the commands fails, do the following commands get executed?

opengrok-tools/src/test/python/test_command_sequence.py
```diff
@@ -116,3 +117,34 @@ def test_project_subst():
...
+                        {"command": ["/usr/bin/touch", file_bar]}]
+        # Running 'cat' on non-existing entry causes it to return 1.
+        cmd_list = [{"command": ["/bin/cat", "/foobar"]}]
+        commands = CommandSequence(CommandSequenceBase("test-cleanup-list",
+                                                       cmd_list,
+                                                       cleanup=cleanup_list))
+        assert commands is not None
+        commands.run()
+        assert os.path.isfile(file_foo)
+        assert os.path.isfile(file_bar)
```
tulinkry: no blocking; could you also assert the retcodes of the commands?

tulinkry:
Hurray!

opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
vladak: they should, let me augment the test case.

PR #2880, opened 2019-07-29
lucene 8.2.0

tarzanek:
Update to lucene 8.2.0
I also expect a problem in tests, they need fixes
(I am not testing locally yet, had to rebuild my home server)


tarzanek:
@vladak @tulinkry @ahornace  review please?

tarzanek:
@vladak are you OK? I basically made 1.1 the last supported version with this commit (since 1.0 is on lucene 6.5 which is unsupported in lucene 8.x )

vladak:
Sounds okay. This will trigger 1.3.0 release.

PR #2881, opened 2019-07-29
add option to insert XML snippet into web.xml

vladak:
This change adds the ability to insert content to the `web.xml` file in the WAR file. This is handy for enabling authentication while deploying.

Will update https://github.com/oracle/opengrok/wiki/Authorization-based-on-HTTP-Basic-Authentication#application-deployment-descriptor once this is merged in.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/deploy.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/deploy.py
@@ -18,7 +18,7 @@
 # CDDL HEADER END
 
 #
-# Copyright (c) 2008, 2018, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2008, 2019, Oracle and/or its affiliates. All rights reserved.
 # Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
 #
 
@@ -27,10 +27,13 @@ import os
 import tempfile
 from shutil import copyfile
 from zipfile import ZipFile
+import xml.etree.ElementTree as ET
 
 from .utils.log import get_console_logger, get_class_basename, \
     fatal
 from .utils.parsers import get_baseparser
+
+WEB_XML = 'WEB-INF/web.xml'
 
 """
  deploy war file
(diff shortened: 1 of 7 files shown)


opengrok-tools/src/main/python/opengrok_tools/deploy.py
```diff
@@ -27,77 +27,135 @@
...
 
 """
 
 
-def repack_war(logger, sourceWar, targetWar, configFile, defaultConfigFile):
+class ProcessingException(Exception):
+    pass
+
+
+def insert_file(input_xml, insert_xml_file):
```
tulinkry: how about moving this function to a separate file in utils?

vladak: I think this is so specific to this script that it is not worth it.

vladak:
Tested with:
```
$ opengrok-deploy -l DEBUG -i tomcat-security.xml opengrok-web/target/source.war /tmp/newsrc.war
$ rm -rf WEB-INF
$ unzip /tmp/newsrc.war WEB-INF/web.xml
$ cat WEB-INF/web.xml
```

where `web.xml` contents were:
```xml
<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
         http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
         version="3.1">

    <display-name>OpenGrok</display-name>
    <description>A wicked fast source browser</description>
    <context-param>
        <description>Full path to the configuration file where OpenGrok can read its configuration</description>
        <param-name>CONFIGURATION</param-name>
        <param-value>/var/opengrok/etc/configuration.xml</param-value>
    </context-param>


</web-app>
```
and `tomcat-security.xml`:
```xml
<web-app>
    <security-constraint>
        <web-resource-collection>
            <web-resource-name>API endpoints are checked separately by the web app</web-resource-name>
            <url-pattern>/api/*</url-pattern>
        </web-resource-collection>
    </security-constraint>

    <security-constraint>
        <web-resource-collection>
            <web-resource-name>In general everything needs to be authenticated</web-resource-name>
            <url-pattern>/api/v1/search</url-pattern> <!-- protect search endpoint whitelisted above -->
            <url-pattern>/api/v1/suggest/*</url-pattern> <!-- protect suggest endpoint whitelisted above -->
            <url-pattern>/*</url-pattern> <!-- protect the whole application -->
        </web-resource-collection>

        <auth-constraint>
            <role-name>*</role-name>
        </auth-constraint>

        <user-data-constraint>
            <transport-guarantee>NONE</transport-guarantee>
        </user-data-constraint>
    </security-constraint>

    <security-role>
        <role-name>*</role-name>
    </security-role>

    <login-config>
        <auth-method>BASIC</auth-method>
    </login-config>

</web-app>
```

opengrok-tools/src/main/python/opengrok_tools/deploy.py
```diff
@@ -27,77 +27,135 @@
...
+
+    root = ET.fromstring(input_xml)
+    insert_tree = ET.parse(insert_xml_file)
+    insert_root = insert_tree.getroot()
+    index = len(root)
 
-    WEB_XML = 'WEB-INF/web.xml'
+    for elem in list(insert_root.findall('.')):
+        for e in list(elem):
+            root.insert(index, e)
```
tulinkry: there must be something like append, no?

tulinkry:
> One thing to keep in mind is that the contents of the file to be inserted must be fully formed XML.

In case, this is not true, the code blows with an exception, doesn't it?

vladak:
Yes, that's why it is handled:
```
$ opengrok-deploy -i invalid.xml opengrok-web/target/source.war /tmp/newsrc.war 
Repacking opengrok-web/target/source.war with custom configuration path to /tmp/OpenGroktmpWarhcdym5cc.war
Cannot parse file 'invalid.xml' as XML
```

vladak:
Test case added.

opengrok-tools/src/main/python/opengrok_tools/deploy.py
vladak: good call, there is `extend()`: https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.extend

opengrok-tools/src/main/python/opengrok_tools/utils/xml.py
```diff
@@ -0,0 +1,58 @@
...
+        raise XMLProcessingException("Cannot read file '{}'".
+                                     format(insert_xml_file)) from e
+
+    insert_root = insert_tree.getroot()
+
+    for elem in list(insert_root.findall('.')):
+        root.extend(list(elem))
+
+    return '<?xml version="1.0" encoding="UTF-8"?>\n' + \
+           ET.tostring(root, encoding="unicode")
```
tulinkry: nb, isn't there a method of the ET which can print the string including the header? Looks to me like something we shouldn't think about.

vladak: There is: https://stackoverflow.com/questions/15356641/how-to-write-xml-declaration-using-xml-etree-elementtree

PR #2890, opened 2019-08-05
Upgrade jQuery to 3.4.1

ahornace:
Hi,

we've recently had one CVE regarding the jQuery version and I've noticed that OpenGrok is still using an old one so I've updated it.

Thanks :)


vladak:
Thanks and keep it coming ! :-)

tulinkry:
Did you test all the dependent features, namely:

 - differ
 - navigate window
 - scopes window
 - message window
 - hash auto scrolling
 - suggestions
 - select all/deselect/clear on search box
 - restoreFocusAfterSearchSubmit
 - toggling lines
 - toggling annotations
 
?

ahornace:
In release notes they say that it is safe to upgrade from `3.x`.

I've played with it a little and tried some features. However, I did not go into many details since we do not have any UI tests and there is no testing scenario so I would be shooting blanks mostly. We can create step by step wiki for situations like these. I prefer selenium tests though.

tulinkry:
So can you please test the mentioned features that they remain working after the upgrade?

PR #2891, opened 2019-08-06
Minify js and css files

ahornace:
Hi,

I've added maven plugin which minifies `js` and `css` files. I don't think there will be a noticable improvement since our js and css files are quite small.

`jslint` produces a few warnings so I will create an issue for that as soon as this is merged.

If there is a need to debug the `js` code, then simply `debug=true` query param can be appended to the URL.

Changes in the `util.js` file fixed errors where `class` should be enclosed in `"` -> the same can be seen on jQuery page as well: https://api.jquery.com/jQuery/#creating-new-elements

Thanks :)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
@@ -40,7 +40,7 @@ public class Scripts implements Iterable<Scripts.Script> {
 
     private static final String DEBUG_SUFFIX = "-debug";
 
-    enum ScriptType {
+    enum Type {
         MINIFIED, DEBUG
     }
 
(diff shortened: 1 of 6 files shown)


vladak:
Looks okay to me. The https://github.com/oracle/opengrok/wiki/Debugging will need an update once this is merged in.

vladak:
Needs rebase. Also, I would like to hear from @tulinkry.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
```diff
@@ -103,11 +109,15 @@ public String toHtml() {
         SCRIPTS.put("jquery", new FileScript("js/jquery-3.4.1.min.js", 10));
         SCRIPTS.put("jquery-ui", new FileScript("js/jquery-ui-1.12.1-custom.min.js", 11));
         SCRIPTS.put("jquery-tablesorter", new FileScript("js/jquery-tablesorter-2.26.6.min.js", 12));
-        SCRIPTS.put("tablesorter-parsers", new FileScript("js/tablesorter-parsers-0.0.1.js", 13));
+        SCRIPTS.put("tablesorter-parsers", new FileScript("js/tablesorter-parsers-0.0.1-min.js", 13));
```
tulinkry: i've seen the general consent about the nameing is the filename.min.js - I'm talking about `-min.js` suffix.
examples:
- https://code.jquery.com/jquery-3.4.1.min.js
- https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.7/umd/popper.min.js
- https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/js/bootstrap.min.js

tulinkry: applies to the others as well

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
```diff
@@ -38,6 +38,12 @@
  */
 public class Scripts implements Iterable<Scripts.Script> {
 
+    private static final String DEBUG_SUFFIX = "-debug";
+
+    enum ScriptType {
+        MINIFIED, DEBUG
+    }
```
tulinkry: this could be a separate class or the naming could be just Type as when it's nested it's sort of redundant

tulinkry:
Does it work in IDEA out of the box when developing? That was my main concern since I was thinking about this few years ago.

ahornace:
> Does it work in IDEA out of the box when developing? That was my main concern since I was thinking about this few years ago.

What do you mean exactly? I would say it works as expected. If you run the `opengrok-web` from the IDEA then it has minified versions. You need to add `debug=true` for debug versions to be included.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
ahornace: I've wanted to do that but did not find the configuration on the plugin web :D

tulinkry:
> What do you mean exactly? I would say it works as expected. If you run the opengrok-web from the IDEA then it has minified versions. You need to add debug=true for debug versions to be included.

Yeah, that's what I meant. So as a developer I don't need to run any extra commands to get it working.

ahornace:
> Yeah, that's what I meant. So as a developer I don't need to run any extra commands to get it working.

I don't think there is an easy way to do that unless we want to do `debug` builds. I don't mind using query param. It could also be easily solved by using debug cookie.

PR #2894, opened 2019-08-07
Change plugins file structure to follow a standard way

ahornace:
Hi,

`plugins` source files still were not located in `src/main/java`, `src/test/java` directories.

Thanks :)


vladak:
Is this going to have impact on authorization stack configuration ?

ahornace:
> Is this going to have impact on authorization stack configuration ?

I don't think so. Package names stayed the same.

PR #2895, opened 2019-08-08
Add information about mount data folder and configuration folder.

wy193777:
Also fix a little bit typo


vladak:
Thanks !

wy193777:
I have a question, on `docker/index.sh`, when run indexer, both `OPS=${INDEXER_FLAGS:='-H -P -S -G'}` and `$INDEXER_OPT` been passed to indexer. But on `docker/README.md`, only `$INDEXER_OPT` is mentioned. Can `$INDEXER_OPT` actually overwrite `$OPS`?

vladak:
It can. This is sort of a "hidden feature". Either it should be removed or documented.

PR #2896, opened 2019-08-08
Add additional checkstyle checks

ahornace:
Hi,

fixes #2653 

few other checks I find quite useful.

Thanks :)


vladak:
There are still some warnings.

vladak:
Could you summarize the main new style changes ?

vladak:
Perhaps we should have a coding style document..

ahornace:
> Could you summarize the main new style changes ?

- `JavadocStyle` - checks that first sentence ends with `.` and that tags have proper start/end
- `LineLength` - standard is now 120 chars per line but I kept it relaxed due to @tulinkry 's request
- `EmptyBlock` - empty `{}` in `if`, `for`, etc.
- `HideUtilityClassConstructor` - class with only `static` members should have `private` constructor -> there is no need to instantiate it

Other rules are checking for spaces around operators, before/after `(),|` and proper placement of `{}`.

All of these are pretty standard and IDEA will automatically format it in that way.

> Perhaps we should have a coding style document..

No one follows such things. Everyone has their own style and will ignore/forget it unless you have it enforced somehow (e.g. with checkstyle failing builds).

ahornace:
Also, this one would be awesome https://checkstyle.sourceforge.io/config_javadoc.html#InvalidJavadocPosition but there is a lot of classes which break it and I did not want to start another flame war.

vladak:
Actually, some people are quite passionate about coding style guides :-) Maybe `style.xml` could be used as a sort of documentation except it does not give any rationale.

Anyhow, this looks good to me. @tarzanek, @tulinkry, @idodeclare ?

idodeclare:
Looks good to me — but there was some content deleted which I would want preserved, either as tweaked Javadoc or moved to a regular comment. E.g. Chandan’s “Created on...” comment and the unescaped regex text (since raw strings or text blocks are future ware in Java). I’ll add PR comments.

ahornace:
I would argue that `Created on...` is completely redundant. It does not tell me anything useful. I don't care if the class was created in 2005 or 2006. If I need that sort of information, I can always use annotate.

ahornace:
`unescaped regex text` -> I don't think this should be part of the javadoc as well. You can easily use some tool to escape/unescape it for you.

tarzanek:
this is a hard debate and there are no correct answers
sacrifice readability and duplicate stuff for easy access? (unescaped regexps?)

that said I am OK with this push(and I think the deleted stuff won't be missed)
+1 from my side

tarzanek:
I won't merge until @idodeclare is OK though, since I checked all deleted text, but if Chris will find something really worthy getting back (for me neither Created on or regexp are worthy, I like dedup) we can argue about it

idodeclare:
Oops, Lubos, you set up a deadlock. —No jk I'm good.

@ahornace why did `CustomSloppy` and others have to get all the new suppressions? For these new rules, shouldn't they be very trivially revised?

ahornace:
> @ahornace why did CustomSloppy and others have to get all the new suppressions? For these new rules, shouldn't they be very trivially revised?

`CustomSloppyPhraseScorer`, `CustomPhraseQuery`, `PhrasePositions` -> these classes are virtually copied from Apache Lucene with small modifications. I wanted to keep them the way they are for easy comparison with newer versions from Lucene.

`JFlexNonXref` and other in `EmptyBlock` are there due to `while (yylex()...) {}` statements

`MellonHeaderDecoder` -> I've encountered some issue where CheckStyle was complaining about `</NameIDFormat>` in `{@code <NameIDFormat>urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress</NameIDFormat>}` -> probably a bug in CheckStyle

vladak:
> I would argue that `Created on...` is completely redundant. It does not tell me anything useful. I don't care if the class was created in 2005 or 2006. If I need that sort of information, I can always use annotate.

+1 if anyone wants to have a glimpse at the history, it's in the SCM logs.

vladak:
As for the unescaped regexps, I feel that by leaving them in there is a risk of divergence. So +1 for removing them from JavaDoc comments.

All in all, I think these concerns are pretty minor. If I expected anything, it would be some sort of debate over the actual selection of stylistic properties. Since there was no such thing, I'd just merge this in.

vladak:
No objections for the past day, merging.

PR #2901, opened 2019-08-19
make sub-repository detection more flexible

vladak:
A bit hacky solution to allow sub-repositories of different types. Follow-up to #2817.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -93,8 +93,10 @@ public class GitRepository extends Repository {
      */
     private static final Version MINIMUM_VERSION = new Version(2, 1, 2);
 
+    protected static final String repoType = "git";
+
     public GitRepository() {
-        type = "git";
+        type = repoType;
         /*
          * This should match the 'iso-strict' format used by
          * {@code getHistoryLogExecutor}.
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -93,8 +93,10 @@
      */
     private static final Version MINIMUM_VERSION = new Version(2, 1, 2);
 
+    protected static final String repoType = "git";
```
ahornace: I'd follow constant naming convention: `REPO_TYPE`

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -402,62 +402,73 @@ public boolean hasAnnotation(File file) {
                 continue;
             }
 
-            String path;
-            try {
-                path = file.getCanonicalPath();
+            if (types == null) {
```
ahornace: I'd personally refactor this a bit (I don't like `add(null)` part), for example:
```
private List<Repository> getRepositories(final File file, final List<String> types) {
    if (types == null || types.isEmpty()) {
        return Collections.singletonList(RepositoryFactory.getRepository(file, null));
    }
    return types.map(t -> RepositoryFactory.getRepository(file, t)).collect(Collectors.toList()).
}
```
And then I would iterate over the repositories.

vladak:
closing in favour of #2912

PR #2906, opened 2019-08-27
Fix some lgtm warnings

ahornace:
Hi,

fixed some simple stuff reported by lgtm: https://lgtm.com/projects/g/oracle/opengrok

Thanks :)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/executables/ELFAnalyzer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/executables/ELFAnalyzer.java
@@ -30,7 +30,6 @@ import java.nio.ByteOrder;
 import java.nio.MappedByteBuffer;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -120,7 +119,6 @@ public class ELFAnalyzer extends FileAnalyzer {
             return null;
         }
 
-        HashMap<String, Integer> sectionMap = new HashMap<String, Integer>();
         ELFSection[] sections = new ELFSection[eh.e_shnum];
         int[] readables = new int[eh.e_shnum];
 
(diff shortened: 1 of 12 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Statistics.java
```diff
@@ -201,55 +201,55 @@ public synchronized void addRequestTime(String category, long v) {
         return timingAvg;
     }
 
-    public synchronized void setRequestCategories(Map<String, Long> requestCategories) {
```
vladak: the complaint in lgtm says:
```json
        {                                                                       
          "ruleId": "com.lgtm/java-queries:java/unsynchronized-getter",         
          "ruleIndex": 2,                                                       
          "message": {                                                          
            "text": "This get method is unsynchronized, but the corresponding [set method](1) is synchronized."
          },                                                                    
          "locations": [                                                        
            {                                                                   
              "physicalLocation": {                                             
                "artifactLocation": {                                           
                  "uri": "opengrok-indexer/src/main/java/org/opengrok/indexer/web/Statistics.java",
                  "uriBaseId": "%SRCROOT%",                                     
                  "index": 4                                                    
                },                                                              
                "region": {                                                     
                  "startLine": 151,                                             
                  "startColumn": 30,                                            
                  "endLine": 151,                                               
                  "endColumn": 50,                                              
                  "charOffset": 5186,                                           
                  "charLength": 20                                              
                }                                                               
              }                                                                 
            }                                                                   
          ],                              
```
Could this have unintended side effects ?

vladak:
I went through all the defects by using exported list from lgtm.com (JSON) and besides the single one all the changes look good.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Statistics.java
ahornace: I've thought about it as well. But then decided to follow the advice. The reasons to do so:
- using `synchronized` in method definition is considered bad because you are exposing the lock outside - you should always have an inner lock, e.g.:
```
private final Object o = new Object();
...
public void m() {
    synchronized (o) {
        ...
    }
}
```
- the attributes are not `volatile` which means that the value can actually be cached and not propagated properly
- it seems that these methods are used for serialization/deserialization and are called only from tests

vladak:
I meant that the synchro was there for a reason so I am not comfortable
removing it for good.

Dne st 28. 8. 2019 8:42 uživatel Adam Hornáček <notifications@github.com>
napsal:

ahornace:
> I meant that the synchro was there for a reason so I am not comfortable removing it for good.

Can you please elaborate that reason?

vladak:
see #1563

ahornace:
> see #1563

This is still not sufficient for me. But I've started integrating https://metrics.dropwizard.io/4.0.0/getting-started.html which should replace `Statistics` completely so I'll abide by your request and add it back.

Issue #2885, opened 2019-07-31
How to add procedure(plsql) definition type to navigate window?


elkin5:
I need to get the functions and procedure of a plsql file and then take them to an api rest that allows to be obtained

vladak:
Not sure I understand the question fully - could you rephrase it as a list the steps perhaps ?

elkin5:
The navigate window show this (variable, package, function):

![Capture](https://user-images.githubusercontent.com/4118198/62370482-dc73a900-b4f7-11e9-81f6-43e1cdc286bd.PNG)

i need add procedure to navigate window.

otherside, is it posible get this elements from the navigate window?

vladak:
Okay, these are 2 different questions. Better handle one question in single Issue only.

The problem seems to be that `PROCEDURE` entries are not recognized as definitions. This is done by Universal ctags - just make you are using recent version. According to #2293 this should work in recent OpenGrok releases.

As for retrieving all definitions in a document(s) via RESTful API, I don't think it works currently - tried searching for a wildcard returns list of files however the definitions are represented as a single `...` entry for each file. Please file a separate issue to track this.

vladak:
Also, it depends on the suffix of the filename: https://github.com/oracle/opengrok/blob/a391bead371079d671483425934a74aad8ec4344/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/sql/PLSQLAnalyzerFactory.java#L33-L40

vladak:
Anyow, some of the definitions are recognized so this is perhaps just a matter of adding more. Trying locally, I cannot even get ctags to recognize definitions of functions. Can you reduce the file to generic minimal contents ?

elkin5:
The suffix (pkg) was added in the .java files.
I work on the latest version of opengrok.
The code viewer recognizes the procedure as definition but the navigation window does not show these definitions (procedure), how could I do it?

vladak:
How do you know that the procedure is recognized as a definition ? If you do a search for definition, is it found ?

vladak:
i.e. what are the contents of the `tags` file if you run `ctags --langmap=sql:+.pkg` on the file in question ? (assuming the file indeed has the `.pkg` suffix and `ctags` is the same ctags program as is used for running the Indexer)

vladak:
[...]

Issue #2856, opened 2019-07-04
SVN History does no work


elkin5:
Istallation: Docker

Source: /opengrok/src
Svn Project: /opengrok/src/pro
Svn Checkout: svn checkout svn://serverSrc.project.com.co/project pro/

it is indexed but the svn history does no work.

vladak:
What does the indexer log say ?

elkin5:
what is the path of the indexer log?. Docker Installation

vladak:
see https://github.com/oracle/opengrok/blob/master/docker/README.md#indexer-logs

elkin5:
This is the most relevant information:

org.opengrok.indexer.history.SubversionRepository.getInfoDocument Failed to execute svn info for [/opengrok/src/SAO]. Repository disabled.
04-Jul-2019 19:56:33.544 SEVERE [invalidate-repos-1441] org.opengrok.indexer.util.Executor.exec Failed to read from process: /usr/bin/svn
        java.io.IOException: An error occurred while parsing the xml output
                at org.opengrok.indexer.history.SubversionHistoryParser.processStream(SubversionHistoryParser.java:194)
                at org.opengrok.indexer.util.Executor.exec(Executor.java:222)
                at org.opengrok.indexer.history.SubversionHistoryParser.parse(SubversionHistoryParser.java:165)
                at org.opengrok.indexer.history.SubversionRepository.getHistory(SubversionRepository.java:292)
                at org.opengrok.indexer.history.SubversionRepository.determineCurrentVersion(SubversionRepository.java:428)
                at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:171)
                at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:210)
                at org.opengrok.indexer.history.HistoryGuru$3.run(HistoryGuru.java:953)
                at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
                at java.util.concurrent.FutureTask.run(FutureTask.java:266)
                at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
                at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
                at java.lang.Thread.run(Thread.java:748)
        Caused by: org.xml.sax.SAXParseException; lineNumber: 3; columnNumber: 1; XML document structures must start and end within the same entity.
[...]

Issue #2833, opened 2019-06-25
Support for .PKG extension files


OscarGelvez:
What's up friend?
I would like to know if the .pkg extension files are supported? Or if it is possible to add support for this type of files.
I need it for a project that I'm doing where these types of files are worked.
These files have the following structure:

/ ********* Specification ************* /
Create or replace namePackage
Procedure1 ();
Procedure2 ();
Function ();
Procedure3 ();
END namePackage;

/ ********* Body ************* /
Create or replace namePackage
// Global Attributes
// Global variables

Procedure1 () Begin / * code * / End Procedure1;
Procedure1 () Begin / * code * / End Procedure2;
Function1 () Begin / * code * / End Function1;
Procedure4 () Begin / * code * / End Procedure3;

END namePackage;

The methods or functions that are in the block specification can be accessed publicly.

I appreciate your collaboration
regards

vladak:
What language is the .pkg extension tied with ?

OscarGelvez:
Hi friend
In PL / SQL.

OscarGelvez:
I modified the opengrok-indexer / src / main / java / org / opengrok / indexer / analysis / Ctags.java class to support the extension of .PKG file and also modify the opengrok-indexer / src / main / java / org / opengrok / indexer / analysis / sql / plsql / PLSQLAnalyzerFactory class by adding the PKG suffix.

I would like to know how you could modify your own behavior when it comes to a file with this extension.

I need the syntactic analyzer to recognize that there can be 2 times the declaration of a method due to the structure of the files that I exposed previously.

idodeclare:
@OscarGelvez , release 1.3.4 support dynamically reconfiguring `Ctags` operation when `-A,--analyzer` is used to map an extension. (Your private build statically maps SQL to .PKG already of course, but I was just informing.)

As to recognizing twice-declared methods, that would be a requirement of Universal Ctags. Please could you get the latest version of u-ctags, and run it on your file (making sure to map SQL to .PKG on the command-line) to see what tags it finds from your test file.

vladak:
Closing since .pkg is too generic and a way how to configure indexer to provide the mapping exists.

PR #2908, opened 2019-08-27
Solve issues

elkin5:
- #2833 
- #2856
- #2885


vladak:
There are build failures, you need to solve them first.

elkin5:
I do not know what is the problem. 
I have not compiled yet the project with unit tests

vladak:
The build log in Wercker says:
```
[ERROR] Failures: 
[ERROR]   PascalXrefTest.sampleTest:40->XrefTestBase.writeAndCompare:62->XrefTestBase.writeAndCompare:90 org.opengrok.indexer.analysis.pascal.PascalAnalyzerFactory xref--should have no diffs
```

vladak:
My guess is that one of the changes, likely the addition of `procedure` changed the output of xref for Pascal (which seems credible given that `src/test/resources/analysis/pascal/sample.pas` used in `PascalXrefTest` contains some procedure declarations). This might not be bad, just the test needs to be fixed. If you want to be sure this was the change that caused the test failure, just stash away the changes done to `XrefStyle.java` and re-run the test(s).

vladak:
Otherwise, I have no problem with the rest of the changes.

From formal point of view, it would be nice if relevant changeset commit messages said `fixes #XYZ` to close the related issues automatically on merge.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/sql/PLSQLAnalyzerFactory.java
```diff
@@ -36,7 +36,8 @@
         "PLD",
         "PLB",
         "PKB",
-        "PCK"
+        "PCK",
+        "PKG"
```
tarzanek: PKG is too generic (and mostly used for packages, not for sql) ... I wouldn't do this, from my side this can go to your own config per repo, but not for all users.
So please remove this, most users would be only bugged by this.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -148,6 +148,7 @@ private void initialize() throws IOException {
         command.add("--langmap=sql:+.pks"); // RFE #19208 ?
         command.add("--langmap=sql:+.pkb"); // # 1763
         command.add("--langmap=sql:+.pck"); // # 1763
+        command.add("--langmap=sql:+.pkg"); // # 2833
```
tarzanek: same as above, pkg extension is too generic, not valid for most users

tarzanek:
Hi @elkin5 
the change looks good, however I think .pkg taking to SQL is too big and 90% of the users will be a bit surprised by this.
You can add this extension manually at your site, but we shouldn't force all users to this mapping.
And there is basically only one hiccup there - the ctags support for mapped extension to an analyser - so that would be space for improvement (supporting that when you add extension, it will also be added to ctags).

vladak:
I pulled the changes for #2856 in, for the #2885 changes please submit new PR once you deal with the Pascal test changes. Thanks.

PR #2911, opened 2019-08-29
Metrics support

ahornace:
Hi,

this change introduces https://github.com/dropwizard/metrics

The only thing that is missing is metrics persistence. However, Dropwizard metrics have support for many tools (e.g. Graphite, Prometheus, ...) that we should leverage for this purpose. I think we should have a discussion about which tool to use first before I delve into the implementation. I've tested Graphite and it worked seamlessly.

Also, I did not implement any health checks. It's a plan to do so in the future.

Last thing, `AdminServlet` exposes thread dumps and some processor information that I find unnecessary. I think we should remove them.  What do you guys think? 

Note: `HealthCheckServletContextListener` had to be added for `AdminServlet` to work properly.

Thanks :)


vladak:
As I mentioned over the coffee yesterday, I am mostly worried about the lack of persistent statistics. Maybe that's not a big problem, certainly depends on whether this will make the operations significantly more difficult. I'd like to try statsd + Graphana.

tarzanek:
w00t, prometheus + grafana are nice (we use them too)
let me review this

tarzanek:
lgtm, could @tulinkry review too?

vladak:
There are conflicts that need to be resolved.

Also, I would like to see some how-to on how to use this in practice.

ahornace:
I've wanted to look at prometheus but did not have much time lately :( 

I will try this week :)

vladak:
Also, any documentation (Apiary, wikis) that needs to be changed as a result ?

tarzanek:
I guess lots of stuff will need to be documented, this would be a start
setting up prometheus and grafana is easy, check what we do in https://github.com/scylladb/scylla-monitoring
(then basically prometheus downloads over http metrics from target, so it's like pull model, where opengrok would expose the metrics)
also prometheus exporter support is easy to add to those metrics - so a config variable should be exposed to enable such exporter

tulinkry:
I was thinking the same (to introduce these metrics here)!

> could @tulinkry review too?
I will.

opengrok-indexer/src/test/java/org/opengrok/indexer/authorization/AuthorizationFrameworkReloadTest.java
```diff
@@ -153,7 +146,7 @@ public void run() {
         }
 
         // Double check that at least one reload() was done.
-        reloads = stats.getRequest("authorization_stack_reload") - reloads;
+        long reloads = Metrics.getInstance().counter("authorization_stack_reload").getCount();
```
tulinkry: doesn't it interfere with other tests? the previous version expected a change in reloads

opengrok-web/pom.xml
```diff
@@ -148,6 +148,16 @@ Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
...
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-jersey2</artifactId>
+            <version>4.1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-servlets</artifactId>
+            <version>4.1.0</version>
```
tulinkry: a candidate to property

opengrok-web/src/main/java/org/opengrok/web/AuthorizationFilter.java
```diff
@@ -62,38 +69,38 @@ public void doFilter(ServletRequest sr, ServletResponse sr1, FilterChain fc) thr
...
+                }
+
+                httpRes.sendError(HttpServletResponse.SC_FORBIDDEN, "Access forbidden");
                 return;
             }
-
-            httpRes.sendError(HttpServletResponse.SC_FORBIDDEN, "Access forbidden");
-            return;
+        } finally {
+            timerCtx.stop();
```
tulinkry: this records even non-forbidden requests, doesn't it?

tulinkry: btw you can use the timer in try-with-resources (not here perhaps)

opengrok-web/src/main/java/org/opengrok/web/HealthCheckServletContextListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.web;
+
+import com.codahale.metrics.health.HealthCheckRegistry;
+import com.codahale.metrics.servlets.HealthCheckServlet;
+
+public class HealthCheckServletContextListener extends HealthCheckServlet.ContextListener {
```
tulinkry: is this needed?

opengrok-web/src/main/java/org/opengrok/web/api/v1/RestApp.java
```diff
@@ -34,6 +36,7 @@
     
     public RestApp() {
         register(new SuggesterAppBinder());
+        register(new InstrumentedResourceMethodApplicationListener(Metrics.getInstance()));
```
tulinkry: in short, what does it do? `InstrumentedResourceMethodApplicationListener`

opengrok-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -233,6 +239,14 @@
         <url-pattern>/eforbidden</url-pattern>
         <!-- FORBIDDEN -->
     </servlet-mapping>
+    <servlet>
+        <servlet-name>metrics</servlet-name>
+        <servlet-class>com.codahale.metrics.servlets.AdminServlet</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>metrics</servlet-name>
+        <url-pattern>/metrics/*</url-pattern>
```
tulinkry: can this be part of the rest api?
also authorization applies or not?

tulinkry:
Did you consider adding some configuration options to configure, for example, graphite reporter?

opengrok-web/src/main/webapp/WEB-INF/web.xml
vladak: what about the Statistics endpoint documentation in Apiary ? This will probably need to be updated.

vladak:
rebase ?

opengrok-indexer/pom.xml
```diff
@@ -194,6 +194,11 @@ Portions Copyright (c) 2017-2019, Chris Fraire <cfraire@me.com>.
             <artifactId>jackson-annotations</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-core</artifactId>
```
vladak: On https://dropwizard.io/en/stable/manual/core.html I see that `metrics-core` package depends on whole lot of other packages. Some of them are already delivered. How does it work e.g. with Jersey/Jackson version ?

ahornace: Good catch. Maven resolves this somehow, I believe by including the newest version. However, I can specify that these dependencies are already provided and the ones from Dropwizard should be ignored.

vladak: As long as both work.. :-)

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SuggesterController.java
```diff
@@ -98,6 +99,7 @@
     @GET
     @Authorized
     @CorsEnable
+    @Timed
```
vladak: what does this do ?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/StatsController.java
```diff
@@ -1,59 +0,0 @@
...
-import javax.ws.rs.GET;
-import javax.ws.rs.PUT;
-import javax.ws.rs.Path;
-import javax.ws.rs.Produces;
-import javax.ws.rs.core.MediaType;
-import java.io.IOException;
-
-import static org.opengrok.indexer.util.StatisticsUtils.loadStatistics;
-
-@Path("/stats")
```
vladak: The Apiary document file needs to be updated it seems.

opengrok-web/src/main/webapp/WEB-INF/web.xml
vladak: As for the authentication/authorization this is definitely a concern since the metrics included names of projects along with their success/failure authorization counts. Surely we do not want to leak project names without authorization.

vladak: According to https://metrics.dropwizard.io/3.1.0/manual/servlets/ `AdminServlet` basically aggregates bunch of servlets, including the `MetricServlet` that exposes the `/metrics` endpoint which presents the metricts. Maybe put this behind `LocalhostFilter` ?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SuggesterController.java
```diff
@@ -98,6 +99,7 @@
...
     @Produces(MediaType.APPLICATION_JSON)
     public Result getSuggestions(@Valid @BeanParam final SuggesterQueryData data) throws ParseException {
```
vladak: This will create metric named `org.opengrok.web.api.v1.controller.SuggesterController.getSuggestions` - not sure we want to expose the full class names in metrics.

vladak:
@ahornace given your current time constraints I will probably merge as is and then will do some adjustments myself before next release.

opengrok-web/src/main/webapp/WEB-INF/web.xml
vladak: Actually, you can leave out the project names from the meters altogether since it is not good practice anyway - it would cause "cardinality explosion" (https://youtu.be/deBSrmskpcY?t=1728)

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
+import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
+import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
+import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
+import io.micrometer.core.instrument.binder.system.ProcessorMetrics;
+import io.micrometer.prometheus.PrometheusConfig;
+import io.micrometer.prometheus.PrometheusMeterRegistry;
+
+public final class Metrics {
```
vladak, line 33: This is currently specific to the webapp so it should probably live in the `org.opengrok.web` package. I'd like to gather indexer statistics in the future however it will probably use different registry - I think statsD is more suitable for indexer since it allows sending the data rather than waiting for it to be scraped.

opengrok-web/src/main/java/org/opengrok/web/api/v1/RestApp.java
vladak: After the changes for Micrometer this file remains unchanged (remove the copyright change).

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SuggesterController.java
vladak: Not relevant anymore (remove the copyright change).

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/StatsController.java
vladak: This still holds after the Micrometer changes.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SuggesterController.java
vladak: No longer relevant.

opengrok-web/src/main/webapp/WEB-INF/web.xml
vladak: Once the project names are out, the endpoint does not have to be covered by authentication/authorization I think. Otherwise #3222 would help.

opengrok-web/src/main/java/org/opengrok/web/servlet/MetricsServlet.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+@WebServlet("/metrics/prometheus")
+public class MetricsServlet extends HttpServlet {
```
vladak, line 35: IDEA complains that this serializable class is lacking `serialVersionUID`

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
ahornace: It's also used in `AuthorizationFramework` which is a part of indexer so moving it under `org.opengrok.web` won't work.

vladak: You're right. The `AuthorizationFramework` is also part of `RuntimeEnvironment` which makes it even harder to disentangle. Let's leave it like that for now. With the addition of other registries it will have to be made more generic.

vladak:
Thanks ! Will fix the remaining issue(s).

vladak:
created https://github.com/oracle/opengrok/wiki/Monitoring

Issue #2330, opened 2018-08-31
HistoryGuru#updateRepositories() is dead code


vladak:
The code for updating (mirroring) repository contents in `HistoryGuru` (and `update()` methods in various repositories) has never been used AFAIK - it is only called from `HistoryGuruTest`. This duplicates the work currently done by `mirror.py`.

It should be decided how mirroring will be done and either enhance the update code (it lacks e.g. support for proxies, proper logging or hooks) and make it available via RESTful API or rip it out and live with `mirror.py`.

vladak:
Given that in the mean time a fairly elaborate set of scripts was introduced that facilitates the work done by update, let's remove it.

PR #2912, opened 2019-08-31
Fixes #2900 : Add Repository isNestable() property

### LINKED ISSUE #2900: nested git repositories not found
I'm using the latest OpenGrok 1.3.1 release, and I have an issue with the repository detection for the history creation.

Our project is using repo for the file versioning, allowing to handle multiple Git repositories. The folder structure looks like:

```
baseDirectory/.repo
baseDirectory/anotherDirectory
baseDirectory/anotherDirectory/.git
```
The grok source dir has a symbolic link to the baseDirectory.

The problem is that the git repositories are not detected and therefore history is not working. Whereas it is working if I create a symlink in the grok source dir directly to the git repositories. But not as a subdir. The strange thing is that we used opengrok version 1.0 before and we had no problems with that. I've already played with parameter --depth but it didn't have any effect.
Indexing itself is working fine, only the history is missing.

idodeclare:
@vladak, this PR is a different approach from your #2901 to fixing #2900 where instead of modifying a repository to know of others' type, a property, `isNestable()` (default `false`), indicates if a `Repository` subclass can be contained within other repositories. I think this is simpler to achieve your goal in #2817 of "efficiency (esp. those repository types which do not just look for special files/directories but execute a command to detect a repository)."

Only Git and Mercurial are marked as `isNestable() == true`. (N.b. only Git, Mercurial, and Repo are marked as `supportsSubRepositories() == true`.)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2005, 2019, Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017-2018, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2017-2019, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.history;
 
@@ -387,13 +387,15 @@ public final class HistoryGuru {
      * @param files list of files to check if they contain a repository
      * @param ignoredNames what files to ignore
      * @param recursiveSearch whether to use recursive search
-     * @param type type of the repository to search for or {@code null}
      * @param depth current depth - using global scanningDepth - one can limit
      * this to improve scanning performance
+     * @param isNested a value indicating if a parent {@link Repository} was
+     * already found above the {@code files}
      * @return collection of added repositories
      */
     private Collection<RepositoryInfo> addRepositories(File[] files,
-            IgnoredNames ignoredNames, boolean recursiveSearch, String type, int depth) {
+            IgnoredNames ignoredNames, boolean recursiveSearch, int depth,
+            boolean isNested) {
 
         List<RepositoryInfo> repoList = new ArrayList<>();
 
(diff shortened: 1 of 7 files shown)


vladak:
It this going to work for the `RepoRepository` nesting a `GitRepository` ? I wanted to add a positive test case to cover that (one of the reasons my original PR was not progressing) to make sure the original issue is indeed fixed.

Also, it seems that the cvs checkout is failing in Travis for some reason.

vladak:
Actually, the Mercurial/Git test should still pass - we should not allow nesting of repositories of different types I think even though individually they can be nested.

idodeclare:
> we should not allow nesting of repositories of different types

Well going back to your original motivation for #2817, you wrote that it was "for efficiency (esp. those repository types which do not just look for special files/directories but execute a command to detect a repository)." But as we know that broke `RepoRepository`.

I would argue that "we should not allow nesting of repositories of different types" is a solution in search of a problem — to which you have to create a hacky loophole in order to avoid the bug of #2900. I'm just proposing we support your original desire for efficiency in a more straight-forward fashion by having `Repository` subclasses describe whether they are nestable. Only those which have a cheap and high-confidence check (e.g. `.git` or `.hg`) are thus described as nestable in this patch.

Yes that means if someone checked out Mercurial within Git then there would be two repositories — which OpenGrok supports with no problem of course.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -540,39 +536,6 @@ public void updateRepositories() {
         }
     }
 
-    /**
-     * Update the source contents in given repositories.
-     *
-     * @param paths A list of files/directories to update
-     */
-    public void updateRepositories(Collection<String> paths) {
```
vladak: how is this removal related to the repository scan ?

idodeclare: While working in `HistoryGuru`, IDEA pointed out unused methods, and as we don't reflect into that class, it seemed a valid analysis — so I accepted to delete the methods.

vladak: Yep, definitely dead code. Better do this in separate changeset. There might be an issue tracking the removal but I cannot find it at the moment. Also, the individual repositories need to be cleaned up, e.g. `MercurialRepository#update()` etc.

idodeclare: OK undone

vladak: this is tracked by #2330

vladak:
Yes, that is reasonable. I agree that the original requirement to nest only repositories of the same type is artificial.

PR #2918, opened 2019-09-05
Unittest + parallelization

vladak:


tulinkry:
A downside of this approach is that `-s` has no effect when running the test with xdist, so it makes debugging the tests with occasional "prints" harder.

PR #2920, opened 2019-09-06
Bugfix/symlink laxness

idodeclare:
Hello,

Please consider for integration this patch to fix and enhance symlink handling:

* `acceptSymlink()` properly checks previously-seen symlinks ~~(`acceptedNonlocalSymlinks`)~~.
* `HistoryGuru` no longer continues descending into a path after a `ForbiddenSymlinkException`.
* add ~~-C,~~`--canonicalRoot` to allow whitelisting a canonical target, implicitly allowing all symlinks that target a canonical object below the ~~-C,~~`--canonicalRoot` without having to specify `-N,--symlink` for all the links.
* restore historical `-N` option A.K.A. for `--symlink`.
* `IndexDatabase` stores symlinks data for use by `list.jsp` to adjust when necessary a query for #Lines and LOC. (`IndexAnalysisSettings` is automatically upgraded to version 3, but the index afterward is not compatible with previous releases.)

Thank you.


idodeclare:
Yay the build fixed. I didn't [change anything](https://github.com/oracle/opengrok/compare/dae9ef7..2950b31); just pushed to re-start the build to see if the macOS failure was transient and also to reword to fix an issue#.

Z4rc:
This is a very useful feature but for me "#Lines" and "LOC" are not working anymore in the webapp.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexAnalysisSettings3.java
```diff
@@ -0,0 +1,270 @@
+/*
```
tarzanek: I want scala or kotlin :-)

tarzanek:
I had a quick look and it looks OK

idodeclare:
Thank you, @tarzanek

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexAnalysisSettings3.java
```diff
@@ -0,0 +1,270 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Represents a serializable gathering of some top-level metadata concerning the
+ * operation of {@link IndexDatabase} -- and persisted therein too -- which are
+ * re-compared upon each indexing run since changes to them might require
+ * re-indexing particular files or in certain cases all files.
+ */
+public final class IndexAnalysisSettings3 implements Serializable {
```
tulinkry: why 3?

idodeclare: First there was `IndexAnalysisSettings` with `INDEX_ANALYSIS_SETTINGS_OBJVER` set to 1. Then `IndexAnalysisSettings2` with version set to 2. Then someone renamed `IndexAnalysisSettings2` to `IndexAnalysisSettings`. Now this patch has `IndexAnalysisSettings3` with version set to 3.

idodeclare:
Rebased (no conflicts) to pick up 14a8a101695 so Wercker build does not die.

idodeclare:
Rebased for a pom.xml conflict

Issue #2852, opened 2019-07-04
use pygit2 in the scm package


vladak:
PR #2851 added pygit2 for testing. It might be well used in `opengrok_tools.scm.git` as well.

tulinkry:
I wonder why you chose pygit2 (which even requires some brew install as I saw in your PR) over others:

https://github.com/gitpython-developers/GitPython
https://pypi.org/project/python-git/

vladak:
No particular reason besides pretty clean API - at least for what I needed.

tulinkry:
As I understood the pygit2 uses native api calls and not "exec" style like the others. libgit2 is installed on unix systems? (as you only added it to the osx builds)

tulinkry:
Cause suddenly it does not work out of the box on osx. I have to install libgit2 manually and it is not mentioned anywhere that we have this hidden dependency for development.

tulinkry:
when we used the other packages (with the "exec" style) I'd say you dont need anything than python

vladak:
Good point, I should have added a note to the Readme as I experienced the
same.

Dne po 8. 7. 2019 10:35 uživatel Kryštof Tulinger <notifications@github.com>
napsal:

tulinkry:
Is speed of the calls the concern here? Cause I'd prefer not to install any external dependency - as you plan to use it not only for testing but for the git repo.

vladak:
No, pygit2 looked attractive because of the clarity of the API. You just go `init_repository()`, `clone_repository()` and boom, that's it. While with executing the porcelain commands, one has to care about return codes, error string matching etc.

You are probably right that with this non-implicit dependency this is awkward to use. I will keep this as is for now and document it in the development section of the README.

vladak:
But still, the dependency should be investigates in more detail - how come this works out of the box on (some) Linux distros but not macOS. I would have thought that `pip install` should handle this.

vladak:
I just got bitten by this again, trying to run `python setup.py test` on fresh Solaris install. In hindsight, choosing pygit2 was wrong. We need something that does not require compilation.

vladak:
https://pypi.org/project/python-git/ has very limited API, it lacks clone/init, seems to be not very mature in general.

vladak:
The downside of using Git Python module is that command timeouts would have to implemented differently. Now it is provided by the `Command` class automatically (modulo bugs like #3022).

vladak:
GitPython should be excluded as well because the README says:
[...]

PR #2922, opened 2019-09-10
use GitPython instead of pygit2

vladak:
The reasons are explained in #2852.


vladak:
GitPython is mature, it is frequently updated, has extensive API documentation (http://gitpython.readthedocs.org), has exemplary test coverage, Docker image, builds in various CI/CD environments etc. and especially does not require pulling pre-requisite binaries outside of the `setup.py`. Easy choice.

PR #2934, opened 2019-09-23
revise auth framework locking in RuntimeEnvironment

vladak:
Tested with `RuntimeEnvironment:setConfiguration()` modified to sleep for 5 minutes. First, the method was triggered:
```
curl -X PUT -H 'Content-Type: application/xml' -d @/var/opengrok/etc/configuration.xml 'http://localhost:8080/source/api/v1/configuration?reindex=true'
```
and then I did bunch of requests (`curl  http://localhost:8080/source/`) to make sure they are not blocked anymore.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -44,7 +44,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -1595,7 +1595,7 @@ public final class RuntimeEnvironment {
      *
      * @return the framework
      */
-    public synchronized AuthorizationFramework getAuthorizationFramework() {
+    public AuthorizationFramework getAuthorizationFramework() {
         try {
             authFrameworkLock.readLock().lock();
             if (authFramework == null) {
@@ -1613,7 +1613,7 @@ public final class RuntimeEnvironment {
      *
      * @param fw the new framework
      */
-    public synchronized void setAuthorizationFramework(AuthorizationFramework fw) {
+    public void setAuthorizationFramework(AuthorizationFramework fw) {
         try {
             authFrameworkLock.writeLock().lock();
             if (this.authFramework != null) {
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -132,7 +132,7 @@ public final class RuntimeEnvironment {
         lzSearchExecutor = LazilyInstantiate.using(() -> newSearchExecutor());
     }
 
-    /** Instance of authorization framework and its lock.*/
+    // Instance of authorization framework and its lock.
     private AuthorizationFramework authFramework;
     private final ReentrantReadWriteLock authFrameworkLock;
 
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -125,14 +125,16 @@
...
         configLock = new ReentrantReadWriteLock();
+        authFrameworkLock = new ReentrantReadWriteLock();
         watchDog = new WatchDogService();
         lzIndexerParallelizer = LazilyInstantiate.using(() ->
                 new IndexerParallelizer(this));
         lzSearchExecutor = LazilyInstantiate.using(() -> newSearchExecutor());
     }
 
-    /** Instance of authorization framework.*/
+    /** Instance of authorization framework and its lock.*/
```
ahornace:
```suggestion
    // Instance of authorization framework and its lock.
```
If you want to have `and its lock` info here, then I would advise to remove `/**` so javadoc is not confusing.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -125,14 +125,16 @@
...
     private AuthorizationFramework authFramework;
+    private final ReentrantReadWriteLock authFrameworkLock;
```
ahornace:
```suggestion
    private final ReadWriteLock authFrameworkLock;
```
Program to an interface paradigm.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1593,11 +1595,16 @@ public void setStatisticsFilePath(String path) {
...
      * @return the framework
      */
-    public synchronized AuthorizationFramework getAuthorizationFramework() {
-        if (authFramework == null) {
-            authFramework = new AuthorizationFramework(getPluginDirectory(), getPluginStack());
+    public AuthorizationFramework getAuthorizationFramework() {
+        try {
+            authFrameworkLock.readLock().lock();
+            if (authFramework == null) {
+                authFramework = new AuthorizationFramework(getPluginDirectory(), getPluginStack());
```
ahornace: Theoretically, this can be accessed by multiple threads. Is it okay? If we truly want a singleton, then maybe we should revise this.

vladak: true, we do not want multiple instances of `AuthorizationFramework` hanging around (for efficiency).

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -125,14 +126,16 @@
     private RuntimeEnvironment() {
         configuration = new Configuration();
         configLock = new ReentrantReadWriteLock();
+        authFrameworkLock = new ReentrantLock();
```
ktulinger: i think with this you can just use `new Object()` and synchronize on that object

tulinkry: i think with this you can just use new Object() and synchronize on that object

vladak: well, yes. The advantage is that the `synchronized` block looks nicer however using plain `Object` seems strange to me. Is this a fairly common pattern ?

vladak: Also, could there be some unintended side effects ? As I understand it the `synchronized` will use in-object lock. Is there some other operation that could use the same lock ?

tulinkry: you mean like in
https://github.com/oracle/opengrok/blob/ff44f24ac10337e272510634599e287cb25d0e9c/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java#L64 or
https://github.com/oracle/opengrok/blob/ff44f24ac10337e272510634599e287cb25d0e9c/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java#L71
?

object is the simplest reference and that's all you need to synchronize

tulinkry: there's no side effect, it is the same like using `this` but instead you use custom object

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -125,14 +125,16 @@
...
         watchDog = new WatchDogService();
         lzIndexerParallelizer = LazilyInstantiate.using(() ->
                 new IndexerParallelizer(this));
         lzSearchExecutor = LazilyInstantiate.using(() -> newSearchExecutor());
     }
 
-    /** Instance of authorization framework.*/
+    // Instance of authorization framework and its lock.
     private AuthorizationFramework authFramework;
+    private final Object authFrameworkLock;
```
tulinkry: you can initialize it inline here

PR #2940, opened 2019-09-27
--doprint fixes

vladak:
This contains bunch of changes:
- chage `--doprint` to be a toggle
- restore logging to the console by default in `opengrok-indexer`
- change the `doprint` logging to use a logger rather than writing to standard output

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/config_merge.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/config_merge.py
@@ -27,7 +27,7 @@ import sys
 
 from .utils.java import Java
 from .utils.log import fatal
-from .utils.parsers import get_javaparser
+from .utils.parsers import get_java_parser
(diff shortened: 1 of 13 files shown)


vladak:
review from eric.saint.etienne (Oracle):
in `str2bool`:
•  you could cache the value of `v.lower()`
• are you sure you want to use `bool=t` for `true` and `bool=f` for `false` when the user can use `y`/`n` already?
• the argument to `str2bool` is guaranteed to be a string (see `type` in https://docs.python.org/3/library/argparse.html#the-add-argument-method) so no need to check if it's type `bool`... but if you except anything else than a `str` then you also need to make sure it's a `str` before calling `lower()` unconditionally.
• The test `test_str2bool`doesn't test for uppercase
• the test `test_str2bool_exception`doesn't test against niceties like empty strings and unicode strings (non ASCII)

PR #2941, opened 2019-09-27
disambiguate when tag is also name of a subdir

louie0817:
https://github.com/oracle/opengrok/issues/2936
Please accept my written acceptance of the OCA for this PR. I will submit signed document in parallel.


vladak:
Thanks !

PR #2945, opened 2019-09-30
Enable leading wildcard by default

tarzanek:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -435,6 +435,7 @@ public final class Configuration {
     public Configuration() {
         // This list of calls is sorted alphabetically so please keep it.
         cmds = new HashMap<>();
+        setAllowLeadingWildcard(true);
         setAllowedSymlinks(new HashSet<>());
         setAuthorizationWatchdogEnabled(false);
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -494,6 +494,7 @@ public Configuration() {
         setUserPageSuffix("");
         setWebappLAF("default");
         // webappCtags is default(boolean)
+        setAllowLeadingWildcard(true);
```
idodeclare: Hi. See above @tulinkry’s comment at line 436.

tarzanek:
thanks @idodeclare , fixed the order ;-)

vladak:
indexer in Docker uses the Indexer option. Didn't you want to remove that as well ?

vladak:
So, what is the plan with `--leadingWildCards` going forward ?

tarzanek:
we can leave it for some time as backwards compatibility(since I assume lots of folks do use it in their scripts) and then remove it?

vladak:
file a issue to track it then.

idodeclare:
Maybe let’s leave in perpetuity for folks who want to prevent the execution of costly leading-wildcard queries?

vladak:
Can be. Would be nice to turn the command line option into toggle then.

PR #2947, opened 2019-10-02
report duration of PendingFileCompleter

vladak:
Lately I found that indexer spends inordinate amount of time in `PendingFileCompleter#complete()` (also there were lots of files under the `xref` directory with the `org_opengrok` suffix) which made me wonder how much time is actually spend there. This change adds simple reporting.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -46,6 +46,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.TandemPath;
 
@@ -174,17 +176,23 @@ class PendingFileCompleter {
         Instant start = Instant.now();
         int numDeletions = completeDeletions();
         LOGGER.log(Level.FINE, "deleted {0} file(s) (took {1})",
-                new Object[]{numDeletions, Duration.between(start, Instant.now())});
+                new Object[]{numDeletions, DurationFormatUtils.
+                        formatDurationWords(Duration.between(start, Instant.now()).toMillis(),
+                        true, true)});
 
         start = Instant.now();
         int numRenamings = completeRenamings();
         LOGGER.log(Level.FINE, "renamed {0} file(s) (took {1})",
-                new Object[]{numRenamings, Duration.between(start, Instant.now())});
+                new Object[]{numRenamings, DurationFormatUtils.
+                        formatDurationWords(Duration.between(start, Instant.now()).toMillis(),
+                true, true)});
 
         start = Instant.now();
         int numLinkages = completeLinkages();
         LOGGER.log(Level.FINE, "affirmed links for {0} path(s) (took {1})",
-                new Object[]{numLinkages, Duration.between(start, Instant.now())});
+                new Object[]{numLinkages, DurationFormatUtils.
+                        formatDurationWords(Duration.between(start, Instant.now()).toMillis(),
+                        true, true)});
 
         return numDeletions + numRenamings + numLinkages;
     }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -42,7 +42,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -34,6 +34,8 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
@@ -166,12 +168,21 @@ class PendingFileCompleter {
      * @throws java.io.IOException if an I/O error occurs
      */
     public int complete() throws IOException {
+        Instant start = Instant.now();
         int numDeletions = completeDeletions();
-        LOGGER.log(Level.FINE, "deleted {0} file(s)", numDeletions);
+        LOGGER.log(Level.FINE, "deleted {0} file(s) (took {1})",
+                new Object[]{numDeletions, Duration.between(start, Instant.now())});
+
+        start = Instant.now();
         int numRenamings = completeRenamings();
-        LOGGER.log(Level.FINE, "renamed {0} file(s)", numRenamings);
+        LOGGER.log(Level.FINE, "renamed {0} file(s) (took {1})",
+                new Object[]{numRenamings, Duration.between(start, Instant.now())});
+
+        start = Instant.now();
         int numLinkages = completeLinkages();
-        LOGGER.log(Level.FINE, "affirmed links for {0} path(s)", numLinkages);
+        LOGGER.log(Level.FINE, "affirmed links for {0} path(s) (took {1})",
+                new Object[]{numLinkages, Duration.between(start, Instant.now())});
+
         return numDeletions + numRenamings + numLinkages;
     }
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -49,9 +49,8 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.opengrok.indexer.logger.LoggerFactory;
+import org.opengrok.indexer.util.Progress;
 import org.opengrok.indexer.util.TandemPath;
-
-import static org.opengrok.indexer.util.Progress.printProgress;
 
 /**
  * Represents a tracker of pending file deletions and renamings that can later
@@ -216,13 +215,12 @@ class PendingFileCompleter {
             new PendingFileRenamingExec(f.getTransientPath(),
                 f.getAbsolutePath())).collect(
             Collectors.toList());
+        Progress progress = new Progress(LOGGER, "pending renames", numPending);
 
-        AtomicInteger currentCount = new AtomicInteger();
         Map<Boolean, List<PendingFileRenamingExec>> bySuccess =
             pendingExecs.parallelStream().collect(
             Collectors.groupingByConcurrent((x) -> {
-                printProgress(LOGGER, "pending renames",
-                        currentCount.incrementAndGet(), numPending);
+                progress.incrementAndLog();
                 try {
                     doRename(x);
                     return true;
@@ -265,13 +263,12 @@ class PendingFileCompleter {
             parallelStream().map(f ->
             new PendingFileDeletionExec(f.getAbsolutePath())).collect(
             Collectors.toList());
+        Progress progress = new Progress(LOGGER, "pending deletions", numPending);
 
-        AtomicInteger currentCount = new AtomicInteger();
         Map<Boolean, List<PendingFileDeletionExec>> bySuccess =
             pendingExecs.parallelStream().collect(
             Collectors.groupingByConcurrent((x) -> {
-                printProgress(LOGGER, "pending deletions",
-                        currentCount.incrementAndGet(), numPending);
+                progress.incrementAndLog();
                 try {
                     doDelete(x);
                     return true;
@@ -319,13 +316,12 @@ class PendingFileCompleter {
             linkages.parallelStream().map(f ->
                 new PendingSymlinkageExec(f.getSourcePath(),
                         f.getTargetRelPath())).collect(Collectors.toList());
+        Progress progress = new Progress(LOGGER, "pending renames", numPending);
 
-        AtomicInteger currentCount = new AtomicInteger();
         Map<Boolean, List<PendingSymlinkageExec>> bySuccess =
             pendingExecs.parallelStream().collect(
                 Collectors.groupingByConcurrent((x) -> {
-                    printProgress(LOGGER, "pending renames",
-                            currentCount.incrementAndGet(), numPending);
+                    progress.incrementAndLog();
                     try {
                         doLink(x);
                         return true;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -42,11 +42,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
```diff
@@ -241,9 +266,12 @@ private int completeDeletions() throws IOException {
             new PendingFileDeletionExec(f.getAbsolutePath())).collect(
             Collectors.toList());
 
+        AtomicInteger currentCount = new AtomicInteger();
         Map<Boolean, List<PendingFileDeletionExec>> bySuccess =
             pendingExecs.parallelStream().collect(
             Collectors.groupingByConcurrent((x) -> {
+                printProgress(LOGGER, "pending deletions",
```
ahornace: It is possible due to thread intertwining that you will print higher progress before the lower progress, right? E.g. 100% will be printed and after that only 50%. Is it okay?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package org.opengrok.indexer.util;
+
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class Progress {
```
ahornace: Maybe make it a bit more object-oriented? E.g. `Progress` class would be initialized by `logger`, `totalCount`, etc. and then you could just call a method like `incrementAndPrint()`?

vladak: all right, it looks nicer now.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
vladak: not sure how big problem is this. My guess is that if anyone runs with `--progress` then he also bumps the log level higher value so the interleaving would not be as disturbing as with the extreme 100/50 case.

The solution would be to spawn a reporting thread I guess.

ahornace: Or maybe make `printProgress` `synchronized`?

vladak: That would be more expensive that the atomic integer and could slow down the overall operation.

ahornace: I would still pick consistency over it. Very roughly, let's say lock costs `100ns`, even though you would have a thousands of operations, it still should be nothing compared to indexing time. But I'm okay with it. Should not happen that often and still, it's just a log.

vladak: added simple logger thread.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,127 @@
...
+
+        // nag the thread.
+        synchronized (sync) {
+            sync.notify();
+        }
+    }
+
+    private void logLoop() {
+        // Assuming printProgress configuration setting cannot be changed on the fly.
+        if (!RuntimeEnvironment.getInstance().isPrintProgress()) {
```
ahornace: Why not move this check to the constructor and return `DummyProgress` which will not have a thread overhead?

vladak: wanted to do it and then forgot, thanks for the reminder.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class Progress implements AutoCloseable {
+    private Logger logger;
```
idodeclare: `final` for some of these?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+    private long totalCount;
+    private String prefix;
+
+    private AtomicLong currentCount = new AtomicLong();
+    private Thread loggerThread = null;
+    private AtomicBoolean run = new AtomicBoolean();
```
idodeclare: The usual signal for this type of worker thread coordination is `volatile Boolean` since it is updated by only one thread and read by the other and no further synchronization is necessary — since there’s no harm to loop an additional time

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+        if (loggerThread != null) {
+            // nag the thread.
+            synchronized (sync) {
+                sync.notify();
+            }
+        }
+    }
+
+    private void logLoop() {
+        while (run.get()) {
```
idodeclare: I would recommend `true` here and put the return only in the loop.

Also I recommend keeping a local `int` counter and only logging if the new `currentCount` is actually greater than the locally tracked value. This way regardless of the timing of the thread signal the highest count will be logged before the thread ends and also there won’t be a duplication of logging for certain timings of the thread signaling.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            if (logger.isLoggable(currentLevel)) {
+                logger.log(currentLevel, "Progress: {0} ({1}%) for {2}",
+                        new Object[]{currentCount, currentCount * 100.0f /
+                                totalCount, prefix});
+            }
+
+            if (currentCount >= totalCount) {
+                return;
```
idodeclare: Superfluous given the above suggestions

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            // wait for event
+            try {
+                synchronized (sync) {
+                    if (!run.get()) {
```
idodeclare: Do you think you might accidentally forgo logging the final entry by returning here?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+            } else if (currentCount % 10 == 0) {
+                currentLevel = Level.FINER;
+            } else {
+                currentLevel = Level.FINEST;
+            }
+
+            if (logger.isLoggable(currentLevel)) {
+                logger.log(currentLevel, "Progress: {0} ({1}%) for {2}",
+                        new Object[]{currentCount, currentCount * 100.0f /
+                                totalCount, prefix});
```
idodeclare: The original implementation had a guard against divide by zero

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,135 @@
...
+            } else if (currentCount % 50 == 0) {
+                currentLevel = Level.FINE;
+            } else if (currentCount % 10 == 0) {
+                currentLevel = Level.FINER;
+            } else {
+                currentLevel = Level.FINEST;
+            }
+
+            if (logger.isLoggable(currentLevel)) {
+                logger.log(currentLevel, "Progress: {0} ({1}%) for {2}",
```
idodeclare: “Prefix” is something “coming before” but `prefix` is only ever used at the end

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: oh yes, for all

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: okay, thanks. changed.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: restored

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: It's actually deadlock prevention. If the thread calling close() notifies the loop thread while it is still executing the loop body (before the wait() call) they will wait for each other indefinitely.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: Interestingly that's how I implemented the loop initially. Also, I wanted to do the increment check however then forgot about it or maybe I was too lazy/distracted :-)

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,142 @@
...
+                                totalCount, suffix});
+            }
+
+            cachedCount = currentCount;
+
+            // wait for event
+            try {
+                synchronized (sync) {
+                    if (!run) {
+                        return;
```
idodeclare: >> Do you think you might accidentally forgo logging the final entry by returning here?

>It's actually deadlock prevention. If the thread calling close() notifies the loop thread while it is still executing the loop body (before the wait() call) they will wait for each other indefinitely.

I meant maybe it's better to `continue` instead of `return` to be sure to do the final logging. I recommend moving the `run` check of line 82 right before this try block — i.e. check `run` right after logging and in an unsynchronized context. Still keep this `synchronized` check, and it's a nice double-checked locking pattern.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,142 @@
...
+            if (!run) {
+                return;
+            }
+
+            long currentCount = this.currentCount.get();
+            Level currentLevel;
+
+            // Do not log if there was no progress.
+            if (cachedCount >= currentCount) {
+                continue;
```
idodeclare: [Spurious wakeups](http://tutorials.jenkov.com/java-concurrency/thread-signaling.html#spurious-wakeups) are a thing. If so, and you get here, then you have a tight loop with a `continue` burning CPU. I recommend just making the logging below conditional on `currentCount > cachedCount` and proceeding to `wait` again.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: I see, thanks.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
vladak: I actually encountered the active wait myself when adding a test.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+            for (int i = 0; i < totalCount; i++) {
+                progress.increment();
+                // Give the logger thread some time to log.
+                while (progress.getLoggerThread().getState() != Thread.State.WAITING) {
+                    TimeUnit.MILLISECONDS.sleep(10);
+                }
+            }
+        }
+
```
idodeclare: Might it be worth it to confirm that the logged thread terminated?

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
```diff
@@ -0,0 +1,90 @@
...
+            for (int i = 0; i < totalCount; i++) {
+                futures.add(executor.submit(progress::increment));
+            }
+        }
+
+        System.out.println("Waiting for threads to finish");
+        futures.forEach(future -> {
+            try {
+                future.get();
```
idodeclare: Might the `AutoCloseable` `close()` have already finished? Should the futures waits happen inside the `try`?

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
```diff
@@ -0,0 +1,90 @@
...
+            } catch (Exception e) {
+                Assert.fail();
+            }
+        });
+        executor.shutdown();
+        System.out.println("Verifying");
+        Mockito.verify(logger, atLeast(1)).log(any(), anyString(), any(Object[].class));
```
idodeclare: If the futures waits are changed to happen before `close()`, might this be strictly `times(totalCount)` as with the previous test?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,144 @@
...
+        if (loggerThread == null) {
+            return;
+        }
+
+        try {
+            run = false;
+            synchronized (sync) {
+                sync.notify();
+            }
+            loggerThread.join(0);
```
idodeclare: Is perhaps `join()` more understandable by not having to remember that 0 means "wait forever"?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
```diff
@@ -0,0 +1,144 @@
...
+            } else if (currentCount % 50 == 0) {
+                currentLevel = Level.FINE;
+            } else if (currentCount % 10 == 0) {
+                currentLevel = Level.FINER;
+            } else {
+                currentLevel = Level.FINEST;
+            }
+
+            // Do not log if there was no progress.
+            if (logger.isLoggable(currentLevel) && (cachedCount < currentCount)) {
```
idodeclare: Nit: if `cachedCount` isn't less than `currentCount`, might the above `currentCount` arithmetic be short-circuited too?

vladak: sure, good catch.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
vladak: good idea

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
vladak: yes, that would be more robust.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
vladak: this case does not wait for the logger thread to reach the waiting state so it usually is less that total count.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/ProgressTest.java
```diff
@@ -0,0 +1,94 @@
...
+            for (int i = 0; i < totalCount; i++) {
+                progress.increment();
+                // Give the logger thread some time to log.
+                while (loggerThread.getState() != Thread.State.WAITING) {
+                    TimeUnit.MILLISECONDS.sleep(10);
+                }
+            }
+        }
+
+        Assert.assertSame(loggerThread.getState(), Thread.State.TERMINATED);
```
idodeclare: Not sure if the thread will 100% have `TERMINATED` so immediately after the `close()` but I see it did pass once through Travis.

vladak: okay, added a loop.

PR #2949, opened 2019-10-03
LDAP uid case (in)sensitivity fix for LdapFilter plugin

vladak:
Adds basic attribute transforms to `LdapFilterPlugin`.

Sample config:
```xml
    <void method="add">                                                         
            <void property="name">                                              
                <string>opengrok.auth.plugin.LdapFilterPlugin</string>          
            </void>                                                             
            <void property="flag">                                              
                <string>REQUIRED</string>                                       
            </void>                                                             
            <void property="setup">                                             
                <void method="put">                                             
                    <string>configuration</string>                              
                    <string>/opengrok/auth/config/ldap-plugin-config-pdit.xml</string>
                </void>                                                         
                <void method="put">                                             
                    <string>filter</string>                                     
                    <string>(&amp;(objectclass=posixGroup)(cn=foo*)(memberUid=%uid%))</string>
                </void>                                                         
                <void method="put">                                             
                    <string>transforms</string>                                 
                    <string>uid:toLowerCase</string>                            
                </void>                                                         
            </void>                                                             
        </object>                                                               
    </void>                
```

Will update https://github.com/oracle/opengrok/wiki/Authorization-plugins once this is in.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
@@ -71,7 +71,7 @@ public class FilterUtil {
     }
 
     /**
-     * see expandUserFilter
+     * see expandUserFilter.
      */
     public static String expandUserFilter(User user, String filter) {
         return expandUserFilter(user, filter, null);
--- a/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
@@ -27,6 +27,7 @@ import opengrok.auth.plugin.entity.User;
 
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -51,9 +52,9 @@ public class FilterUtil {
     static String doTransform(String value, String transform) {
         switch (transform) {
             case LOWER_CASE:
-                return value.toLowerCase();
+                return value.toLowerCase(Locale.ROOT);
             case UPPER_CASE:
-                return value.toUpperCase();
+                return value.toUpperCase(Locale.ROOT);
             default:
                 throw new UnsupportedOperationException(String.format("transform '%s' is unsupported", transform));
         }
--- a/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
@@ -25,14 +25,56 @@ package opengrok.auth.plugin.util;
 
 import opengrok.auth.plugin.entity.User;
 
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
 public class FilterUtil {
 
     private FilterUtil() {
         // utility class
     }
 
-    public static String replace(String filter, String name, String value) {
+    private static final String LOWER_CASE = "toLowerCase";
+    private static final String UPPER_CASE = "toUpperCase";
+
+    public static void checkTransforms(Map<String, String> transforms) {
+        Set<String> possibleTransforms = new HashSet<>(Arrays.asList(LOWER_CASE, UPPER_CASE));
+        for (String transform : transforms.values()) {
+            if (!possibleTransforms.contains(transform)) {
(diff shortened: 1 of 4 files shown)


vladak:
@tulinkry could you take a look ?

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
```diff
@@ -25,12 +25,58 @@
...
+            if (!possibleTransforms.contains(transform)) {
+                throw new UnsupportedOperationException(String.format("invalid transform: %s", transform));
+            }
+        }
+    }
+
+    static String doTransform(String value, String transform) {
+        switch (transform) {
+            case LOWER_CASE:
+                return value.toLowerCase();
```
idodeclare: Specify `Locale.ROOT` (to avoid the problem of #1768)

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
```diff
@@ -25,12 +25,58 @@
...
+            case UPPER_CASE:
+                return value.toUpperCase();
```
idodeclare: Specify `Locale.ROOT`

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
vladak: oo, thanks !

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
```diff
@@ -25,12 +25,59 @@
...
+    }
+
+    static String doTransform(String value, String transform) {
+        switch (transform) {
+            case LOWER_CASE:
+                return value.toLowerCase(Locale.ROOT);
+            case UPPER_CASE:
+                return value.toUpperCase(Locale.ROOT);
+            default:
+                throw new UnsupportedOperationException(String.format("transform '%s' is unsupported", transform));
```
tulinkry: did you try what happens when you try to load/use plugin with such settings?

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
```diff
@@ -25,12 +25,59 @@
...
+            if ((transform = transforms.get(name)) != null) {
+                value = doTransform(value, transform);
+            }
+        }
+
+        return filter.replaceAll("(?<!\\\\)%" + name + "(?<!\\\\)%", value);
+    }
+
+    /**
+     * see expandUserFilter.
```
tulinkry: `@see`

plugins/src/main/java/opengrok/auth/plugin/util/FilterUtil.java
vladak: you mean testing the valid values or the exception case ?

vladak: tested both `toUpperCase` and invalid transform. The plugin loaded with the former just fine, the latter failed with exception as expected:
```
14-Oct-2019 14:12:55.976 SEVERE [http-nio-8080-exec-564] org.opengrok.indexer.authorization.AuthorizationPlugin.load Plugin "opengrok.auth.plugin.LdapFilterPlugin" has failed while loading with exception:
        java.lang.UnsupportedOperationException: invalid transform: toFoo
                at opengrok.auth.plugin.util.FilterUtil.checkTransforms(FilterUtil.java:46)
                at opengrok.auth.plugin.LdapFilterPlugin.loadTransforms(LdapFilterPlugin.java:105)
                at opengrok.auth.plugin.LdapFilterPlugin.load(LdapFilterPlugin.java:91)
                at org.opengrok.indexer.authorization.AuthorizationPlugin.load(AuthorizationPlugin.java:121)
                at org.opengrok.indexer.authorization.AuthorizationStack.load(AuthorizationStack.java:148)
                at org.opengrok.indexer.authorization.AuthorizationStack.load(AuthorizationStack.java:148)
                at org.opengrok.indexer.authorization.AuthorizationFramework.loadAllPlugins(AuthorizationFramework.java:293)
                at org.opengrok.indexer.authorization.AuthorizationFramework.afterReload(AuthorizationFramework.java:357)
                at org.opengrok.indexer.framework.PluginFramework.reload(PluginFramework.java:379)
...
```

PR #2955, opened 2019-10-11
make getDiffData() parallel

vladak:
This change should speed up diff view for large (in terms of history) repositories.

For renamed files the approach taken (parallelize `getRevision()` in `getDiffData()`) is sub-optimal as the most expensive work (retrieve the history of the file across renames) is done twice (but in parallel as opposed to the previous state). @DaliborSkrobak told me he has maybe an idea how to fix this better.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -57,6 +57,7 @@ import org.apache.lucene.search.SearcherManager;
 import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.util.NamedThreadFactory;
 import org.opengrok.indexer.authorization.AuthorizationFramework;
 import org.opengrok.indexer.authorization.AuthorizationStack;
 import org.opengrok.indexer.history.HistoryGuru;
@@ -91,6 +92,7 @@ public final class RuntimeEnvironment {
     private final ReentrantReadWriteLock configLock;
     private final LazilyInstantiate<IndexerParallelizer> lzIndexerParallelizer;
     private final LazilyInstantiate<ExecutorService> lzSearchExecutor;
+    private final LazilyInstantiate<ExecutorService> lzRevisionExecutor;
     private static final RuntimeEnvironment instance = new RuntimeEnvironment();
 
     private final Map<Project, List<RepositoryInfo>> repository_map = new ConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -255,16 +261,28 @@ public DiffData getDiffData() {
             InputStream[] in = new InputStream[2];
             try {
                 // Get input stream for both older and newer file.
+                ExecutorService executor = Executors.newFixedThreadPool(2,
```
idodeclare: Is a private pool necessary? I wonder why not use `IndexerParallelizer` `getHistoryExecutor()` and `CountDownLatch` with timeout (and cancellation) to track the work?

vladak: While I agree permanent pool (for the whole web app) would be better, I have several objections to using `IndexerParallelizer` pool:
  - nit: the name suggests it should be used for indexing
  - the pools do not name the threads (which I find useful when debugging) - this can be fixed though
  - the `lzHistoryExecutor` has number of threads set to `env.getHistoryParallelism()` which I am not sure translates well into web app runtime

idodeclare:
> nit: the name suggests it should be used for indexing

Our namespace is `org.opengrok.indexer`.

vladak:
> > nit: the name suggests it should be used for indexing
> 
> Our namespace is `org.opengrok.indexer`.

:-) Fair enough.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -255,16 +261,27 @@ public DiffData getDiffData() {
...
                     if (in[i] == null) {
                         data.errorMsg = "Unable to get revision "
                                 + Util.htmlize(data.rev[i]) + " for file: "
                                 + Util.htmlize(getPath());
+                        executor.shutdownNow();
                         return data;
                     }
                 }
+                // The Executor used by given repository will enforce the timeout.
+                executor.awaitTermination(0, TimeUnit.SECONDS);
```
idodeclare: I'm not sure that a `shutdownNow()`/`awaitTermination()` is apt for a shared, long-running thread pool during a user request.

vladak: oops. Thanks for spotting this.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -153,6 +157,20 @@ public Thread newThread(Runnable runnable) {
...
+    }
+
+    private ExecutorService newRevisionExecutor() {
+        return Executors.newFixedThreadPool(this.getMaxRevisionThreadCount(),
+                new NamedThreadFactory("get-revision"));
+    }
+
+    public void shutdownRevisionExecutor() throws InterruptedException {
+        getRevisionExecutor().shutdownNow();
+        getRevisionExecutor().awaitTermination(0, TimeUnit.SECONDS);
```
idodeclare: Is there an effective difference between awaiting `0` or just not awaiting?

vladak: for some reason I thought this will wait indefinitely. Will use some sane value.

vladak:
Thanks Chris !

PR #2956, opened 2019-10-11
fix scrolling to symbol in Chrome

### LINKED ISSUE #2943: Desktop Chrome: Click local symbol doesn't scroll to its definition but top of the file
**Describe the bug**
When clicking a local symbol, it scrolls to top of the file instead of its definition position. It happens for desktop Chrome (Ubuntu and Windows7), but none of Microsoft IE, FireFox, Safari, iOS Chrome has such problem with the same server. I'm not sure whether there's some compatiblity issue with latest desktop Chrome.

OS: Ubuntu 18.04
OpenGrok: 1.3.1
JDK: 1.8.0_222
Webapp server: Tomcat8
Desktop Chrome: Version 77.0.3865.90 (Official Build) (64-bit)

**To Reproduce**
1. Index with this command:
sudo opengrok-indexer -J=-Djava.util.logging.config.file=/var/opengrok/logging.properties -a lib/opengrok.jar -- -c /usr/local/bin/ctags -s /sourcecode -d /var/opengrok/data -H -P -S -G  -W /var/opengrok/etc/configuration.xml -U http://localhost:8080/source
2. Restart tomcat
3. Browser code in desktop Chrome and click local symbol to scroll

**Expected behavior**
When clicking symbol, if it is defined in the same file, it should scroll to its definition.

**Screenshots**
N/A

**Additional context**
I also tried previous non-python version (1.1-rc34) and got the same issue. I do not remember there was such issue before, since I'm keeping using OpenGrok for years.

vladak:
fixes #2943

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
@@ -112,8 +112,8 @@ public class Scripts implements Iterable<Scripts.Script> {
         SCRIPTS.put("tablesorter-parsers", new FileScript("js/tablesorter-parsers-0.0.2.min.js", 13));
         SCRIPTS.put("tablesorter-parsers" + DEBUG_SUFFIX, new FileScript("js/tablesorter-parsers-0.0.2.js", 13));
         SCRIPTS.put("searchable-option-list", new FileScript("js/searchable-option-list-2.0.7.min.js", 14));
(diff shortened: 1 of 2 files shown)


vladak:
Forgot to commit, should be fixed now.

vladak:
I tested this in Chrome 75.0.3770.100 on Ubuntu, however I cannot replicate the problem there. This needs testing in Firefox too.

PR #2957, opened 2019-10-13
Feature/typescript

idodeclare:
Hello,

Please consider for integration this patch to add TypeScript support based on TypeScript 3.6 and ECMAScript 10th Edition, June 2019.

* Add ECMAScript.lexh and ECMAScriptProductions.lexh.
* `.TS` extension is re-mapped to `TypeScriptAnalyzerFactory`.
* `AnalyzerGuru` version is bumped so that existing, indexed `.TS` files are re-analyzed by the new `TypeScriptAnalyzer`.
* Refactor `JavaScriptAnalyzer` also to use the 10th Edition ECMAScript*.lexh files.
* `JavaScriptAnalyzer`version is bumped so that existing, indexed `.JS` files are re-analyzed.

Thank you.


opengrok-indexer/src/test/resources/analysis/typescript/sample.ts
```diff
@@ -0,0 +1,226 @@
+/*
+ * Copyright 2019 GitHub
```
tarzanek: hmm, below is MIT ? if yes I am OK

tarzanek:
nice, thank you very much for this, I'd think it can be merged, assuming all licensing is in order
@vladak what do you think?

opengrok-indexer/src/test/resources/analysis/typescript/sample.ts
idodeclare: Yes. I scrounge for substantial samples that are CDDL-compatible

tarzanek:
thank you @idodeclare !

idodeclare:
@tarzanek , my pleasure!

Issue #2917, opened 2019-09-05
JSON Injection


QiAnXinCodeSafe:
opengrok-master/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
![image](https://user-images.githubusercontent.com/39950310/64309337-d37e6780-cfce-11e9-88a9-40c0b245f0ba.png)

opengrok-master/opengrok-indexer/src/main/java/org/opengrok/indexer/util/ClassUtil.java
![image](https://user-images.githubusercontent.com/39950310/64309380-e42edd80-cfce-11e9-974a-c70c1253a80a.png)
![image](https://user-images.githubusercontent.com/39950310/64309394-edb84580-cfce-11e9-865e-3b8ac09c22f1.png)

The method writes unvalidated input into JSON. This call could allow an attacker to inject arbitrary elements or attributes into the JSON entity.

vladak:
Could you expand on why this is a problem ? The `value` in `ProjectsController` can be indeed arbitrary (although currently this administrative interface is protected by localhost filter) however there has to be a setter for the field and Jackson has to be able to form an object from the string value. So, the validation happens when JSON is converted to Java object.

vladak:
As suggested in https://github.com/oracle/opengrok/pull/2962#discussion_r336719935 a DTO design pattern should be used to access data via API. This should help to get rid of any "JSON injection".

PR #2962, opened 2019-10-18
add API to get RepositoryInfo fields

vladak:
The existing endpoint to get the `type` was folded into the new endpoint:
```
$ curl --silent http://localhost:8080/source/api/v1/repositories/property/type?repository=/testrepo; echo
Mercurial
```
and other work too:
```
$ curl --silent http://localhost:8080/source/api/v1/repositories/property/datePatterns?repository=/testrepo; echo
["yyyy-MM-dd hh:mm ZZZZ"]
$ curl --silent http://localhost:8080/source/api/v1/repositories/property/parent?repository=/testrepo; echo
ssh://foo.us.bar.com/foo
$ curl --silent http://localhost:8080/source/api/v1/repositories/property/working?repository=/testrepo; echo
true 
```


vladak:
The trouble of returning `NOT_FOUND` (404) is that web.xml has the `/enoent` handler which is somewhat inappropriate for API.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/RepositoriesController.java
```diff
@@ -24,28 +24,45 @@
...
+    public Object get(@QueryParam("repository") final String repositoryPath, @PathParam("field") final String field)
+            throws IOException {
+
+        RepositoryInfo ri = getRepositoryInfo(repositoryPath);
+        if (ri == null) {
+            throw new WebApplicationException("cannot find repository with path: " + repositoryPath,
+                    Response.Status.NOT_FOUND);
+        }
+
+        return ClassUtil.getFieldValue(ri, field);
```
idodeclare: I object to exposing data got via encapsulation-breaking reflection through web APIs like this. It was done for `Configuration`, and I would hope to see it undone.

The safer way to my thinking is to create a transfer object with public fields where an OpenGrok developer has to explicitly determine what is safely copied out of `RepositoryInfo` (or `Configuration`) to be exposed via `getFieldValue` through a web API.

vladak:
> May I ask why you're not running tests locally to confirm your pull requests are good to go?
> 
> `mvn clean && mvn test`

Because it is faster to run it in Wercker than on my laptop. Time to order new one I guess..

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/RepositoriesController.java
vladak: Same thing is done also in `ProjectsController`. I realize this is a bit of a lazy approach, wonder if there is a way how to create/generate such object using annotations.

vladak: Anyway, added DTO sub-class just to see how it would look like. Is this something you had in mind ?

vladak: Also, I see this approach should fix #2917.

vladak: Using bean generator from cglib works fine. I like it better than duplicated getters/setters. This approach can be pushed further - collection of name/type pairs, annotations, etc.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryInfo.java
```diff
@@ -89,6 +90,30 @@ public RepositoryInfo(RepositoryInfo orig) {
         this.currentVersion = orig.currentVersion;
     }
 
+    /**
+     * @return Data Transfer Object for RepositoryInfo.
+     * It will have the same members as in RepositoryInfo except datePatterns.
+     */
+    public Object getRepositoryInfoData() {
```
idodeclare: I think it's perhaps unusual for classes to create their own DTOs. For OpenGrok maybe it should be `org.opengrok.indexer.web` that should own the DTO logic, external to e.g. `org.opengrok.indexer.history` or `org.opengrok.indexer.configuration`?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/BeanBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ */
+
+package org.opengrok.indexer.util;
+
+import net.sf.cglib.beans.BeanGenerator;
```
idodeclare: I had eyeballed [ModelMapper](http://modelmapper.org/getting-started/) for possible usefulness, and I see on the [cglib wiki](https://github.com/cglib/cglib/wiki) that ModelMapper uses cglib. If you get a chance, I'm curious what you think.

vladak: No matter how I tried to use ModelMapper it was always returning the DTO object with all members being null. Then I realized its major downside - it assumes getters and setters are already in place in the DTO object.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryInfo.java
vladak: Both approaches have some sense I think. Having the DTO definition logic near web app is possible a bit more neat.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/BeanBuilder.java
vladak: In the end I combined modelmapper with cglib and annotations to achieve the desired effect.

opengrok-web/src/main/java/org/opengrok/web/util/DTO.java
```diff
@@ -0,0 +1,55 @@
...
+        // ModelMapper assumes getters/setters so use BeanGenerator to provide them.
+        BeanGenerator beanGenerator = new BeanGenerator();
+        for (Field field : object.getClass().getDeclaredFields()) {
+            if (field.isAnnotationPresent(DTOElement.class)) {
+                beanGenerator.addProperty(field.getName(), field.getType());
+            }
+        }
+        Object bean = beanGenerator.create();
+
+        ModelMapper modelMapper = new ModelMapper();
```
idodeclare: I read that `ModelMapper` is thread-safe, and "unless you need different mappings between the same types, then it’s best to re-use the same `ModelMapper` instance ... as a singleton."

vladak: okay, reused.

PR #2967, opened 2019-10-23
Extract symbol-tokenizer test code

idodeclare:
Hello,

Please consider for integration this patch to extract a method from code duplicated in symbol tokenizer tests.

Thank you.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/ada/AdaSymbolTokenizerTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/ada/AdaSymbolTokenizerTest.java
@@ -19,18 +19,17 @@
 
 /*
  * Copyright (c) 2010, 2018, Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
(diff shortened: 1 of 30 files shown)


idodeclare:
Rebased to also apply to new `SQLSymbolTokenizerTest` and `PLSQLSymbolTokenizerTest`.

tarzanek:
nice cleanup, merging, thank you

idodeclare:
Thank you for looking, @tarzanek

tarzanek:
well I merged and mac build now fails, I did a second rebuild, but I guess the problem will be elsewhere than in mac test flakiness

idodeclare:
I just know that `brew update` is not recommended by Travis-CI, but we do that. I think it consumes a larger portion of the build, and then we hit the 50 minute limit

tarzanek:
33mins ... <sigh> ... ok, so we shouldn't do brew update ...  @vladak any reason why we do brew update?
resp. if we need certain packages updated, we should only update those and not the whole thing it seems

tarzanek:
I removed the brew update, so let's see what will break :-)

idodeclare:
Also try `export HOMEBREW_NO_AUTO_UPDATE=1` before any `brew install`

tarzanek:
ok, this seems to have worked, we're now close to linux+jdk11 times

PR #2968, opened 2019-10-28
Filter subclasses override Match() instead of creating Ignore()

idodeclare:
Hello,

Please consider for integration this minor revision to `@Override match()` instead of creating `Ignore()` methods for `Filter` subclasses, which improves clarity.

Thank you.


idodeclare:
Thank you, @tulinkry and @tarzanek

Issue #2589, opened 2018-12-12
folder names "CVS" and "CVSROOT" wont index


jvaneck:
I have a folder called "CVS" that refuses to index. I've removed the ignoredNames property from the configuration file and they just get added right back when I re-index. 

Why do they add themselves back in? How do I index a folder called "CVS" ? 

lines removed from configuration.xml file: 
     <void method="add">
      <string>CVS</string>
     </void>
     <void method="add">
      <string>CVSROOT</string>
     </void>

vladak:
Looks like #1458. I don't think this is possible given that `CVSrepository` constructor does this:
```java
78          ignoredFiles.add(".cvsignore");
79          ignoredDirs.add("CVS");
80          ignoredDirs.add("CVSROOT");
```

vladak:
Okay, it is just a coincidence, not a CVS repository metadata directory.

vladak:
`CVSrepository#isRepositoryFor()` is
```java

138          if (file.isDirectory()) {
139              File cvsDir = new File(file, "CVS");
140              return cvsDir.isDirectory();
141          }
142          return false;
```
It should probably check also some of the contents of the `CVS` directory.

jvaneck:
Is there any way to work around this issue and get the folder to be indexed?

I'm confused about this being an enhancement without a workaround but who am I to argue.

vladak:
I cannot think of any workaround.

There is not much difference whether an issue is marked as bug or enhancement. It only counts whether someone is working on it :-)

PR #2969, opened 2019-10-28
Add --disableRepository handling

idodeclare:
Hello,

Please consider for integration this patch to add `--disableRepository <type_name>` handling, which is an easier workaround for #2589 (via `--disableRepository cvs`) versus enhancing ignored-directory handling to support qualifying only a `CVS` directory containing e.g. `cvswrappers`. (A `Configuration` serialization DSL to support this nuanced ignoring did not seem prudent compared to a user-specified disabling of `CVSRepository`.)

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryFactory.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryFactory.java
@@ -27,11 +27,17 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.opengrok.indexer.configuration.Configuration;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
+import org.opengrok.indexer.index.IgnoredNames;
 import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.ForbiddenSymlinkException;
 
@@ -62,20 +68,43 @@ public final class RepositoryFactory {
         new SSCMRepository()
     };
 
+    private static final Map<String, Class<? extends Repository>> byName = new HashMap<>();
+
+    static {
+        final String REPOSITORY = "Repository";
+        for (Repository repository : repositories) {
+            Class<? extends Repository> clazz = repository.getClass();
+            String repoName = clazz.getSimpleName();
+            byName.put(repoName, clazz);
+            byName.put(repoName.toLowerCase(Locale.ROOT), clazz);
+            if (repoName.endsWith(REPOSITORY)) {
+                String shortName = repoName.substring(0, repoName.length() - REPOSITORY.length());
+                if (shortName.length() > 0) {
+                    byName.put(shortName, clazz);
+                    byName.put(shortName.toLowerCase(Locale.ROOT), clazz);
+                }
+            }
+        }
+    }
+
+    /** Private to enforce static. */
     private RepositoryFactory() {
-        // Factory class, should not be constructed
     }
 
     /**
      * Get a list of all available repository handlers.
      *
-     * @return a list which contains none-{@code null} values, only.
+     * @return a list that contains non-{@code null} values only
      */
     public static List<Class<? extends Repository>> getRepositoryClasses() {
-        ArrayList<Class<? extends Repository>> list
-                = new ArrayList<>(repositories.length);
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+        ArrayList<Class<? extends Repository>> list = new ArrayList<>(repositories.length);
         for (int i = repositories.length - 1; i >= 0; i--) {
-            list.add(repositories[i].getClass());
+            Class<? extends Repository> clazz = repositories[i].getClass();
+            if (isEnabled(clazz, env)) {
+                list.add(clazz);
+            }
         }
         
         return list;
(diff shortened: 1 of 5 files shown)


idodeclare:
[Fixed link typo](https://github.com/oracle/opengrok/compare/2e79e9e..141d8ae)

vladak:
https://github.com/oracle/opengrok/wiki/Indexer-configuration will need to be updated with `disabledRepositories` after this is merged in.

vladak:
I was going to suggest to make it configurable via read-only configuration however as you said this is probably so rare that it does not justify.

vladak:
..in which case just get rid of the changes in `Configuration` and do this purely in `RuntimeEnvironment`.

idodeclare:
> ..in which case just get rid of the changes in Configuration and do this purely in RuntimeEnvironment.

`history.jsp` executes `RepositoryFactory.getRepository` via `HistoryGuru`. We wouldn't want a repository disabled at indexing time to be enabled later in the web runtime.

vladak:
Oh, okay. So there is a way how to serialize the changes for read-only configuration after all.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
```diff
@@ -43,32 +53,78 @@
...
+
+    @ConditionalRun(RepositoryInstalled.MercurialInstalled.class)
+    @Test
+    public void testNormallyEnabledMercurialRepository() throws IllegalAccessException,
+            InvocationTargetException, ForbiddenSymlinkException, InstantiationException,
+            NoSuchMethodException, IOException {
+
+        File root = new File(repository.getSourceRoot(), "mercurial");
+        env.setSourceRoot(root.getAbsolutePath());
+        assertNotNull("should get MercurialRepository",
```
vladak: nit: the message describes different assertion than the one made here

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
```diff
@@ -43,32 +53,78 @@
...
+    public void testMercurialRepositoryWhenDisabled() throws IllegalAccessException,
+            InvocationTargetException, ForbiddenSymlinkException, InstantiationException,
+            NoSuchMethodException, IOException {
+
+        env.setDisabledRepositories(new HashSet<>(
+                Collections.singletonList("MercurialRepository")));
+
+        File root = new File(repository.getSourceRoot(), "mercurial");
+        env.setSourceRoot(root.getAbsolutePath());
+        assertNull("should not get MercurialRepository if disabled",
```
vladak: ditto

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -162,6 +163,9 @@ public static void main(String[] argv) {
                 pauseToAwaitProfiler();
             }
 
+            disabledRepositories.addAll(cfg.getDisabledRepositories());
```
vladak: perhaps logprint (`FINEST` ?) all disabled repositories at the end of `parseOptions()` for debugging purposes ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryFactory.java
```diff
@@ -62,20 +68,43 @@
         new SSCMRepository()
     };
 
+    private static final Map<String, Class<? extends Repository>> byName = new HashMap<>();
+
+    static {
+        final String REPOSITORY = "Repository";
```
vladak: hopefully everyone will name respository classes with this suffix. Assert this in a test ?

idodeclare: OK added `testRepositoryFactoryEveryImplIsNamedAsRepository`

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
idodeclare: OK added

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
idodeclare: Isn't think OK since the message is an expectation description — i.e. "what _should_ have happened?" — displayed on failure?

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
idodeclare: Isn't think OK since the message is an expectation description — i.e. "what _should_ have happened?" — displayed on failure?

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
vladak: As I said this is probably nit picking, however the test verifies the non-null property rather than checking that the returned object is actually an instance of `MercurialRepository`.

vladak:
As for the serialization, this works when used in read-only configuration:
```xml

  <void property="disabledRepositories">
   <void method="add">
    <string>MercurialRepository</string>
   </void>
   <void method="add">
    <string>CVSRepository</string>
   </void>
  </void>

 </object>
</java>
```

opengrok-indexer/src/test/java/org/opengrok/indexer/history/RepositoryFactoryTest.java
```diff
@@ -105,8 +105,12 @@ public void testNormallyEnabledMercurialRepository() throws IllegalAccessExcepti
 
         File root = new File(repository.getSourceRoot(), "mercurial");
         env.setSourceRoot(root.getAbsolutePath());
-        assertNotNull("should get MercurialRepository",
+        assertNotNull("should get repository for mercurial/",
                 RepositoryFactory.getRepository(root));
+
+        List<Class<? extends Repository>> clazzes = RepositoryFactory.getRepositoryClasses();
```
vladak: nice addition

vladak:
Thanks !

PR #2971, opened 2019-11-04
Fix log suffix to read "linkages"

idodeclare:
Hello,

Please consider for integration this patch to fix the log message for pending linkages from the recent PR #2947.

Thank you.


idodeclare:
Thank you!

PR #2975, opened 2019-11-11
add disabled project handling to mirror.py

vladak:
This change adds optional command to be called/executed for disabled projects. Like with sync.py this supports both RESTful API calls as well as command execution. This allows for instance to tag the disabled projects with Messages so they are annotated in the UI.

The disabled command is configured globally and will vary based on project thanks to pattern substitution/append. If this is found not to be flexible enough, it can be modified to allow per-project disabled commands however for the time being I find this good enough.

Examples:
### API call:
```yml
disabled_command:                                                               
  command:                                                                     
    - http://localhost:8080/source/api/v1/messages                             
    - POST                                                                     
    - cssClass: info                                                           
      duration: PT1M                                                           
      tags: ['%PROJECT%']                                                      
      text: disabled project                                                   
projects:
  foo:
    disabled: true
```
### command exec:
```yml
disabled_command:                                                               
  command: [cat] 
projects:
  foo:
    disabled: true
```

Any failures in disabled command processing are logged and do not change the overall result of the mirroring command.

https://github.com/oracle/opengrok/wiki/Repository-synchronization will be updated once this is in.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -71,7 +71,11 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
 
 ### adds message to a system [POST]
 
-Usable values for `cssClass` are: `info`, `warning`, `error`
+Usable values for `cssClass` are: `info`, `warning`, `error`.
+This will affect CSS class of the displayed message, leading to different coloring.
+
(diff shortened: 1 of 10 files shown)


vladak:
Tomas Heran (@oracle) says the changes looks good to him.

PR #2976, opened 2019-11-12
reset ctags to last known good changeset

vladak:
Reset Universal ctags to latest known working version.

### DIFF
--- a/dev/install-universal_ctags.sh
+++ b/dev/install-universal_ctags.sh
@@ -2,6 +2,8 @@
 
 git clone https://github.com/universal-ctags/ctags.git
 cd ctags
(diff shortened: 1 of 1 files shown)


tarzanek:
were you able to figure out the upstream bug?

vladak:
Not yet. It might be something we need to fix.

vladak:
This is just a hot fix.

idodeclare:
The only thing I saw from commit messages is perhaps it’s defunct now to use option such as `—C-kinds` rather than `—kinds-C`. No computer still so I can’t test yet.

vladak:
filed #2977

PR #2963, opened 2019-10-20
Feature/ctags opts

idodeclare:
Hello,

Please consider for integration this patch to support customizing `Ctags` language mapping (inclusion or exclusion) per `-A,--indexer` so that Universal Ctags operation is changed too when users specify additional extensions or prefixes to map to analyzers (or unmap from them).

Thank you.

PR #2979, opened 2019-11-17
Feature/uctags

idodeclare:
Hello,

Please consider for integration this patch to accommodate the recent Universal Ctags strictness of language definitions by running an additional `ctags --list-languages` during the verification stage and only declaring `--langdef` if required by the Ctags version.

I also switched away from the newly deprecated syntax `--LANG-kinds` to `--kinds-LANG`.

To aid initial debugging, I updated `Indexer --help` to show the Ctags command-line when an extra level of detail is requested. This changes `--help --detailed` to instead show more details when `-h,--help` is specified an increasing number of times. I'm open to suggestions if this seems clunky.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -148,17 +148,17 @@ public class Ctags implements Resettable {
 
         command = new ArrayList<>();
         command.add(env.getCtags());
-        command.add("--c-kinds=+l");
+        command.add("--kinds-c=+l");
 
         // Workaround for bug #14924: Don't get local variables in Java
         // code since that creates many false positives.
         // CtagsTest : bug14924 "too many methods" guards for this
         // universal ctags are however safe, so enabling for them
-        command.add("--java-kinds=+l");
+        command.add("--kinds-java=+l");
 
-        command.add("--sql-kinds=+l");
-        command.add("--Fortran-kinds=+L");
-        command.add("--C++-kinds=+l");
+        command.add("--kinds-sql=+l");
+        command.add("--kinds-Fortran=+L");
+        command.add("--kinds-C++=+l");
         command.add("--file-scope=yes");
         command.add("-u");
         command.add("--filter=yes");
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -144,8 +144,9 @@ public class Ctags implements Resettable {
     }
 
     private void initialize() {
-        command = new ArrayList<>();
+        env.validateUniversalCtags();
 
+        command = new ArrayList<>();
         command.add(env.getCtags());
         command.add("--c-kinds=+l");
 
@@ -250,7 +251,9 @@ public class Ctags implements Resettable {
     }
 
     private void addRustSupport(List<String> command) {
-        command.add("--langdef=rust");
+        if (!env.getCtagsLanguages().contains("Rust")) { // Built-in would be capitalized.
+            command.add("--langdef=rust"); // Lower-case if user-defined.
+        }
         defaultLangMap.add(".RS", "rust"); // Upper-case file spec
 
         // The following are not supported yet in Universal Ctags b13cb551
@@ -264,9 +267,12 @@ public class Ctags implements Resettable {
     }
 
     private void addPowerShellSupport(List<String> command) {
-        command.add("--langdef=powershell");
+        if (!env.getCtagsLanguages().contains("PowerShell")) { // Built-in would be capitalized.
+            command.add("--langdef=powershell"); // Lower-case if user-defined.
+        }
         defaultLangMap.add(".PS1", "powershell"); // Upper-case file spec
         defaultLangMap.add(".PSM1", "powershell"); // Upper-case file spec
+
         command.add("--regex-powershell=/\\$(\\{[^}]+\\})/\\1/v,variable/");
         command.add("--regex-powershell=/\\$([[:alnum:]_]+([:.][[:alnum:]_]+)*)/\\1/v,variable/");
         command.add("--regex-powershell=/^[[:space:]]*(:[^[:space:]]+)/\\1/l,label/");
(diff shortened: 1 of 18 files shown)


tarzanek:
:-) looks like you had fun Chris (and this PR is one more reason for us to start moving to scala/kotlin)
merging, thank you

opengrok-indexer/src/test/java/org/opengrok/indexer/util/CtagsUtilTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.opengrok.indexer.condition.ConditionalRun;
+import org.opengrok.indexer.condition.CtagsInstalled;
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+
+import java.util.List;
+
+/**
+ * Represents a container for tests of {@link CtagsUtil}.
+ */
+@ConditionalRun(CtagsInstalled.class)
```
vladak: `CtagsInstalled` is basically `RuntimeEnvironment.getInstance().validateUniversalCtags()` which relies on `CtagsUtil` functionality which is the class under test. This sounds little bit fragile to me.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CtagsUtil.java
```diff
@@ -54,6 +60,49 @@ public static boolean validate(String ctagsBinary) {
...
+            if (lang.length() > 0) {
+                result.add(lang);
+            }
+        }
+        return result;
+    }
+
+    /**
+     * Creates a new instance, and attempts to configure it from the
+     * environment.
```
vladak: should probably explain how it differs to `Ctags()` constructor and when/where to use it.

vladak:
I wonder if all these `addXYZSupport` methods from `Ctags` class can be moved their respective `XYZAnalyzerFactory` classes (or maybe some other class type that will hold such info) so that they are there together with suffixes/magics etc. and collected from there similarly to how `registerAnalyzer()` works.

vladak:
@tarzanek seriously, if you want to move to another language (for the Analyzers), finish #2660 first and then we can talk.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -136,22 +134,31 @@ public void close() {
...
+     * Gets the command-line arguments used to run ctags.
+     * @return a defined (immutable) list
+     */
+    public List<String> getArgv() {
+        initialize();
+        return Collections.unmodifiableList(command);
+    }
+
+    private void initialize() {
+        env.validateUniversalCtags();
```
vladak: is there a point of continuing if this returns False ?

vladak: I am thinking if we should be more strict w.r.t. ctags validation and just terminate the indexer right in the beginning if no valid ctags implementation is found.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -1116,6 +1121,36 @@ private static void pauseToAwaitProfiler() {
...
+        StringBuilder result = new StringBuilder();
+        Ctags ctags = CtagsUtil.newInstance(env);
+        List<String> argv = ctags.getArgv();
+        for (int i = 0; i < argv.size(); ++i) {
+            if (i > 0) {
+                result.append("\t");
+            }
+            String arg = argv.get(i);
+            if (arg == null) {
+                result.append("UNDEFINED");
```
vladak: what will this do ? and how could it happen actually ?

tarzanek:
sorry for too quick merge - @idodeclare will you be able to address @vladak comments in a new PR?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
vladak: Also, the work done by `initialize()` is repeated in every worker which seems unnecessary given all of them will share the same configuration/argv.

vladak:
Most of my comments relate to pre-existing stuff so should be part of distinct PR anyway.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/CtagsUtilTest.java
idodeclare: Haha I hadn't realized the convolution. Probably `CtagsInstalled` has outlived its usefulness, right?

idodeclare:
> I wonder if all these `addXYZSupport` methods from `Ctags` class can be moved their respective `XYZAnalyzerFactory` classes (or maybe some other class type that will hold such info) so that they are there together with suffixes/magics etc. and collected from there similarly to how `registerAnalyzer()` works.
> 
> As it is currently, e.g. `addKotlinSupport()` duplicates the suffixes from `KotlinAnalyzerFactory`.

I like the idea of removing the suffix duplication, but I'm not sure how strong is the case to relocate the `--regex` definitions.

For a long time I thought it would be necessary to do such radical refactoring of `Ctags` in order to customize ctags operation when OpenGrok `-A,--analyzer` is used for suffix/prefix management. I recently realized that the Ctags `LANG` keywords are sufficient to coordinate ctags operation across OpenGrok classes — and #2963 was the result.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
idodeclare: `Indexer` is already strict when ctags validation fails.

I probably should have added a call to `validateUniversalCtags()` to opengrok-web instead of putting it into `Ctags.initialize()`, since my intention anyway was to ensure `Ctags` operates for `--webappCtags on` the same as during indexing.

idodeclare: Regarding the repetition of `initialize()`, we already minimize `Ctags` construction with `CtagsObjectFactory`, so a small bit of redundant initialization seemed harmless.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
idodeclare: I wrote that when `Ctags` initially had a `setBinary()` method and started with `null` binary. I retired that method in a subsequent commit but didn't remove the `"UNDEFINED"` handling.

PR #2981, opened 2019-11-19
fix API call to get repository type

vladak:
Forgot to update the scripts after the API endpoint changes in cset e03dc428.


vladak:
> LGTM. (No tests though?)

yeah. I need to figure out a way how to run the webapp for Python tests or construct mock server based on Apiary documentation.

Issue #930, opened 2015-04-07
Feature request: Support assembly


claunia:
Add support for major assembly languages: 6502, 8080, Z80, 8086, IA-32, x86-64, PowerPC, 68000, MIPS, ARM, Sparc.

tarzanek:
this would be very easy to do ... just write lex grammar parsers for symbols and xrefs ... see recent pushes for haskell, or older for scala/python/...
ctags even support this out of box!

tarzanek:
https://github.com/OpenGrok/OpenGrok/blob/master/doc/ctags.config
has some enhancement we use in Solaris dev to recognize asm functions ...

claunia:
It surprises me you use OpenGrok in Solaris development but have not implemented at least a SPARC and/or x86 support.

tarzanek:
generic full search is already powerful
I will have a deeper look how we handle asm
but above ctag file is probably good enough for people
but I agree it could be better!

vladak:
Currently assembly files are handled in C analyzer, according to suffix (e.g. `*.s`) in `CAnalyzerFactory.java`.

tarzanek:
I got a patch for this from Greg, so I will try to push this soon ...

vladak:
The major question is what exactly is wanted, especially w.r.t. support for different architectures. I'd assume one would like to have at least:
  - function entry points
  - labels as symbols

On the other hand, some functionality of C analyzer should be preserved, e.g. assembly files often contain includes. Currently one can click on the include file name to trigger file based search.

As for the labels, the analyzer would have to understand various notations, e.g. stuff like on x86:

  `jne	4f;`

is forward jump - to label defined after this instruction (similarly '4b' would be backward jump).

vladak:
What kind of patch exactly ? What architectures does it support ?

vladak:
Another notable thing is that ctags-5.8 generates tags for labels etc. for files with  "asm", "ASM", "s", "S" etc. suffixes, in its simplest form it should be just a matter of consuming the tags.

tarzanek:
that's why I say it should be doable ;)

vladak:
@tarzanek Patch from Greg Onufer I guess. Is it (the patch) still floating around somewhere ?

PR #2983, opened 2019-11-21
Feature/uctags2

idodeclare:
Hello,

Please consider for integration this patch as a follow-on to #2979, incorporating feedback from @vladak.

The file extension mapping redundancy in `Ctags` is fixed, and mappings are done based on analyzer prefixes or suffixes during `registerAnalyzer()`. This created a test error for an assembly language test file, `bug19195/test.s`, which was formerly analyzed by the `CAnalyzer` but implicitly tagged by the Ctags Asm parser. I cloned `CAnalyzer` as `AsmAnalyzer`, mapping `.S` and `.ASM` to it.

Thank you.


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -479,10 +501,11 @@ public static void addExtension(String extension, AnalyzerFactory factory) {
     /**
      * Gets an unmodifiable view of the language mappings resulting from
      * {@link #addExtension(String, AnalyzerFactory)} and
-     * {@link #addPrefix(String, AnalyzerFactory)}.
+     * {@link #addPrefix(String, AnalyzerFactory)} merged with default language
+     * mappings of OpenGrok's analyzers.
      */
     public static LangMap getLangMap() {
-        return langMap.unmodifiable();
+        return langMap.mergeSecondary(defaultLangMap).unmodifiable();
```
vladak: I like this design pattern. Is it called somehow ?

vladak:
Nice ! Paves the way for #930 I guess.

idodeclare:
Thank you for reviewing, @vladak.

PR #2987, opened 2019-11-26
Rewrite Dockerfile to build the project

ahornace:
I've seen that it is quite common to build projects directly in `Dockerfile` so I thought I will give it a try. It is then very useful in `docker-compose` since you can build many things by a single command.

If the maven repositories are cached, then it takes about 6 minutes to build the whole docker image. But if travis does not use cache and builds it all from scratch, then it may significantly prolong the travis execution time which might not be that wise.

What do you guys think?

If you do not like it, then feel free to close it :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,59 @@
-FROM debian:stable-slim as fetcher
+# Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
 
-# TODO copy just the 'distribution' directory, not all source code
+FROM ubuntu:bionic as build
+
+RUN apt-get update && apt-get install -y openjdk-8-jdk python3 maven python3-venv
+
+# Create a first layer to cache the "Maven World" in the local repository.
+# Incremental docker builds will always resume after that, unless you update the pom
+WORKDIR /mvn
+COPY pom.xml /mvn/
+COPY opengrok-indexer/pom.xml /mvn/opengrok-indexer/
+COPY opengrok-web/pom.xml /mvn/opengrok-web/
+COPY plugins/pom.xml /mvn/plugins/
+COPY suggester/pom.xml /mvn/suggester/
+
+# distribution and opengrok-tools do not have dependencies to cache
+RUN sed -i 's:<module>distribution</module>::g' /mvn/pom.xml
+RUN sed -i 's:<module>opengrok-tools</module>::g' /mvn/pom.xml
+
+RUN mkdir -p /mvn/opengrok-indexer/target/jflex-sources
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp
+COPY opengrok-indexer/jflex-tt.txt /mvn/opengrok-indexer/jflex-tt.txt
+COPY opengrok-indexer/jflex-tt-end.txt /mvn/opengrok-indexer/jflex-tt-end.txt
+COPY opengrok-indexer/jflex-code.txt /mvn/opengrok-indexer/jflex-code.txt
+COPY opengrok-indexer/jflex-code-end.txt /mvn/opengrok-indexer/jflex-code-end.txt
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp/WEB-INF/ && touch /mvn/opengrok-web/src/main/webapp/WEB-INF/web.xml
+
+# dummy build to cache the dependencies
+RUN mvn package -DskipTests -Dcheckstyle.skip -Dmaven.antrun.skip
+
+# build the project
 COPY ./ /opengrok-source
 WORKDIR /opengrok-source
 
-# update the system
-RUN apt-get update
-
-# find most recent package file
-RUN cp `ls -t distribution/target/*.tar.gz | head -n1 |awk '{printf("%s",$0)}'` /opengrok.tar.gz
+RUN mvn -DskipTests=true -Dmaven.javadoc.skip=true -B -V package
+RUN cp distribution/target/*.tar.gz /opengrok.tar.gz
 
 FROM tomcat:9-jre8
-LABEL maintainer "opengrok-dev@yahoogroups.com"
-
-# prepare OpenGrok binaries and directories
-COPY --from=fetcher opengrok.tar.gz /opengrok.tar.gz
-RUN mkdir -p /opengrok /opengrok/etc /opengrok/data /opengrok/src && \
-    tar -zxvf /opengrok.tar.gz -C /opengrok --strip-components 1 && \
-    rm -f /opengrok.tar.gz
+LABEL maintainer="opengrok-dev@yahoogroups.com"
 
 # install dependencies and Python tools
-RUN apt-get update && apt-get install -y git subversion mercurial unzip inotify-tools python3 python3-pip python3-venv && \
-    python3 -m pip install /opengrok/tools/opengrok-tools*
+RUN apt-get update && apt-get install -y git subversion mercurial unzip inotify-tools python3 python3-pip python3-venv
 
 # compile and install universal-ctags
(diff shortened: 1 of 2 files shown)


vladak:
> If the maven repositories are cached, then it takes about 6 minutes to build the whole docker image. But if travis does not use cache and builds it all from scratch, then it may significantly prolong the travis execution time which might not be that wise.

What actually happens now ? Is the caching in Travis builds used at all ?

Dockerfile
```diff
@@ -1,34 +1,60 @@
-FROM debian:stable-slim as fetcher
+# Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
 
-# TODO copy just the 'distribution' directory, not all source code
+FROM ubuntu:bionic as build
```
vladak: is this going to change the resulting OS in the image ?

ahornace: No, it's just for the build

ahornace:
> What actually happens now ? Is the caching in Travis builds used at all ?

I did not have time to check and play with it. However, is docker build even needed to be in travis? It is basically checking only `Dockerfile` correctness and takes a lot of time.

vladak:
> > What actually happens now ? Is the caching in Travis builds used at all ?
> 
> I did not have time to check and play with it. However, is docker build even needed to be in travis? It is basically checking only `Dockerfile` correctness and takes a lot of time.

It was added to catch any Docker related problems early. Too late to do it during release build. Long running builds/tests are not a problem I'd say - as long as they do not hit the CI timeouts.

Dockerfile
```diff
@@ -1,34 +1,60 @@
...
+RUN mkdir -p /mvn/opengrok-indexer/target/jflex-sources
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp
+COPY opengrok-indexer/jflex-tt.txt /mvn/opengrok-indexer/jflex-tt.txt
+COPY opengrok-indexer/jflex-tt-end.txt /mvn/opengrok-indexer/jflex-tt-end.txt
+COPY opengrok-indexer/jflex-code.txt /mvn/opengrok-indexer/jflex-code.txt
+COPY opengrok-indexer/jflex-code-end.txt /mvn/opengrok-indexer/jflex-code-end.txt
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp/WEB-INF/ && touch /mvn/opengrok-web/src/main/webapp/WEB-INF/web.xml
+
+# dummy build to cache the dependencies
+RUN mvn package -DskipTests -Dcheckstyle.skip -Dmaven.antrun.skip
```
vladak:
```suggestion
RUN mvn -DskipTests -Dcheckstyle.skip -Dmaven.antrun.skip package
```

Dockerfile
```diff
@@ -1,34 +1,60 @@
-FROM debian:stable-slim as fetcher
+# Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
 
-# TODO copy just the 'distribution' directory, not all source code
+FROM ubuntu:bionic as build
+
+RUN apt-get update && apt-get install -y openjdk-8-jdk python3 maven python3-venv
```
vladak:
```suggestion
RUN apt-get update && apt-get install -y openjdk-8-jdk maven python3 python3-venv
```

ahornace:
> It was added to catch any Docker related problems early. Too late to do it during release build. Long running builds/tests are not a problem I'd say - as long as they do not hit the CI timeouts.

That reminds me not to use `latest` tag which is a good practice. In that case there is basically nothing that can make the docker build to fail except change in the `mvn clean package` command which will probably stay the same for a long long time.

But probably better safe than sorry.

Dockerfile
ahornace: I like it the other way round but okay :D

Dockerfile
vladak: So how the resulting OS will get set ? (or where from)

ahornace: From this command: `FROM tomcat:9-jre8`

Dockerfile
```diff
@@ -1,34 +1,60 @@
...
+RUN sed -i 's:<module>distribution</module>::g' /mvn/pom.xml
+RUN sed -i 's:<module>opengrok-tools</module>::g' /mvn/pom.xml
+
+RUN mkdir -p /mvn/opengrok-indexer/target/jflex-sources
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp
+COPY opengrok-indexer/jflex-tt.txt /mvn/opengrok-indexer/jflex-tt.txt
+COPY opengrok-indexer/jflex-tt-end.txt /mvn/opengrok-indexer/jflex-tt-end.txt
+COPY opengrok-indexer/jflex-code.txt /mvn/opengrok-indexer/jflex-code.txt
+COPY opengrok-indexer/jflex-code-end.txt /mvn/opengrok-indexer/jflex-code-end.txt
+RUN mkdir -p /mvn/opengrok-web/src/main/webapp/WEB-INF/ && touch /mvn/opengrok-web/src/main/webapp/WEB-INF/web.xml
```
vladak: I think this section here actually makes it necessary to run Docker builds for every push/PR as it relies on internal structure of the project.

ahornace: Yes, you are right :D I always forget how overcomplicated the build is :D

Dockerfile
```diff
@@ -1,34 +1,60 @@
...
+WORKDIR /mvn
+COPY pom.xml /mvn/
+COPY opengrok-indexer/pom.xml /mvn/opengrok-indexer/
+COPY opengrok-web/pom.xml /mvn/opengrok-web/
+COPY plugins/pom.xml /mvn/plugins/
+COPY suggester/pom.xml /mvn/suggester/
+
+# distribution and opengrok-tools do not have dependencies to cache
+RUN sed -i 's:<module>distribution</module>::g' /mvn/pom.xml
+RUN sed -i 's:<module>opengrok-tools</module>::g' /mvn/pom.xml
```
vladak: why is this necessary ?

Dockerfile
vladak: more than before I mean.

Dockerfile
ahornace: Because if they are present, then `mvn` call will try to execute the plugins in those submodules which will fail the build. And I did not want to create other "dummy" files for it to pass with no added value.

Dockerfile
vladak: I never liked the GNU style of arguments (mixing options with arguments). In this case I think the options are global and not specific to the `package` sub command therefore it makes sense to state them first.

vladak:
> > It was added to catch any Docker related problems early. Too late to do it during release build. Long running builds/tests are not a problem I'd say - as long as they do not hit the CI timeouts.
> 
> That reminds me not to use `latest` tag which is a good practice. In that case there is basically nothing that can make the docker build to fail except change in the `mvn clean package` command which will probably stay the same for a long long time.

using the latest tag where ?

vladak:
> I've seen that it is quite common to build projects directly in `Dockerfile` so I thought I will give it a try. It is then very useful in `docker-compose` since you can build many things by a single command.

could you give an example ? what do you do with docker compose and OpenGrok exactly ?

ahornace:
> could you give an example ? what do you do with docker compose and OpenGrok exactly ?

At work I have a lot of projects that are being built by Dockerfile: 
- ui
- backend
- grafana with configuration

and other dependencies: db, kafka, prometheus

All I need to do is run `docker-compose build && docker-compose up` from any machine with docker and it just runs.

And at the moment I have OpenGrok with metrics being pushed to Graphite and shown in Grafana but it needs some final touches.

ahornace:
> using the latest tag where ?

E.g. I should not use `ubuntu:bionic` because it seems it is being updated regularly but a specific tag which will stay the same forever, e.g. `ubuntu:bionic-20191029`.

PR #2988, opened 2019-11-27
Also validate canonicalRoot when deserializing Configuration

idodeclare:
Hello,

Please consider for integration this small patch that applies the same `--canonicalRoot` validation upon deserializing `Configuration` as is done for `Indexer`.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/CanonicalRootValidator.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/CanonicalRootValidator.java
@@ -1,0 +1,66 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/CanonicalRootValidator.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Represents a validator of a --canonicalRoot value.
+ */
+public class CanonicalRootValidator {
+
+    /**
+     * Validates the specified setting, returning an error message if validation
+     * fails.
+     * @return a defined instance if not valid or {@code null} otherwise
```
vladak: I'd say this sort of return value semantics is a bit unusual however (I'd expect an exception thrown by static class) it is quite clearly marked in the code that uses it (`problem = validate(...)`) so that's fine with me. I think Golang error return style has something to it.

idodeclare: Thank you for looking. Yes since it was going to be used in two different manners — once for a fatal termination and another time to just skip over an entry — I just made it return a problem description.

vladak: Okay. I realize not having to handle an exception makes the code less verbose.

PR #2990, opened 2019-12-01
Format argv logging to aid copy/paste execution

idodeclare:
Hello,

Please consider for integration this patch to adjust the format of `Executor` logging not to use the default `List` serialization that uses commas but to use a specialized serialization that formats commands as would be required to execute on the platform. I.e. the commands are logged more clearly and also could be copy/pasted to a shell for execution (as when debugging).

E.g. formerly the following would appear in a log:

```
Executing command [git, log, -1, --pretty=%cd#%h %an %s, --date=iso8601-strict]
```

and would now be formatted as:

```
Executing command [git log -1 --pretty=$'%cd#%h %an %s' --date=iso8601-strict]
```

I also included `os.name`-dependent formatting for Windows:

```
Executing command [git log -1 --pretty="%cd#%h %an %s" --date=iso8601-strict]
```

Thank you.


vladak:
The Python utilities should get similar treatment.

PR #2998, opened 2019-12-04
Fix #2997 : redirect (conditionally) before writing partially

### LINKED ISSUE #2997: Encountered `Cannot call sendRedirect() after the response has been committed` when navigate search result
****
Encountered `Cannot call sendRedirect() after the response has been committed` when navigate to the search result after perform a search. check the list.jsp, seems it is performing a `response.sendRedirect()` but the JSP generated content before that statement is exceeded 8kb default JSP buffer size and got flushed to the client, so cannot perform `response.sendRedirect()` call.

**To Reproduce**
Setup brand new OpenGrok, index some source, perform a search, navigate into the search result.
Encountered this error.

**Expected behavior**
Can access the search result

**Screenshots**
opengrok-1 3 5

**Environment**
- Ubuntu 18.04.3 LTS
- opengrok-1.3.5
- apache-tomcat-9.0.29 & apache-tomcat-8.5.49 (tested in those application servers, got same result)
- project source code using SVN (1.9.7)

idodeclare:
Hello,

Please consider for integration this patch to relocate the conditional calls to `sendRedirect()` to happen before any response is partially constructed.

Thank you.

### DIFF
--- a/opengrok-web/src/main/webapp/list.jsp
+++ b/opengrok-web/src/main/webapp/list.jsp
@@ -56,6 +56,8 @@ org.opengrok.web.DirectoryListing,
 org.opengrok.indexer.web.SearchHelper"
 %>
 <%
+final String DUMMY_REVISION = "unknown";
+
 {
     // need to set it here since requesting parameters
     if (request.getCharacterEncoding() == null) {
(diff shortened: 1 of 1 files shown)


opengrok-web/src/main/webapp/list.jsp
```diff
@@ -65,6 +67,30 @@ org.opengrok.indexer.web.SearchHelper"
...
+             * revision string so that xref can be generated from the resource
+             * file directly.
+             */
+            location = cfg.getRevisionLocation(DUMMY_REVISION);
+            response.sendRedirect(location);
+            return;
+        }
+    }
+
     Annotation annotation = cfg.getAnnotation();
```
vladak: I wonder if annotate works still after the change, especially the combination with revision (get a historical revision and annotate that).

idodeclare: Yes, the conditional redirect-to-self was and is dependent on `rev.length == 0`, so there is no redirect if a historical rev is specified.

vladak: okay, let's try the fix then.

PR #2999, opened 2019-12-06
use __str__ of the Command class for logging consistently

vladak:
This change should deal with most of the log entries for the `Command` class, e.g.:
```
2019-12-06 21:16:20,068    DEBUG opengrok_tools.utils.command | Command '/bin/false 1 opengrok' took 0 seconds
2019-12-06 21:16:20,069    ERROR opengrok_tools.utils.commandsequence | command '/bin/false 1 opengrok' for project opengrok failed with code 1, breaking
```

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
@@ -83,16 +83,10 @@ class CommandSequence(CommandSequenceBase):
         self.logger = logging.getLogger(__name__)
         self.logger.setLevel(base.loglevel)
 
-    def run_command(self, command):
+    def run_command(self, cmd):
         """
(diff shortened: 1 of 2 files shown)


opengrok-tools/src/main/python/opengrok_tools/utils/commandsequence.py
```diff
@@ -128,14 +128,14 @@ def run(self):
                         if not self.driveon:
                             self.logger.debug("command '{}' for project {} "
                                               "requested break".
-                                              format(self.name, command))
+                                              format(command, self.name))
```
idodeclare: Was this a transposition bug (and below)? Should a brief mention about a fix be included in the commit message?

vladak: It was.

PR #3000, opened 2019-12-07
Feature/fixup args

idodeclare:
Hello,

Please consider for integration this patch with some revisions to arg handling:

* Remove some defunct code.
* Do not silently skip over unexpected option definitions.
* Fix alphabetic ordering of a few `Indexer` options.
* Tweak line spacing at a couple of places for`--help`.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -376,44 +376,6 @@ public final class Indexer {
     }
 
     /**
-     * Web address consisting of webappURI and port of a web address.
-     */
-    public static class WebAddress {
-        private String host;
-        private int port;
-
-        WebAddress(String host, String port) throws NumberFormatException {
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -453,9 +417,6 @@ public static WebAddress parseWebAddress(String webAddr) {
             });
         });
 
-        // An example of how to add a data type for option parsing
-        OptionParser.accept(WebAddress.class, Indexer::parseWebAddress);
```
vladak: I wonder if this was meant to be used for -U ?

idodeclare: Perhaps. The validation of -U elsewhere in `Indexer` is superior.

vladak: Okay then.

PR #3002, opened 2019-12-12
Bugfix SQL and PLSQL

idodeclare:
Hello,

Please consider for integration this patch with some fixes for SQL and PLSQL:

* Extract symbol tokenizers for SQL and PLSQL so they are not just using `PlainSymbolTokenizer`. Share as much of the definitions as possible, and just specialize SQL.lexh and PLSQL.lexh.
* Move content of sql2003reserved.dat, sql2008reserved.dat, sql2011reserved.dat, and plsql2011reserved.dat into `sql.Consts` and `PLSQLConsts`.
* Fix `PLSQLXref` to use `PLSQLConsts` instead of just `sql.Consts`.
* Add SQL and PLSQL symbol tokenizer tests.

Thank you.


vladak:
Nice, thanks !

tarzanek:
hmm, I always just took latest reserved keywords from whatever SQL norm has them and dropped into the analyzer dir
this include of "data" into code itself makes the analyzer more isolated, true, but you need to diff the new sql norm and add/remove whatever changed, when it comes up, which makes the analyser maintainer life harder
what was the reason that triggered this change? (just curious, I don't have a real problem with this merge)

idodeclare:
Lubos,

This came up during my review of casing mismatches in analyzers. I found the following for sql:

* SQL and PLSQL analyzers were properly indicated as `%ignorecase` (`%caseless`) in JFlex, but their keyword matching was technically case-sensitive against the all-caps words in those .dat files and the all-lower-case transformations also entered into the sets. Mixed-case versions would not be identified as keywords.

* `PLSQLXref` was not using `PLSQLConsts`. No class had ever referenced `PLSQLConsts`.

As far as maintainer-life, for every language it's a continual challenge to reconcile OpenGrok's handling against a current version of a language. E.g. when I added TypeScript support, I reconciled the current keywords defined in `javascript.Consts` against the ECMAScript 2019 definitions. Of course I use scripting to help.

I just quickly converted `sql.Consts` and `PLSQLConsts` using my script to integrate successively the 2003, 2008 and 2011 .dat files. (I've never heard of a language standard removing keywords. What's an example?)

tarzanek:
perfect, that explains the background to me,
so the case matching stuff bug is a good catch!

if your scripts are publishable (i.e. don't contain private stuff), feel free to dump them to dev folder https://github.com/oracle/opengrok/tree/master/dev

thank you for all the catches Chris, appreciate it

PR #3011, opened 2020-01-03
Messages class rework

vladak:
This change makes message class levels first class citizens and propagates them to the UI so that the color reflects the highest message level for given project.

Most of the changes were done in 2019 so leaving the copyright year as is.

Example with 2 messages (info + error):
Screenshot from 2020-01-03 14-43-47

I also verified search results case and directory listing view.


vladak:
asked @tulinkry to have a look.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -50,7 +50,44 @@
...
+         */
+        SUCCESS("success"), INFO("info"), WARNING("warning"), ERROR("error");
+
+        private String cssClassString;
+
+        CssClassType(String str) {
+            cssClassString = str;
+        }
+
+        public static CssClassType StringToCssClassType(String val) throws IllegalArgumentException {
```
ktulinger: I think I've never seen a java method starting with capital letter

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -50,7 +50,44 @@
...
+            if (val == null) {
+                return INFO;
```
ktulinger: this should be illegal argument as well no? the caller should check for null and assing default INFO

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -50,7 +50,44 @@
...
+            }
+            throw new IllegalArgumentException("class type does not match any known value");
+        }
+
+        @Override
+        public String toString() {
+            return cssClassString;
+        }
+
+        public static int compare(CssClassType c1, CssClassType c2) {
```
ktulinger: there is compareTo of a Comparable interface

ktulinger: you should implement it

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -230,4 +231,32 @@ private static String messagesToJson(Group group, String... additionalTags) {
...
+    /**
+     * @param tags message tags
+     * @return name of highest cssClass of messages present in the system or null.
+     */
+    public static String getCssClass(String... tags) {
+        Set<MessagesContainer.AcceptedMessage> messages = new TreeSet<>();
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+        for (String tag : tags) {
+            messages.addAll(env.getMessages(tag));
```
ktulinger: sounds like a job for flatMap

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
vladak: This was probably implementation ~convenience~laziness.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
vladak: This is needed for `max()` which requires `Comparator` rather than `Comparable`. However, the method needs to be static so it can be used via method reference hence the class cannot implement `Comparator`.

vladak: perhaps using `Comparator` as a static member is the cleanest way.

ktulinger:
Actually, when you changed it to this message levels, this field no longer descibes a css class but it's rather a MessageLevel (like LogLevel). This css class could be misleading because you don't translate it to a css class at all ... you make it `note-<class>`

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -78,9 +79,7 @@ public String toString() {
             return cssClassString;
         }
 
-        public static int compare(CssClassType c1, CssClassType c2) {
-            return Integer.compare(c1.ordinal(), c2.ordinal());
-        }
+        public static Comparator<CssClassType> valueComparator = Comparator.comparingInt(Enum::ordinal);
```
ktulinger: can be final and thus with constant naming scheme - capital letters

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
vladak: okay, I converted the code for the sake of learning stream API :-)

vladak:
> Actually, when you changed it to this message levels, this field no longer descibes a css class but it's rather a MessageLevel (like LogLevel).

True, I always wanted to get rid of this naming.

apiary.apib
```diff
@@ -84,7 +84,7 @@ see https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java
 
         {
           "tags": ["main"],
-          "cssClass": "class",
+          "messageLevel": "class",
```
ktulinger: use valid value

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -100,7 +100,7 @@ private Message() { // needed for deserialization
     public Message(
             final String text,
             final Set<String> tags,
-            final String cssClass,
+            final String messageLevel,
```
ktulinger: why not use the MessageLevel type?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -115,16 +115,16 @@ public Message(
...
     }
 
     public Set<String> getTags() {
         return tags;
     }
 
-    public String getCssClass() {
-        return cssClass.toString();
+    public String getMessageLevel() {
+        return messageLevel.toString();
```
ktulinger: why not use the MessageLevel type?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -250,9 +250,9 @@ public String getText() {
             return message.getText();
         }
 
-        @JsonProperty("cssClass")
-        public String getCssClass() {
-            return message.getCssClass();
+        @JsonProperty("messageLevel")
+        public String getMessageLevel() {
```
ktulinger: why not use the MessageLevel type?

vladak:
@tulinkry all the comments should be addressed by now.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -50,7 +52,39 @@
...
+         */
+        SUCCESS("success"), INFO("info"), WARNING("warning"), ERROR("error");
+
+        private final String messageLevelString;
+
+        MessageLevel(String str) {
+            messageLevelString = str;
+        }
+
+        public static MessageLevel stringToMessageLevel(String val) throws IllegalArgumentException {
```
ktulinger: non blocking, I found usualy these as `fromString`

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -157,6 +191,28 @@ public int compareTo(final Message o) {
...
+        public MessageLevel deserialize(final JsonParser parser, final DeserializationContext context)
+                throws IOException {
+            try {
+                return MessageLevel.stringToMessageLevel(context.readValue(parser, String.class));
+            } catch (DateTimeParseException e) {
+                throw new IOException(e);
+            }
+        }
+    }
+
```
ktulinger: I just wonder, serializer is not needed?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesUtils.java
```diff
@@ -230,4 +234,32 @@ private static String messagesToJson(Group group, String... additionalTags) {
...
     }
+
+    /**
+     * @return name of highest cssClass of messages present in the system or null.
+     */
+    static String getHighestMessageLevel(Collection<MessagesContainer.AcceptedMessage> messages) {
+        return messages.
+                stream().
+                map(MessagesContainer.AcceptedMessage::getMessageLevel).
+                max(VALUE_COMPARATOR).
```
ktulinger: I'd qualify here with the class name so the name does not seem a little out of context.

opengrok-web/src/main/webapp/default/style.css
```diff
@@ -1249,6 +1236,34 @@ h6.message-group-caption {
...
+    background-color: #31708f;
+}
+
+.note-warning {
+    background-color: #8a6d3b;
+}
+
+.note-error {
+    background-color: #a94442;
+}
```
ktulinger: a question here is if you want to restrict these classes to be used only for `important-note` or just without any restriction (one can use it later for completely unrelated styling with the given colour).

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
vladak: nope, it works out of the box.

opengrok-web/src/main/webapp/default/style.css
vladak: Is there a way how to limit the `note-*` styles for both `important-note` and `.important-note.important-note-rounded`. The thing is, `.important-note` is used both with and without `important-note-rounded`.

vladak: Initially, I wanted to unite the `note-*` styles with `.message-group-item.*` styles so the color definitions are defined just once however the latter use `color` instead of `backgound-color`.

vladak: By elegant I mean something like this:
```CSS
.important-note {
    text-align: center;
    display: inline-block;
    max-height: 100%;
}

.important-note.note-success {
    background-color: #3c763d;
}

... // more .note styles
```
and then it would be nice if I can say something like this:
```CSS
.important-note.note-*.important-note-rounded {
    border-radius: 50%;
    max-width: 100%;
    min-width: 20px;
}
```
however I am not sure this is possible in CSS.

vladak: Follow-up on our Slack discussion: it turns out that I misread the CSS selectors guide - the `.class1.class2` selector is not strictly hierarchical so there can be style that can only be used with `important-note` while being also usable with both `important-note.note` and `important-note-rounded`.

PR #3012, opened 2020-01-03
Lucene 8.4.0

tarzanek:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -58,7 +58,7 @@ import javax.ws.rs.core.Response;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
-import org.apache.lucene.codecs.lucene80.Lucene80Codec;
+import org.apache.lucene.codecs.lucene84.Lucene84Codec;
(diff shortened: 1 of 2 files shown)


vladak:
is this backward compatible with 8.3 in terms of index ?

tarzanek:
I think a reindex will be needed

idodeclare:
> I think a reindex will be needed

Shouldn't it be OK without reindex? "[File formats are back-compatible between major versions. Version X.N should be able to read indexes generated by any version after and including version X&#8209;1.0...."](https://cwiki.apache.org/confluence/display/lucene/BackwardsCompatibility)

tarzanek:
> 
> 
> > I think a reindex will be needed
> 
> Shouldn't it be OK without reindex? "[File formats are back-compatible between major versions. Version X.N should be able to read indexes generated by any version after and including version X‑1.0...."](https://cwiki.apache.org/confluence/display/lucene/BackwardsCompatibility)

the reason is just due to codec update, and I think we did 2 jumps in lucene version recently anyways (8.3 and now 8.4, so we are past 2 last versions in a short time from where I stand)

idodeclare:
But the [PR](https://github.com/apache/lucene-solr/pull/973) for the postings format change included in 8.4 shows the previous 8.0 codec moved to backward-codecs.

backward-codecs for 8.4 include `org.apache.lucene.codecs.lucene70.Lucene70Codec` (and now `org.apache.lucene.codecs.lucene80.Lucene80Codec`) which aligns with supporting "version X&#x2011;1.0".

tarzanek:
then it should work :-)
(but I never tested anything special and the test Vlad wrote just do basic opening of old index (which works))

Issue #3020, opened 2020-01-13
run Travis builds periodically


vladak:
This will help to detect changes in the build environment.

vladak:
Travis supports this via cron jobs (https://github.com/travis-ci/travis-ci/issues/582) or there is https://nightli.es/

vladak:
Scheduled to run weekly and with _Do not run if there has been a build in the last 24h_ set to on.

ref: https://www.coursera.org/lecture/continuous-integration/schedules-and-setting-up-example-schedules-in-travis-ci-tpxUo

Issue #3019, opened 2020-01-13
HistoryGuruTest.testGetRevision fails in Travis Windows environment


vladak:
The `HistoryGuruTest.testGetRevision` fails in Travis Windows env due to:
```
Jan 13, 2020 8:17:15 AM org.opengrok.indexer.util.Executor exec

WARNING: Non-zero exit status -1 from command [blame Makefile] in directory C:\Users\travis\AppData\Local\Temp\source975669849041833016\rcs_test

[ERROR] Tests run: 7, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 16.187 s <<< FAILURE! - in org.opengrok.indexer.history.HistoryGuruTest

[ERROR] testGetRevision  Time elapsed: 0.449 s  <<< FAILURE!

java.lang.AssertionError: Failed to get revision 1.1 of C:\Users\travis\AppData\Local\Temp\source975669849041833016\razor\Razor-Simple\SimpleCProgram\main.c

	at org.junit.Assert.fail(Assert.java:88)

	at org.junit.Assert.assertTrue(Assert.java:41)

	at org.junit.Assert.assertNotNull(Assert.java:712)

	at org.opengrok.indexer.history.HistoryGuruTest.testGetRevision(HistoryGuruTest.java:110)

	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)

	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

	at java.lang.reflect.Method.invoke(Method.java:498)

	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)

	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)

	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)

	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)

	at org.junit.rules.RunRules.evaluate(RunRules.java:20)

	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)

	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)

[...]

Issue #3018, opened 2020-01-13
XrefTest.java tests fail in Travis Windows environment


vladak:
In Travis Windows environment all `*XrefTest.java` tests fail, e.g.:
```
AdaXrefTest.sampleTest:38->XrefTestBase.writeAndCompare:62->XrefTestBase.writeAndCompare:90 org.opengrok.indexer.analysis.ada.AdaAnalyzerFactory xref--should have no diffs
```
with:
```
[INFO] Running org.opengrok.indexer.analysis.ada.AdaXrefTest
@@ -1,50 +1,50 @@
- <!DOCTYPE html>

- <html lang="en">

- <head>

- <meta charset="UTF-8">

- <title>sampleFile - OpenGrok cross reference for /sampleFile</title></head><body>

- <a class="l" name="1" href="#1">1</a><span class="c">--</span>

- <a class="l" name="2" href="#2">2</a><span class="c">-- CDDL HEADER START</span>

- <a class="l" name="3" href="#3">3</a><span class="c">--</span>

- <a class="l" name="4" href="#4">4</a><span class="c">-- The contents of this file are subject to the terms of the</span>

- <a class="l" name="5" href="#5">5</a><span class="c">-- Common Development and Distribution License (the &quot;License&quot;).</span>

- <a class="l" name="6" href="#6">6</a><span class="c">-- You may not use this file except in compliance with the License.</span>

- <a class="l" name="7" href="#7">7</a><span class="c">--</span>

- <a class="l" name="8" href="#8">8</a><span class="c">-- See LICENSE.txt included in this distribution for the specific</span>

- <a class="l" name="9" href="#9">9</a><span class="c">-- language governing permissions and limitations under the License.</span>

- <a class="hl" name="10" href="#10">10</a><span class="c">--</span>

- <a class="l" name="11" href="#11">11</a><span class="c">-- When distributing Covered Code, include this CDDL HEADER in each</span>

- <a class="l" name="12" href="#12">12</a><span class="c">-- file and include the License file at LICENSE.txt.</span>

- <a class="l" name="13" href="#13">13</a><span class="c">-- If applicable, add the following below this CDDL HEADER, with the</span>

- <a class="l" name="14" href="#14">14</a><span class="c">-- fields enclosed by brackets &quot;[]&quot; replaced with your own identifying</span>

- <a class="l" name="15" href="#15">15</a><span class="c">-- information: Portions Copyright [yyyy] [name of copyright owner]</span>

[...]

Issue #3017, opened 2020-01-12
test_binaries/test_versions do not work in Travis Windows environment


vladak:
The `test_binaries`/`test_versions` tests fail when run in Windows Travis environment. Initially, the command return value was `None` instead of 0. When I tried to follow the advice on https://stackoverflow.com/questions/5658622/python-subprocess-popen-environment-path and set `env` passed to `Popen` in `Command` to `environ.copy()` (Maven sets the environment to parent environment with the virtual environment binary path) and `shell=True` the return value changed to 1 and the error output contained e.g. `'opengrok-projadm' is not recognized as an internal or external command, operable program or batch file.` so there is still something wrong. I verified that the `PATH` environment variable contains the Python directory - this is often the cure suggested for such problems.

vladak:
This may be actually a bug in the functionality of the `Command` class rather than tests.

vladak:
ref https://travis-ci.org/vladak/OpenGrok/jobs/636083127

PR #3021, opened 2020-01-13
run Windows builds in Travis only

vladak:
This change unites all testing to be done in Travis. The major advantage is that Windows test results will be now available for pull requests. I noticed that some of the recently integrated changes broke Windows builds only by taking a casual look at AppVeyor build logs.

The builds are a bit fragile since e.g. the Python and Maven versions are hardcoded in the paths. We should be running the builds not only for changes but periodically (#3020).

Some problematic tests were identified, most of them I disabled for Windows:
  - #3016 : not a Windows problem per se, the test needs to be more robust
  - #3017 : problem with `subprocess.Popen()` on Windows w.r.t. env vars/path
  - #3018 : Windows/Travis related problems (newlines/locale)
  - #3019 : possibly Travis env problem

I think it is beneficial to include this change even with the above disabled tests so that we get more timely idea when something gets broken.


vladak:
The JDK is currently set to Java SE 8, maybe we should use OpenJDK + more versions.

Issue #2580, opened 2018-12-06
Feature Request:  Select multiple file types


gtoph:
Would be nice if we could select multiple file types to search through, such as looking for a string in all "C" and "python" type files for example.

vladak:
I think the underlying service can support multiple types just fine, it is just a matter of changing the UI.

tulinkry:
Yes this is supported, but not selectable in the ui.

<img width="603" alt="screenshot 2019-02-17 at 14 06 26" src="https://user-images.githubusercontent.com/6997160/52913321-463b2600-32bd-11e9-8b64-e8f44c38de37.png">

gtoph:
There was a 'multiple' option I found in one of the jsp files that turned the type list into a checkbox from radio and allowed multiple selections, but it needs a little work as far as ui behavior and actually sending those multiple selections to the searching framework.

vladak:
The trouble with the UI is that the language selector is implemented via searchable-option-list (SOL) and therefore is supplemented by rectangles - just like the project picker. Having another rectangle "cloud" next to the search form si unwanted, I think.

ahornace:
Closing as duplicate in favour of #1537

PR #3029, opened 2020-01-29
Clear the type field on clear form event

### LINKED ISSUE #3028: clear button does not clear file type field
When clicking clear, all search fields are cleared but not the file type drop down.  Should clear not also reset that?

tulinkry:
fixes #3028

I haven't tested this in browser. Can someone pull the code and test it? It should be working.


vladak:
> I haven't tested this in browser. Can someone pull the code and test it? It should be working.

does not seem to do the trick for me.

tulinkry:
Right. The code in the type picker will clear it only when `config.multiple` is allowed (which is not for the type field).

vladak:
> Right. The code in the type picker will clear it only when `config.multiple` is allowed (which is not for the type field).

that's tracked by #2580 BTW.

vladak:
Superseded by #3152.

Issue #1984, opened 2018-01-22
For a single search result redirect to the file immediately


tulinkry:
May be a useful feature.

Iff there is a single search result we can maybe immediately redirect to that xref.

Use case:
Clicking on class name in some Java file leads directly to the file with that class (assuming unique name per project) rather than to a search results with one result.

vladak:
@tulinkry feel free to submit a PR for this.

tulinkry:
Looks quite challenging as `totalHits` after the search means total files which matches the query (and not the particular lines). There isn't a evident simple way to get the total line count out of the search.

PR #3043, opened 2020-02-19
Bugfix/redirect 1 def

idodeclare:
Hello,

Please consider for integration this patch to remedy the broken functionality to automatically link to a definition from an xref file when the definition is the single search result.

While researching an approach for #1984 I found there was a former single-definition redirect that was no longer working but could be fixed with a bit of special handling.

Thank you.


idodeclare:
Closing for now, as I'm near a solution for all (non-`HIST`) single-result redirection.

PR #3045, opened 2020-02-19
implement history API endpoint

vladak:
This change adds the API endpoint for history. It supports pagination and history for directories.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -6,6 +6,14 @@ OpenGrok RESTful API documentation. The following endpoints are accessible under
 
 Besides `/suggester` and `/search` endpoints, everything is accessible from `localhost` only.
 
+## Authorization framework reload [/configuration/authorization/reload]
+
+### reloads authorization framework [POST]
+
++ Request (application/text)
+
++ Response 204
+
 ## Configuration [/configuration]
 
 ### return XML representation of configuration [GET]
@@ -59,11 +67,68 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
 
 + Response 204
 
-## Authorization framework reload [/configuration/authorization/reload]
+## History [/history]
 
-### reloads authorization framework [POST]
+### get history entries [GET]
 
-+ Request (application/text)
++ Parameters
+  + path (string) - path of file/directory to get history for
+  + withFiles (optional, string) - whether to include list of files
+  + start (optional, int) - start index
+  + max (optional, int) - number of entries to get (default value 1000)
+
++ Response 200 (application/json)
+  + Body
+
+        {
+          "entries": [
+            {
+              "revision": "86b0ab6b",
+              "date": 1565163646000,
+              "author": "Adam Hornacek <adam.hornacek@oracle.com>",
+              "tags": null,
+              "message": "Try to use mvnw in CI",
+              "files": [
+                "/opengrok/docker/README.md"
+              ]
+            },
+            {
+              "revision": "a391bead",
+              "date": 1564745178000,
+              "author": "Adam Hornacek <adam.hornacek@oracle.com>",
+              "tags": null,
+              "message": "Fix docker readme typo",
+              "files": [
+                "/opengrok/docker/README.md"
+              ]
+            },
+            {
+              "revision": "33551f14",
+              "date": 1563531391000,
+              "author": "Vladimir Kotal <vlada@devnull.cz>",
+              "tags": "1.3.0, 1.2.25, 1.2.24",
+              "message": "document docker-compose",
+              "files": [
+                "/opengrok/docker/README.md"
+              ]
+            }
+          ]
+        }
+
+## Include files reload [/system/includes/reload]
+
+### reloads all include files for web application [PUT]
+
++ Response 204
+
+## Index searchers refresh [/system/refresh]
+
+### refreshes index searchers for specified project [PUT]
+
++ Request (text/plain)
+  + Body
+
+            project name to refresh
 
 + Response 204
 
@@ -121,6 +186,18 @@ see https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java
             "text": "test&nbsp;message"
           }
         ]
+
+## Path descriptions [/system/pathdesc]
+
+### updates path descriptions for web application [POST]
+
++ Request (text/plain)
+  + Body
+
+            /foo    bar
+            /foo/foo    bar bar
+
++ Response 204
 
 ## Projects [/projects]
 
@@ -266,7 +343,7 @@ The repository path is relative to source root.
   + hist (optional, string) - history field value to search for
   + type (optional, string) - type of the files to search for
   + projects (optional, string) - projects to search in
-  + maxresults (optional, string) - maximum number of documents whose hits will be returned
+  + maxresults (optional, string) - maximum number of documents whose hits will be returned (default 1000)
   + start (optional, string) - start index from which to return results
 
 + Response 200 (application/json)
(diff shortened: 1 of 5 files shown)


vladak:
I re-ordered the Apiary definitions to be sorted alphabetically (best effort).

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/HistoryController.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.SortedSet;
+
+@Path("/history")
+public class HistoryController {
+
+    private RuntimeEnvironment env = RuntimeEnvironment.getInstance();
```
tulinkry: final, static?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/HistoryController.java
```diff
@@ -0,0 +1,157 @@
...
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(revision, date, author, tags, message, files);
+        }
+    }
+
+    static class HistoryDTO implements JSONable {
+        @JsonProperty
+        private List<HistoryEntryDTO> entries;
```
tulinkry: final

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/HistoryController.java
```diff
@@ -0,0 +1,157 @@
...
+        @JsonProperty
+        private Date date;
+        @JsonProperty
+        private String author;
+        @JsonProperty
+        private String tags;
+        @JsonProperty
+        private String message;
+        @JsonProperty
+        private SortedSet<String> files;
```
tulinkry: all can be final

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/HistoryController.java
```diff
@@ -0,0 +1,157 @@
...
+        public int hashCode() {
+            return Objects.hash(entries);
+        }
+    }
+
+    static HistoryDTO getHistoryDTO(List<HistoryEntry> historyEntries) {
+        List<HistoryEntryDTO> entries = new ArrayList<>();
+        for (HistoryEntry entry : historyEntries) {
+            entries.add(new HistoryEntryDTO(entry));
+        }
```
tulinkry: nb, looks like a task for streaming api

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/HistoryControllerTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    public void testHistoryGet() throws Exception {
+        final String path = "git";
+        Response response = target("history")
+                .queryParam("path", path)
+                .request()
+                .get();
+        HistoryDTO history = response.readEntity(new GenericType<HistoryDTO>() {});
+        History repoHistory = HistoryGuru.getInstance().getHistory(new File(repository.getSourceRoot(), path));
+        assertEquals(history, getHistoryDTO(repoHistory.getHistoryEntries()));
```
tulinkry: can you assert something more - contanins some string, minimal length?
it relies on the serialization and when it's broken you wouldn't catch it.

apiary.apib
```diff
@@ -59,11 +67,68 @@ Besides `/suggester` and `/search` endpoints, everything is accessible from `loc
...
-## Authorization framework reload [/configuration/authorization/reload]
+## History [/history]
 
-### reloads authorization framework [POST]
+### get history entries [GET]
 
-+ Request (application/text)
++ Parameters
+  + path (string) - path of file/directory to get history for
+  + withFiles (optional, string) - whether to include list of files
```
tulinkry: string?

vladak:
Authnz checks added, however they can only be used in servlet container per https://stackoverflow.com/a/35868654/11582827 , i.e. the Grizzly test will skip them (and produce bunch of `java.lang.NoSuchMethodException: Could not find a suitable constructor in javax.servlet.http.HttpServletRequest class.` exceptions to the log).

apiary.apib
tulinkry: it's boolean not a string

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/HistoryController.java
vladak: they would have to be initialized in the plain constructor first I guess.

tulinkry: I get it, it's the serialization issue

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/HistoryControllerTest.java
vladak: better now ? :-)

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/HistoryControllerTest.java
```diff
@@ -105,12 +105,19 @@ public void testHistoryDTOEquals() {
...
+        int start = 2;
         Response response = target("history")
                 .queryParam("path", path)
+                .queryParam("max", size)
+                .queryParam("start", start)
                 .request()
                 .get();
         HistoryDTO history = response.readEntity(new GenericType<HistoryDTO>() {});
+        assertEquals(size, history.getEntries().size());
+        assertEquals("Kryštof Tulinger <krystof.tulinger@oracle.com>", history.getEntries().get(0).getAuthor());
```
tulinkry: oh yeah, thank you

tulinkry:
Last question, why would reusing HistoryEntry and making it jsonable be a bad decision? Why you return an object of {entries: []} instead of just plain array [] ?

vladak:
> Last question, why would reusing HistoryEntry and making it jsonable be a bad decision?

It's partly caused by the process - initially I augmented the members of `HistoryEntry` and `History` with `@DTOElement` annotations however I figured out that `DTOUtils` does not work recursively and also has problems with `StringBuffer` (used for `message`) so I went for separate DTO object and finally used `JSONable`. Using dedicated DTO object is less prone to problems with crossing the boundary so I stick to that.

tulinkry:
Ok then. And the second question?

tulinkry:
I'm thinking of it ... when you used object as a returned type instead of an array. You could include additional metadata like "next" or "prev" (page) with the link to the next page and don't set them when it's the last page. Also something like total number would be nice or the input parameters (start and max). I imagine that they can be useful for someone developing against it.

vladak:
> Why do you choose returning an object of `{entries: []}` instead of just plain `array []` ?

For easier testing. It is not strictly necessary. I just pushed a changeset that removes `HistoryDTO` and the endpoint returns a list of history entries.

However, it made me wonder about pagination, e.g. how does one know how many entries in total there are ?

vladak:
Okay, resurrected `HistoryDTO`, added the counters in.

tulinkry:
> However, it made me wonder about pagination, e.g. how does one know how many entries in total there are ?

Just like with github api, one does not know. One tries to get some records and stops when the array becomes empty.

tulinkry:
Let's go

vladak:
> > However, it made me wonder about pagination, e.g. how does one know how many entries in total there are ?
> 
> Just like with github api, one does not know. One tries to get some records and stops when the array becomes empty.

That's a bit hapless I think, especially when the underlying data changes while iterating. Having the total count helps in this regards a bit.

tulinkry:
> That's a bit hapless I think, especially when the underlying data changes while iterating. Having the total count helps in this regards a bit.

I'm not saying it isn't. Just that some giants make it that way.

vladak:
Anyway, thanks for the review !

tulinkry:
You are very welcome.

jbaek7023:
@vladak I just want to say I really appreciate the quick turnaround.

Issue #3042, opened 2020-02-18
API endpoints for Annotate, History, file listing


jbaek7023:
Looking at the API documentation for Opengrok here (https://opengrok.docs.apiary.io/),

1) Is this official API documentation? Is this official API documentation? 

I'm building a Frontend web application that can talk with OpenGrok API. 
I see History, Annotate and the file detail page such as http://androidxref.com/9.0.0_r3/xref/frameworks/base/packages/Keyguard/proguard.flags promising. 

2) Does OpenGrok application expose every available feature as API endpoints?

Thanks a lot!

tulinkry:
Hello,

1. Yes, this is official documentation. @vladak it's not versioned right? It contains what is currently in master.
2. No, it does not. The original purpose of the API is to make the system accessible by administrator and the opengrok indexer. Public features include searching and suggesting.

vladak:
>     1. Yes, this is official documentation. @vladak it's not versioned right? It contains what is currently in master.

Yes. In theory it should work both ways - whatever is edited on Apiary gets propagated to the Github repo (https://github.com/oracle/opengrok/blob/master/apiary.apib) and the other way.

vladak:
>     1. Does OpenGrok application expose every available feature as API endpoints?

Well, expose "every feature" would be probably too much :-) What are you after specifically ?

jbaek7023:
@vladak @tulinkry Thanks for the quick turnaround!!! 
I am building a Front-End application (ReactJS) which can interact with OpenGrok API endpoints. 

Here are the descriptions:
**(1) "Search Detail" endpoint** When I click the filename, it opens the file.
![Screenshot from 2020-02-18 02-56-56](https://user-images.githubusercontent.com/23069682/74730314-b941ea00-51fa-11ea-8d31-b0c157bc7bb7.png)

![Screenshot from 2020-02-18 03-00-37](https://user-images.githubusercontent.com/23069682/74730461-00c87600-51fb-11ea-8e36-c37d8a3649bf.png)
example endpoint
```
{
"body": [{"line": 1, "content": "/*"} , {"line":2, "contentL": "* Copyright (C) ....}
"filename": "/android-4.2.2_r1.2/external/webkit/Source/WebKit/chromium/test/CCThreadTaskTest.cpp"
...
]
}
```

**(2) "Annotate" endpoint. When I click the annotation, it returns the annotation with the line**
[...]

PR #3050, opened 2020-02-24
file content/genre API

vladak:
This change implements API endpoints for retrieving file content and genre. There is no paging support.

I refactored the authorization introduced for the history API endpoint so that it can be used via annotations.

Demo:
```
$ curl -D headers -s -X GET 'http://localhost:8080/source/api/v1/file/content?path=opengrok/dev/checkstyle/fileheader.txt' | jq
[
  {
    "line": "/*",
    "number": 1
  },
  {
    "line": " * CDDL HEADER START",
    "number": 2
  },
  {
    "line": " *",
    "number": 3
  },
  {
    "line": " * The contents of this file are subject to the terms of the",
    "number": 4
  },
  {
    "line": " * Common Development and Distribution License (the \"License\").",
    "number": 5
  },
  {
    "line": " * You may not use this file except in compliance with the License.",
    "number": 6
  },
  {
    "line": " *",
    "number": 7
  },
  {
    "line": " * See LICENSE.txt included in this distribution for the specific",
    "number": 8
  },
  {
    "line": " * language governing permissions and limitations under the License.",
    "number": 9
  },
  {
    "line": " *",
    "number": 10
  },
  {
    "line": " * When distributing Covered Code, include this CDDL HEADER in each",
    "number": 11
  },
  {
    "line": " * file and include the License file at LICENSE.txt.",
    "number": 12
  },
  {
    "line": " * If applicable, add the following below this CDDL HEADER, with the",
    "number": 13
  },
  {
    "line": " * fields enclosed by brackets \"[]\" replaced with your own identifying",
    "number": 14
  },
  {
    "line": " * information: Portions Copyright [yyyy] [name of copyright owner]",
    "number": 15
  },
  {
    "line": " *",
    "number": 16
  },
  {
    "line": " * CDDL HEADER END",
    "number": 17
  },
  {
    "line": " */",
    "number": 18
  },
  {
    "line": "",
    "number": 19
  }
]
$ curl -D headers -s -X GET 'http://localhost:8080/source/api/v1/file/genre?path=opengrok/dev/checkstyle/fileheader.txt'
PLAIN
$ curl --head -s -X GET 'http://localhost:8080/source/api/v1/file/content?path=opengrok/testdata/sources/java/main.jar' 
HTTP/1.1 406 
Set-Cookie: JSESSIONID=A9E0888D024C309CBAD44C507A7BBAB0; Path=/source; HttpOnly
Content-Type: text/html;charset=utf-8
Content-Language: en
Content-Length: 1212
Date: Mon, 24 Feb 2020 11:17:16 GMT

$ curl -s -X GET 'http://localhost:8080/source/api/v1/file/genre?path=opengrok/testdata/sources/java/main.jar' 
XREFABLE
```


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        Document doc;
+        if ((doc = getDocument(file)) == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot get document for file");
+            return null;
+        }
+
+        String fileType = doc.get(QueryBuilder.T);
+        if (!fileType.equals(AbstractAnalyzer.Genre.PLAIN.typeName())) {
```
idodeclare: Recommend transposing to avoid possible NPE:

`!AbstractAnalyzer.Genre.PLAIN.typeName().equals(fileType)`

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+                           @QueryParam("path") final String path) throws IOException, ParseException {
+
+        File file = getFile(path, response);
+        if (file == null) {
+            return null;
+        }
+
+        Document doc;
+        if ((doc = getDocument(file)) == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot find document for file");
```
idodeclare: Line 121 reads "Cannot get document for file"

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+    @PathAuthorized
+    @Path("/genre")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getGenre(@Context HttpServletRequest request,
+                           @Context HttpServletResponse response,
+                           @QueryParam("path") final String path) throws IOException, ParseException {
+
+        File file = getFile(path, response);
+        if (file == null) {
+            return null;
```
idodeclare: Maybe copy comment from line 115 that was helpful there?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        Document doc;
+        if ((doc = getDocument(file)) == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot find document for file");
+            return null;
+        }
+
+        return AbstractAnalyzer.Genre.get(doc.get(QueryBuilder.T)).toString();
```
idodeclare: Possibly `AbstractAnalyzer.Genre.get(doc.get(QueryBuilder.T))` could be null, leading to NPE on `toString()`

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1588,6 +1611,7 @@ public static Definitions getDefinitions(File file)
             return null;
```
idodeclare: Comment above seems inapt with this refactoring

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1588,6 +1611,7 @@ public static Definitions getDefinitions(File file)
...
         }
 
+        Document doc;
```
idodeclare: Maybe move the `return` as the last statement of the try, so this declaration and assignment can also be moved together?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1596,22 +1620,18 @@ public static Definitions getDefinitions(File file)
                 // No hits, no definitions...
                 return null;
             }
-            Document doc = searcher.doc(top.scoreDocs[0].doc);
+            doc = searcher.doc(top.scoreDocs[0].doc);
             String foundPath = doc.get(QueryBuilder.PATH);
 
             // Only use the definitions if we found an exact match.
```
idodeclare: Comment seems inapt with this refactoring

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1596,22 +1620,18 @@ public static Definitions getDefinitions(File file)
                 // No hits, no definitions...
```
idodeclare: Comment seems inapt with this refactoring

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+import static org.opengrok.indexer.index.IndexDatabase.getDocument;
+
+@Path(FileController.PATH)
+public class FileController {
+
+    public static final String PATH = "/file";
+
+    private static final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    private ArrayList<String> lines = new ArrayList<>();
```
idodeclare: Any benefit to defining this here when it's only used by `/content` and not by `/genre`?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/FileControllerTest.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.ws.rs.core.Application;
+import javax.ws.rs.core.GenericType;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static org.junit.Assert.assertEquals;
+
+public class FileControllerTest extends JerseyTest {
```
idodeclare: Any benefit to testing conditions that result in 400s or 404s?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+
+        return file;
+    }
+
+    @GET
+    @CorsEnable
+    @PathAuthorized
+    @Path("/content")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Object getContent(@Context HttpServletRequest request,
```
idodeclare: Not sure the utility of a web method that just splits a file using standard Java `readline()` and returns it encoded in JSON. Why not just return the raw content more efficiently?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+
+        public int getNumber() {
+            return this.number;
+        }
+    }
+
+    private static File getFile(String path, HttpServletResponse response) throws IOException {
+        if (path == null) {
+            if (response != null) {
+                response.sendError(HttpServletResponse.SC_NOT_ACCEPTABLE, "Missing path parameter");
```
tulinkry: not acceptable is related to content negotiation headers, this is more likely a bad request

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+
+        Document doc;
+        if ((doc = getDocument(file)) == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot get document for file");
+            return null;
+        }
+
+        String fileType = doc.get(QueryBuilder.T);
+        if (!fileType.equals(AbstractAnalyzer.Genre.PLAIN.typeName())) {
+            response.sendError(HttpServletResponse.SC_NOT_ACCEPTABLE, "Not a text file");
```
tulinkry: you could employ the accept header here and offer the file
1. as json when application/json
2. as plain text when text/plain
3. as binary data when application/octet-stream

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+            return this.line;
+        }
+
+        public int getNumber() {
+            return this.number;
+        }
+    }
+
+    private static File getFile(String path, HttpServletResponse response) throws IOException {
+        if (path == null) {
```
tulinkry: this can't happen given you have the filter always before this controller, right?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,169 @@
...
+            while ((line = br.readLine()) != null) {
+                lines.add(line);
+            }
+        }
+
+        int count = 1;
+        List<LineDTO> linesDTO = new ArrayList<>();
+        for (String line: lines) {
+            LineDTO l = new LineDTO(line, count++);
+            linesDTO.add(l);
```
tulinkry: I see creating one unnecessary array list `lines`, or am I missing something?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/FileControllerTest.java
vladak: The problem is that Grizzly used by JerseyTest does not set neither `request` or `response`.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
vladak: I was wondering about this myself. This was originally suggested in #3042 by @jbaek7023 , perhaps there is some value in that for him.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
vladak: It seems during tests the filter is not used, possibly because the context is not set by Grizzly.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
vladak: I think @tulinkry's suggestion to use accept headers makes sense.

vladak:
> Could you test with curl also on escaped path parameter? `file/content?path=opengrok%2Fsrc%2Fmain%2Ffile`

works fine.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
vladak: Good idea. I did not want to access the request for the header (since it is `null` in `JerseyTest`) so there is a bit of code duplication.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
idodeclare: If it’s returning a split meant to align with OpenGrok’s analysis, then it should be using `SourceSpllitter` with any necessary project-configured `ExpandTabsReader` via a `ReaderWrapper`. (There is a TODO to comprehensively accommodate the various line/paragraph/page white space separators. E.g. see Erlang.)

Otherwise I recommend to defer the JSON format and let the client handle the raw source text on its own.

In this patch it’s only `readline()`, and that is not what OpenGrok uses. Also `readline()` does not preserve end of line space, so the sum of the lines will not equal the original document, and the client will have no way to know original line endings. (With some languages like Erlang they could be varied in the original.)

vladak: True, JSON removed.

tulinkry: > and the client will have no way to know original line endings

Not true if implemented with content negotiation for both raw format and json format. When the client cares, he can choose the raw format. When he does not, he just likes to have a list of lines, he chooses json with split lines.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -0,0 +1,172 @@
...
+            return null;
+        }
+
+        Document doc;
+        if ((doc = getDocument(file)) == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot get document for file");
+            return null;
+        }
+
+        return transfer(file);
```
tulinkry: when this throws FileNotFoundException, it should possibly answer with 404

PR #3053, opened 2020-02-25
Feature/redirect 1 result

idodeclare:
Hello,

Please consider for integration this patch to redirect directly to a file when a search produces a single hit.

* Fix existing but broken functionality to automatically link to a definition from an xref file (e.g. `#function1`) when the definition is the single search result. The previous functionality used only an HTML fragment identifier, which was broken when list.jsp was updated to redirect so as to include a revision ID in the URL (e.g. `?r=04f4d4a`). As a fragment identifier is client-side only, it is lost in a redirect without special handling (i.e. a mediating parameter, `?fi=function1`).
* Add general support for redirecting directly to a file when a search produces a single hit. A mediating parameter for a match offset (e.g. `?mo=5733`) is used so that an offset can be translated to a fragment identifier (e.g. `?fi=23`) while still allowing the revision ID production (e.g. `?r=04f4d4a`).
* ~Add `QueryParameters` class, and extract named literals for OpenGrok's web query parameters.~ Add several literals to support this patch, including `FRAGMENT_IDENTIFIER` and `MATCH_OFFSET`.
* Share some duplicated code between `SourceSplitter` and `LineBreaker`, and share logic to convert an offset to a line index. This is needed for translating e.g. `?mo=5733` to `?fi=23`.
* Optimize the `Context` class to avoid creating some large allocations when not usually needed.
* Revise some OpenGrok-internal APIs to rationalize some inconsistencies with the use of "offset" and "position". "Offset" should be used to accord with Lucene's definition as the number of characters from the beginning of a document. "Position" means an ordinal designation among lists of tokens where the token list matches the order in source code. "Index" can be used to mean ordinal designation in a general collection.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
@@ -20,7 +20,7 @@
 /*
  * Copyright (c) 2005, 2019, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright 2011 Jens Elkner.
- * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.search.context;
 
@@ -56,12 +56,13 @@ import org.opengrok.indexer.web.Util;
  */
 public class Context {
 
+    static final int MAXFILEREAD = 1024 * 1024;
+
     private static final Logger LOGGER = LoggerFactory.getLogger(Context.class);
 
     private final Query query;
     private final QueryBuilder qbuilder;
     private final LineMatcher[] m;
-    static final int MAXFILEREAD = 1024 * 1024;
     private char[] buffer;
     PlainLineTokenizer tokens;
     String queryAsURI;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
@@ -63,7 +63,7 @@ public class Context {
     private final Query query;
     private final QueryBuilder qbuilder;
     private final LineMatcher[] m;
-    String queryAsURI;
+    private final String queryAsURI;
 
     /**
      * Map whose keys tell which fields to look for in the source file, and
@@ -95,7 +95,9 @@ public class Context {
         QueryMatchers qm = new QueryMatchers();
         m = qm.getMatchers(query, TOKEN_FIELDS);
         if (m != null) {
-            buildQueryAsURI(qbuilder.getQueries());
+            queryAsURI = buildQueryAsURI(qbuilder.getQueries());
+        } else {
+            queryAsURI = "";
         }
     }
 
@@ -232,10 +234,9 @@ public class Context {
      *
      * @param subqueries a map containing the query text for each field
      */
-    private void buildQueryAsURI(Map<String, String> subqueries) {
+    private String buildQueryAsURI(Map<String, String> subqueries) {
         if (subqueries.isEmpty()) {
-            queryAsURI = "";
-            return;
+            return "";
         }
         StringBuilder sb = new StringBuilder();
         for (Map.Entry<String, String> entry : subqueries.entrySet()) {
@@ -245,7 +246,7 @@ public class Context {
                 .append('&');
         }
         sb.setLength(sb.length() - 1);
-        queryAsURI = sb.toString();
+        return sb.toString();
     }
 
     private boolean alt = true;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
@@ -63,7 +63,6 @@ public class Context {
     private final Query query;
     private final QueryBuilder qbuilder;
     private final LineMatcher[] m;
-    private char[] buffer;
     PlainLineTokenizer tokens;
     String queryAsURI;
 
@@ -99,7 +98,6 @@ public class Context {
         if (m != null) {
             buildQueryAsURI(qbuilder.getQueries());
             //System.err.println("Found Matchers = "+ m.length + " for " + query);
-            buffer = new char[MAXFILEREAD];
             tokens = new PlainLineTokenizer((Reader) null);
         }
     }
@@ -213,7 +211,7 @@ public class Context {
 
         try {
             List<String> fieldList = qbuilder.getContextFields();
-            String[] fields = fieldList.toArray(new String[fieldList.size()]);
+            String[] fields = fieldList.toArray(new String[0]);
 
             String res = uhi.highlightFieldsUnion(fields, query, docId,
                 linelimit);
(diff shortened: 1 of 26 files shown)


vladak:
rebase please

idodeclare:
Rebased with straight-forward resolution for Util.java, which @vladak and I both modified.

tarzanek:
from a quick look it looks OK
I'd prefer to have separate PR for the refactoring of constants, but I guess I can live with this
merge?

vladak:
I did take a cursory look but wanted to understand it in a bit more 
detail before it is merged.

idodeclare:
> I'd prefer to have separate PR for the refactoring of constants....

No problem, @tarzanek . I raised that PR.

idodeclare:
> I'd prefer to have separate PR for the refactoring of constants...

@tarzanek , @vladak approved and merged that PR. (Thank you again, @vladak )

tulinkry:
> Add general support for redirecting directly to a file when a search produces a single hit

This is number 1. Can't wait for release.

tarzanek:
thank you Chris & Vlad

opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
```diff
@@ -212,7 +210,7 @@ public boolean getContext2(RuntimeEnvironment env, IndexSearcher searcher,
 
         try {
             List<String> fieldList = qbuilder.getContextFields();
-            String[] fields = fieldList.toArray(new String[fieldList.size()]);
+            String[] fields = fieldList.toArray(new String[0]);
```
tarzanek: Why 0?
Above looked like optimization so toArray won't have to allocate new Array, since we know how many elements we will have.
I think if 0 is correct, then in case of above you could just directly call .toArray() with no params

opengrok-indexer/src/main/java/org/opengrok/indexer/util/LineBreaker.java
```diff
@@ -65,65 +63,52 @@ public void reset(StreamSource src, ReaderWrapper wrapper)
...
-            switch (c) {
-                case '\r':
-                    c = reader.read();
-                    if (c == -1) {
-                        newOffsets.add(length);
-                        break;
-                    } else {
+
+            redo_c:
+            while (true) {
```
tarzanek: seriously?
do we want to go for endless cycles with goto?
back to Basic? :-D

(yes, it's easy for me to complain, but this raises eyebrows, so there has to be a better way)
(the first thing that came to my mind was a jflex lexer, but not sure if we don't have the same problem with line # there too)

tarzanek: couldn't resist to ping @kahatlen here as the EOL expert ;-)
(hey Knuth, how are you these days?)

opengrok-indexer/src/main/java/org/opengrok/indexer/util/SourceSplitter.java
```diff
@@ -189,74 +150,49 @@ public void reset(StreamSource src, ReaderWrapper wrapper)
...
-                                bld.setLength(0);
-                                break;
-                            default:
-                                slist.add(bld.toString());
-                                bld.setLength(0);
 
-                                bld.append((char) c);
-                                break;
+            redo_c:
+            while (true) {
```
tarzanek: same as above
-1 for endless cycles

tarzanek:
this looks ok refactoring wise and redir wise
I don't like the endless cycles however, the concern here is the endless loading of a web page (where I'd rather see a weird error than a page not replying forever in case something goes wrong)

tarzanek:
above said, thank you for the fixe and for the refactoring, this made reviewing so much easier, especially for bigger diff as this one

opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
idodeclare: IntelliJ recommends 0

tulinkry: I remember @ahornace telling me that there is no optimization on using size and 0.

ahornace: Taken directly from idea:
```
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).
```

`.toArray(T[] t)` is necessary so JVM can create array of proper type because of type erasure

opengrok-indexer/src/main/java/org/opengrok/indexer/util/LineBreaker.java
idodeclare: > seriously?
> do we want to go for endless cycles with goto?
> back to Basic? :-D
> 
> (yes, it's easy for me to complain, but this raises eyebrows, so there has to be a better way)

You've never hand-written a scanner? 😄 But you're right it's generally not a good idea; imagine code reviewing a hand-written `PerlXref`. But this is a teeny one, and look at that clear `break` at the end of the inner.... (Anyway search for `while (true)` in opengrok....)

> (the first thing that came to my mind was a jflex lexer, but not sure if we don't have the same problem with line # there too)

I wrote these classes originally when it was a big pain to add additional .lex files, especially in packages where they weren't already used. But now we can add them anywhere easily (e.g. `XrefSourceTransformer`). Also I confess if I had used JFlex originally I wouldn't have had the deficient handling that I'm updating here....

How about if we decide at some point to update OpenGrok `STANDARD_EOL` as we probably should do, then I rewrite these to use JFlex? But for now I do think I'm updating and testing sufficiently in this patch.

idodeclare: OK I spent some time and replaced with a couple of new JFlex scanners

opengrok-indexer/src/main/java/org/opengrok/indexer/search/context/Context.java
tarzanek: ok, so this means javadocs suck, I mean, they are outdated :-)

tarzanek:
hehe, so if there are more while(true) in opengrok I hope my name is next to them :-D
will figure out what is STANDARD_EOL (my last memories are how Knut revamped the EOL handling a bit, which made a mess around xrefers and lexers :-( )

tarzanek:
thank you Chris, you're the jflex master!
from my side this looks OK, @vladak @ahornace ?

tarzanek:
thank you Chris, you're the jflex master!
from my side this looks OK, @vladak @ahornace ?

idodeclare:
> thank you Chris, you're the jflex master!
> from my side this looks OK....

Thank you, Lubos!

idodeclare:
While coding the speedup of SOL, I noticed in code the `#sbox` auto-submit that OpenGrok does when you × a project on the search page. I hadn't realized that behavior existed.

I tested some, and it is jarring when this patch redirects to a single search result when all you've done is × a project. Let me try for a workaround.

idodeclare:
Rebased first on master

idodeclare:
[Pushed fixes](https://github.com/oracle/opengrok/compare/4a68f31..01dfb76) for the unexpected redirect described above and also for a `docID` adjustment that is sometimes necessary depending on Lucene `LeafReaderContext`.

tarzanek:
lgtm, thanks for the compressed fix
@vladak merge?

tulinkry:
what about merging?

tulinkry:
@vladak

tarzanek:
let me merge, that said I am quite sure I won't have time to fix any issues that will arise from this merge this week :-(

idodeclare:
Thank you, Lubos!

PR #3056, opened 2020-02-26
Annotation API

vladak:
This change add simple annotation API.

Limitations:
  - no support for paging
  - description is HTML encoded

While testing I found out that Git annotations are broken so fix for that is included.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -6,6 +6,31 @@ OpenGrok RESTful API documentation. The following endpoints are accessible under
 
 Besides `/suggester` and `/search` endpoints, everything is accessible from `localhost` only.
 
+## Annotation [/annotation{?path}]
+
+### Get annotation for a file [GET]
+
++ Parameters
+  + path (string) - path of file, relative to source root
+
++ Response 200 (application/json)
+  + Body
+
+            [
+              {
+                "revision": "c55d5891",
+                "author": "Adam Hornáček",
+                "description": "changeset: c55d5891\nsummary: Rewrite README.txt to use markdown syntax\nuser: Adam Hornáček <adam.hornacek@oracle.com>\ndate: Wed Aug 30 17:42:12 CEST 2017",
+                "version": "1/15"
+              },
+              {
+                "revision": "5e0c6b22",
+                "author": "Vladimir Kotal",
+                "description": "changeset: 5e0c6b22\nsummary: bump year\nuser: Vladimir Kotal <vlada@devnull.cz>\ndate: Thu Jul 18 14:43:01 CEST 2019",
+                "version": "14/15"
+              }
+            ]
+
 ## Authorization framework reload [/configuration/authorization/reload]
 
 ### reloads authorization framework [POST]
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
@@ -28,7 +28,6 @@ import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.Color;
 import org.opengrok.indexer.util.LazilyInstantiate;
 import org.opengrok.indexer.util.RainbowColorGenerator;
-import org.opengrok.indexer.web.Util;
 
 import java.io.IOException;
 import java.io.StringWriter;
@@ -53,7 +52,7 @@ public class Annotation {
     private static final Logger LOGGER = LoggerFactory.getLogger(Annotation.class);
 
     private final List<Line> lines = new ArrayList<>();
-    private final Map<String, String> desc = new HashMap<>();
+    private final Map<String, String> desc = new HashMap<>(); // revision to description
     private final Map<String, Integer> fileVersions = new HashMap<>(); // maps revision to file version
     private final LazilyInstantiate<Map<String, String>> colors = LazilyInstantiate.using(this::generateColors);
     private int widestRevision;
@@ -163,7 +162,7 @@ public class Annotation {
     }
 
     void addDesc(String revision, String description) {
-        desc.put(revision, Util.encode(description));
+        desc.put(revision, description);
     }
 
     public String getDesc(String revision) {
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -389,7 +389,6 @@ public class GitRepository extends Repository {
                 ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK),
                 "rev-list",
                 "--reverse",
-                "--max-count=1",
                 "HEAD",
                 "--",
                 fullpath
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -47,6 +47,7 @@ import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.BufferSink;
 import org.opengrok.indexer.util.Executor;
+import org.opengrok.indexer.util.HeadHandler;
 import org.opengrok.indexer.util.StringUtils;
 import org.opengrok.indexer.util.Version;
 
@@ -396,23 +397,17 @@ public class GitRepository extends Repository {
 
         Executor executor = new Executor(Arrays.asList(argv), new File(getDirectoryName()),
                 RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
-        int status = executor.exec();
+        HeadHandler headHandler = new HeadHandler(1);
+        int status = executor.exec(false, headHandler);
 
-        try (BufferedReader in = new BufferedReader(
-                new InputStreamReader(executor.getOutputStream()))) {
-            String line;
-
-            if ((line = in.readLine()) != null) {
-                return line.trim();
-            }
+        String line;
+        if (headHandler.count() > 0 && (line = headHandler.get(0)) != null) {
+            return line.trim();
         }
 
-        if (status != 0) {
-            LOGGER.log(Level.WARNING,
-                    "Failed to get first revision for: \"{0}\" Exit code: {1}",
-                    new Object[]{fullpath, String.valueOf(status)});
-            return null;
-        }
+        LOGGER.log(Level.WARNING,
+                "Failed to get first revision for: \"{0}\" Exit code: {1}",
+                new Object[]{fullpath, String.valueOf(status)});
 
         return null;
     }
@@ -449,10 +444,10 @@ public class GitRepository extends Repository {
         cmd.add("--");
         cmd.add(getPathRelativeToCanonicalRepositoryRoot(file.getCanonicalPath()));
 
-        Executor exec = new Executor(cmd, new File(getDirectoryName()),
+        Executor executor = new Executor(cmd, new File(getDirectoryName()),
                 RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
         GitAnnotationParser parser = new GitAnnotationParser(file.getName());
-        int status = exec.exec(true, parser);
+        int status = executor.exec(true, parser);
 
         // File might have changed its location if it was renamed.
         // Try to lookup its original name and get the annotation again.
@@ -468,10 +463,10 @@ public class GitRepository extends Repository {
             }
             cmd.add("--");
             cmd.add(findOriginalName(file.getCanonicalPath(), revision));
-            exec = new Executor(cmd, new File(getDirectoryName()),
+            executor = new Executor(cmd, new File(getDirectoryName()),
                     RuntimeEnvironment.getInstance().getInteractiveCommandTimeout());
             parser = new GitAnnotationParser(file.getName());
-            status = exec.exec(true, parser);
+            status = executor.exec(true, parser);
         }
 
         if (status != 0) {
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
@@ -1,0 +1,109 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ */
+
+package org.opengrok.indexer.util;
+
+import java.io.BufferedInputStream;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The purpose of this class is to provide {@code StreamHandler} that limits the output
+ * to specified number of lines. Compared to {@code SpoolHandler} it consumes
+ * limited amount of heap.
+ */
+public class HeadHandler implements Executor.StreamHandler {
+    private final int maxLines;
+
+    private final List<String> lines = new ArrayList<>();
+    private final Charset charset;
+
+    private static final int bufferedReaderSize = 200;
+
+    /**
+     * Charset of the underlying reader is set to UTF-8.
+     * @param maxLines maximum number of lines to store
+     */
+    public HeadHandler(int maxLines) {
+        this.maxLines = maxLines;
+        this.charset = StandardCharsets.UTF_8;
+    }
+
+    /**
+     * @param maxLines maximum number of lines to store
+     * @param charset character set
+     */
+    public HeadHandler(int maxLines, Charset charset) {
+        this.maxLines = maxLines;
+        this.charset = charset;
+    }
+
+    /**
+     * @return number of lines read
+     */
+    public int count() {
+        return lines.size();
+    }
+
+    /**
+     * @param index index
+     * @return line at given index. Will be non {@code null} for valid index.
+     */
+    public String get(int index) {
+        return lines.get(index);
+    }
+
+    // for testing
+    static int getBufferedReaderSize() {
+        return bufferedReaderSize;
+    }
+
+    @Override
+    public void processStream(InputStream input) throws IOException {
+        try (BufferedInputStream bufStream = new BufferedInputStream(input);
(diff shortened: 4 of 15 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -24,6 +24,7 @@
 
 package org.opengrok.indexer.history;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
```
tulinkry: I don't see this being used in this changeset, was it a missing import?

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.File;
+import java.io.IOException;
+
+public class FileUtil {
+
+    private static final RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+
+    public static File getFile(String path, HttpServletResponse response) throws IOException {
```
tulinkry: Could you refactor this function to throw related exceptions and map these exceptions in exception mapper? It would make it obvious and without side effects. If I imported this function from a library I would not know what it does ... it takes a path and a response for some reason and returning a file.

Also a javadoc on public util method would not hurt.

apiary.apib
```diff
@@ -6,6 +6,31 @@ OpenGrok RESTful API documentation. The following endpoints are accessible under
...
+
++ Parameters
+  + path (string) - path of file, relative to source root
+
++ Response 200 (application/json)
+  + Body
+
+            [
+              {
+                "number": 1,
```
tulinkry: what is the number?

apiary.apib
```diff
@@ -6,6 +6,31 @@ OpenGrok RESTful API documentation. The following endpoints are accessible under
...
+                "revision": "e2736ff6",
+                "author": "Vladimir Kotal",
+                "desc": "changeset:&nbsp;e2736ff6&lt;br/&gt;summary:&nbsp;add&nbsp;microbadges&lt;br/&gt;user:&nbsp;Vladimir&nbsp;Kotal&nbsp;&#60;vlada@devnull.cz&#62;&lt;br/&gt;date:&nbsp;Fri&nbsp;Jun&nbsp;28&nbsp;21:33:41&nbsp;CEST&nbsp;2019"
```
tulinkry: can have long naming "description"

apiary.apib
```diff
@@ -6,6 +6,31 @@ OpenGrok RESTful API documentation. The following endpoints are accessible under
...
+              },
```
tulinkry: we also utilize the version number in all version of the file (like version 4 out of 22), can you include that as well?

apiary.apib
tulinkry: I wouldn't expect a html code inside json in a description field, is it easy to get rid of it?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: yes

apiary.apib
vladak: just entry number, I guess that's not necessary.

apiary.apib
vladak: I don't understand. Could you rephrase or provide example ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -389,7 +389,6 @@ private String getFirstRevision(String fullpath) throws IOException {
                 ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK),
                 "rev-list",
                 "--reverse",
-                "--max-count=1",
```
idodeclare: No big problem here but some food for thought:

* `Executor.exec()` and `Executor.exec(boolean)` actually spool the entire output of a program into memory. Probably that's not a burdensome amount with this removal of `--max-count`, given this is a `rev-list` of a single file.
* When the spooled bytes are read later at line 402, they are read with the default Java charset, which is locale dependent on the OS installation. For hashes of a `rev-list`, that is perhaps always OK — but I'm just remarking since elsewhere in `GitRepository`, specific effort is made to use a correct charset (see `newLogReader()`).
* Technically when read later at line 402, a `BufferedReader` is overkill, given that the bytes are already fully spooled in-memory.
* Indeed all the output of a `Process` (via `exec()`) must be read or else a deadlock could happen as one awaits the process exit value. The default spool does consume everything.

I wonder if this is an opportunity to create a `HeadHandler` implementation of `Executor.StreamHandler` that is line-oriented (e.g. characters out of bytes), that demands a specified encoding instead of allowing the default Java charset, and that reads and stores a maximum number of lines and then just consumes and discards bytes (without bothering with character conversion) to exhaust the stream and avoid allocating heap.

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.opengrok.web.util;
+
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.File;
+import java.io.IOException;
+
+public class FileUtil {
```
idodeclare: Should the constructor be made private for a static utility class?

vladak: yep, actually the `checkstyle` complains about it.

apiary.apib
tulinkry: Snímek obrazovky 2020-02-27 v 17 19 50

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
```diff
@@ -0,0 +1,73 @@
...
+     * @throws NoPathException
+     */
+    public static File getFile(String path, HttpServletResponse response) throws NoPathException, NoFileException {
+        if (path == null) {
+            throw new NoPathException("Missing path parameter");
+        }
+
+        File file = new File(env.getSourceRootFile(), path);
+        if (!file.isFile()) {
+            throw new NoFileException("File not found");
```
tulinkry: There is already FileNotFound (or similar) in java core

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    /**
+     * @param path path relative to source root
+     * @return file object corresponding to the file under source root
+     * @throws NoFileException
+     * @throws NoPathException
+     */
+    public static File getFile(String path, HttpServletResponse response) throws NoPathException, NoFileException {
```
tulinkry: I think I wrote it somewhere else too, I'm little bit against this naming. But if you feel comfortable, you can disregard.

tulinkry: response is not needed

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
tulinkry: Definitely good idea, I'd go for LineHandler (or similar) and then derive the HeadHandler out of it. For simplicity a lot of code could benefit from this.

idodeclare: Thank you, @tulinkry .

Maybe the reverse though: a `LineHandler` might be an unbounded `HeadHandler`. The latter would need to interact low-level with the stream in order to be able to read and dump bytes when a line limit is reached.

apiary.apib
tulinkry: you can include it but give it better name :-D

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
vladak: toFile() ?

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
vladak: aha ! thanks.

apiary.apib
vladak: Oh that. I never noticed it before :-)

opengrok-web/src/main/java/org/opengrok/web/util/FileUtil.java
vladak: Makes sense, thanks for pointing out the existence of exception mappers.

apiary.apib
vladak: it would make sense to have it if pagination was supported.

apiary.apib
vladak: it turns out it is.

opengrok-web/src/main/java/org/opengrok/web/api/v1/NoPathExceptionMapper.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class NoPathExceptionMapper implements ExceptionMapper<NoPathParameterException> {
+    @Override
+    public Response toResponse(NoPathParameterException e) {
+        return Response.status(Response.Status.NOT_ACCEPTABLE).build();
```
tulinkry: https://github.com/oracle/opengrok/pull/3050#discussion_r383695289

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: Thanks guys. It occurred to me that this might be a problem but pushed the resolution aside. Will look into it.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -163,10 +163,14 @@ void addLine(String revision, String author, boolean enabled) {
...
     void addDesc(String revision, String description) {
-        desc.put(revision, Util.encode(description));
+        desc.put(revision, description);
     }
 
     public String getDesc(String revision) {
+        return Util.encode(desc.get(revision));
+    }
+
+    public String getDescRaw(String revision) {
```
idodeclare: Any possibility to make this class non-HTML aware and move the encoding to `-web` or `.web`?

vladak: Sure; thought the `getDesc()` is used more widely.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -713,82 +713,87 @@ public static void readableLine(int num, Writer out, Annotation annotation, Stri
...
+        if (enabled) {
+            out.write(anchorClassStart);
+            out.write("r");
+            out.write("\" style=\"background-color: ");
+            out.write(annotation.getColors().getOrDefault(r, "inherit"));
+            out.write("\" href=\"");
+            out.write(URIEncode(annotation.getFilename()));
+            out.write("?a=true&amp;r=");
+            out.write(URIEncode(r));
+            String msg = Util.encode(annotation.getDesc(r));
```
idodeclare: Given the pre-existing `if (msg != null)` check nearby below, should perhaps the new `Util.encode()` be done in that if-block?

vladak: definitely. Should have done the testing. Thanks.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: The latest changes have the basic `HeadHandler` (sans encoding).

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final List<String> lines = new ArrayList<>();
+
+    // TODO add encoding
+    public HeadHandler(int numlines) {
+        this.numlines = numlines;
+    }
+
+    public List<String> getLines() {
+        return lines;
```
idodeclare: I recommend to avoid breaking encapsulation (though it's rampant in OpenGrok). E.g. rather than returning a private list, how about:

```
public int count() {
...
}

public String get(int index) {
...
}
```
```

File: opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java [hunk 14]
```diff
@@ -0,0 +1,73 @@
...
+            if (line == null) { // EOF
+                return;
+            }
+            lines.add(line);
+            linenum++;
+        }
+        // Read and forget the rest. Need to use read() since it is not guaranteed
+        // that newline will ensue.
+        // TODO: would readline() get the job done ?
+        while (reader.read() != -1) {
```
idodeclare: Technically this is doing needless `byte` to `char` conversion. E.g. were it a multi-byte encoding, then decoding is not worthwhile.

My original thinking on `HeadHandler` was to have the principal buffering at the byte-level. E.g. a `BufferedReader(InputStreamReader, /*tiny*/200)` atop a `BufferedInputStream`. Then when all the needed `readLine()` calls are done to just `read()` and discard from the `BufferedInputStream` — probably not a byte at a time but with say a very small `byte[1024]` buffer.

(Using `readLine()` here would defeat the goal of avoiding needless heap allocation.)

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
vladak: I must admit OpenGrok code base has been primary source of my Java programming influence :-)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -389,23 +390,19 @@ private String getFirstRevision(String fullpath) throws IOException {
...
-                new InputStreamReader(executor.getOutputStream()))) {
-            String line;
+        HeadHandler headHandler = new HeadHandler(1);
+        int status = executor.exec(false, headHandler);
 
-            if ((line = in.readLine()) != null) {
-                return line.trim();
-            }
+        String line;
+        if ((line = headHandler.get(0)) != null) {
```
idodeclare: I think the API requires `if (headHandler.count() > 0) {`

vladak: I see, for some reason I thought it will return `null`.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
vladak: I see. Now, `BufferedReader.read()` will not read a single byte from the underlying stream - it will call `fill()` to fill the internal buffer.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -389,31 +390,24 @@ private String getFirstRevision(String fullpath) throws IOException {
...
-                new InputStreamReader(executor.getOutputStream()))) {
-            String line;
+        HeadHandler headHandler = new HeadHandler(1);
+        int status = executor.exec(false, headHandler);
 
-            if ((line = in.readLine()) != null) {
-                return line.trim();
-            }
+        String line;
+        if (headHandler.size() > 0 && (line = headHandler.get(0)) != null) {
```
idodeclare: Nit: I think your implementation does ensure non-null for a valid index. Maybe that could be stated firmly in the Javadoc?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
vladak: Anyhow, the suggested layering sounds good so I have implemented that, together with charset choice.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return line at given index
+     */
+    public String get(int i) {
+        return lines.get(i);
+    }
+
+    @Override
+    public void processStream(InputStream input) throws IOException {
+        BufferedInputStream bufStream = new BufferedInputStream(input);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(bufStream, this.charset), 200);
```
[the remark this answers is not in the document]
vladak: definitely. Wanted to do it but then forgot.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HeadHandler.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+            lines.add(line);
+            lineNum++;
+        }
+
+        // Read and forget the rest.
+        byte[] buf = new byte[1024];
+        int off = 0;
+        int len;
+        while ((len = bufStream.read(buf, off, buf.length)) != -1) {
```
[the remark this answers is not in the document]
vladak: okay, sounds reasonable.

vladak:
> LGTM

Thanks, this was nice learning experience.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/HeadHandlerTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+        RandomInputStream rndStream = new RandomInputStream(totalCount, lines);
+        HeadHandler handler = new HeadHandler(headLines);
+        assertTrue(totalCount > handler.getBufferedReaderSize());
+        handler.processStream(rndStream);
+        assertEquals(headLines, handler.size());
+        assertEquals(totalCount + 1, rndStream.getCount());
+        int linesLen = 0;
+        for (int i = 0; i < headLines; i++) {
+            linesLen += handler.get(i).length();
```
idodeclare: With this approach using randomness, I wonder if it would be informative to compare the line content to what was generated.

tulinkry: For me it seems a little much work for a test; I'd prepare some files as resources (perhaps with this generator) and assert the content.

vladak: I wanted the test to adapt automatically based on changes to `HeadHandler` - in particular the internal buffer size.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/HeadHandlerTest.java
```diff
@@ -0,0 +1,91 @@
...
+            letters = letters + "0123456789";
+        }
+
+        public int getCount() {
+            return count;
+        }
+
+        @Override
+        public int read() throws IOException {
+            if (++count < maxcount) {
```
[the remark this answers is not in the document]
vladak: yes. was rough.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/HeadHandlerTest.java
```diff
@@ -0,0 +1,91 @@
...
+                // Want the newlines to appear within the first half.
+                if (count % ((maxcount / 2) / maxlines) == 0) {
```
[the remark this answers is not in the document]
vladak: good idea. I should randomize the test cases' parameters too.

PR #3059, opened 2020-02-28
do not ignore directories starting with . when scanning for projects

vladak:
This change allows projects to have any name that can be represented on the file system.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright 2011 Jens Elkner.
  * Portions Copyright (c) 2017-2019, Chris Fraire <cfraire@me.com>.
  */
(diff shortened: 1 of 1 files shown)


idodeclare:
As I wrote on the linked issue, I'm not sure it was correctly diagnosed, so I recommend holding off on this change.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -946,7 +946,7 @@ public void prepareIndexer(RuntimeEnvironment env,
                     p.setName(name);
                     p.completeWithDefaults();
                     projects.put(name, p);
-                } else if (!name.startsWith(".") && file.isDirectory()) {
+                } else if (file.isDirectory()) {
```
tulinkry: what about `.` and `..` ?

vladak: https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles() says:
> an array of File objects is returned, one for each file or directory in the directory. Pathnames denoting the directory itself and the directory's parent directory are not included in the result.

vladak:
> As I wrote on the linked issue, I'm not sure it was correctly diagnosed, so I recommend holding off on this change.

removed the issue reference in the changeset.

vladak:
I hope this is non-controversial so could be merged.

idodeclare:
It's controversial because the handling is a deliberate and because it seems the genesis of this patch was a misdiagnosed issue. Who truly desires this to be removed? Could some user be broken for no good benefit?

vladak:
Okay, I don't feel strongly about this so closing.

PR #3061, opened 2020-03-04
Bugfix/perforce

idodeclare:
Hello,

Please consider for integration this patch to optimize Perforce executions to get history data.

* Run a `p4 changes` and `p4 filelog`, and integrate the results, so only two p4 commands are needed for `History` of a directory tree.
* Parse file names from p4 commands so that `HistoryEntry` instances have defined files.
* Recognize binary file log for annotations.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
@@ -20,9 +20,13 @@
 /*
  * Copyright (c) 2008, 2018, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2019, Chris Ross <cross@distal.com>.
+ * Portions Copyright (c) 2020, Chris Fraire <cfraire@me.com>.
  */
 
 package org.opengrok.indexer.history;
+
+import static org.opengrok.indexer.history.PerforceRepository.protectPerforceFilename;
+import static org.opengrok.indexer.history.PerforceRepository.unprotectPerforceFilename;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -31,6 +35,7 @@ import java.io.Reader;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
```diff
@@ -40,36 +41,38 @@
...
-@ConditionalRun(RepositoryInstalled.PerforceInstalled.class)
 public class PerforceRepositoryTest {
 
     @Rule
     public ConditionalRunRule rule = new ConditionalRunRule();
     
     private static boolean skip;
     private static List<File> files;
-    private static final File root = new File("/export/opengrok_p4_test");
+    private static final File root = new File("/var/opengrok/src/p4foo");
```
vladak: I assume this needs some setup steps to populate the root somehow. Could you perhaps update the test case with the steps ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryFactory.java
```diff
@@ -51,21 +51,30 @@
...
+            new SCCSRepository(),
+            new RazorRepository(),
+            new RCSRepository(),
+            new CVSRepository(),
+            new SSCMRepository(),
+            /*
+             * The following do expensive checks to determine isRepositoryFor(),
+             * so put them at the end of the repositories array.
+             */
+            new AccuRevRepository(),
```
vladak: good idea !

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
```diff
@@ -78,18 +78,31 @@ static String protectPerforceFilename(String name) {
         return t;
     }
 
+    static String unprotectPerforceFilename(String name) {
+        String t = name.replace("%40", "@");
+        t = t.replace("%23", "#");
+        t = t.replace("%2A", "*");
+        t = t.replace("%25", "%");
+        if (!name.equals(t)) {
+            LOGGER.log(Level.FINEST,
```
vladak: use isLoggable() also here ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
```diff
@@ -271,4 +304,56 @@ private static Date newDate(int year, int month, int day, int hour, int minute,
         cal.set(year, month - 1, day, hour, minute, second);
         return cal.getTime();
     }
+
+    private static HistoryEntry parseLedeLine(List<HistoryEntry> entries, HistoryEntry entry,
```
vladak: what does `Lede` stand for ?

vladak:
Nice, finally someone who can competently change Perforce code. I have only bunch of nits.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/PerforceRepositoryTest.java
idodeclare: OK done

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
idodeclare: Yes

opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceHistoryParser.java
idodeclare: I renamed as `parseEntryLine()`.

vladak:
Seems like Perforce will finally start working fine. Thanks !

PR #3063, opened 2020-03-07
Upgrade to JFlex 1.8.2

idodeclare:
Hello,

Please consider for integration this patch to upgrade to the recent JFlex 1.8.2, with the best feature the reduced memory footprint for Unicode scanners from multiple megabytes to scores of kilobytes.

This patch accommodates JFlex's widening at @vladak's [request](https://github.com/jflex-de/jflex/issues/536) the `yychar` field to `long` from `int`. After a position of `Integer.MAX_VALUE`, however, Lucene would not be able to index postings for any matches, so this patch just silently drops any such matches.

This patch also removes the regex logic rewrite for a few scanners (Java, JavaScript, JSON, Plain, PlainFull) to limit tokens to the Lucene limit (32,766) and updates `JFlexTokenizer` generally for all scanners to truncate tokens larger than the limit (and to silently drop above). (I question the utility of tokens of this huge, limited size; but leading wildcards possibly make huge tokens still useful.)

BTW version 1.8 includes a [PR of mine](https://github.com/jflex-de/jflex/pull/693) 😄 so we can edit `lexh` include files, and the JFlex plugin will recognize the modified include files and regenerate any affected lexers. With version 1.7 we have to touch the `lex` files too to get the lexers rebuilt.

I see v1.8.x also includes some other updates requested by @vladak, such as Javadoc fixes which make it no longer necessary to rewrite `<tt>` and `<code>`.

Thank you.


idodeclare:
[Pushed a JavaDoc fix](https://github.com/oracle/opengrok/compare/497aef0..b158749) that errors out strictly on Java 11

tarzanek:
lgtm, merge?

idodeclare:
I put this back in Draft since PR #3053 has some new scanners (`SourceSplitterScanner` and `LineBreakerScanner`) that will need to be updated for the Lucene 1.8.3 API.

tarzanek:
I merged 3053, so I guess I just broke this,
will you be able to rebase and fix @idodeclare ?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/TextAnalyzer.java
```diff
@@ -91,10 +91,13 @@ public Xrefer writeXref(WriteXrefArgs args) throws IOException {
...
+    /**
+     * Gets a BOM-stripped {@link Reader} (default UTF-8 charset) of the
+     * specified {@code stream}, wrapped in a {@link ZeroReader}.
+     */
     protected Reader getReader(InputStream stream) throws IOException {
-        // SRCROOT is read with UTF-8 as a default.
-        return IOUtils.createBOMStrippedReader(stream,
-            StandardCharsets.UTF_8.name());
+        // sourceRoot is read with UTF-8 as a default.
+        return new ZeroReader(IOUtils.createBOMStrippedReader(stream,
```
vladak, line 100: why is this suddenly needed ? What Reader suffers from the 0 return problem specifically ?

vladak:
This needs the change I had in my (now abandoned) PR to switch JFlex Windows builds explicitly to use UTF-8: https://github.com/oracle/opengrok/pull/3174/files#diff-c51990a9d80395538c85816a557fca5eR216

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/JFlexTokenizer.java
```diff
@@ -119,18 +128,44 @@ public void sourceCodeSeen(SourceCodeSeenEvent evt) {
...
+     */
+    protected void clearAttributesEtc() {
         clearAttributes();
-        //FIXME increasing below by one(default) might be tricky, need more analysis
-        // after lucene upgrade to 3.5 below is most probably not even needed        
-        this.posIncrAtt.setPositionIncrement(1);
-        this.termAtt.setEmpty();
-        this.termAtt.append(str);
-        this.offsetAtt.setOffset(start, end);
+        didSetAttribsValues = false;
```
vladak, line 169: nit/question: why not use `this.didSetAttribsValues` like in `setAttribs()` ?

vladak:
Modulo the comments I had this looks good; I like we got rid of the generated source modifications.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/TextAnalyzer.java
idodeclare: JFlex just advises that on some JRE implementations, they found readers returning 0 sometimes. JFlex prior to 1.6.1 did a guard against this, but afterward JFlex offered `ZeroReader` as the guard, which needs to be explicitly used. I'm putting `ZeroReader` then into OpenGrok, since the workaround is desirable.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/JFlexTokenizer.java
idodeclare: Oh I preserved the former syntax by the previous author. In methods I don't normally use `this.` unless disambiguation is needed — e.g. often within setters. In the new `clearAttributesEtc()` there was no ambiguity.

idodeclare:
> This needs the change I had in my (now abandoned) PR to switch JFlex Windows builds explicitly to use UTF-8: https://github.com/oracle/opengrok/pull/3174/files#diff-c51990a9d80395538c85816a557fca5eR216

Included

idodeclare:
> I put this back in Draft since PR #3053 has some new scanners (`SourceSplitterScanner` and `LineBreakerScanner`) that will need to be updated for the JFlex 1.8.2 API.

I forced push to include updated handling given the JFlex-ing of `SourceSplitterScanner` and `LineBreakerScanner` in #3053

vladak:
The failing tests in Travis build are due to #3176, otherwise looks good.

vladak:
Thanks !

PR #3073, opened 2020-03-17
report directory traversal time

vladak:
When discussing slow indexer time in https://github.com/oracle/opengrok/issues/3071#issuecomment-600007923 I found that the traversal time is not reported even though the `Statistics` object is created before hand.


vladak:
Thanks Chris.

PR #3075, opened 2020-03-17
Feature/check port

idodeclare:
Let's see if some port probing improves the reliability of tests on Windows.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/PortChecker.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/PortChecker.java
@@ -1,0 +1,75 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/PortChecker.java
```diff
@@ -0,0 +1,75 @@
...
+        if (port < MIN_PORT_NUMBER || port > MAX_PORT_NUMBER) {
+            throw new IllegalArgumentException("Port out of valid range: " + port);
+        }
+
+        ServerSocket ss = null;
+        DatagramSocket ds = null;
+        try {
+            ss = new ServerSocket(port);
+            ss.setReuseAddress(true);
+            ds = new DatagramSocket(port);
```
vladak: I wonder why it needs to check DGRAM socket as well.

idodeclare: Apache code states port as available only if not TCP nor UDP. I borrowed as-is

PR #3076, opened 2020-03-18
Extract to QueryParameters named literals

idodeclare:
@tarzanek ,

This is that commit you wanted as a separate PR.

Regards.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
@@ -20,7 +20,7 @@
 /*
  * Copyright (c) 2005, 2019, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright 2011 Jens Elkner.
- * Portions Copyright (c) 2017-2019, Chris Fraire <cfraire@me.com>.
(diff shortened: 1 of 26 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -1369,17 +1390,18 @@ public Void apply(Integer page) {
                         // append request parameters
                         if (request != null && request.getQueryString() != null) {
                             String query = request.getQueryString();
-                            query = query.replaceFirst("(\\?|&amp;|&|)n=\\d+", "");
-                            query = query.replaceFirst("(\\?|&amp;|&|)start=\\d+", "");
-                            query = query.replaceFirst("^(\\?|&amp;|&)", "");
+                            query = query.replaceFirst(RE_Q_E_A_A_COUNT_EQ_VAL, "");
+                            query = query.replaceFirst(RE_Q_E_A_A_START_EQ_VAL, "");
+                            query = query.replaceFirst(RE_A_ANCHOR_Q_E_A_A, "");
```
vladak: these constants are a little bit cryptic on the first sight but they actually make sense.

vladak:
nice and tidy, thanks !

idodeclare:
@vladak , thank you for reviewing

PR #3084, opened 2020-03-23
check the value before putting to average key lengths map

vladak:
This change sanitizes the values before putting them to `averageLengths` map in suggester project data, thus preventing the `IllegalArgumentException` exception in `initSearchCountMap()`.

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
+++ b/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
```diff
@@ -261,8 +261,10 @@ private WFSTCompletionLookup build(final IndexReader indexReader, final String f
         WFSTCompletionLookup lookup = createWFST();
         lookup.build(iterator);
 
-        double averageLength = (double) iterator.termLengthAccumulator / lookup.getCount();
-        averageLengths.put(field, averageLength);
+        if (iterator.termLengthAccumulator > 0) {
```
ahornace: I'd prefer the condition to check `lookup.getCount()` but that's not a big deal.

Issue #2704, opened 2019-02-27
testComplexSuggestions2, testGetPopularityDataAll sometimes fails


vladak:
```
[ERROR]   SuggesterControllerTest.testComplexSuggestions2:407 
Expected: iterable over ["main", "mutablelistof"] in any order
     but: No item matches: "main", "mutablelistof" in []
```

vladak:
like in https://travis-ci.org/oracle/opengrok/jobs/499311368 but I have also seen the failures in AppVeyor builds.

ahornace:
From my experience it fails because of this race condition which is bugging us for a long time: https://github.com/oracle/opengrok/issues/2030

The link you provided is still running and I see no logs :/

vladak:
Sorry, just restarted the build because this was a release build.

Dne st 27. 2. 2019 17:01 uživatel Adam Hornáček <notifications@github.com>
napsal:

vladak:
Getting more failure in `SuggesterControllerTest` on Windows:
  - `testGetPopularityDataDifferentField`: https://travis-ci.org/oracle/opengrok/jobs/637835530
  - `testGetPopularityDataAll`: https://travis-ci.org/oracle/opengrok/jobs/639408758

vladak:
The `before` method of the test has timeout set to 15 seconds:
```java
          await().atMost(15, TimeUnit.SECONDS).until(() ->
                  getSuggesterProjectDataSize() == env.getProjectList().size());
```
which makes me wonder whether this is enough and how/if Awaitility provides any feedback/log if the timeout expired and the condition is not met.

vladak:
Also, for the failed case `testGetPopularityDataDifferentField` the setup of the test reported:
```
Jan 16, 2020 9:09:10 AM org.opengrok.suggest.SuggesterProjectData initSearchCountMap
SEVERE: Could not create ChronicleMap, most popular completion disabled, if you are using JDK9+ make sure to specify: --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports java.base/jdk.internal.misc=ALL-UNNAMED --add-exports java.base/sun.nio.ch=ALL-UNNAMED
java.lang.IllegalArgumentException: Average key size must be a positive, finite number
	at net.openhft.chronicle.map.ChronicleMapBuilder.checkAverageSize(ChronicleMapBuilder.java:314)
[...]

PR #3086, opened 2020-03-23
make suggester popularity tests wait on rebuild

vladak:
This change fixes the popularity data related test failures tracked by #2704. After making `increaseSearchCount()` testable and adding logging it became evident that the failures stem from bailing on waiting on suggester rebuild.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestCasePrinterRule.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestCasePrinterRule.java
@@ -1,0 +1,88 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ */
+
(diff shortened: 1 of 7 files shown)


vladak:
Also, I removed awaitility based check of suggester initialization from the suggester controller test. Firstly, awaitility seems like a glorified active waiting in this context, secondly it required use of reflection to get inside the suggester private data.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -234,6 +249,11 @@ public void rebuild(final Collection<NamedIndexDir> indexDirs) {
             return;
         }
 
+        // for testing
```
ahornace: Is it needed?

vladak: you mean the comment ?

ahornace: Yes, it is in the PR at least 3 times. This kind of comment indicates that the code should be removed in production code.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestCasePrinterRule.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    public final Statement apply(Statement statement, Description description) {
+        final String testDescription = getClassBasename(description.getClassName()) + "#" +  description.getMethodName();
+        beforeContent = String.format("\n[TEST START: %s]\n", testDescription);
+        afterContent =  String.format("[TEST ENDED: %s] ", testDescription);
+        return printer.apply(statement, description);
+    }
+}
```
ahornace: Missing newline.

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -87,6 +91,12 @@
 
     private final int rebuildParallelismLevel;
 
+    private boolean rebuilding;
```
ahornace: Should be `volatile`.

vladak: I assume this is done so that it is not optimized away ?

opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestCasePrinterRule.java
vladak: it will be printed in `after()`.

PR #3088, opened 2020-03-24
re-enable testGetSuggestionsMultipleProjects()

vladak:
with #2030 fixed it should be possible to re-enable the `testGetSuggestionsMultipleProjects()` test.


vladak:
will fold into #3086

PR #3091, opened 2020-03-25
Feature/revved css

idodeclare:
Hello,

Please consider for integration this patch to specify a revision too for the OpenGrok look-and-feel CSS files — for cache busting when updated.

I also removed a redundant searchable-option-list-2.0.3.min.css (given that there is a minify step now), after reconciling for differences and finding one different color that was coming through in an OpenGrok release.

Thank you.


vladak:
looks sane, thanks !

PR #3094, opened 2020-03-26
switch Docker image to JDK 11

vladak:
This change switches the Docker image to JDK 11. It helps with growing RSS of the Tomcat process (presumably by using different GC implementation).


Dockerfile
```diff
@@ -67,7 +67,7 @@ ENV CATALINA_HOME /usr/local/tomcat
 ENV CATALINA_BASE /usr/local/tomcat
 ENV CATALINA_TMPDIR /usr/local/tomcat/temp
 ENV PATH $CATALINA_HOME/bin:$PATH
-ENV JRE_HOME /usr
+# ENV JRE_HOME /usr
```
idodeclare: Maybe leave out since Git history has it?

vladak: yeah, forgot to delete it.

Issue #3093, opened 2020-03-26
Date/Time for subversion repositories can be incorrect due to parsing issue


harrisric:
For a subversion repository the date typically appears in the format yyyy-MM-dd'T'HH:mm:ss.SSSxxx'Z' (where x are microseconds) e.g.
2020-03-24T17:11:35.545818Z

Currently the date parsing for subversion repos tries the formats;
            "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
            "yyyy-MM-dd'T'HH:mm:ss.'Z'",
            "yyyy-MM-dd'T'HH:mm:ss'Z'"
Currently this means the date 
2020-03-24T17:11:35.545818Z 
is reported as 
Tue Mar 24 17:20:40 GMT 2020
(or 1585070440818 milliseconds using date.getTime())
because the parsing is treating the fractions of a second as milliseconds and therefore adding an extra 545 seconds (9m5s) to the time as well as storing the wrong number of milliseconds.

PR #3095, opened 2020-03-26
Fixes #761,  #3093 (svn history on renamed files, svn date parsing)

### LINKED ISSUE #761: Cannot view history of moved / renamed SVN files
Hi,

I am currently using OpenGrok as a front end for Subversion. I am using RANCID to archive device configurations that are stored in Subversion. I check out a copy of the repo locally on the same server and index with OpenGrok. There are no issues with the setup.

My current problem is when I rename a file in Subversion using the svn move command and commit the result, the history for that file is no longer visible in OpenGrok after a checkout and index. I can go to the command line and view revisions using svn commands so the history is still there. The history for the top level /SVN in OpenGrok has a set svn:ignores after my move, commit, index. I am not sure if this is a Subversion issue or how OpenGrok is interpreting the move. Any information about this process would be greatly appreciated.

Regards,
Scotty

harrisric:
This fixes the problem with svn history by tracking files that are renamed. Also fixes a bug with the date handling for SVN repos.

I accept the Oracle Contributor Agreement.


vladak:
Nice ! Please add your copyright to the changed source code files.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
```diff
@@ -89,7 +102,12 @@ public void endElement(String uri, String localName, String qname) throws SAXExc
                 entry.setAuthor(s);
             } else if ("date".equals(qname)) {
                 try {
-                    entry.setDate(repository.parse(s));
+                    // need to strip microseconds off - assume final character is Z otherwise invalid anyway.
+                    String dateString = s;
+                    if (s.length() > 24) {
```
idodeclare: Perhaps add e.g. `static final String SVN_MILLIS_DATE_EG` with the known-good format and then use `SVN_MILLIS_DATE_EG.length()` to avoid the unnamed literals here and the next line?

(This reminds of a TODO to refactor out of the old `DateFormat` for thread safety — and you've now made apparent also for microseconds handling.)

harrisric: Good idea.
Also thought it would be good to get rid of java.util.Date generally, but that seemed quite invasive for a first contribution!

vladak: I assume you mean this comment: https://github.com/oracle/opengrok/blob/52fc2931f25089055b5cf1dea11f846521d748ef/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java#L620

Having TODO's just in the source code is not good, filed  #3096 to track this.

idodeclare:
@harrisric , there are some Travis build test failures for you to look at. (Mac and Windows fail often enough for spurious reasons that those don't usually block integration; but Linux needs to run reliably.)

Were the test suites succeeding for you locally with `mvn test`?

harrisric:
> @harrisric , there are some Travis build test failures for you to look at. (Mac and Windows fail often enough for spurious reasons that those don't usually block integration; but Linux needs to run reliably.)
> 
> Were the test suites succeeding for you locally with `mvn test`?

I'd been running the tests locally in Eclipse. I'd built the zip file manually to get me started, without realising the ant-mechanism that had been set up here and the trick to make sure the various repo-files were found.

idodeclare:
Ah I see. Travis is good (and essential) but _slow_, so I usually use some form of the following too:

#### Full, clean build and test
`$ ./mvnw clean test`

#### Incremental test
`$ ./mvnw  -DfailIfNoTests=false -Dtest='Subversion*' test`

harrisric:
@idodeclare - I think I've figured out the problem with my new tests - when it runs on my local machine in windows it is fine because the repo that the new test directory is connected to is also on my machine. When I run it in the WSL on my machine it can't connect to the repo and therefore fails to get any history. 
Based on that I think even if I include the repository there will be no way that the copy will be able to remain linked to it. So I will need to look at a way of either building the repo during the test either using svnadmin load from an svndump or else by constructing the sequence of commits the test is looking for.

idodeclare:
Making progress, I see. Just one remaining error it seems.

(I don’t see the link to the test build. Not sure why. I had to go to Travis from the “build” link on the OpenGrok GitHub home.)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
```diff
@@ -99,11 +123,14 @@ public void endElement(String uri, String localName, String qname) throws SAXExc
...
                 if (s.startsWith(prefix) && !s.equals(prefix)) {
-                    File file = new File(home, s.substring(prefix.length()));
-                    String path = file.getAbsolutePath().substring(length);
+                  File file = new File(home, s.substring(prefix.length()));
+                  String path = file.getAbsolutePath().substring(length);
                     // The same file names may be repeated in many commits,
                     // so intern them to reduce the memory footprint.
                     entry.addFile(path.intern());
+                    if (isRenamed) {
+                      renamedFiles.add(file.getAbsolutePath().substring(home.length() + 1));
```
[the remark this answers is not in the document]
harrisric: Thanks, but unfortunately not as simple as that - the use of File also means that we switch convention from forward-slash to back-slash as well.
I tried to write a unit test which showed this but that seems to show up the limitations of the SubversionHistoryParserTest using a fixed 'home' of '/' (which on windows might e.g. become 'C:\' and therefore even with the current code appears to leave the leading-slash in place).

I've made the other formatting corrections suggested.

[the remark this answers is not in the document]
idodeclare: Oh I see. Thank you for explaining. I forget the Windows considerations.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
```diff
@@ -56,13 +61,20 @@
 
     private static class Handler extends DefaultHandler2 {
 
+        /**
+         * Example of the longest date format that we should except - SimpleDateFormat cannot cope with micro/nano seconds.
```
vladak, line 65:
```suggestion
         * Example of the longest date format that we should accept - SimpleDateFormat cannot cope with micro/nano seconds.
```

opengrok-indexer/src/test/java/org/opengrok/indexer/history/FileHistoryCacheTest.java
```diff
@@ -607,6 +610,134 @@ public void testRenamedFilePlusChangesBranched() throws Exception {
...
+     * Make sure produces correct history where several files are renamed in a single commit.
+     */
+    @ConditionalRun(RepositoryInstalled.SubversionInstalled.class)
+    @Test
+    public void testMultipleRenamedFiles() throws Exception {
+        File reposRoot = new File(repositories.getSourceRoot(), "subversion");
+        History updatedHistory;
+
+        // The test expects support for renamed files.
+        RuntimeEnvironment.getInstance().setHandleHistoryOfRenamedFiles(true);
```
vladak, line 624: set this to `false` at the end of the test ?

opengrok-indexer/src/test/java/org/opengrok/indexer/history/FileHistoryCacheTest.java
```diff
@@ -607,6 +610,134 @@ public void testRenamedFilePlusChangesBranched() throws Exception {
...
+     * Make sure produces correct history for a renamed and moved file in Subversion
+     */
+    @ConditionalRun(RepositoryInstalled.SubversionInstalled.class)
+    @Test
+    public void testRenamedFile() throws Exception {
+        File reposRoot = new File(repositories.getSourceRoot(), "subversion");
+        History updatedHistory;
+
+        // The test expects support for renamed files.
+        RuntimeEnvironment.getInstance().setHandleHistoryOfRenamedFiles(true);
```
vladak, line 683: ditto

vladak:
I only have couple of nits otherwise this is good to go. It is surprising that it required so few changes.

vladak:
Also, a question: what happens when you are browsing the history of renamed file and want to display its version before the rename(s) ? In Mercurial/Git it is necessary to lookup the original name of the file to actually retrieve the content in such revision. Is this the case for Subversion ?

harrisric:
> 
> 
> Also, a question: what happens when you are browsing the history of renamed file and want to display its version before the rename(s) ? In Mercurial/Git it is necessary to lookup the original name of the file to actually retrieve the content in such revision. Is this the case for Subversion ?

I'm not sure what specifically happens under the covers, but (even without my changes) when you are in the file view you can rewind the version number to before a rename and it handles it without any problems.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/FileHistoryCacheTest.java
harrisric: Good point - there is already some resetting being done in the tearDown - I'll make sure that all variables are reset to their defaults. Unfortunately not a clean way to do this that I can see.

vladak:
Thanks !

PR #3104, opened 2020-04-08
P4 filelog parsing update

gtoph:
Adding -s option to limit the history to only include integrations that affect the file as there can be hundreds of branches listed where where it's not touched and shouldn't be included in the history.
Because the current regex can find p4 path information in the comment sections, only search for P4 paths at the start of the line.


gtoph:
Ah whoops.  Been tweaking the code for a while now, but it's my first time actually doing a pull request.  Hopefully all good now.

vladak:
I don't see you listed on https://www.oracle.com/technetwork/community/oca-486395.html#c

Please initiate the OCA process and/or state that you are okay with terms.

gtoph:
I'm fine with it.  I emailed the form just yesterday so they probably just havent gotten to it yet.

vladak:
Sounds good, merging. Thanks !

PR #3106, opened 2020-04-09
introduce the --nolock option to sync.py

vladak:
This change supports the following workflow:
  - a project is added using `projadm.py`
  - it is indexed using `sync.py` (so that the same set of per project indexer options is used as is for periodic reindex using the same `sync.py` configuration)

This fixes a problem which happens when the project `sync.py` fails because the periodic `sync.py` is already running. The exclusion using the locking for periodic sync is in order however in this case it causes problems (project is added but not indexed).

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/sync.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/sync.py
@@ -19,7 +19,7 @@
 #
 
 #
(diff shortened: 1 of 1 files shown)


ktulinger:
I don't understand why it is a problem to wait for the next periodic run of the sync.py

vladak:
This is for adding a new project and indexing it. Next round of global 
indexing can happen in many hours.

PR #3107, opened 2020-04-09
add username and timeout to toString()

vladak:
This change adds more info to the `toString()` method of `LdapServer`. This aids debugging when authorization framework is reloaded or when a server is reconnecting.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/ldap/LdapServer.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/ldap/LdapServer.java
@@ -67,6 +67,13 @@ public class LdapServer implements Serializable {
         this.url = server;
     }
 
+    public LdapServer(String server, String username, String password) {
+        this(prepareEnv());
+        this.url = server;
+        this.username = username;
+        this.password = password;
+    }
+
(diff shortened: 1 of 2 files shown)


plugins/src/main/java/opengrok/auth/plugin/ldap/LdapServer.java
```diff
@@ -134,7 +141,7 @@ public synchronized boolean isWorking() {
      * @return the new connection or null
      */
     private synchronized LdapContext connect() {
-        LOGGER.log(Level.INFO, "Server {0} connecting", this.url);
+        LOGGER.log(Level.INFO, "Connecting to LDAP server {0} ", this.toString());
```
idodeclare, line 144: Nit: is `toString()` superfluous here and below (and if not, should these logging computations be short-circuited by convention with e.g. `if (LOGGER.isLoggable(Level.INFO))`)?

tarzanek:
lgtm, I agree with the nitpick, but this is I guess the same debate of explicit versus implicit
that said I like Scala a lot (although the way how the language is implicit can drive you crazy :-D 
but then it's easy to read without all the boilerplate code)

PR #3108, opened 2020-04-12
Add HCL and Terraform analyzers

idodeclare:
Hello,

Please consider for integration this patch to add an HCL analyzer and a specialization of it for Terraform.

This patch also fixes so that `mvn test` can be repeated successfully without needing `mvn clean`. 796b45ab66 left out some needed cleanup from the setup of `-create-svn-repository`.

Thank you.


tarzanek:
nice , lgtm and thanks for dividing the work in the patches, I guess this is an option to divide it into PRs if the diff is small
merge?
@vladak any comments?

idodeclare:
> nice , lgtm 

@tarzanek , thank you for reviewing

vladak:
quite a piece of work, thanks !

idodeclare:
@vladak , thank you for reviewing

PR #3109, opened 2020-04-12
Fix rare uncounted Ruby LOC

idodeclare:
Hello,

Please consider for integration this patch with a fix for a rare uncounted Ruby LOC. I spotted it when studying my Ruby HERE-doc support to apply it for HCL and Terraform.

Thank you.


tarzanek:
lgtm, +1 from me for merge

PR #3110, opened 2020-04-12
Squash YYYY and some other Sonar-identified bugs

idodeclare:
Just went through Sonar and squashed several identified bugs.


vladak:
Nice work ! I have to admit I lost Sonar from the radar a bit :-)

PR #3112, opened 2020-04-13
Fix #1665 : searchable-option-list bulk operations are slow

### LINKED ISSUE #1665: select all is very slow with a large number of projects
The select all operation (and clear/invert) is very slow when there are a large number of projects in the list.  In my example I have about 2000 projects and select all takes about 15 seconds on Chrome and much longer on IE.  

I'm using opengrok-1.0 and I think the issue has to do with the code in searchable-option-list that updates the selected display area once per project (see _addSelectionDisplayItem/_removeSelectionDisplayItem).

idodeclare:
Hello,

Please consider for integration this patch to fix the bulk operations that are slow when the project count in the thousands (`Select-All`, `Invert-Selection`, `Clear`) by adding a `Map` by `val()` and also avoiding creating HTML elements when exceeding a defined `maxShow`.

When a bulk operation is in-progress, the `X-Items-Selected` update is deferred and only done at the end of the operation.

I encountered one testing problem which I need advice for: I was running with a thousand test projects, but my system could not actually support a query for all 1000 projects, and it showed a page with just `0`. On hitting Back, the page was not in a consistent state, and `Select-All`, `Invert-Selection`, and `Clear` did not work as expected. Any advice?

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
@@ -19,7 +19,7 @@
 
 /*
  * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
- * Portions Copyright (c) 2017, Chris Fraire <cfraire@me.com>.
+ * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.web;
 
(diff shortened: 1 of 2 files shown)


tarzanek:
lgtm, +1 for merge

idodeclare:
Ohhhhhhhhhh sorry I introduced a bug. When the SOL is being bound by `maxShow`, and you X a shown item, another selected item should be shown — but I removed that bit not understanding. Luckily this is for the non-bulk UI, so let me see about effectively restoring that behavior.

PR #3113, opened 2020-04-13
Remove superfluous boolean

idodeclare:
@tarzanek , I realized the `boolean` I added was superfluous. Please would you take a look when you get a moment.

Thank you.


tarzanek:
merging, thnx

PR #3114, opened 2020-04-14
Bugfix/SOL max-show

idodeclare:
This fixes the bug I accidentally introduced in 5cac8be5c7 when `maxShow` is in effect and a user clicks × to remove an item. Another item should appear but does not do that.

The meat of this fix is in `_removeSelectionDisplayItem`, which Github-diff shows fine below.

Github-diff is showing a much bigger diff than this patch truly is because its diff algorithm is not finding good alignment from my extraction of `_buildSelectionDisplayItem` from `_addSelectionDisplayItem`, and also it shows the whitespace changes.

Following is a picture of a better diff showing just the straight-forward extraction of `_buildSelectionDisplayItem`:
image

Thank you.


vladak:
LGTM. where does the diff in the screenshot come from ?

idodeclare:
Thank you, @vladak . That's a screen shot of a Beyond Compare diff view.

PR #3119, opened 2020-04-17
fix broken links in diff view

vladak:
Easiest fix.


vladak:
No worries. The UI is not very well covered by tests so things are expected to break.

idodeclare:
I found some more of the pattern you found. Let me get a PR.

PR #3120, opened 2020-04-17
Fix more broken links due to indentation

idodeclare:
@vladak , this is a PR for a few more fixes of the bugs you identified.


vladak:
Great, thanks !

PR #3121, opened 2020-04-18
Huge text handling

idodeclare:
Hello,

Please consider for integration this patch to add Huge Text file handling.

`Indexer` and `Configuration` get two new settings, `hugeTextThresholdBytes` (default 1_000_000) and `hugeTextLimitCharacters` (default 5_000_000). The threshold determines when OpenGrok will override a `PLAIN` genre file as a `hugetext` `DATA` file instead. The character limit determines how much to read and index for `hugetext` (with contextless truncation); the limit may be zero.

`hugeTextThresholdBytes` is checked for applicable files with each run, while no state for `hugeTextLimitCharacters` is stored. Changing `hugeTextLimitCharacters` after indexing would require touching affected source code files to revise the index.

For affected gzip and bzip2 files, changes to either `hugeTextThresholdBytes` or `hugeTextLimitCharacters` would require touching affected compressed files to revise the index.

Thank you.


tarzanek:
this actually looks good, I just quickly skimmed through

maybe only one concern, I think we should WARN instead of FINE print if a file is skipped because of limits ... (unless FINE is printed by default to log file ... but then I'd like to see those as WARN on console too ... )

vladak:
What happens if index is created with particular limits and then the limits are changed ?

idodeclare:
> maybe only one concern, I think we should WARN instead of FINE print if a file is skipped because of limits ... (unless FINE is printed by default to log file ... but then I'd like to see those as WARN on console too ... )

No file is skipped. It is still included but under the `HugeTextAnalyzer` where the user chooses how much to analyze (as low as 0). It also gets no xref.

idodeclare:
> What happens if index is created with particular limits and then the limits are changed ?

I tried to describe that above, but to clarify:

You can change `hugeTextThresholdBytes` and then re-index, and (as applicable) `PLAIN` files will be reclassified as `hugetext` `DATA` or `hugetext` `DATA` will be re-classified to `PLAIN`.

If you change `hugeTextLimitCharacters` and re-index, then nothing will happen because no state is stored related to that limit.

`PLAIN` files inside gzip or bzip2 are handled by `HugeTextAnalyzer`, but changing `hugeTextThresholdBytes` and re-indexing does nothing because the uncompressed size is not known when `checkSettings` would need to decide. (Changing `hugeTextLimitCharacters` and reindexing also does nothing for gzip or bzip2, for the same reason above.)

tarzanek:
sorry, I meant "trimmed" down, not skipped
anyways, logs should say it loud that file x and y are now not considered fully
I don't expect this output will flood the screen, but it should be a warning to get printed on console

idodeclare:
@tarzanek , that's done.

idodeclare:
> `PLAIN` files inside gzip or bzip2 are handled by `HugeTextAnalyzer`, but changing `hugeTextThresholdBytes` and re-indexing does nothing because the uncompressed size is not known when `checkSettings` would need to decide.

I suppose it would be straight-forward to store a value for uncompressed size in the `Document` so that we could compare against `hugeTextThresholdBytes` in `checkSettings` for gzip/bzip2. Worth it?

idodeclare:
> > `PLAIN` files inside gzip or bzip2 are handled by `HugeTextAnalyzer`, but changing `hugeTextThresholdBytes` and re-indexing does nothing because the uncompressed size is not known when `checkSettings` would need to decide.
> 
> I suppose it would be straight-forward to store a value for uncompressed size in the `Document` so that we could compare against `hugeTextThresholdBytes` in `checkSettings` for gzip/bzip2. Worth it?

Oh but that would mean decompressing entirely. Probably not a good idea.

tarzanek:
lgtm, so +1 from my side
but since this PR is big, I'd like to get at least one other review to be merged
@vladak , @tulinkry  ?

idodeclare:
Just rebased on master since this needed revision to accommodate the `Configuration` API changes of #3127 (but would not have shown up as having any Git conflicts)

vladak:
I will take a look; also needs rebase.

idodeclare:
Just trivial conflicts upon rebase

idodeclare:
Just rebasing for trivial conflicts related to R analyzer and then again after parallel detection merged

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -432,7 +432,18 @@ public static void main(String[] argv) {
...
+         *
+         *    8888888888  .d8888b.                888
+         *          d88P d88P  Y88b               888
+         *         d88P         888               888
+         *        d88P        .d88P       .d8888b 88888b.   8888b.  888d888 .d8888b
+         *     88888888   .od888P"       d88P"    888 "88b     "88b 888P"   88K
+         *      d88P     d88P"           888      888  888 .d888888 888     "Y8888b.
+         *     d88P      888"            Y88b.    888  888 888  888 888          X88
+         *    d88P       888888888        "Y8888P 888  888 "Y888888 888      88888P'
+         */
```
vladak, line 446: :-D

I wonder if there is a way how to enforce via test case.

opengrok-indexer/src/test/java/org/opengrok/indexer/index/HugeTextTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.opengrok.indexer.history.RepositoryFactory;
+import org.opengrok.indexer.util.TestRepository;
+import org.opengrok.indexer.web.Util;
+
+import java.io.IOException;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * @author Trond Norbye
```
vladak, line 48: seems like a new class ?

idodeclare: > seems like a new class ?

I copied an existing test class because I needed the assertions around added files. CDDL dictates "You may not remove or alter ... any descriptive text giving attribution to any Contributor or the Initial Developer."

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
idodeclare: The parser definition and `helpUsage` is encapsulated in `Indexer` and not available for testing easily I think.

idodeclare:
Rebased for trivial conflict in search.jsp

idodeclare:
Rebased for PageConfig.java re-lo, and git automatic-merge took care of it

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: will be fixed via #3968

PR #3126, opened 2020-04-22
add log entry and throw exception when message limit is exceeded

vladak:
When trying to track down a problem with Messages, I realized that when message limit is hit is is not properly messaged (sic!).

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
@@ -127,8 +127,8 @@ public class MessagesContainer {
             }
 
             if (isMessageLimitExceeded()) {
-                LOGGER.log(Level.WARNING, "cannot add message to the system, limit of {0} messages exceeded." +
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/MessagesContainer.java
```diff
@@ -121,7 +127,9 @@ public void addMessage(final Message m) {
             }
 
             if (isMessageLimitExceeded()) {
-                return;
+                LOGGER.log(Level.WARNING, "cannot add message to the system, limit of {0} messages exceeded." +
+                        "consider setting the ''messageLimit'' tunable.", messageLimit);
```
idodeclare, line 131: Odd spacing.

Also maybe it might be more terse as `"exceeded Configuration messageLimit of {0}"`?

vladak: yes, that reads better.

tarzanek:
lgtm, merge

PR #3127, opened 2020-04-23
Minimize reflection into Configuration

idodeclare:
Hello,

Please consider for integration this patch to minimize reflection into `Configuration` and shift that logic to `ConfigurationController` where it's more difficult to avoid given the REST-exposure of all configuration through a single web method.

`RuntimeEnvironment` can use functional interfaces to achieve the equivalent of the former reflection without having to identify methods by `String`; thereby we get compile-time checks.

I also imported an auto-disposable wrapper around `Lock` so that try-with-resources can be used instead of try-finally.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -19,13 +19,11 @@
 
  /*
   * Copyright (c) 2006, 2019, Oracle and/or its affiliates. All rights reserved.
-  * Portions Copyright (c) 2017-2019, Chris Fraire <cfraire@me.com>.
+  * Portions Copyright (c) 2017-2020, Chris Fraire <cfraire@me.com>.
   */
 package org.opengrok.indexer.configuration;
 
 import static org.opengrok.indexer.configuration.Configuration.makeXMLStringAsConfiguration;
-import static org.opengrok.indexer.util.ClassUtil.getFieldValue;
-import static org.opengrok.indexer.util.ClassUtil.setFieldValue;
 
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -46,8 +44,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -70,10 +67,12 @@ import org.opengrok.indexer.index.IgnoredNames;
 import org.opengrok.indexer.index.IndexDatabase;
 import org.opengrok.indexer.index.IndexerParallelizer;
 import org.opengrok.indexer.logger.LoggerFactory;
+import org.opengrok.indexer.util.CloseableReentrantReadWriteLock;
 import org.opengrok.indexer.util.CtagsUtil;
 import org.opengrok.indexer.util.ForbiddenSymlinkException;
 import org.opengrok.indexer.util.LazilyInstantiate;
 import org.opengrok.indexer.util.PathUtils;
+import org.opengrok.indexer.util.ResourceLock;
 import org.opengrok.indexer.web.Prefix;
 import org.opengrok.indexer.web.Statistics;
 import org.opengrok.indexer.web.Util;
@@ -92,7 +91,7 @@ public final class RuntimeEnvironment {
     private static final String URL_PREFIX = "/source" + Prefix.SEARCH_R + "?";
 
     private Configuration configuration;
-    private final ReentrantReadWriteLock configLock;
+    private final CloseableReentrantReadWriteLock configLock;
     private final LazilyInstantiate<IndexerParallelizer> lzIndexerParallelizer;
     private final LazilyInstantiate<ExecutorService> lzSearchExecutor;
     private final LazilyInstantiate<ExecutorService> lzRevisionExecutor;
@@ -103,7 +102,7 @@ public final class RuntimeEnvironment {
 
     private String configURI;
     private Statistics statistics = new Statistics();
-    public IncludeFiles includeFiles = new IncludeFiles();
+    IncludeFiles includeFiles = new IncludeFiles();
     private final MessagesContainer messagesContainer = new MessagesContainer();
 
     private static final IndexTimestamp indexTime = new IndexTimestamp();
@@ -134,7 +133,7 @@ public final class RuntimeEnvironment {
      */
     private RuntimeEnvironment() {
         configuration = new Configuration();
-        configLock = new ReentrantReadWriteLock();
+        configLock = new CloseableReentrantReadWriteLock();
         watchDog = new WatchDogService();
         lzIndexerParallelizer = LazilyInstantiate.using(() ->
                 new IndexerParallelizer(this));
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1996,14 +1843,45 @@ public File getDtagsEftar() {
...
+
+    /**
+     * Performs the specified operation which is provided the runtime
+     * configuration and the specified argument, after first having obtained the
+     * configuration write-lock (and releasing afterward).
+     * @param consumer a defined consumer
+     * @param v the input argument
+     * @param <V> the type of the input to the operation
+     */
+    public <V> void syncWriteConfiguration(ConfigurationValueConsumer<V> consumer, V v) {
```
ktulinger, line 1876: I'd swap the parameter order cause at it seems more natural to me like this

```
syncWriteConfiguration(groups, (c, g) -> {
      populateGroups(g, new TreeSet<>(getProjects().values()));
      c.setGroups(g);
});
```

as its "value -> consume" that value, rather then "consumer of a value -> value"

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CloseableReentrantReadWriteLock.java
```diff
@@ -0,0 +1,85 @@
...
+        writeLock().lock();
+        return unlocker;
+    }
+
+    private ResourceLock newReadUnlocker() {
+        return () -> CloseableReentrantReadWriteLock.this.readLock().unlock();
+    }
+
+    private ResourceLock newWriteUnlocker() {
+        return () -> CloseableReentrantReadWriteLock.this.writeLock().unlock();
```
ktulinger, line 83: why `this.writeLock().unlock()` is not enough?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1996,14 +1843,45 @@ public File getDtagsEftar() {
...
+        try (ResourceLock resourceLock = configLock.writeLockAsResource()) {
+            //noinspection ConstantConditions to silence unreference auto-closeable
+            assert resourceLock != null;
```
ktulinger, line 1879: why do you assert this when you're explicitely checking this in write|readLockAsResource?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CloseableReentrantReadWriteLock.java
```diff
@@ -0,0 +1,85 @@
...
+            unlocker = newReadUnlocker();
+            // No synchronization necessary since overwrite is of no matter.
+            readUnlocker = unlocker;
+        }
+        readLock().lock();
+        return unlocker;
+    }
+
+    /**
+     * @return a defined {@link ResourceLock} once the {@link #writeLock()} ()}
```
ktulinger, line 56: there are some extra brackets

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
```diff
@@ -91,4 +94,37 @@ public void reloadAuthorization() {
...
+        Object result = env.syncReadConfiguration(configuration -> {
+            try {
+                return ClassUtil.getFieldValue(configuration, fieldName);
+            } catch (IOException ex) {
+                capture[EXCEPTION_INDEX] = ex;
+                return null;
+            }
+        });
+        if (capture[EXCEPTION_INDEX] != null) {
+            throw capture[EXCEPTION_INDEX];
```
ktulinger, line 109: the exception here is only translated to `throw new WebApplicationException(e, Response.Status.BAD_REQUEST)`

can you thro that one here and remove the `try` block in the caller?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
```diff
@@ -91,4 +94,37 @@ public void reloadAuthorization() {
...
+        }
+        return result;
+    }
+
```
ktulinger, line 113: extra line?

ktulinger:
This is a very good idea with the configuration consumer!

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
idodeclare: That's to silent an annoying Java warning of the auto-closeable not being referenced in the block which was a blocker in IDEA when it runs tests with warnings-as-errors

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CloseableReentrantReadWriteLock.java
idodeclare: Oh that was from the imported implementation. Yes it seems superfluous

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
idodeclare: That would mean a line like the following:

`syncWriteConfiguration(flag, Configuration::setListDirsFirst);`

which seems to invert the important part.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CloseableReentrantReadWriteLock.java
idodeclare: Got it

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
idodeclare: I don't understand

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
idodeclare: Yes

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
ktulinger: I don't know :-D I somehow saw two lines. Nevermind.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
idodeclare: I'll rename the vague arguments and transpose it as you suggest

idodeclare:
Thank you for taking a look, @tulinkry

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
ktulinger: what about `@SuppressWarnings("unused")` ?

idodeclare: No it’s a very focused warning for auto-closeable

idodeclare:
@ktulinger, were the changes satisfactory?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: thank you

idodeclare:
Thank you, @tulinkry

idodeclare:
Just rebasing to ensure it still builds OK, since any new `Configuration` properties might be Git-conflict free but still need to be updated for this PR's `Configuration` API changes.

tarzanek:
can Creative Commons Attribution-ShareAlike be merged to CDDL project?
if the answer is yes, I'll merge
@vladak FYI

idodeclare:
@tarzanek , yes I believe so. It's the license which Stack Overflow uses

idodeclare:
> can Creative Commons Attribution-ShareAlike be merged to CDDL project?
> if the answer is yes, I'll merge

@tarzanek, see e.g. [OGKTextField.java](https://github.com/oracle/opengrok/blob/0826ee034ae906baad43b9524d25da960f3360d1/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/OGKTextField.java#L2).

tarzanek:
in such case this can be merged,
thank you Chris, you rock! :-)

idodeclare:
Thank you, Lubos! 😄

PR #3129, opened 2020-04-25
log LDAP search parameters for some of the failure cases

vladak:
This change aids debugging of some LDAP failures.


vladak:
Thanks !

PR #2732, opened 2019-03-25
Feature/search uhighlighter

idodeclare:
Hello,

Please consider for integration this patch to add `HitFormatter` to extend `OGKUnifiedHighlighter` for use by `SearchEngine`.

There are some known, pending migrations to `uhighlight` remaining in OpenGrok. `Hit` generation was one area, and `HistoryContext` continues to be a legacy highlighter.

@wy193777 requested in #2612 that search results not be HTML-ized. This patch mostly addresses the request by enhancing `SearchEngine` to return code-matching results for `full`|`defs`|`refs` and to add substring offsets to the `Hit` results in lieu of the former HTML-emphasis.

`hist` searches will still return legacy, HTML-fragment results via `HistoryContext`.

For testing, I uncommented `SearchEngineTest.testSearch()` but set it to `@Ignore`. The test is still useful even with its race condition bugs.

Thank you.

(@vladak, I see you're squash-merging again; please may I ask you don't subject my PRs to that.)

PR #3130, opened 2020-04-28
Extract SettingsHelper from SearchHelper

idodeclare:
Hello,

Please consider for integration this patch to extract `SettingsHelper` from `SearchHelper`. This is from open PR #2732 and is a pre-requisite to a number of ready features, including search context and directory \#LinesLOC.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
@@ -67,13 +67,10 @@ public class SettingsHelper {
      * Gets any mapped symlinks (after having called {@link #getSettings(String)}).
      * @return either a defined map or {@code null}
      */
-    public Map<String, IndexedSymlink> getSymlinks(String projectName) {
-        if (mappedIndexedSymlinks == null) {
-            throw new IllegalStateException("getSettings() not yet called");
-        }
-
-        String k = projectName != null ? projectName : "";
-        Map<String, IndexedSymlink> indexSymlinks = mappedIndexedSymlinks.get(k);
+    public Map<String, IndexedSymlink> getSymlinks(String projectName) throws IOException {
+        getSettings(projectName);
+        String projectKey = projectName != null ? projectName : "";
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    public IndexAnalysisSettings3 getSettings(String projectName) throws IOException {
+        if (mappedAnalysisSettings == null) {
+            IndexAnalysisSettingsAccessor dao = new IndexAnalysisSettingsAccessor();
+            IndexAnalysisSettings3[] setts = dao.read(reader, Short.MAX_VALUE);
+            map(setts);
+        }
+
+        String k = projectName != null ? projectName : "";
+        return mappedAnalysisSettings.get(k);
```
tulinkry, line 118: k is not very descriptive

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
```diff
@@ -0,0 +1,145 @@
...
+        this.reader = reader;
+    }
+
+    /**
+     * Gets any mapped symlinks (after having called {@link #getSettings(String)}).
+     * @return either a defined map or {@code null}
+     */
+    public Map<String, IndexedSymlink> getSymlinks(String projectName) {
+        if (mappedIndexedSymlinks == null) {
+            throw new IllegalStateException("getSettings() not yet called");
```
tulinkry, line 72: is it problem to call it here `getSettings()`?

idodeclare: No doesn't seem so

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
idodeclare: Revised

tarzanek:
lgtm, merging

idodeclare:
Thank you for the reviews, @ktulinger and @tarzanek

PR #3134, opened 2020-05-04
do not highlight path in masthead when Message is in effect

vladak:
In the end I chose the simplest solution rather than juggling CSS (https://stackoverflow.com/questions/5135019/css-opacity-only-to-background-color-not-the-text-on-it) and it makes more sense UI wise I think as it is aligned with the rest (project picker, project listing).


ktulinger:
screenshot please?

vladak:
Screenshot from 2020-05-04 15-38-35

vladak:
Thanks for the review !

PR #3135, opened 2020-05-04
append string to Message text for disabled messages

vladak:
In order to make operational aspect easier, it would help if the cause of disabled projects could be propagated to the UI when using the `opengrok-mirror` script. The `disabled: true` behavior will be preserved however if the value is a string, it will be appended to the message. I.e. when running the tool with this config:
```yaml
disabled_command:
  command:
    - http://localhost:8080/source/api/v1/messages
    - POST
    - messageLevel: info
      duration: PT5M
      tags: ['%PROJECT%']
      text: disabled project
projects:
  opengrok:
    disabled: true
  opengrok-1.0:
    disabled: true
    disabled-reason: "foo bar"
```
the `opengrok` project's disabled message will be:
```json
{"messageLevel": "info", "duration": "PT5M", "tags": ["opengrok"], "text": "disabled project"}
```
while `opengrok-1.0`'s will be:
```json
{"messageLevel": "info", "duration": "PT5M", "tags": ["opengrok-1.0"], "text": "disabled project: foo bar"}
```

### DIFF
--- a/opengrok-tools/requirements.txt
+++ b/opengrok-tools/requirements.txt
@@ -11,4 +11,4 @@ requests==2.20.0
 Resource==0.2.1
 six==1.11.0
 urllib3==1.23
-GitPython==3.0.2
+GitPython==3.0.6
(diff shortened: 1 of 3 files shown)


ktulinger:
What about another configuration property like `disabled-reason` or similar instead of introducing this variable duality?

vladak:
had to bump GitPython version due to https://github.com/gitpython-developers/GitPython/issues/983

vladak:
> What about another configuration property like `disabled-reason` or similar instead of introducing this variable duality?

yes, that would be cleaner.

vladak:
Thanks for the review !

vladak:
will update https://github.com/oracle/opengrok/wiki/Repository-synchronization

PR #3138, opened 2020-05-06
Directories #Lines and LOC

idodeclare:
Hello,

Please consider for integration this patch to compute and store aggregated directory #Lines and LOC.

This patch will do a quick computation from file #Lines and LOC to update an existing index and thereafter recompute using deltas.

~`IndexAnalysisSettings4` is bumped....~

Thank you.

image


idodeclare:
Is Travis working OK?

idodeclare:
@tulinkry, there are OpenGrok users with deployments that take a week or more to fully index (and multiple weeks pre-1.1), so avoiding “reindex needed” should be paramount whenever possible.

It’s why I added `AnalyzerGuru.getVersionNo` and `FileAnalyzer.getVersionNo` and related handling. It’s why we ask whether reindex is needed for a Lucene update.

tulinkry:
I understand and that's why we introduced the versioning scheme like we have it. The need for reindex can come from other changes than Lucene.

 I no here to decide what is better or not, just wanted to point the stuff for the others who reads this (and can decide), that it might by little over to introduce a non-trivial amount of code handling a index versions transitions.

tulinkry:
Does it affect the index time?

idodeclare:
It's not a lot of logic to support the index transition. We just need to query existing `NUML` and `LOC` from Lucene and then run the numbers through the new aggregator.

> Does it affect the index time and size?

Yes, there's a new `Document` for every directory, but after the first run, for each incremental indexing, only directories with changed files (and parent directories thereof up to the root) need to be modified.

idodeclare:
Rebased with a fix for the issue from the last Travis test run

opengrok-indexer/src/main/java/org/opengrok/indexer/util/NumberUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
+
+    /**
+     * Parses the specified {@code value} without throwing a checked exception,
+     * returning a default 0.
+     */
+    public static long tryParseLongPrimitive(String value) {
```
tulinkry, line 45: i think the default should be a parameter for this so the caller can specify what it expects in case of a non-long string

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/NullableNumLinesLOC.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public Long getNumLines() {
+        return numLines;
+    }
+
+    public Long getLOC() {
+        return loc;
+    }
+}
```
tulinkry, line 52: I don't like this duplication, any way around this with Optionals, or simply any other way than just duplicating class?

idodeclare: `NullableNumLinesLOC` is a renamed `FileExtra` where possible nullability is acknowledged for reporting purposes (e.g. for files that have no #Lines/LOC such as ZIP files). For the analysis and aggregation of this data, it's useful to leave out nullability, so `NumLinesLOC` exists with `long` and not `Long`.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/NumberUtil.java
idodeclare: Shouldn't we wait for that use case?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/NullableNumLinesLOC.java
ktulinger: Understood, and what about forexample method `supportsLoc` that would return `true` for the cases you mentioned and then you could read the primitive values and `false` when reading these values wouldn't be meaningful. Basically - using only one class and hiding the ugly logic (`loc != null && lines != null`) inside it?

tulinkry: Understood, and what about forexample method supportsLoc that would return true for the cases you mentioned and then you could read the primitive values and false when reading these values wouldn't be meaningful. Basically - using only one class and hiding the ugly logic (loc != null && lines != null) inside it?

tulinkry: Just an idea

opengrok-indexer/src/main/java/org/opengrok/indexer/util/NumberUtil.java
tulinkry: alright

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -505,6 +534,17 @@ public void update() throws IOException {
                         }
                     }
 
+                    /*
+                     * Since we are avoiding an IndexAnalysisSettings bump to
+                     * store needed state related to #Lines/LOC, we need to
+                     * ensure that at least one D-document gets saved. Record
+                     * zeroes for a .foo under sourceRoot to do this.
+                     */
+                    countsAggregator.register(new NumLinesLOC("/.foo", 0, 0));
```
ktulinger, line 543: I'm not sure if I follow this logic. Could you extract the filename to a constant?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/NullableNumLinesLOC.java
idodeclare: Your idea would increase cyclomatic complexity everywhere the current non-nullable class is used. I don't support

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
idodeclare: I don't understand how a constant will help anyone follow if that long comment cannot do.

tulinkry: These two weren't connected.

idodeclare: Oh. OK I extracted a constant and updated the comment hopefully for clarity.

tulinkry: Understood, nice.

idodeclare:
Just trivial conflicts upon rebase

idodeclare:
Rebased for giggles

vladak:
Thanks Chris and sorry it took so long.

PR #3139, opened 2020-05-06
Fix #3078 : record user intent for projects

### LINKED ISSUE #3078: strange behavior of 'select all', 'invert' and 'clear' buttons under project(s)
I find that buttons to select all projects, invert/clear selection don't do what I expect.

**To Reproduce**

On OG home page:

1. click select all (projects)
2. click invert (project selection)
3. type search text in full text search
4. click search

Should: no results because of no project selected
Actual: search results from all projects

OG version 1.3.8

idodeclare:
Hello,

Please consider for integration this patch that tracks user intent on the search window so that a user who removes all projects or clears the project selection ends up doing a search against zero projects instead of unexpectedly reviving the project list from cookies in that case.

Also this patch adds options to searchable-option-list-XXX.js to avoid hard-coding OpenGrok-specific elements. This should allow upstreaming OpenGrok updates.

Thank you.


opengrok-web/src/main/webapp/js/searchable-option-list-2.0.13.js
```diff
@@ -1093,9 +1097,14 @@
                         /*
                          * Modified for OpenGrok in 2017.
                          */
-                        if (isOnSearchPage()) {
-                            $('#xrd').val("1"); // no redirect
-                            $('#sbox').submit();
+                        if (self.config.quickDeleteForm) {
+                            if (self.config.quickDeletePermit) {
```
tulinkry, line 1101: I saw the original author using

`$.isFunction(this.config.events.onChange)`

is it worth to keep it consistent?

idodeclare: I'm a JS noob. Can you clarify?

idodeclare:
Ugh and I left a comma off accidentally. Bad merge. It's hard with these JS files changing names.

opengrok-web/src/main/webapp/js/searchable-option-list-2.0.13.js
tulinkry: I mean that lower below this code the original author wrote a check for function existence like

```
                if ($.isFunction(this.config.events.onChange)) {
                    this.config.events.onChange.call(this, this, $changedInputs);
                }
```

so perhaps we want to use it as well like

```
if ($.isFunction(self.config.quickDeletePermit))
```

to be consistent with the author's style

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -923,10 +923,6 @@ private static void splitByComma(String value, List<String> result) {
         TreeSet<String> projectNames = new TreeSet<>();
         List<Project> projects = getEnv().getProjectList();
 
-        if (projects == null) {
-            return projectNames;
-        }
-
```
tulinkry, line 929: why is this no longer needed?

opengrok-web/src/main/webapp/js/searchable-option-list-2.0.13.js
tulinkry: however, looking at the doc https://api.jquery.com/jQuery.isFunction/ the function is deprecated, so keep it as you have it

tulinkry:
The comments go wild a little, I hope you catch where it belongs.

opengrok-web/src/main/webapp/js/searchable-option-list-2.0.13.js
idodeclare: OK thank you

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
idodeclare: `projects` is always defined because `RuntimeEnvironment` does a `new ArrayList`.

idodeclare:
Rebased and fixed that missing comma

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
tulinkry: ok, nice

idodeclare:
Thank you for reviewing, @tulinkry

idodeclare:
Approved by @tulinkry on 7 May; what do you require for a merge?

tarzanek:
same as #3140 , I've restarted travis and will merge shortly
thank you Chris, appreciate it and sorry for delays :-)

idodeclare:
Thank you, Lubos!

tarzanek:
had a quick look
and the only regression I see is how will OpenGrok restore the last searched projects from cookies ...
or this use case is no longer valid?
(search for multiple projects, close browser, start browser to get your projects restored)

idodeclare:
That behavior you describe would not be affected.

This is just when the user has actively revised projects and selected none (possibly inadvertently) that cookie-projects won’t be restored to search against. OpenGrok will instead advise “You must select a project.”

(You could click Search a second time, and cookie projects [if available] would then be used.)

PR #3140, opened 2020-05-06
Regular readable time

idodeclare:
Hello,

Please consider for integration this patch so `PendingFileCompleter` uses the regular OpenGrok `getReadableTime()` instead of `DurationFormatUtils` so that sub-second detail is provided.

`StringUtils.getReadableTime()` is updated to use `StringBuilder` and to avoid some pointless high-precision (e.g., "1 day 35 ms") when for other intervals milliseconds are not shown. It is also updated not just to report "0" for near-instant intervals but "0 ms".

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/StringUtils.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/StringUtils.java
@@ -124,45 +124,53 @@ public final class StringUtils {
         return javaClassPattern.matcher(s).matches();
     }
 
-  /**
-   * Convert value in milliseconds to readable time.
-   * @param time_ms delta in milliseconds
-   * @return human readable string   
-   */
-  public static String getReadableTime(long time_ms) {
-      String output = "";
-      long time_delta = time_ms;
+    /**
+     * Convert value in milliseconds to readable time.
+     * @param time_ms delta in milliseconds
+     * @return human readable string
+     */
+    public static String getReadableTime(long time_ms) {
+        StringBuilder output = new StringBuilder();
+        long time_delta = time_ms;
 
-      int milliseconds = (int) (time_delta % 1000);
-      time_delta /= 1000;
-      int seconds = (int) (time_delta % 60);
-      time_delta /= 60;
-      int minutes = (int) (time_delta % 60);
-      time_delta /= 60;
-      int hours = (int) (time_delta % 24);
-      int days = (int) (time_delta / 24);
+        int milliseconds = (int) (time_delta % 1000);
+        time_delta /= 1000;
+        int seconds = (int) (time_delta % 60);
+        time_delta /= 60;
+        int minutes = (int) (time_delta % 60);
+        time_delta /= 60;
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/StringUtils.java
```diff
@@ -116,53 +116,61 @@ public static boolean isOnlyWhitespace(String str) {
...
-    //    - class must be qualified with a package name
-    //    - only letters in package name, starting with lower case
-    //    - class name must be in CamelCase, starting with upper case
-    return javaClassPattern.matcher(s).matches();
-  }
+        // Only match a small subset of possible class names to prevent false
+        // positives:
+        //    - class must be qualified with a package name
+        //    - only letters in package name, starting with lower case
+        //    - class name must be in CamelCase, starting with upper case
```
tulinkry, line 123: consider moving this comment to the javadoc as it's valuable for the caller

opengrok-indexer/src/test/java/org/opengrok/indexer/util/StringUtilsTest.java
```diff
@@ -39,12 +39,12 @@ public void testValues() {
...
-            1075634299
+            1075634299, 86480001
         };
         String[] expected = {
-            "0", "100 ms", "1.0 seconds", "1.500 seconds", "0:01:04",
-            "0:02:04", "1:00:51", "1 day", "1 day 34.349 seconds",
-            "12 days 10:47:14"
+            "0 ms", "100 ms", "1.0 seconds", "1.500 seconds", "0:01:04",
+            "0:02:04", "1:00:51", "1 day", "1 day",
+            "12 days 10:47:14", "1 day 00:01:20"
```
tulinkry, line 47: could you add a `86400001` (one day + one ms) too?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/StringUtils.java
idodeclare: OK done

opengrok-indexer/src/test/java/org/opengrok/indexer/util/StringUtilsTest.java
idodeclare: OK

idodeclare:
Approved by @tulinkry on 6 May; what do you require for a merge?

tarzanek:
from my point of view only the travis mac build failed, let me rerun, if it will be green, me or Vlada can merge (I'd still love Oracle to give back the repos, so we can add more people with merge permissions)

tarzanek:
ok, I've rerun the job, but I doubt it has anything to do with your changes (was a download error)

vladak:
The last Travis build seems to be okay, merging.

vladak:
Thanks !

idodeclare:
Thank you, @tarzanek and @vladak !

PR #3143, opened 2020-05-06
Fix to restore "submit" type

idodeclare:
Tested also to make sure the `#xrd` functionality still works with this patch.


vladak:
Thanks guys !

PR #3144, opened 2020-05-07
fix equivalence check in Command class

vladak:
This change fixes some equivalence checks (as opposed to identity checks) as reported by lgtm.com.


vladak:
Thanks !

PR #3152, opened 2020-05-23
Fix #3028 : on Clear, reset the radio to Any

### LINKED ISSUE #3028: clear button does not clear file type field

idodeclare:
Hello,

Please consider for integration this patch to reset the `#type` radio to Any when one clicks to Clear.

Thank you.


vladak:
It seems that this solution relies on the fact that there is just one radio button based droplist in the UI, that is fine as there are no plans to extend the UI.

This will have to be adapted for #2580.

idodeclare:
> It seems that this solution relies on the fact that there is just one radio button based droplist in the UI, that is fine as there are no plans to extend the UI.

That's incorrect. See the patch where this is applied only to `$("#type")`.

> This will have to be adapted for #2580.

It's fairer to write that whoever (ever?) implements #2580 will have to ensure that "clear file type" is not broken if this PR is accepted. That should be straight-forward since `deselectAll()` is well-exercised (see the Projects drop-down).

tarzanek:
from my side feel free to merge

idodeclare:
Rebased to bump JS file names after #3139

tarzanek:
merging, thank you Chris!

PR #3156, opened 2020-05-28
Redirect when 1 hit for query of just PATH

idodeclare:
Hello,

Please consider for integration this patch to redirect directly to a file when a search by `PATH` alone produces a single hit.

I noticed this when clicking on OpenGrok's linking of something file-like or path-like (in a comment for example), and I was taken to a search result with one item. This feels weird given the recent #3053.

Thank you.


vladak:
Thank you. The redirect feature is boon.

idodeclare:
Thank you for reviewing, @vladak.

tulinkry:
Yes, it's awesome!

idodeclare:
Glad you like it!

Issue #3153, opened 2020-05-25
Docker: OPENGROK_WEBAPP_CONTEXT doesn't have full effect


timschumi:
**Describe the bug**

When starting a Docker container, OPENGROK_WEBAPP_CONTEXT does not have full effect and the webpage is not available at the chosen path.

**To Reproduce**

1. Set up Docker using [this](https://gist.githubusercontent.com/timschumi/b5bb13111efc7aa2195defcc482a5d61/raw/fc4f0acc6d8296c761b09df98e7a09931e4fff03/docker-compose.yml) docker-compose file.
2. Create `src`, `data` and `etc` folders in the same directory.
3. Start the container.
4. Access http://localhost:8080/opengrok
5. Receive a "File not found" page due to OpenGrok still using / as its root.

**Expected behavior**
OpenGrok fully honors the OPENGROK_WEBAPP_CONTEXT variable and shows the main page when accessing /opengrok (or whatever was set).

**Screenshots**
[Main page](https://user-images.githubusercontent.com/16820960/82769096-3cdf2880-9e33-11ea-8daa-ab73806e0e11.png) and [/opengrok page](https://user-images.githubusercontent.com/16820960/82769100-48caea80-9e33-11ea-936e-127e598b95a8.png) when setting OPENGROK_WEBAPP_CONTEXT="/opengrok". The port has just been used during my testing, the linked docker-compose.yml launches the service on the default port 8080.

**Additional context**

I've tested that the OPENGROK_WEBAPP_CONTEXT variable is successfully propagated to the application. The issue rather is that [in the Dockerfile](https://github.com/oracle/opengrok/blob/52fc2931f25089055b5cf1dea11f846521d748ef/Dockerfile#L75), the OpenGrok .war file is explicitly installed as ROOT.war. Hardcoding that to something different moves the application to the corresponding directory (but this obviously isn't very portable and the Dockerfile isn't intended for that). One solution would be to move/link that file at runtime, but I'm unsure whether that is desired or how that should be done.

vladak:
The `opengrok-deploy` command `RUN` should honor the `OPENGROK_WEBAPP_CONTEXT` variable. There is also a bit of a trickery done there with the redirect from `/source` to `/` , not sure this is necessary:

https://github.com/oracle/opengrok/blob/1c8c7ecb11894fe72ab40888ce85e2dcfb8752a4/Dockerfile#L72-L77

[...]

Issue #3054, opened 2020-02-25
Setting up OPENGROK_WEBAPP_CONTEXT doesn't change the root directory


jbaek7023:
I made a change to the Dockerfile to server OpenGrok on `/source` instead of `/`. But it doesn't work.

```
# environment variables
ENV SRC_ROOT /opengrok/src
ENV DATA_ROOT /opengrok/data
ENV OPENGROK_WEBAPP_CONTEXT /source <<<<<<<<<<<<<<<<<<<<<<<<<My change here
ENV OPENGROK_TOMCAT_BASE /usr/local/tomcat
ENV CATALINA_HOME /usr/local/tomcat
ENV CATALINA_BASE /usr/local/tomcat
ENV CATALINA_TMPDIR /usr/local/tomcat/temp
ENV PATH $CATALINA_HOME/bin:$PATH
ENV JRE_HOME /usr
ENV CLASSPATH /usr/local/tomcat/bin/bootstrap.jar:/usr/local/tomcat/bin/tomcat-juli.jar

# custom deployment to / with redirect from /source
# RUN rm -rf /usr/local/tomcat/webapps/* && \
#    opengrok-deploy -c /opengrok/etc/configuration.xml \
#        /opengrok/lib/source.war /usr/local/tomcat/webapps/source.war && \
#    mkdir "/usr/local/tomcat/webapps/source" && \
#    echo '<% response.sendRedirect("/"); %>' > "/usr/local/tomcat/webapps/source/index.jsp"
```

I changed `ENV OPENGROK_WEBAPP_CONTEXT /` to `ENV OPENGROK_WEBAPP_CONTEXT /source`. But it's still redirecting to `/` and the default URL is still `/`

I commented out the `RUN rm -rf ... opengrok-deploy` but it's now showing 404.

jbaek7023:
Here is the change I made to the Dockerfile

```
# environment variables
ENV SRC_ROOT /opengrok/src
ENV DATA_ROOT /opengrok/data
ENV OPENGROK_WEBAPP_CONTEXT /source
ENV OPENGROK_TOMCAT_BASE /usr/local/tomcat
ENV CATALINA_HOME /usr/local/tomcat
ENV CATALINA_BASE /usr/local/tomcat
ENV CATALINA_TMPDIR /usr/local/tomcat/temp
ENV PATH $CATALINA_HOME/bin:$PATH
ENV JRE_HOME /usr
[...]

PR #3160, opened 2020-06-02
docker: Fix OPENGROK_WEBAPP_CONTEXT behaviour.

timschumi:
This allows the user to correctly specify the URL subfolder that OpenGrok should run on,
and the container appropriately reacts without having to fully rebuild it.

Fixes: #3153

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -5,10 +5,29 @@ if [ -z "$REINDEX" ]; then
 	REINDEX=10
 fi
 
+if [ "${OPENGROK_WEBAPP_CONTEXT}" = "/" ]; then
+	WAR_NAME="ROOT.war"
+else
+	WAR_NAME="${OPENGROK_WEBAPP_CONTEXT#/}.war"
+fi
+
+if [ ! -f "/usr/local/tomcat/webapps/${WAR_NAME}" ]; then
+	date +"%F %T Deployment path changed. Redeploying..."
+
+	# Delete old deployment and (re)deploy
+	rm -rf /usr/local/tomcat/webapps/*
+	opengrok-deploy -c /opengrok/etc/configuration.xml \
+            /opengrok/lib/source.war /usr/local/tomcat/webapps/${WAR_NAME}
+
+	# Set up redirect from /source
+	mkdir "/usr/local/tomcat/webapps/source"
+	echo "<% response.sendRedirect(\"${OPENGROK_WEBAPP_CONTEXT}\"); %>" > "/usr/local/tomcat/webapps/source/index.jsp"
+fi
+
 indexer(){
 	# Wait for Tomcat startup.
 	date +"%F %T Waiting for Tomcat startup..."
-	while [ "`curl --silent --write-out '%{response_code}' -o /dev/null 'http://localhost:8080/'`" == "000" ]; do
+	while [ "`curl --silent --write-out '%{response_code}' -o /dev/null 'http://localhost:8080/${OPENGROK_WEBAPP_CONTEXT#/}'`" == "000" ]; do
 		sleep 1;
 	done
 	date +"%F %T Startup finished"
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -17,7 +17,7 @@ if [ ! -f "/usr/local/tomcat/webapps/${WAR_NAME}" ]; then
 	# Delete old deployment and (re)deploy
 	rm -rf /usr/local/tomcat/webapps/*
 	opengrok-deploy -c /opengrok/etc/configuration.xml \
-            /opengrok/lib/source.war /usr/local/tomcat/webapps/${WAR_NAME}
+            /opengrok/lib/source.war "/usr/local/tomcat/webapps/${WAR_NAME}"
 
 	# Set up redirect from /source
 	mkdir "/usr/local/tomcat/webapps/source"
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -5,19 +5,19 @@ if [ -z "$REINDEX" ]; then
 	REINDEX=10
 fi
 
-if [[ "${OPENGROK_WEBAPP_CONTEXT}" = *" "* ]]; then
+if [[ "${URL_ROOT}" = *" "* ]]; then
 	date +"%F %T Deployment path contains spaces. Deploying to root..."
-	export OPENGROK_WEBAPP_CONTEXT="/"
+	export URL_ROOT="/"
 fi
 
 # Remove leading and trailing slashes
-OPENGROK_WEBAPP_CONTEXT="${OPENGROK_WEBAPP_CONTEXT#/}"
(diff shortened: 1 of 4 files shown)


timschumi:
I hereby accept the [Oracle Contributor Agreement](https://www.oracle.com/technetwork/oca-405177.pdf) for this Pull Request. A properly signed version of this form should be on it's way.

timschumi:
As an afterthought, since the `OPENGROK_WEBAPP_CONTEXT` was (as far as I know) deprecated and isn't in use anymore, maybe a different variable name should be used to avoid confusion.

docker/start.sh
```diff
@@ -5,10 +5,29 @@ if [ -z "$REINDEX" ]; then
...
+	WAR_NAME="${OPENGROK_WEBAPP_CONTEXT#/}.war"
+fi
+
+if [ ! -f "/usr/local/tomcat/webapps/${WAR_NAME}" ]; then
+	date +"%F %T Deployment path changed. Redeploying..."
+
+	# Delete old deployment and (re)deploy
+	rm -rf /usr/local/tomcat/webapps/*
+	opengrok-deploy -c /opengrok/etc/configuration.xml \
+            /opengrok/lib/source.war /usr/local/tomcat/webapps/${WAR_NAME}
```
idodeclare, line 20: Nit: perhaps quote the variable expansion?

docker/start.sh
```diff
@@ -5,10 +5,29 @@ if [ -z "$REINDEX" ]; then
 	REINDEX=10
 fi
 
+if [ "${OPENGROK_WEBAPP_CONTEXT}" = "/" ]; then
+	WAR_NAME="ROOT.war"
+else
+	WAR_NAME="${OPENGROK_WEBAPP_CONTEXT#/}.war"
```
idodeclare, line 11: In Tomcat, one might specify a context path of e.g. `/source/opengrok` with an archive named `source#opengrok.war`. Might that be supported with substitution here?

docker/start.sh
```diff
@@ -5,10 +5,29 @@ if [ -z "$REINDEX" ]; then
 	REINDEX=10
 fi
 
+if [ "${OPENGROK_WEBAPP_CONTEXT}" = "/" ]; then
```
idodeclare, line 8: If `OPENGROK_WEBAPP_CONTEXT` is empty or contains whitespace, should it be reset earlier to `/` to benefit this conditional and later usage?

docker/start.sh
timschumi: Of course. Not sure how I missed that. I'll fix it with a later patch set.

docker/start.sh
timschumi: I'm unsure whether there are any advantages to that compared to simply filling in the URL-subfolder as-is (with minor changes to unify the format).

I'd also prefer the path-based version personally, since it's easier to handle and understand (in my opinion).

docker/start.sh
timschumi: Sounds good, but we'll have to take a look at format restrictions for the variable later anyways (leading slash, trailing slash, etc.).

docker/start.sh
idodeclare: My question should have been more general: what will happen if the user specifies `/source/opengrok` as a context?

timschumi: I just noticed that what I intended doesn't actually work. I thought that tomcat resolves the used .war file by the full path (i.e. that a path of `/usr/share/tomcat/webapp/source/opengrok.war` would result in OpenGrok being accessible at `/source/opengrok`, although `/source` is a whole other story), but that apparently only works for `ROOT` and first-level files.

I'll do some testing with substitutions and add a commit if everything seems fine.

timschumi:
One thing that I definitely wanted to suggest/discuss is changing the variable name to a different one.

`OPENGROK_WEBAPP_CONTEXT` has been deprecated/has not been in use for some time, so we might want to rename it to something different.

idodeclare:
If `OPENGROK_WEBAPP_CONTEXT` was not even functional in the `docker/*` area, then I do not suppose anyone would feel strongly about your retiring it in this PR and proposing a new name.

timschumi:
Those build failures seem unrelated to my changes, do they?

idodeclare:
The Mac and Windows builds provide useful info but are unreliable because of environmental issues. Often they succeed, but sometimes they do not.

I agree that the failure seems unrelated to your changes.

tarzanek:
I remember opengrok-deploy should have an option to deploy to a different webapp root, @tulinkry ?
So some parts of this code might not be needed ...

timschumi:
> I remember opengrok-deploy should have an option to deploy to a different webapp root, @tulinkry ?
> So some parts of this code might not be needed ...

As far as I can tell, this was only a thing with the old bash scripts (that was precisely what `OPENGROK_WEBAPP_CONTEXT` was used for).

The new python scripts don't seem to have that functionality, handling everything through the raw .war path that is given by the user (see the [Python transition wiki page](https://github.com/oracle/opengrok/wiki/Python-scripts-transition-guide#no-replacement) and vladak's comments on #3054 for more information).

timschumi:
Just went through my older unread E-Mails, it appears that my signed OCA has now been properly registered. Is there anything that remains to get this ready for submission?

docker/start.sh
```diff
@@ -5,10 +5,38 @@ if [ -z "$REINDEX" ]; then
...
+if [ ! -f "/usr/local/tomcat/webapps/${WAR_NAME}" ]; then
+	date +"%F %T Deployment path changed. Redeploying..."
+
+	# Delete old deployment and (re)deploy
+	rm -rf /usr/local/tomcat/webapps/*
+	opengrok-deploy -c /opengrok/etc/configuration.xml \
+            /opengrok/lib/source.war "/usr/local/tomcat/webapps/${WAR_NAME}"
+
+	# Set up redirect from /source
+	mkdir "/usr/local/tomcat/webapps/source"
```
tarzanek, line 32: I guess these commands will print errors when run multiple times (e.g. stop/start/stop/start)
would it make sense to make them idempotent?

tarzanek: (while above was run only once when the docker container was created)

tarzanek:
but still it can be merged, the errors would be only cosmetic, so feel free to merge, @vladak ?

docker/start.sh
timschumi: Like this it's only running those commands if the .war file isn't located at the expected location (i.e. first start or after URL_ROOT has been changed), isn't it? Or am I missing something?

vladak: Looks quite okay to me with these nits:
  - if you restart the container with different `URL_ROOT` value the old one will probably be still lying around (not a big deal, just a note)
  - the exit codes of the commands are not checked (might as well run the script with `set -e`)

vladak:
Merging as is, thanks !

PR #3162, opened 2020-06-03
faster Git tags

idodeclare:
Hello,

Please consider for integration this patch to use @louie0817 bulk Git log-tags command and to fix the issue @louie0817 identified where for multiple tags for the same Git commit, only one was incorporated and the rest silently dropped.

Thank you.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -566,6 +566,37 @@ public class GitRepository extends Repository {
         return true;
     }
 
+    /**
+     * Builds a Git tag list by querying Git commit hash, commit time, and tag
+     * names.
+     * <p>Repository technically relies on the tag list to be ancestor ordered.
+     * <p>For a version control system that uses "linear revision numbering"
+     * (e.g. Subversion or Mercurial), the natural ordering in the
+     * {@link TreeSet} is by ancestor order and so
+     * {@link TagEntry#compareTo(HistoryEntry)} always determines the correct
+     * tag.
+     * <p>For {@link GitTagEntry} that does not use linear revision numbering,
+     * the {@link TreeSet} will be ordered by date. That does not necessarily
+     * align with ancestor order. In that case,
+     * {@link GitTagEntry#compareTo(HistoryEntry)} that compares by date can
+     * find the wrong tag.
+     * <p>Linus Torvalds: [When asking] "'can commit X be an ancestor of commit
+     * Y' (as a way to basically limit certain algorithms from having to walk
+     * all the way down). We've used commit dates for it, and realistically it
+     * really has worked very well. But it was always a broken heuristic."
+     * <p>"I think the lack of [generation numbers] is literally the only real
+     * design mistake we have [in Git]."
+     * <p>"We discussed adding generation numbers about 6 years ago [in 2005].
+     * We clearly *should* have done it. Instead, we went with the hacky `let's
+     * use commit time', that everybody really knew was technically wrong, and
+     * was a hack, but avoided the need."
+     * <p>If Git ever gets standard generation numbers,
+     * {@link GitTagEntry#compareTo(HistoryEntry)} should be revised to work
+     * reliably in all cases akin to a version control system that uses "linear
+     * revision numbering."
+     * @param directory a defined directory of the repository
+     * @param interactive true if in interactive mode
+     */
     @Override
     protected void buildTagList(File directory, boolean interactive) {
         this.tagList = new TreeSet<>();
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -568,68 +566,74 @@ boolean hasFileBasedTags() {
...
-            TagEntry tagEntry = buildTagEntry(directory, tag, interactive);
-            // Reverse the order of the list
-            this.tagList.add(tagEntry);
+            // In case of partial success, do not null-out tagList here.
         }
+
+        /*
+         * Note that the former algorithm ran `git tag` first. `git tag` lists
+         * tags in lexicographic order, so the former algorithm with its final
+         * reverse() would have produced a result that was reverse-
```
vladak: nit: the reversal was done via traversal of the acquired list and appending to the end of the new list.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -568,68 +566,74 @@ boolean hasFileBasedTags() {
...
+         * lexicographically ordered.
+         *
+         * Repository technically relies on the tag list to be reverse-ancestor
```
vladak: I wonder if this could be turned into javadoc with proper references to `Repository` etc.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -568,68 +566,74 @@ boolean hasFileBasedTags() {
...
                     new Object[]{directory.getAbsolutePath(), String.valueOf(status)});
-            this.tagList = null;
-        }
-        
-        // Now get hash & date for each tag.
-        for (String tag : tagsList) {
-            TagEntry tagEntry = buildTagEntry(directory, tag, interactive);
-            // Reverse the order of the list
-            this.tagList.add(tagEntry);
+            // In case of partial success, do not null-out tagList here.
```
vladak, line 597: This will basically lead to incomplete history information that cannot be restored later on. Better than the previous approach however makes me wondering if tearing the whole thing down would be in order.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitTagParserTest.java
```diff
@@ -0,0 +1,233 @@
...
+    public void shouldParseOpenGrokTagsLog() throws IOException {
+        final byte[] OGK_LOG_BYTES = OGK_LOG.getBytes(StandardCharsets.UTF_8);
+        ByteArrayInputStream bytesIn = new ByteArrayInputStream(OGK_LOG_BYTES);
+
+        TreeSet<TagEntry> tagList = new TreeSet<>();
+        GitTagParser parser = new GitTagParser(tagList);
+        parser.processStream(bytesIn);
+
+        /*
+         * The following comparison works because there is no trailing space in
```
vladak, line 206: nit: the real output from `git log --tags ..` has newline at the end

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitTagEntry.java
```diff
@@ -19,22 +19,23 @@
...
- * HistoryEntry.
+ * {@link HistoryEntry}.
  *
  * @author Stanislav Kozina
  */
-public class GitTagEntry extends TagEntry {
+class GitTagEntry extends TagEntry {
 
-    private final String hash;
+    final String hash;
```
vladak, line 36: introduce a getter ?

vladak:
This looks good, especially thanks for the extensive comments. I only had couple of nits.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
idodeclare: Thank you for taking a look, @vladak . Your review didn't include the commit I pushed last night after I realized that `tagList` is not actually a `List` but a `SortedSet`. So my analysis was incorrect. I updated it to retract that the old and new algorithm produce any different result. They produce just about the same `SortedSet` — and there is still the Git-related issue of `date` as a "broken heuristic" for ancestry.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
idodeclare: Throwing an exception would not (in current implementation) avoid the subsequent indexing stage of the project, so the end result would still be missing history information.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitTagParserTest.java
idodeclare: Oh you're right

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: True, that's a long present problem. Just wanted to mention for completeness - I am not remotely sure that terminating the whole history due to tag problem is the right thing to do.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: That's right, I just clicked on the initial changeset and forgot about the rest.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
idodeclare: OK

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitTagEntry.java
idodeclare: OK

vladak:
Thanks guys !

tarzanek:
oh this is perfect!
thank you guys, I will retest if this fixes the lengthy history problem on my (lots of git repos) setup too! :-)

idodeclare:
Cool! Fingers crossed, @tarzanek

Issue #1167, opened 2016-08-18
git history ignores merge commits


tulinkry:
Git history ignores the merge commits for particular files. When the merge commit did an actual change to the code then there is an inconsistent view on the file - file contains the correct current version but it's not visible which commit did the current version.

An example is file `src/org/opensolaris/opengrok/analysis/c/CXref.lex` in the OpenGrok.
The xref contains current:

```
120 <SCOMMENT> {
121 {WhiteSpace}*{EOL}      {
122     out.write(yytext());
123     yypop();
124     startNewLine();}
125 }
```

However according to the last history the content should be this:

```
119 <SCOMMENT> {
120 {WhiteSpace}*{EOL}      { yybegin(YYINITIAL); out.write("</span>");
121                  startNewLine();}
122 }
```

Merge commits are present in the git log command however they contain empty file list. Therefore they are not assigned to particular files and not displayed in the app.

Is there any intention in this?

vladak:
Until `git log` can somehow display the filenames for merge commits, there is probably not much we can do about this.

tulinkry:
Switches `-c --cc` kinda solves this for me, at least it seems like that.

vladak:
Looks like `-c -cc` is for `git diff` only ?

tulinkry:
Closed by #3167

PR #3167, opened 2020-06-14
Use git log -m, and handle octopus merges too

idodeclare:
Hello,

Please consider for integration this patch to use the `-m` switch with `git log` to support octopus merges as well as to properly report if new files come into existence during a merge.

(This is an extract from a branch of mine actually named `bugfix/git_parsing` from last year which I realize now was intended at first to support the above but where I had gone down a deep rabbit hole trying to speed up history for the FreeBSD repo and arrived at several time-consuming but lackluster results. I put the branch on the back burner forgetting its original purpose.)

Users will not be forced to reindex, but any merge history previously hidden will require a reindex to be visible.

Thank you.


vladak:
What happens when diffing a merge changeset and regular changeset in history view ?

vladak:
Also, seems like there are some Windows build failures related to the changes.

idodeclare:
> What happens when diffing a merge changeset and regular changeset in history view ?

@vladak , it's no different from any other diff. Reminder that OpenGrok starts two processes (in parallel thanks to #2955) to get the diff data for the two commit hashes. A merge has a commit hash too.

Also I pushed a change that hopefully fixes the Windows build.

vladak:
Sounds good. Thanks !

tulinkry:
I think this touches also #1167, @idodeclare do you think this resolves it?

idodeclare:
> I think this touches also #1167, @idodeclare do you think this resolves it?

@tulinkry, yes I think it does.

tulinkry:
Thanks, closing the issue

PR #3174, opened 2020-07-14
JFlex 1.8.2

vladak:
Upgrades JFlex to the most recent version. The most visible change there is the conversion of `yychar` from `int` to `long`. This bubbles down to quite a lot of places. In the end this is not so useful because of interface limitations taking `int` only (like `String.substring()` or Lucene's `OffsetAttribute.setOffset()`), so casting has to be used.


vladak:
There are some unrelated failures in `tools`, namely in the `test_mirror.py::test_incoming_retval` test case. This will be investigated separately.

idodeclare:
I already had a PR #3063 for this, but with a more targeted approach (24 files touched) to avoid infecting APIs with `long` where it could do no good and also without this PRs unchecked casting to `int`.

vladak:
Oh, closing then. Will look at your PR.

Issue #3176, opened 2020-07-20
test_mirror.py::test_incoming_retval fails with git 2.27


vladak:
As seen e.g. in https://travis-ci.org/github/oracle/opengrok/jobs/709508016 the `test_mirror.py::test_incoming_retval` sometimes fails with:
```
>               assert opengrok_tools.mirror.main() == CONTINUE_EXITVAL
E               assert 1 == 2
E                 +1
E                 -2
src/test/python/test_mirror.py:176: AssertionError
----------------------------- Captured stdout call -----------------------------
output of ['/usr/bin/git', 'pull', '--dry-run']:
Synchronizing repository /tmp/tmp0bijaz9s/cloned_repo
output of ['/usr/bin/git', 'pull', '--ff-only']:
Your configuration specifies to merge with the ref 'refs/heads/master'
from the remote, but no such ref was fetched.
----------------------------- Captured stderr call -----------------------------
failed to perform pull: command ['/usr/bin/git', 'pull', '--ff-only'] in directory /tmp/tmp0bijaz9s/cloned_repo exited with 1
failed to synchronize repository /tmp/tmp0bijaz9s/cloned_repo
=========================== short test summary info ============================
FAILED src/test/python/test_mirror.py::test_incoming_retval - assert 1 == 2
```

vladak:
This problem surfaced with new Git version. I was not able to reproduce this problem on Ubuntu 19.10 with git 2.20.1. The Travis worker uses git 2.27. Once I upgraded to git 2.28 I can reproduce the problem locally.

vladak:
This actually happens also when working with Git on command line so this is not related to GitPython:
```
$ mkdir repo
$ cd repo/
$ git init
Initialized empty Git repository in /tmp/repo/.git/
$ cd ..
$ git clone repo clone
Cloning into 'clone'...
warning: You appear to have cloned an empty repository.
done.
$ git remote -v
origin	/tmp/repo (fetch)
origin	/tmp/repo (push)
$ git pull --ff-only
Your configuration specifies to merge with the ref 'refs/heads/master'
from the remote, but no such ref was fetched.
```

vladak:
When the parent repository is populated with at least one changeset the problem goes away.

PR #3177, opened 2020-07-20
lucene 8.6

tarzanek:
@vladak @idodeclare are you OK with reflection for picking the latest codec?
fwiw the tools build fails in travis on linux system,
mac and windows are ok

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -57,8 +57,6 @@ import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.Response;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.standard.StandardAnalyzer;
-import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
-import org.apache.lucene.codecs.lucene84.Lucene84Codec;
 import org.apache.lucene.document.DateTools;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
@@ -410,13 +408,6 @@ public class IndexDatabase {
             IndexWriterConfig iwc = new IndexWriterConfig(analyzer);
             iwc.setOpenMode(OpenMode.CREATE_OR_APPEND);
             iwc.setRAMBufferSizeMB(env.getRamBufferSize());
-            /**
-             * Most data in OpenGrok is indexed but not stored, so use the best
-             * compression on the minority of data that is stored, since it
-             * should not have a detrimental impact on overall throughput.
-             */
-            iwc.setCodec(new Lucene84Codec(
-                Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
             writer = new IndexWriter(indexDirectory, iwc);
             writer.commit(); // to make sure index exists on the disk
             completer = new PendingFileCompleter();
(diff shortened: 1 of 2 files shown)


vladak:
The tools build failure is tracked by #3176, will look into it.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -415,8 +417,18 @@ public void update() throws IOException {
              * compression on the minority of data that is stored, since it
              * should not have a detrimental impact on overall throughput.
              */
-            iwc.setCodec(new Lucene84Codec(
-                Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
```
vladak, line 419: So, why is this no longer accessible ? And why do we actually need Lucene 5.0 related value ?

tarzanek: good question,
it is still needed, I am just doing it with dynamic variable/class instead of hardcoded LuceneXXCodec

tarzanek: but I didn't find a latest way for Lucene50StoredFieldsFormat getting of compression Mode

vladak: I mean - could we replace with something else ?

vladak: the documentation still mentions it: https://lucene.apache.org/core/8_4_1/core/org/apache/lucene/codecs/lucene84/Lucene84Codec.html#Lucene84Codec-org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat.Mode-

idodeclare: Some parts of the Lucene API are stable over major releases, so `Lucene50StoredFieldsFormat` is still the "latest."

There is a public constructor for the 8.6 codec, so I don't believe reflection would be necessary:

```
public Lucene86Codec(Lucene50StoredFieldsFormat.Mode mode) {
```

vladak: exactly. I am still puzzled why to use reflection here.

idodeclare: ... and the codec is itself public:

```
public class Lucene86Codec extends Codec {
```

tarzanek: the goal is to get rid of version specific hard-codes and have upgrade automatic

idodeclare: Isn't it still effectively hard-coded since it's reflecting to seek a constructor taking `lucene50.Lucene50StoredFieldsFormat`?

Maybe I was thinking backward to write "Most data in OpenGrok is indexed but not stored, so use the best compression...." Since it's the very minority of data, `BEST_SPEED` sounds just fine too in hindsight. Perhaps no real need to `setCodec()`?

vladak: Right. I don't think we care too much about size reduction either.

tarzanek: then this is even better, we can just drop this whole section :-)

tarzanek:
cleaned the unneeded codec
the checks still fail on python, so I guess this can be merged,
@vladak ?

vladak:
is full reindex needed ?

tarzanek:
format did change from 84 to 86 so yes

PR #3185, opened 2020-08-13
Add mirroring support for Perforce repositories

bobdoah:
Although opengrok supports Perforce for history, the opengrok-mirror tool does not support syncing the repositories. It has some prerequisites, but only the same as for the history functionality. 

# OCA
I am signing on behalf of myself as an individual and no other person or entity, including my employer, has or will have rights with respect my contributions.


vladak:
@bobdoah please add your Copyright to both files so that I can merge this in.

bobdoah:
How's that @vladak?

vladak:
Looks good. Thanks !

PR #3187, opened 2020-08-18
fix incoming check and test for newer git verions

vladak:
This fixes a build failure which is preventing a release.

The `git config` command works with both old and new version of Git.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/scm/git.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/scm/git.py
@@ -46,7 +46,7 @@ class GitRepository(Repository):
                               env_vars=self.env, logger=self.logger)
         cmd.execute()
         if cmd.getretcode() != 0 or cmd.getstate() != Command.FINISHED:
-            cmd.log_error("failed to perform pull")
+            cmd.log_error("failed to configure git pull.ff")
 
     def reposync(self):
         git_command = [self.command, "pull", "--ff-only"]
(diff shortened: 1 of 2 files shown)


opengrok-tools/src/main/python/opengrok_tools/scm/git.py
```diff
@@ -58,7 +67,7 @@ def incoming(self):
                               env_vars=self.env, logger=self.logger)
         cmd.execute()
         self.logger.info("output of {}:".format(git_command))
-        self.logger.info(cmd.geterroutputstr())
+        self.logger.info(len(cmd.geterroutputstr()))
```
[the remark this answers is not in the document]
vladak: just a debugging leftover, removed.

opengrok-tools/src/main/python/opengrok_tools/scm/git.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, logger, path, project, command, env, hooks, timeout):
...
             raise RepositoryException("Cannot get git command")
 
+        # The incoming() check relies on empty output so configure
+        # the repository first to avoid getting extra output.
+        git_command = [self.command, "config", "--local", "pull.ff", "only"]
+        cmd = self.getCommand(git_command, work_dir=self.path,
+                              env_vars=self.env, logger=self.logger)
+        cmd.execute()
+        if cmd.getretcode() != 0 or cmd.getstate() != Command.FINISHED:
+            cmd.log_error("failed to perform pull")
```
idodeclare, line 49: How about `"failed to configure git pull.ff"`?

vladak: fixed

vladak:
Thanks for the review !

PR #3188, opened 2020-08-18
delete ctags temp files at the end of indexing

vladak:
This change should get rid of temporary files left around after killed Ctags processes.

I verified that ctags creates the temporary files with names matching given pattern (on Linux):
```
$ strace -e trace=openat ctags -f - opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/LinkageType.java 2>&1 >/dev/null | grep tags\.
...
openat(AT_FDCWD, "/tmp/tags.Z1MYq4", O_RDWR|O_CREAT|O_EXCL, 0600) = 3
```
and made sure the files are properly cleaned up at the end of the indexing.

Also created entries with name matching the pattern and having extra properties such as:
- non-empty directory
- file owned by `root:root` (while the indexer running as non-root user)

to verify the behavior in case of failure.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/CtagsUtil.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/CtagsUtil.java
@@ -91,7 +91,7 @@ public class CtagsUtil {
         return result;
     }
 
-    public static void deleteTempFiles() throws IOException {
+    public static void deleteTempFiles() {
         String[] dirs = {System.getProperty("java.io.tmpdir"),
                 System.getenv("TMPDIR"), System.getenv("TMP")};
 
@@ -100,7 +100,7 @@ public class CtagsUtil {
         }
     }
 
-    private static void deleteTempFiles(String directoryName) throws IOException {
+    private static void deleteTempFiles(String directoryName) {
         final Pattern pattern = Pattern.compile("tags\\.\\S{6}"); // ctags uses this pattern to call mkstemp()
 
         if (directoryName == null) {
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/CtagsUtil.java
```diff
@@ -86,6 +91,43 @@ public static boolean validate(String ctagsBinary) {
...
+        }
+
+        File dir = new File(directoryName);
+        File[] files = dir.listFiles((dir1, name) -> {
+            Matcher matcher = pattern.matcher(name);
+            return matcher.find();
+        });
+
+        for (File file : files) {
+            if (file.isFile()) {
```
idodeclare, line 121: How about simpler just as `if (!file.delete()) {` and just the log statement inside with no exception handling?

vladak: that's nicer, changed.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/CtagsUtil.java
```diff
@@ -29,10 +29,15 @@
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.logger.LoggerFactory;
 
+import java.io.File;
+import java.io.IOException;
```
idodeclare, line 33: `UnusedImports` here and for `file.Files` (next line) per build

vladak: fixed

PR #3190, opened 2020-08-21
changes to LdapAttrPlugin

vladak:
Seeing failures of `LdapAttrPlugin` in the log where log entries referenced `User` object instead of `LdapUser` object:
```
21-Aug-2020 16:31:47.512 WARNING [http-nio-8080-exec-92] opengrok.auth.plugin.LdapAttrPlugin.fillSession empty records or attribute values ou for user User{id=foo@bar.com, username=null, cookieTimestamp=null, timeouted=false, attrs={}}
```
made me realize this is the wrong user to log as `LdapAttrPlugin` deals with `LdapUser` when getting the attribute.

While at it, I followed IDEA's advice to refactor this part to avoid the else branch.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -159,7 +159,7 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
 
                     records = res.getAttrs();
                 } else {
-                    LOGGER.log(Level.FINE, "no DN for user {0} on {1}",
+                    LOGGER.log(Level.FINE, "no DN for LDAP user {0} on {1}",
                             new Object[]{ldapUser, ldapProvider});
                 }
             } catch (LdapException ex) {
--- a/plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
@@ -128,9 +128,8 @@ public class LdapAttrPlugin extends AbstractLdapPlugin {
     public void fillSession(HttpServletRequest req, User user) {
         updateSession(req, false);
 
-        LdapUser ldapUser;
-        if ((ldapUser = (LdapUser) req.getSession().
-                getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance))) == null) {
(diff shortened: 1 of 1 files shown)


ahornace:
If you are already doing some refactoring, could you please move these declarations to the place of definition/usage? I think it would make the code much easier to read.
```java
boolean sessionAllowed;
LdapUser ldapUser;
Map<String, Set<String>> records = null;
Set<String> attributeValues;
```

Also, could you please rebase?

vladak:
done and done.

plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -126,13 +126,9 @@ String getSessionAllowedAttrName() {
...
-        boolean sessionAllowed;
-        LdapUser ldapUser;
-        Map<String, Set<String>> records = null;
-        Set<String> attributeValues;
-
         updateSession(req, false);
 
+        LdapUser ldapUser;
         if ((ldapUser = (LdapUser) req.getSession().
                 getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance))) == null) {
```
ahornace, line 133: I think it looks better like this:
```suggestion
        LdapUser ldapUser = (LdapUser) req.getSession().                getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance));
        if (ldapUser == null) {
```

plugins/src/main/java/opengrok/auth/plugin/LdapAttrPlugin.java
```diff
@@ -128,9 +128,8 @@ String getSessionAllowedAttrName() {
     public void fillSession(HttpServletRequest req, User user) {
         updateSession(req, false);
 
-        LdapUser ldapUser;
-        if ((ldapUser = (LdapUser) req.getSession().
-                getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance))) == null) {
+        LdapUser ldapUser = (LdapUser) req.getSession().                getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance));
```
ahornace, line 131: I missed the whitespaces :( 
```suggestion
        LdapUser ldapUser = (LdapUser) req.getSession().getAttribute(LdapUserPlugin.getSessionAttrName(ldapUserInstance));
```

PR #3192, opened 2020-08-22
Upgrade tablesorter to 2.31.3; enable directory sort reset/restart

idodeclare:
Hello,

Please consider for integration this patch to upgrade jquery-tablesorter and to enable two options related to sorting for directory listing:

* `sortReset` so that sorting can be disabled with a third click
* `sortRestart` so that sorting always restarts from a column's `sortInitialOrder` (i.e. descending-sort for #Lines, LOC, Size, and Date) when clicking a column to begin sorting.

Thank you.


vladak:
Thanks !

PR #3193, opened 2020-08-22
Fix --help {ctags,repos}

idodeclare:
Hello,

Please consider for integration this patch to fix `--help ctags` and `--help repos`. As `exitWithHelp()` is called before the runtime environment has had its configuration updated with parsed options, `exitWithHelp()` needs to force in some settings for ctags or repos help.

Also this patch shows any user-disabled repositories with `--help repos`.

Thank you.


vladak:
Nice, thanks !

PR #3194, opened 2020-08-22
Feature/coverage

idodeclare:
This is just a patch to add test coverage for many of the zeroes in coveralls.


opengrok-web/src/test/java/org/opengrok/web/util/FileUtilTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+import java.io.FileNotFoundException;
+import java.util.UUID;
+
+/**
+ * Represents a container for tests of {@link FileUtil}.
+ */
+public class FileUtilTest {
+
+    @Test
```
vladak, line 37:
```suggestion
    @Test(expected = NoPathParameterException.class)
```

vladak: or given that we already use Junit 5 it should be possible to use `assertThrows()`

plugins/src/test/java/opengrok/auth/plugin/UserWhiteListPluginTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Before
+    public void setUp() {
+        plugin = new UserWhiteListPlugin();
+    }
+
+    @Test
+    public void shouldThrowOnLoadIfNullArgument() {
+        NullPointerException caughtException = null;
```
vladak, line 86: ditto - use JUnit mechanisms for asserting the exceptions ?

plugins/src/test/java/opengrok/auth/plugin/UserWhiteListPluginTest.java
```diff
@@ -0,0 +1,179 @@
...
+    public static void beforeClass() throws Exception {
+        tempWhitelist = File.createTempFile("UserWhiteListPluginTest", "txt");
+        try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(
+                new FileOutputStream(tempWhitelist), StandardCharsets.UTF_8))) {
+            writer.write(OK_USER);
+            // Don't bother with trailing LF.
+        }
+
+        validPluginParameters = new HashMap<>();
+        validPluginParameters.put(UserWhiteListPlugin.FILE_PARAM, tempWhitelist.getPath());
```
vladak, line 68: would be nice to see what happens if the parameters include path to non-existent file.

vladak:
This looks good, I only had a few nits. Please rebase.

plugins/src/test/java/opengrok/auth/plugin/UserWhiteListPluginTest.java
```diff
@@ -0,0 +1,179 @@
...
+        plugin.load(validPluginParameters);
+
+        DummyHttpServletRequest req = new DummyHttpServletRequest();
+        req.setAttribute(UserPlugin.REQUEST_ATTR, new User(OK_USER));
+
+        Project randomProject = new Project(RandomString.generateUpper(10));
+        boolean projectAllowed = plugin.isAllowed(req, randomProject);
+        assertTrue("should allow OK_USER for random project 1", projectAllowed);
+
+        randomProject = new Project(RandomString.generateUpper(10));
```
vladak, line 127: why test 2 projects if they are both random ?

opengrok-web/src/test/java/org/opengrok/web/util/FileUtilTest.java
idodeclare: Done

plugins/src/test/java/opengrok/auth/plugin/UserWhiteListPluginTest.java
idodeclare: Done

vladak:
Thanks !

PR #3200, opened 2020-08-31
introduce command timeout types

vladak:
This change makes the `Executor` command timeout more generic. Instead of setting the timeout based on interactivity, there will be 4 types of timeout:
  - indexer
  - interactive (think webapp through UI)
  - webapp start (think deploy)
  - RESTful

Each will be tunable.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -172,7 +173,7 @@ public final class RuntimeEnvironment {
 
     public void shutdownRevisionExecutor() throws InterruptedException {
         getRevisionExecutor().shutdownNow();
-        getRevisionExecutor().awaitTermination(getCommandTimeout(), TimeUnit.SECONDS);
+        getRevisionExecutor().awaitTermination(getIndexerCommandTimeout(), TimeUnit.SECONDS);
     }
 
     /**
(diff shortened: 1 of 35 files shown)


vladak:
While at it, I added logging to `indexParallel()` to record times to index individual files. This might help a fellow user to track where the indexer spends the time.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -227,29 +228,59 @@ public void setNestingMaximum(int nestingMaximum) {
...
+                return getIndexerCommandTimeout();
+            case INTERACTIVE:
+                return getInteractiveCommandTimeout();
+            case WEBAPP_START:
+                return getWebappStartCommandTimeout();
+            case RESTFUL:
+                return getRestfulCommandTimeout();
+        }
+
+        throw new InvalidParameterException("invalid command timeout type");
```
ahornace, line 243: Would not `IllegalArgumentException` make more sense? `InvalidParameterException` is from `security` package and is not that commonly used.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Statistics.java
```diff
@@ -28,28 +28,47 @@
...
-  }
+    /**
+     * log a message along with how much time it took since the constructor was called.
+     * @param logger logger instance
+     * @param logLevel log level
+     * @param msg message string
+     */
+    public void report(Logger logger, Level logLevel, String msg) {
+        long stopTime = System.currentTimeMillis();
+        String time_str = StringUtils.getReadableTime(stopTime - startTime);
```
ahornace, line 46: I know it was like this but since you are touching it, could you please rename to `timeStr` to follow Java standard?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Statistics.java
```diff
@@ -28,28 +28,47 @@
 import static org.opengrok.indexer.util.StringUtils.getReadableTime;
 
 public class Statistics {
-        
-  private final long startTime;  
 
-  public Statistics() {
+    private final long startTime;
```
ahornace, line 32: I still prefer using `Instant` and `Duration` classes for computing the duration as it is more high level. If you do not, then feel free to ignore.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Statistics.java
vladak: sure

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
vladak: sounds reasonable, changed.

opengrok-indexer/src/main/java/org/opengrok/indexer/util/Statistics.java
vladak: why not, changed. I like `Duration.between()`.

ahornace:
Could you please fix the following error?
```
[ERROR] /home/travis/build/oracle/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java:32:8: Unused import - java.security.InvalidParameterException. [UnusedImports]
```

vladak:
> Could you please fix the following error?
> 
> ```
> [ERROR] /home/travis/build/oracle/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java:32:8: Unused import - java.security.InvalidParameterException. [UnusedImports]
> ```

fixed

PR #3203, opened 2020-09-01
reduce logging emitted from running Java tests

vladak:
This change limits the amount of logs produced by each build significantly by setting Java log level to SEVERE, avoiding Travis build failure. Some of the indexer tests still produce small output since they use `System.out.println()`, nothing to worry about.

If some of the tests fail, it would be necessary to re-run the build with changed `mvn` invocation in the `dev/main` (i.e. add `-Dtest` per https://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html and remove the `logging.properties` argument).

### DIFF
--- a/dev/logging.properties
+++ b/dev/logging.properties
@@ -1,0 +1,18 @@
+handlers= java.util.logging.ConsoleHandler
+
+# Default global logging level.
+# This specifies which kinds of events are logged across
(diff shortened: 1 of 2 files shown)


vladak:
The problem does not seem to be happening anymore.

Issue #3206, opened 2020-09-03
Suggester does not work on Java 14


stevenschlansker:
**Describe the bug**
OpenGrok 1.4.1, Java 14.0.2

**To Reproduce**
Start up Suggester with recommended `--add-exports` modules.
It fails:
```
[...]

PR #3207, opened 2020-09-03
Suggester: upgrade chronicle-core to 2.20.20 for Java 14 support

stevenschlansker:
Proposed fix for #3206

This change is mechanical and I claim no copyright to it.  I believe I already accepted OCA in the past, but if not, I release this PR under it

Unfortunately, I was not able to build the resulting opengrok -- the build seems to crash for me -- but I did verify the POM is valid.


vladak:
Ok, let's give it a try. If it breaks something we can always back it out. Thanks !

PR #3208, opened 2020-09-03
treat all Command arguments as strings when substituting

vladak:
After adding numeric argument to the `sync.py` configuration, I got this exception:
```
  File "/usr/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/opengrok/dist/bin/venv/lib/python3.7/site-packages/opengrok_tools/sync.py", line 63, in worker
    x.run()
  File "/opengrok/dist/bin/venv/lib/python3.7/site-packages/opengrok_tools/utils/commandsequence.py", line 121, in run
    args_append=[self.name], excl_subst=True)
  File "/opengrok/dist/bin/venv/lib/python3.7/site-packages/opengrok_tools/utils/command.py", line 77, in __init__
    self.fill_arg(args_append, args_subst)
  File "/opengrok/dist/bin/venv/lib/python3.7/site-packages/opengrok_tools/utils/command.py", line 338, in fill_arg
    if pattern in cmdarg:
TypeError: argument of type 'int' is not iterable
```
This change treats all arguments as strings.

### DIFF
--- a/opengrok-tools/src/main/python/opengrok_tools/utils/command.py
+++ b/opengrok-tools/src/main/python/opengrok_tools/utils/command.py
@@ -333,6 +333,7 @@ class Command:
         newcmd = []
         subst_done = -1
         for i, cmdarg in enumerate(self.cmd):
+            cmdarg = str(cmdarg)
             if args_subst:
                 for pattern in args_subst.keys():
                     if pattern in cmdarg:
--- a/opengrok-tools/src/test/python/test_command.py
+++ b/opengrok-tools/src/test/python/test_command.py
@@ -34,10 +34,10 @@ from opengrok_tools.utils.command import Command
 
 
 def test_subst_append_default():
-    cmd = Command(['foo', '=ARG=', 'bar'],
+    cmd = Command(['foo', '=ARG=', 33, 'bar'],
                   args_subst={"ARG": "blah"},
                   args_append=["1", "2"])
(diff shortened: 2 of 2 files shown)


alls0rts (approved):
Using str() to convert numeric argument to a string.
Added test argument (33) to test that it passes.
Looking good.

vladak:
Thanks Stacey !

opengrok-tools/src/main/python/opengrok_tools/utils/command.py
```diff
@@ -333,6 +333,7 @@ def fill_arg(self, args_append=None, args_subst=None):
         newcmd = []
         subst_done = -1
         for i, cmdarg in enumerate(self.cmd):
+            cmdarg = str(cmdarg)
```
ericsaintetienne, line 336: There are other places where items in self.cmd are assumed to be strings. For example `Command.__str__()`
I'd suggest to do this assignment in `Command.__init__()` instead:
```
self.cmd = map(str, self.cmd)
```

vladak: will be done in #3225 , thanks for the neat idea !

PR #3212, opened 2020-09-04
use concurrent HashSet for parallelStream() operations

vladak:
This should help #3210. At least somewhat, there are more calls to `parallelStream()` using the list returned from `Collectors.toList()` that might fall into the same trap.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


idodeclare:
No I think this is mis-diagnosed. The parallel operation in `PendingFileCompleter` doesn't modify the hash set, so it's not the source of the exception.

#3210 logs show that a `IndexAlreadyClosedException` was caught, but then then somehow OpenGrok's special logic to quickly abort in that case — by checking `alreadyClosedCounter` — did not operate as expected, and then `IndexDatabase finishWriting()` very unexpectedly was called. If `PendingFileCompleter` somehow started operating when a still-running `addFile()` thread did its work, it could have modified the `renames` hash set and resulted in the `ConcurrentModificationException`.

I wonder if an out-of-memory condition (not necessarily `OutOfMemoryError`) could have wrought havoc in the operation. The command-line for #3210 doesn't show any options advised by [Tuning for large code bases](https://github.com/oracle/opengrok/wiki/Tuning-for-large-code-bases), so despite being on a 32GB server, wouldn't the process have hit lower limits?

vladak:
I also wondered about low memory condition (esp. in the presence of large number of repositories). Let's move the discussion to the original issue.

Issue #3211, opened 2020-09-04
click the "history" link NullPointerException


mmh891113:
**Describe the bug**
when click the "history" link show the NullPointerException error.

```
There was an error!

null

java.lang.NullPointerException
	at java.base/java.io.File.<init>(File.java:278)
	at org.opengrok.indexer.web.PageConfig.getShortPath(PageConfig.java:1609)
	at org.opengrok.indexer.web.PageConfig.getHistoryTitle(PageConfig.java:1672)
	at org.apache.jsp.history_jsp._jspService(history_jsp.java:196)
	at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:71)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
	at org.apache.jasper.servlet.JspServletWrapper.service(JspServletWrapper.java:477)
	at org.apache.jasper.servlet.JspServlet.serviceJspFile(JspServlet.java:385)
	at org.apache.jasper.servlet.JspServlet.service(JspServlet.java:329)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:231)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.opengrok.web.ResponseHeaderFilter.doFilter(ResponseHeaderFilter.java:42)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.opengrok.web.StatisticsFilter.doFilter(StatisticsFilter.java:59)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.opengrok.web.AuthorizationFilter.doFilter(AuthorizationFilter.java:46)
[...]

PR #3213, opened 2020-09-04
use getPath() for all this.path references

vladak:
Replaces all (sans `getPath()` itself) references to `this.path` in `PageConfig` with call to `getPath()`.

This should help #3211.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
@@ -763,7 +763,6 @@ public final class PageConfig {
      * current path is the "/" directory, otherwise set to an empty string.
      */
     public String[] getSearchOnlyIn() {
-        String path = getPath();
         if (isDir()) {
             return path.length() == 0
                     ? new String[]{"/", "this directory", "disabled=\"\""}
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -763,13 +763,14 @@ public Annotation getAnnotation() {
      * current path is the "/" directory, otherwise set to an empty string.
      */
     public String[] getSearchOnlyIn() {
+        String path = getPath();
```
idodeclare, line 766: How about revising line 768 rather than adding a local variable `path` hiding the field of same name

vladak: oh, that's an omission. fixed.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -766,10 +766,10 @@ public Annotation getAnnotation() {
         if (isDir()) {
             return path.length() == 0
```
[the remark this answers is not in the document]
vladak: It does, however replacing all `path` occurrences makes this easy to check and fosters good future behavior in case any copies the existing code.

idodeclare, line 767: So ... are you going to revise this line to avoid referencing the field?

vladak: um, yes. Thanks.

idodeclare (approved):

PR #3214, opened 2020-09-05
Detect if add() is called while complete() is running

idodeclare:
This will detect the bizarre situation where `add()` is called while `complete()` is running, per my wild theory re #3210.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
@@ -62,8 +62,8 @@ import org.opengrok.indexer.util.TandemPath;
  * <p>
  * No methods are thread-safe between each other. E.g.,
  * {@link #complete()} should only be called by a single thread after all
- * additions of {@link PendingFileDeletion}s and {@link PendingFileRenaming}s
- * are indicated.
+ * additions of {@link PendingSymlinkage}s, {@link PendingFileDeletion}s, and
+ * {@link PendingFileRenaming}s are indicated.
  * <p>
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/PendingFileCompleter.java
```diff
@@ -171,6 +185,15 @@ public boolean add(PendingFileRenaming e) {
...
     public int complete() throws IOException {
+        completing = true;
+        try {
+            return complete2();
+        } finally {
+            completing = false;
+        }
+    }
+
+    private int complete2() throws IOException {
```
vladak, line 196: use some more descriptive name ?

idodeclare: I used to name these like `completeInner()` or `completeWork()` but I now like to keep the name even closer to the public method, and the single `2` character is the smallest difference, yet still communicates sequence. Thoughts?

vladak: perhaps call the outer method `completeWrapper()` ?

vladak: actually, `complete()` and `completeInner()` would be fine to me as well.

vladak: Also, I saw somewhere naming scheme such as `complete()` and `doComplete()`.

idodeclare: done as `completeInner()`

vladak:
Thanks !

Issue #3218, opened 2020-09-10
not able to run indexer for svn due to "Server SSL certificate verification failed: issuer is not trusted" on docker container


rmshnair:
```
opengrok    | 10-Sep-2020 03:57:38.611 SEVERE [invalidate-repos-15] org.opengrok.indexer.util.Executor.exec Failed to read from process: svn
opengrok    |    java.io.IOException: An error occurred while parsing the xml output
opengrok    |            at org.opengrok.indexer.history.SubversionHistoryParser.processStream(SubversionHistoryParser.java:225)
opengrok    |            at org.opengrok.indexer.util.Executor.exec(Executor.java:230)
opengrok    |            at org.opengrok.indexer.history.SubversionHistoryParser.parse(SubversionHistoryParser.java:196)
opengrok    |            at org.opengrok.indexer.history.SubversionRepository.getHistory(SubversionRepository.java:292)
opengrok    |            at org.opengrok.indexer.history.SubversionRepository.determineCurrentVersion(SubversionRepository.java:412)
opengrok    |            at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:234)
opengrok    |            at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:151)
opengrok    |            at org.opengrok.indexer.history.RepositoryFactory.getRepository(RepositoryFactory.java:274)
opengrok    |            at org.opengrok.indexer.history.HistoryGuru$3.run(HistoryGuru.java:877)
opengrok    |            at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
opengrok    |            at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
opengrok    |            at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
opengrok    |            at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
opengrok    |            at java.base/java.lang.Thread.run(Thread.java:834)
opengrok    |    Caused by: org.xml.sax.SAXParseException; lineNumber: 3; columnNumber: 1; XML document structures must start and end within the same entity.
opengrok    |            at java.xml/com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:204)
[...]

PR #3219, opened 2020-09-10
install CA certificates in the opengrok Docker image

vladak:
This might or might not help #3218 however it is good idea to have this package installed in general.


vladak:
Found that the `opengrok/docker:latest` image already has the `ca-certificates` package installed.

PR #3224, opened 2020-09-14
remove project/stack specific authorization metrics

vladak:
follow-up to #2911 - removes metrics that leak project/stack names. This is necessary because:
  - the `/metrics/prometheus` endpoint is currently out in the clear
  - per project/stack metrics would cause cardinality explosion (negative pattern for monitoring)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
@@ -536,16 +536,13 @@ public final class AuthorizationFramework extends PluginFramework<IAuthorization
             lock.readLock().unlock();
         }
 
+        // Update the timers.
         authTimer.record(duration);
         if (overallDecision) {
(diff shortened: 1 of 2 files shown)


vladak:
After the change this is what the list of authorization related metrics look like:
```
$ curl -X GET -s http://localhost:8080/source/metrics/prometheus | grep ^auth
authorization_cache_misses_total 4.0
authorization_sessions_invalidated_total 0.0
authorization_negative_seconds_count 0.0
authorization_negative_seconds_sum 0.0
authorization_negative_seconds_max 0.0
authorization_positive_seconds_max 9.5E-5
authorization_positive_seconds_count 4.0
authorization_positive_seconds_sum 1.19E-4
authorization_stack_reload_total 0.0
authorization_cache_hits_total 0.0
authorization_seconds_count 4.0
authorization_seconds_sum 1.19E-4
authorization_seconds_max 9.5E-5
```

tarzanek:
lgtm
although I'd like to have an option to enable this per project if possible
but this is a new issue I guess

PR #3226, opened 2020-09-16
use tags for metrics

vladak:
This changes metric handling in `AuthorizationFramework` to use tags. Looks like this:
```
$ curl -s -X GET http://localhost:8080/source/metrics/prometheus | grep ^auth
authorization_latency_seconds_max{outcome="negative",} 0.0
authorization_latency_seconds_max{outcome="positive",} 0.0
authorization_latency_seconds_count{outcome="negative",} 0.0
authorization_latency_seconds_sum{outcome="negative",} 0.0
authorization_latency_seconds_count{outcome="positive",} 83.0
authorization_latency_seconds_sum{outcome="positive",} 0.001039
authorization_sessions_invalidated_total 0.0
authorization_stack_reload_total 0.0
authorization_cache_total{what="misses",} 83.0
authorization_cache_total{what="hits",} 58.0
```
which allows for easier creation of all-in-one graphs just like they are available for some of the JVM metrics:

grafana-opengrok-auth

Had to use build pattern to avoid https://github.com/micrometer-metrics/micrometer/issues/877

Also, I changed the metric names to use the dotted notation so they can be converted to native naming conventions used in the respective backend (https://micrometer.io/docs/concepts#_naming_meters).

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
@@ -55,14 +55,27 @@ public final class AuthorizationFramework extends PluginFramework<IAuthorization
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizationFramework.class);
 
-    private final Counter authStackReloadCounter = Metrics.getRegistry().counter("authorization_stack_reload");
-    private final Counter authCacheHits = Metrics.getRegistry().counter("authorization_cache_hits");
-    private final Counter authCacheMisses = Metrics.getRegistry().counter("authorization_cache_misses");
-    private final Counter authSessionsInvalidated = Metrics.getRegistry().counter("authorization_sessions_invalidated");
+    private final Counter authStackReloadCounter = Metrics.getRegistry().
+            counter("authorization.stack.reload");
+    private final Counter authCacheHits = Counter.builder("authorization.cache").
+            description("authorization cache hits").
+            tag("what", "hits").
+            register(Metrics.getRegistry());
+    private final Counter authCacheMisses = Counter.builder("authorization.cache").
+            description("authorization cache misses").
+            tag("what", "misses").
+            register(Metrics.getRegistry());
+    private final Counter authSessionsInvalidated = Metrics.getRegistry().
+            counter("authorization.sessions.invalidated");
 
-    private final Timer authTimer = Metrics.getRegistry().timer("authorization");
-    private final Timer authPositiveTimer = Metrics.getRegistry().timer("authorization_positive");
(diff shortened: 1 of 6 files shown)


vladak:
Also added LDAP lookup timer.

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginClassLoader.java
```diff
@@ -56,7 +56,8 @@
             "org.opengrok.indexer.authorization.plugins.*",
             "org.opengrok.indexer.authorization.AuthorizationException",
             "org.opengrok.indexer.util.*",
-            "org.opengrok.indexer.logger.*"
+            "org.opengrok.indexer.logger.*",
+            "org.opengrok.indexer.Metrics"
```
ahornace, line 60: Could you please explain this to me? I'm probably missing something since I don't know how the plugins in OpenGrok work exactly. I assume that you would like to allow plugins to use the `Metrics` class. However, from my understanding the class is uniquely identified by the name and the classloader that loaded it. Therefore, if the `Metrics` class would be used by the plugin, then it might be a completely different class and the metrics won't be seen/published by the `web` module.

ahornace (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginClassLoader.java
ahornace: Oh, it seems that it is delegating class loading to the tomcat's class loader for some specific cases. Please ignore.

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java
```diff
@@ -55,14 +56,29 @@
...
 
-    private final Counter authStackReloadCounter = Metrics.getRegistry().counter("authorization_stack_reload");
-    private final Counter authCacheHits = Metrics.getRegistry().counter("authorization_cache_hits");
-    private final Counter authCacheMisses = Metrics.getRegistry().counter("authorization_cache_misses");
-    private final Counter authSessionsInvalidated = Metrics.getRegistry().counter("authorization_sessions_invalidated");
-
-    private final Timer authTimer = Metrics.getRegistry().timer("authorization");
-    private final Timer authPositiveTimer = Metrics.getRegistry().timer("authorization_positive");
-    private final Timer authNegativeTimer = Metrics.getRegistry().timer("authorization_negative");
+    MeterRegistry registry = Metrics.getRegistry();
```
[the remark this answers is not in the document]
vladak: fair point, fixed.

ahornace:
Nice work.

PR #3232, opened 2020-09-17
add suggester related meters

vladak:
Add basic suggester meters:
```
# HELP suggester_init_latency_seconds suggester initialization latency
# TYPE suggester_init_latency_seconds summary
suggester_init_latency_seconds_count 1.0
suggester_init_latency_seconds_sum 0.035
# HELP suggester_init_latency_seconds_max suggester initialization latency
# TYPE suggester_init_latency_seconds_max gauge
suggester_init_latency_seconds_max 0.035
# HELP suggester_rebuild_latency_seconds_max suggester rebuild latency
# TYPE suggester_rebuild_latency_seconds_max gauge
suggester_rebuild_latency_seconds_max 0.0
# HELP suggester_rebuild_latency_seconds suggester rebuild latency
# TYPE suggester_rebuild_latency_seconds summary
suggester_rebuild_latency_seconds_count 0.0
suggester_rebuild_latency_seconds_sum 0.0
# HELP suggester_rebuild_total suggester rebuild count
# TYPE suggester_rebuild_total counter
suggester_rebuild_total 0.0
```

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/Suggester.java
+++ b/suggester/src/main/java/org/opengrok/suggest/Suggester.java
@@ -100,8 +100,6 @@ public final class Suggester implements Closeable {
 
     private final CountDownLatch initDone = new CountDownLatch(1);
 
-    private final MeterRegistry registry;
-
     private final Counter suggesterRebuildCounter;
     private final Timer suggesterRebuildTimer;
     private final Timer suggesterInitTimer;
@@ -153,17 +151,16 @@ public final class Suggester implements Closeable {
         this.allowedFields = new HashSet<>(allowedFields);
         this.timeThreshold = timeThreshold;
         this.rebuildParallelismLevel = rebuildParallelismLevel;
-        this.registry = registry;
 
         suggesterRebuildCounter = Counter.builder("suggester.rebuild").
                 description("suggester rebuild count").
-                register(this.registry);
+                register(registry);
         suggesterRebuildTimer = Timer.builder("suggester.rebuild.latency").
                 description("suggester rebuild latency").
-                register(this.registry);
+                register(registry);
         suggesterInitTimer = Timer.builder("suggester.init.latency").
                 description("suggester initialization latency").
-                register(this.registry);
+                register(registry);
     }
 
     /**
--- a/suggester/src/main/java/org/opengrok/suggest/Suggester.java
+++ b/suggester/src/main/java/org/opengrok/suggest/Suggester.java
@@ -22,6 +22,9 @@
  */
 package org.opengrok.suggest;
 
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
 import org.apache.commons.lang3.time.DurationFormatUtils;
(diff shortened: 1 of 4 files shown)


suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -124,8 +134,8 @@ public Suggester(
             final boolean projectsEnabled,
             final Set<String> allowedFields,
             final int timeThreshold,
-            final int rebuildParallelismLevel
-    ) {
+            final int rebuildParallelismLevel,
+            MeterRegistry registry) {
```
ahornace, line 138: Hmm, it seems we are facing telescoping constructor anti-pattern. I'll try to think of something (builder, etc.). Feel free to ignore for now.

ahornace (changes requested):

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -97,6 +100,12 @@
 
     private final CountDownLatch initDone = new CountDownLatch(1);
 
+    private final MeterRegistry registry;
```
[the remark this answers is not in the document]
vladak: fixed

suggester/src/main/java/org/opengrok/suggest/Suggester.java
vladak: okay, ignoring :-)

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -143,6 +151,16 @@ public Suggester(
         this.allowedFields = new HashSet<>(allowedFields);
         this.timeThreshold = timeThreshold;
         this.rebuildParallelismLevel = rebuildParallelismLevel;
+
+        suggesterRebuildCounter = Counter.builder("suggester.rebuild").
+                description("suggester rebuild count").
+                register(registry);
+        suggesterRebuildTimer = Timer.builder("suggester.rebuild.latency").
```
ahornace, line 158: Now that I think about it. Does not `Timer` include also a `Counter` by default? In the code it seems they do: https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusTimer.java

Could you please check if you can plot it in Grafana?

vladak: Indeed it does. I wanted to mention it previously that I'd like to keep it separate.

```
$ curl -X GET -s http://grok-stage.cz.oracle.com/source/metrics/prometheus | grep ^suggest
suggester_init_latency_seconds_count 1.0
suggester_init_latency_seconds_sum 0.035
suggester_init_latency_seconds_max 0.0
suggester_rebuild_latency_seconds_max 0.0
suggester_rebuild_latency_seconds_count 30.0
suggester_rebuild_latency_seconds_sum 0.012
suggester_rebuild_total 30.0
```

ahornace: Okay, no problem :)

PR #3233, opened 2020-09-17
Remove redundant files from project root

ahornace:
I don't remember why `paths.tsv` was there (based on what I found it was supposed to be an example of the parameter) but now it seems that it is only passed to the `doc` directory and I miss the context why it is there.

Secondly, `licenseheader.txt` was used for NetBeans config which is no longer part of repository.

Thanks.


vladak:
The `paths.csv` was the old way how to configure path descriptions. It has been replaced by the `/system/pathdesc` API endpoint.

vladak (approved):

PR #3234, opened 2020-09-18
monitor request latency

vladak:
This change adds general request latency monitoring to the web app.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
@@ -58,7 +58,7 @@ public final class WebappListener
     private static final Logger LOGGER = LoggerFactory.getLogger(WebappListener.class);
 
     private static final String TIMER_ATTR_NAME = "request_start";
-    private static Timer requestTimer;
+    private Timer requestTimer;
 
     /**
      * {@inheritDoc}
--- a/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
@@ -23,7 +23,9 @@
  */
 package org.opengrok.web;
 
+import io.micrometer.core.instrument.Timer;
 import org.opengrok.indexer.Info;
+import org.opengrok.indexer.Metrics;
 import org.opengrok.indexer.analysis.AnalyzerGuru;
 import org.opengrok.indexer.authorization.AuthorizationFramework;
 import org.opengrok.indexer.configuration.CommandTimeoutType;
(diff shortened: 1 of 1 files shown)


opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
```diff
@@ -53,13 +57,19 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WebappListener.class);
 
+    private static final String TIMER_ATTR_NAME = "request_start";
+    private static Timer requestTimer;
```
ahornace, line 61: Why `static`?

opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
```diff
@@ -53,13 +57,19 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WebappListener.class);
 
+    private static final String TIMER_ATTR_NAME = "request_start";
```
ahornace, line 60: From the docs:
> Attribute names should follow the same conventions as package names. Names beginning with java.*, javax.*, and com.sun.*, are reserved for use by Sun Microsystems.

Just a curiosity I found :D  I don't expect you to change it. I think it's fine like this.

opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
vladak: hmm, not necessary, will remove. I don't actually know how the `listener-class` is instantiated.

opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
vladak: Sounds reasonable as it makes it easier to find where given attribute originated from. That said, this is largely not adhered to in the webapp. This is how it looks like usually in `requestDestroyed()`:
```
"request_start" -> {Instant@8437} "2020-09-18T13:53:48.017268Z"
"project_helper_groups" -> {TreeSet@8439}  size = 0
"project_helper_ungrouped_repositories" -> {TreeSet@8441}  size = 8
"project_helper_grouped_projects" -> {TreeSet@8443}  size = 0
"plugin_framework_project_cache" -> {TreeMap@8445}  size = 11
"project_helper" -> {ProjectHelper@8447} 
"project_helper_ungrouped_projects" -> {TreeSet@8449}  size = 3
"project_helper_grouped_repositories" -> {TreeSet@8451}  size = 0
"org.opengrok.indexer.web.PageConfig" -> {PageConfig@8452} 
```

vladak: changed to follow the convention.

Issue #3164, opened 2020-06-08
OpenGrokProject cookie misusing the sameSite attribute


vladak:
FF 76 claims this when loading the index page with developer tab on:

> Cookie “OpenGrokProject” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies

Not sure this is related to my deployment however this is something we might need to investigate.

tulinkry:
This is a new thing introduced lately in the browser's security and enforced by browsers at the moment. It's about limiting the cookies usage in third-party contexts.

vladak:
For the record the cookie is born in `list.jsp` like this:
```java
Cookie cookie = new Cookie(PageConfig.OPEN_GROK_PROJECT, URLEncoder.encode(cookieValue, "utf-8"));
```
and the J2EE object allows to set the `secure` attribute however not `sameSite` it seems.

vladak:
> This is a new thing introduced lately in the browser's security and enforced by browsers at the moment. It's about limiting the cookies usage in third-party contexts. (It's not connected to your setup)

okay, so something we will need to address.

tulinkry:
As this is a new thing, it may require to update the servlet-api (where the Cookie object is). More details is at https://github.com/eclipse-ee4j/servlet-api/issues/175 (wasn't looking all through that, so it may not be resolved yet).

If update is not applicable, we'd need to set the cookie directly, so instead of:
```
response.addCookie(myCookie)
```
we need:
```
response.setHeader("Set-Cookie", "key=value; HttpOnly; SameSite=<sameSiteValue>")
```

tulinkry:
As I was already tackling this in my company, I had a problem when multiple cookies were being set at a time (browser wasn't setting that properly) - instead it is possible to use one Set-Cookie directive and join the cookies with `;;` separator.

vladak:
I went through the debate on the EE4J PR and there seems to be a great deal of complexity w.r.t. getting this API in for such small change. Even if the PR was merged in, we would have to do some bunny hops to get to the new Servlet API. https://github.com/eclipse-ee4j/servlet-api/pull/271#issuecomment-578032601 suggests to configure this on the container level (Tomcat) however I'd like to have the control in the web app.

PR #3238, opened 2020-09-22
make all OpenGrok cookies use the SameSite attribute

vladak:
This change is meant to provide a fix for #3164. This works for cookies produced by the OpenGrok web application as well as for the `JSESSIONID` cookie. FF developer console no longer complains about the cookies.


opengrok-indexer/src/main/java/org/opengrok/indexer/web/PageConfig.java
```diff
@@ -861,6 +862,16 @@ private static void splitByComma(String value, List<String> result) {
...
     }
 
+    /**
+     * Add cookie header to the response along with extra attributes.
+     * @param response servlet response
+     * @param name name of the cookie
+     * @param value value of the cookie
+     */
+    public static void addCookie(HttpServletResponse response, String name, String value) {
+        response.addHeader("Set-cookie", name + "=" + value + "; SameSite=Strict; Secure");
```
vladak, line 872: Even though I did not see repeating headers in the response maybe this should just record the cookie in a map and the actual headers will be added in a filter.

vladak: found a solution to this in https://stackoverflow.com/a/51576089/11582827

vladak:
@tulinkry could you take a look ?

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.ws.rs.core.HttpHeaders;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Enumeration;
+
+/**
+ * Makes sure that all cookies originating from the web application have the Same-site attribute set.
+ */
+public class CookieFilter implements Filter {
+    FilterConfig fc;
```
ahornace, line 41:
```suggestion
    private FilterConfig fc;
```

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) {
+        this.fc = filterConfig;
+    }
+
+    @Override
+    public void destroy() {
+        this.fc = null;
```
ahornace, line 89: Is this necessary? It indicates that the whole filter will be taken out of service (I assume garbage collected) so this is not needed, right?

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
```diff
@@ -0,0 +1,91 @@
...
+            if (firstHeader) {
+                response.setHeader(HttpHeaders.SET_COOKIE, String.format("%s; %s", header, getSuffix(response)));
+                firstHeader = false;
+                continue;
+            }
+            response.addHeader(HttpHeaders.SET_COOKIE, String.format("%s; %s", header, getSuffix(response)));
+        }
+    }
+
+    private String getSuffix(HttpServletResponse response) {
```
ahornace, line 64:
```suggestion
    private String getSuffix() {
```

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
```diff
@@ -0,0 +1,91 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        for (Enumeration<String> e = fc.getInitParameterNames(); e.hasMoreElements();) {
```
ahornace, line 67: Would this work?

```java
StreamSupport.stream(Spliterators.spliteratorUnknownSize(fc.getInitParameterNames(), Spliterator.NONNULL), false)
        .map(attributeName -> {
            String attributeValue = fc.getInitParameter(attributeName);
            if (!attributeValue.isEmpty()) {
                return attributeName + "=" + attributeValue;
            } else {
                return attributeValue;
            }
        }).collect(Collectors.joining("; "));
```

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
vladak: most probably not, will remove.

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
vladak: The first argument of `spliteratorUnknownSize` has to be either primitive number type or `Iterator` which is not available in JDK 8. Also, this sort of conversion to stream seems, uh, unnatural.

ahornace: Yes, however, there are helper methods for this in Guava and maybe apache-commons -> if we have on of them on the classpath then we could leverage that. If not, then creating a simple `toStream` util method would make this nicer.

But you are right, https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Enumeration.html#asIterator() is available since JDK9

opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java
ahornace: It probably needs to be overridden but empty:
```
[ERROR] /home/travis/build/oracle/opengrok/opengrok-web/src/main/java/org/opengrok/web/CookieFilter.java:[40,8] org.opengrok.web.CookieFilter is not abstract and does not override abstract method destroy() in javax.servlet.Filter
```

vladak: fixed

ahornace (approved):
Looks good to me.

vladak:
Thanks for the review !

PR #3242, opened 2020-09-24
Fix README copyright

ahornace:
Fixes a year in copyright as it looks that the project is up-to-date.


vladak:
Crazy year. I still hope this is not the "new normal".

Issue #3243, opened 2020-09-24
FileHistoryCache#store is one big memory hog


vladak:
Observing the RSS/CPU of indexer process that does indexing from scratch of multiple repositories with heavy history (Linux kernel, FreeBSD, ...) running with 16 threads and 48 GiB heap (the machine has 32 CPUs and 256 GB RAM), there is clearly something bad going on. The indexing is in the phase of generating history caches for all projects. The indexer process has ~50 GB RSS, is busy on the CPU (say 60%), the usage grows, stays a bit at the maximum (70%) and then quickly falls down (to low 60's %). This cycle repeats every couple of seconds (assuming the GC is busy collecting and then is either done or gets stopped because it spent too much time on the CPU) while there are Mercurial/Git `log` processes running, getting the history of the whole repository. This happens with 1.4.15.

I have not done any heap analysis yet, however by looking at `FileHistoryCache#store`, this is just asking for trouble. First, the whole repository history is stored in memory (in the form of `History`/`HistoryEntry` objects) which could be quite sizeable of its own (sample Linux kernel repo has 500k+ changsets and 50k+ files on disk) and is then converted to the inverted map:
```java
HashMap<String, List<HistoryEntry>> map = new HashMap<>();
```
When tagged history is enabled (as is the case for the indexer run I am observing), it gets even worse:
```java
466                  /*
467                   * We need to do deep copy in order to have different tags
468                   * per each commit.
469                   */
470                  if (env.isTagsEnabled() && repository.hasFileBasedTags()) {
471                      list.add(new HistoryEntry(e));
472                  } else {
473                      list.add(e);
474                  }
```
In such case there will be distinct `HistoryEntry` object for each changeset that touched given file. In overall, this will lead to explosive growth of `HistoryEntry` objects.

vladak:
Low hanging fruit: one way to avoid the explosion is to move the deep copy of the history entries right before the tags are assigned in `doFileHistory()` and then maybe set them to `null` once the `storeFile()` is done.

vladak:
Actually, the deep copy looks unnecessary since the tags are actually reset for all the cases in `doFileHistory()` before they are assigned to the history entries:
https://github.com/oracle/opengrok/blob/115e69dcd1815be83afdeb5751c860f44e486387/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java#L121-L140

vladak:
[...]

PR #3248, opened 2020-09-25
do not deep copy history entries

vladak:
This was meant to help a bit #3243 however it does not do much for my particular use case as described in the issue. Still, should save some heap memory in case of running the indexer with -G.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
@@ -102,38 +102,32 @@ class FileHistoryCache implements HistoryCache {
             Repository repository, File srcFile, File root, boolean renamed)
             throws HistoryException {
 
-        History hist = null;
+        File file = new File(root, filename);
+        // Only store directory history for the top-level directory.
+        if (file.isDirectory() && !filename.equals(repository.getDirectoryName())) {
+            LOGGER.log(Level.FINE, "Not storing history cache for {0}: not top level directory", file);
+            return;
+        }
 
         /*
          * If the file was renamed (in the changesets that are being indexed),
          * its history is not stored in the historyEntries so it needs to be acquired
          * directly from the repository.
-         * This ensures that complete history of the file (across renames)
-         * will be saved.
+         * This ensures that complete history of the file (across renames) will be saved.
          */
+        History hist;
         if (renamed) {
             hist = repository.getHistory(srcFile);
+        } else {
+            hist = new History(historyEntries);
         }
 
-        File file = new File(root, filename);
-
-        if (hist == null) {
-            hist = new History();
-
-            // File based history cache does not store files for individual
-            // changesets so strip them unless it is history for the repository.
-            for (HistoryEntry ent : historyEntries) {
-                if (file.isDirectory() && filename.equals(repository.getDirectoryName())) {
-                    ent.stripTags();
-                } else {
-                    ent.strip();
-                }
-            }
-
-            // add all history entries
-            hist.setHistoryEntries(historyEntries);
-        } else {
-            for (HistoryEntry ent : hist.getHistoryEntries()) {
+        // File based history cache does not store files for individual
+        // changesets so strip them unless it is history for the repository.
+        for (HistoryEntry ent : hist.getHistoryEntries()) {
+            if (file.isDirectory()) {
+                ent.stripTags();
+            } else {
                 ent.strip();
             }
         }
@@ -143,10 +137,7 @@ class FileHistoryCache implements HistoryCache {
             repository.assignTagsInHistory(hist);
         }
 
-        // Only store directory history for the top-level.
-        if (!file.isDirectory() || filename.equals(repository.getDirectoryName())) {
-            storeFile(hist, file, repository, !renamed);
-        }
+        storeFile(hist, file, repository, !renamed);
     }
 
     private boolean isRenamedFile(String filename, Repository repository, History history)
@@ -462,15 +453,8 @@ class FileHistoryCache implements HistoryCache {
                     list = new ArrayList<>();
                     map.put(s, list);
                 }
-                /*
-                 * We need to do deep copy in order to have different tags
-                 * per each commit.
-                 */
-                if (env.isTagsEnabled() && repository.hasFileBasedTags()) {
-                    list.add(new HistoryEntry(e));
-                } else {
-                    list.add(e);
-                }
+
+                list.add(e);
             }
         }
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
@@ -51,6 +51,8 @@ import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.BufferSink;
 import org.opengrok.indexer.util.Executor;
 
+import org.jetbrains.annotations.NotNull;
+
 /**
  * An interface for an external repository.
  *
@@ -95,6 +97,7 @@ public abstract class Repository extends RepositoryInfo {
      *
      * @return {@code true} if the repository can get history for directories
      */
+    @NotNull
     abstract boolean hasHistoryForDirectories();
 
     /**
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
@@ -276,24 +276,25 @@ public abstract class Repository extends RepositoryInfo {
      * tags to changesets which actually exist in the history of given file.
      * Must be implemented repository-specific.
      *
-     * @see getTagList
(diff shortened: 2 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -103,38 +103,34 @@ private void doFileHistory(String filename, List<HistoryEntry> historyEntries,
...
         /*
          * If the file was renamed (in the changesets that are being indexed),
          * its history is not stored in the historyEntries so it needs to be acquired
          * directly from the repository.
-         * This ensures that complete history of the file (across renames)
-         * will be saved.
+         * This ensures that complete history of the file (across renames) will be saved.
          */
         if (renamed) {
             hist = repository.getHistory(srcFile);
```
ahornace, line 121: What is the contract here? Can `repository.getHistory(srcFile)` return `null`? If not, then maybe we can rewrite it to:
History hist;
```java
if (renamed) {
    hist = repository.getHistory(srcFile);
} else {
    hist = new History(historyEntries);
}
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
```diff
@@ -279,24 +279,25 @@ boolean hasFileBasedTags() {
...
+
         Iterator<TagEntry> it = this.getTagList().descendingIterator();
         TagEntry lastTagEntry = null;
-        // Go through all commits of given file
         for (HistoryEntry ent : hist.getHistoryEntries()) {
             // Assign all tags created since the last revision
-            // Revision in this HistoryEntry must be already specified!
-            // TODO is there better way to do this? We need to "repeat"
-            // last element returned by call to next()
+            // Revision in this HistoryEntry must be already specified !
```
ahornace, line 298: Why the space before `!`? Just curious :D

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -463,15 +456,8 @@ public void store(History history, Repository repository)
                     list = new ArrayList<>();
                     map.put(s, list);
                 }
-                /*
-                 * We need to do deep copy in order to have different tags
```
ahornace, line 467: Could you please explain what was the use case?
```
We need to do deep copy in order to have different tags per each commit.
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
vladak: I thought about the `null` return value too. It seems rather than returning null History the underlying methods will throw `HistoryException`.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
vladak: with the space the whole TODO comment gets highlighted in IDEA.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
vladak: I think the author of the original code forgot that the tags were reset in `doFileHistory()`. Assigning the tags to history entries used shared across files would lead to wrong results.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -103,38 +103,32 @@ private void doFileHistory(String filename, List<HistoryEntry> historyEntries,
             Repository repository, File srcFile, File root, boolean renamed)
             throws HistoryException {
 
-        History hist = null;
+        File file = new File(root, filename);
+        // Only store directory history for the top-level directory.
+        if (file.isDirectory() && !filename.equals(repository.getDirectoryName())) {
+            LOGGER.log(Level.FINE, "Not storing history cache for {0}", file);
```
tulinkry, line 109: a reason in the message could be helpful for debugging

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -103,38 +103,32 @@ private void doFileHistory(String filename, List<HistoryEntry> historyEntries,
...
-                }
-            }
-
-            // add all history entries
-            hist.setHistoryEntries(historyEntries);
-        } else {
-            for (HistoryEntry ent : hist.getHistoryEntries()) {
+        // File based history cache does not store files for individual
+        // changesets so strip them unless it is history for the repository.
+        for (HistoryEntry ent : hist.getHistoryEntries()) {
```
tulinkry, line 128: hist could be null when `hist = repository.getHistory(srcFile)` returns null which was handled in the previous code

ahornace: If it's part of the contract that it can return `null`, then it would be better to rewrite it to return `Optional`.

vladak: Sounds to me like a complication. If the history cannot be retrieve, the `HistoryException` should be thrown.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
vladak: sure, fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
vladak: There's no word about `null` on https://github.com/oracle/opengrok/blob/ecb1ce82e129709e5bdf11c88db805a26b2bd7ef/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java#L92-L98
and looking at a few repository implementations they do throw the `HistoryException`, e.g. https://github.com/oracle/opengrok/blob/ecb1ce82e129709e5bdf11c88db805a26b2bd7ef/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialHistoryParser.java#L76-L101
I wonder if it is possible to enforce the class methods not to return null. The `@NotNull` annotation seems to be meant for something else ? Otherwise, provide a wrapper to check for `null` and throw the exception.

vladak: https://github.com/JetBrains/java-annotations has `@NotNull` that might work.

tulinkry: It's ok, I didn't see this conversation https://github.com/oracle/opengrok/pull/3248#discussion_r494812868

tulinkry (approved):

idodeclare:
@vladak what repo might have required the deep copy? It was quite explicit, so wondering if you’ve accounted for all use-cases.

vladak:
> @vladak what repo might have required the deep copy? It was quite explicit, so wondering if you’ve accounted for all use-cases.

I tried to come up with a reason for this code and could not find any, at least w.r.t. tags.

vladak:
For the record the comment about deep copy comes from 3ba66fbb by @skozina that added tagging capability to history view.

vladak:
For this one I think the risk is worth the benefit.

idodeclare:
> For this one I think the risk is worth the benefit.

I'm skeptical firstly because you're removing code conditional on `repository.hasFileBasedTags()` where only a tiny number of `Repository` subclasses are overriding `hasFileBasedTags()` and secondly because you write that "it does not do much for my particular use case." What benefit is there evidence for?

vladak:
Right, I should have measured this first. Will try once indexer can gather JVM statistics.

vladak:
I think it did not do much for my case because the overall memory overhead just trumped the benefit of this change by large margin.

vladak:
Turned out my hunch was correct. I chose the same working set as in #3243 and same indexer options (obviously including -G) and same source code basis to avoid running `git log` with `-m`.

Here are JVM used memory graphs without and with the fix:
without -m and without fix
without -m and with fix

Both graphs show complete indexing process. This is first such time we have introspection into what is happening there. The first steep fall in the graphs correlate with the finish of the `git log` processes. The second graph spans sightly wider because there was another indexer running on the machine at that time. The history cache generation finished at 17:00 and 19:15. The first indexer finished around 17:40. The second peak of old generation area in the second graph is interesting, I don't have good explanation for it.

In overall this saved some 6 GB of heap on average during the history cache creation phase.

It might impact only a small subset of repository types however Git/Mercurial are widely used so the impact is high.

Side node: beware, exporting indexer metrics over statsd can quickly consume your mobile data plan limits as it can easily generate GBs of traffic.

vladak:
The reason why this did not help with the case described in #3243 is that the effect of this change is visible only after the acquire of history entries has finished and the issue prevents that from completing.

PR #3251, opened 2020-09-26
Fix some warnings

ahornace:
When reviewing https://github.com/oracle/opengrok/pull/3249 I did not like the `Do` method name (capital letter at the beginning due to `do` being a keyword). So I've changed it to `execute` and I've noticed that there were some `@SupressWarnings` annotations that could be easily fixed so I've tried to fix those plus I did some simple refactoring suggested by IDEA.


vladak:
Thanks !

PR #3253, opened 2020-09-30
Add maven dependency check plugin

ahornace:
When doing security cup at work I've noticed this plugin and I found it really helpful. It will list all the dependencies with known CVEs and create a report.

I debated whether to add it among `reporting` plugins but I found the usage easier like this since I could call it directly from IDEA.

Output from console:
```
One or more dependencies were identified with known vulnerabilities in Distribution:

ant-1.10.1.jar (pkg:maven/org.apache.ant/ant@1.10.1, cpe:2.3:a:apache:ant:1.10.1:*:*:*:*:*:*:*) : CVE-2020-1945
source.war: chronicle-wire-2.20.4.jar (pkg:maven/net.openhft/chronicle-wire@2.20.4, cpe:2.3:a:wire:wire:2.20.4:*:*:*:*:*:*:*) : CVE-2018-8909
source.war: jquery-3.4.1.min.js (pkg:javascript/jquery@3.4.1.min) : CVE-2020-11022, CVE-2020-11023
source.war: jstl-1.2.jar (cpe:2.3:a:apache:standard_taglibs:1.2:*:*:*:*:*:*:*) : CVE-2015-0254
```

It will also generate a report in html which looks like this:
Screenshot 2020-09-30 at 10 49 41

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -317,6 +317,7 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
                 <version>6.0.2</version>
                 <configuration>
                     <skipSystemScope>true</skipSystemScope>
+                    <assemblyAnalyzerEnabled>false</assemblyAnalyzerEnabled>
                 </configuration>
                 <executions>
                     <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 
 CDDL HEADER END
 
-Copyright (c) 2010, 2018, Oracle and/or its affiliates. All rights reserved.
+Copyright (c) 2010, 2020, Oracle and/or its affiliates. All rights reserved.
 Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
 
 -->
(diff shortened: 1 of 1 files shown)


tarzanek:
lgtm, but I cannot merge anymore <sigh>
so either your or @vladak will have to merge

vladak:
I played with the plugin before. I wonder if there is a way how to make the report more visible / produce some sort of alert.

ahornace:
> I played with the plugin before. I wonder if there is a way how to make the report more visible / produce some sort of alert.
You can fail the build on CVEs greater than some number but I don't find it reasonable since you can break old builds that were fine otherwise.

I don't know about other option :/

vladak:
Travis Windows build has failed due to the dependency-check analysis.

vladak:
> > I played with the plugin before. I wonder if there is a way how to make the report more visible / produce some sort of alert.
> 
> You can fail the build on CVEs greater than some number but I don't find it reasonable since you can break old builds that were fine otherwise.

It would be nice to be able to fail the (release) build if there are any CVEs with CVSS score higher than given threshold.

vladak:
Thanks !

PR #3254, opened 2020-10-02
parallelize repository scan

vladak:
For the past week or so I have been trying to determine the cause of slowly running reindex from scratch which made me to run it over and over with different settings/versions. Watching the behavior of the initial phase of such indexing made me to address the speed of repository scanning. This change adds parallelism on the level of source root sub-directories. The original behavior should be preserved.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017-2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.history;
@@ -41,6 +41,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -88,15 +89,12 @@ public final class HistoryGuru {
      */
     private final Map<String, String> repositoryRoots = new ConcurrentHashMap<>();
 
-    private final int scanningDepth;
-
     /**
      * Creates a new instance of HistoryGuru, and try to set the default source
      * control system.
      */
     private HistoryGuru() {
         env = RuntimeEnvironment.getInstance();
-        scanningDepth = env.getScanningDepth();
 
         HistoryCache cache = null;
         if (env.useHistoryCache()) {
@@ -435,7 +433,7 @@ public final class HistoryGuru {
                                     "Failed to get sub directories for ''{0}'', " +
                                     "check access permissions.",
                                     file.getAbsolutePath());
-                        } else if (depth <= scanningDepth) {
+                        } else if (depth <= env.getScanningDepth()) {
                             repoList.addAll(addRepositories(subFiles,
                                     allowedNesting, depth + 1, isNested));
                         }
@@ -453,7 +451,7 @@ public final class HistoryGuru {
                             LOGGER.log(Level.WARNING,
                                     "Failed to get sub directories for ''{0}'', check access permissions.",
                                     file.getAbsolutePath());
-                        } else if (depth <= scanningDepth) {
+                        } else if (depth <= env.getScanningDepth()) {
                             // Search down to a limit -- if not: too much
                             // stat'ing for huge Mercurial repositories
                             repoList.addAll(addRepositories(subFiles,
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -480,7 +478,25 @@ public boolean hasAnnotation(File file) {
...
+        List<RepositoryInfo> repoList = new ArrayList<>();
+        futures.forEach(future -> {
+            try {
+                repoList.addAll(future.get());
+            } catch (Exception e) {
+                LOGGER.log(Level.WARNING, "failed to get future");
+            }
+        });
+
+        env.getIndexerParallelizer().bounceRepositorySearchExecutor();
```
idodeclare, line 497: I wouldn't think a bounce would be required if futures are being awaited. Why is it necessary?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -679,6 +683,11 @@ public static void main(String[] argv) {
                     "Run as quietly as possible. Sets logging level to WARNING.").execute(v ->
                     LoggerUtil.setBaseConsoleLogLevel(Level.WARNING));
 
+            parser.on("--repositorySearchThreads", "=number", Integer.class,
+                    "The number of threads to use for scanning for repositories.",
+                    "By default the number of threads will be set to the number of available CPUs.").
```
idodeclare, line 688: See line 435:

```
Limit usage lines to 72 characters for concise formatting.
```

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -679,6 +683,11 @@ public static void main(String[] argv) {
                     "Run as quietly as possible. Sets logging level to WARNING.").execute(v ->
                     LoggerUtil.setBaseConsoleLogLevel(Level.WARNING));
 
+            parser.on("--repositorySearchThreads", "=number", Integer.class,
```
idodeclare, line 686: This is mis-ordered and should appear after `--repository`

idodeclare (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: just didn't want to keep the threads around.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -480,7 +478,25 @@ public boolean hasAnnotation(File file) {
...
+            futures.add(executor.submit(() -> addRepositories(new File[]{file},
+                    env.getNestingMaximum(), 0, false)));
+        }
+
+        List<RepositoryInfo> repoList = new ArrayList<>();
+        futures.forEach(future -> {
+            try {
+                repoList.addAll(future.get());
+            } catch (Exception e) {
+                LOGGER.log(Level.WARNING, "failed to get future");
```
idodeclare, line 493: Maybe a more activity-specific message?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -480,7 +478,25 @@ public boolean hasAnnotation(File file) {
      * @return collection of added repositories
      */
     public Collection<RepositoryInfo> addRepositories(File[] files) {
-        return addRepositories(files, env.getNestingMaximum(), 0, false);
+        ExecutorService executor = env.getIndexerParallelizer().getRepositorySearchExecutor();
```
idodeclare, line 481: May I suggest just using the `-T,--threads` executor? It's not otherwise being used at this stage of a run

idodeclare (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexerParallelizer.java
```diff
@@ -147,6 +151,15 @@ public ExecutorService getCtagsWatcherExecutor() {
         return result;
     }
 
+    /**
+     * @return the ExecutorService used for repository scan
+     */
+    public ExecutorService getRepositorySearchExecutor() {
+        ExecutorService result = lzRepositorySearchExecutor.get();
+        repositorySearchExecutor = result;
```
tulinkry, line 159: what is the point of caching what you can get from the lazy supplier?

Is it that you need to keep track if it has ever been initialized?

idodeclare: Yes, because you want to avoid initializing something expensive if it hasn't been used elsewhere. I have a branch where I've added an `isActive` property to `LazilyInstantiate` to allow simplifying `IndexerParallelizer`. I can extract that part from the branch and raise a narrow PR.

tulinkry: I think it looks cleaner with the PR you just raised. Good job. (#3256)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: yes, that's reasonable.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexerParallelizer.java
vladak: Indeed, rebased.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
idodeclare, line 481: I was suggesting just to use `getFixedExecutor()` since it's fully available at this point during a run. It would be fine to leave it to be used later during indexing. `IndexerParallelizer` could be kept simple, and no additional configuration would be needed.

vladak: I see, makes sense.

idodeclare (approved):
LGTM

vladak:
Thanks guys !

PR #3258, opened 2020-10-04
R language

idodeclare:
Hello,

Please consider for integration this patch to support the R language. Also included is a fix for mistakenly ignoring the keyword table for HCL/Terraform.

Thank you.


vladak:
Thanks ! I like how easy it is to add new analyzer.

Issue #747, opened 2014-02-12
after history for directory fails, individual files are history indexed


vladak:
A colleague of mine has a big Mercurial repo (12 GB) where `hg log` on the top-level directory takes many hours to complete. When such repo is indexed for the first time, the `hg log` command times out and the indexer proceeds to adding every single file in the repository and generates xrefs for them. Unfortunately, as a side effect it also tries to generate history for each file (as there is none in the history cache) sequentially.

In the log it looks like this:

```
2014-01-27 09:19:08.175-0800 WARNING t31 Executor.exec: Non-zero exit status -1 from command [/usr/bin/hg, log, --template, changeset: {rev}:{node|short}\n{branches}{tags}{parents}\nuser: {author}\ndate: {date|isodate}\nfiles: {files}\ndescription: {desc|strip|obfuscate}\nfile_copies: {file_copies}\n] in directory /builds1/psumbera2
2014-01-27 09:19:08.175-0800 WARNING t31 HistoryGuru.createCache: An error occured while creating cache for /builds1/psumbera2/userland-src (MercurialRepository)
org.opensolaris.opengrok.history.HistoryException: Failed to get history for: "/builds1/psumbera2/userland-src" Exit code: -1
        at org.opensolaris.opengrok.history.MercurialHistoryParser.parse(MercurialHistoryParser.java:79)
        at org.opensolaris.opengrok.history.MercurialRepository.getHistory(MercurialRepository.java:509)
        at org.opensolaris.opengrok.history.Repository.createCache(Repository.java:296)
        at org.opensolaris.opengrok.history.HistoryGuru.createCache(HistoryGuru.java:475)
        at org.opensolaris.opengrok.history.HistoryGuru.access$000(HistoryGuru.java:51)
        at org.opensolaris.opengrok.history.HistoryGuru$1.run(HistoryGuru.java:519)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
        at java.util.concurrent.FutureTask.run(FutureTask.java:262)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at java.lang.Thread.run(Thread.java:744)
2014-01-27 09:19:08.176-0800 INFO t1 Indexer.prepareIndexer: Done...
[...]

PR #3261, opened 2020-10-05
do not call createCache() unnecessarily

vladak:
Looking at `ensureHistoryCacheExists()` to fix the double emitted log message during reindex, it seems that there is some unnecessary logic there. If the history cache was created for the repository in the 1st step of indexing (the existence of the history cache directory should be enough to tell), it should not be necessary to call `createCache()` again - only for complete failure to create the history cache.

`ensureHistoryCacheExists()` will probably disappear anyway with the fix for #747.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -747,7 +747,8 @@ public final class HistoryGuru {
             return;
         }
 
-        if (historyCache.hasCacheForDirectory(file, repository)) {
+        if (!repository.hasHistoryForDirectories() ||
+                historyCache.hasCacheForDirectory(file, repository)) {
             return;
         }
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -730,9 +730,7 @@ public final class HistoryGuru {
 
     /**
      * Ensure that we have a directory in the cache. If it's not there, fetch
-     * its history and populate the cache. If it's already there, and the cache
-     * is able to tell how recent it is, attempt to update it to the most recent
-     * revision.
+     * its history and populate the cache.
      *
      * @param file the root path to test
      * @throws HistoryException if an error occurs while accessing the history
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -731,24 +729,16 @@ public void ensureHistoryCacheExists(File file) throws HistoryException {
 
         Repository repository = getRepository(file);
         if (repository == null) {
-            // no repository -> no history :(
```
tulinkry, line 734: :(

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -731,24 +729,16 @@ public void ensureHistoryCacheExists(File file) throws HistoryException {
...
+            // no repository -> no history
             return;
         }
 
-        String sinceRevision = null;
-
-        if (historyCache.hasCacheForDirectory(file, repository)) {
-            sinceRevision = historyCache.getLatestCachedRevision(repository);
```
tulinkry, line 741: but what if it has history but for older revision?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: depends on point of view I guess. Also, emojis are now the cool thing to have in (UTF-8 encoded) source code.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: That would fall through. However I think this is not the place to catch this.

idodeclare:
With the previous logic, the second call to `createCache()` in `ensureHistoryCacheExists()` effectively meant that full or incremental retrieval of history could fail "in the 1st step of indexing" but might be given a chance to succeed in the second call.

With this patch, the only thing `ensureHistoryCacheExists()` might try to do is a full retrieval of history after a first full retrieval fails. A failed incremental retrieval will no longer be retried.

I wonder whether the likelihood of success of either of those second tries is actually so low that expunging `ensureHistoryCacheExists()` now (before #747) would bother no one.

vladak:
> I wonder whether the likelihood of success of either of those second tries is actually so low that expunging `ensureHistoryCacheExists()` now (before #747) would bother no one.

Had the same thought. I think it can go.

idodeclare (approved):
LGTM

vladak:
Thanks !

PR #3262, opened 2020-10-06
Fix to count all EOLs. Simplify to avoid yyline

idodeclare:
Please consider for integration this fix for files that use Mac Classic line breaks (`\r`), where the analyzer parsing was detecting the breaks, but the xref counting was not incrementing without the presence of a `\n`. Also the xref was on a single, very long line.

This also removes almost all use of the JFlex `yyline` field, since OpenGrok almost entirely does not rely on `%line` counting, and using `yyline` cluttered the \*.lex\* files.

Thank you.


vladak:
Looks good, thanks !

PR #3266, opened 2020-10-06
remove 'Completed in <duration>' message from the UI

vladak:
This change removes the 'Completed in <duration>' message from the search results page. I think it is not necessary and clutters the UI. If the user is interested in the latency for debugging purposes, he can always use e.g. Web console in Firefox or such. For admins this value is available via the monitoring endpoint.

### DIFF
--- a/opengrok-web/src/main/webapp/search.jsp
+++ b/opengrok-web/src/main/webapp/search.jsp
@@ -71,8 +71,6 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
 
-    long starttime = System.currentTimeMillis();
-
     SearchHelper searchHelper = cfg.prepareSearch();
     // N.b. searchHelper.destroy() is called via
     // WebappListener.requestDestroyed() on presence of the following
@@ -207,8 +205,6 @@ include file="menu.jspf"
             <li>Try more general keywords.</li>
             <li>Use 'wil*' cards if you are looking for partial match.</li>
         </ul>
-        <p><b>Completed in <%= System.currentTimeMillis() - starttime
-            %> milliseconds</b></p>
 	<%
     } else {
         int start = searchHelper.start;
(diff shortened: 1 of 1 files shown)


ahornace (approved):
Even Google shows the time on the search -> I think it might be sometimes insightful. However, in OpenGrok this line was too big and I completely agree that it's nicer without it.

opengrok-web/src/main/webapp/search.jsp
```diff
@@ -71,8 +71,6 @@ include file="projects.jspf"
 {
     PageConfig cfg = PageConfig.get(request);
 
-    long starttime = System.currentTimeMillis();
```
ahornace, line 74: The build failed -> you probably will need to add this line back :)

vladak: forgot to remove all references to `starttime`, should be fixed in next build.

vladak:
> Even Google shows the time on the search -> I think it might be sometimes insightful. However, in OpenGrok this line was too big and I completely agree that it's nicer without it.

I've been trying to make the UI minimal over the course of years and this makes sense to me.

tulinkry:
I think this change is ok, the ui user is most of the time not really interested how much time a query took at the server. The round trip is the important number and that is in the browser console.

vladak:
> I think this change is ok, the ui user is most of the time not really interested how much time a query took at the server. The round trip is the important number and that is in the browser console.

Yeah, I think the number can serve as a sort of advertisement ("look how fast this service is", except when it is not) however that is not our goal I think. The latency should be a concern to whoever is running the service and there is monitoring for that. If anyone is really interested in making sure the latency is within a limit, various monitoring tools on various levels offer this sort of capability (Micrometer can configure Meters with SLA, Prometheus has alerting capability as well as Grafana etc.).

PR #3268, opened 2020-10-06
Chocolatey switched to Python 3.9

vladak:
This change is needed to fix Windows builds.

### DIFF
--- a/dev/main
+++ b/dev/main
@@ -31,7 +31,7 @@ if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
 	export PATH=${PATH}:${JAVA_HOME}
 
 	# needs to be before the default Python 2
-	export PATH=/c/Python38:${PATH}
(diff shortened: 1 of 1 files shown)


dev/main
```diff
@@ -30,8 +30,13 @@ if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
 	export PATH=${PATH}:/c/Program\ Files\ \(x86\)/Subversion/bin
 	export PATH=${PATH}:${JAVA_HOME}
 
+	PYTHON_PATH="/c/Python39"
+	if [[ ! -d $PYTHON_PATH ]]; then
+		echo "Not a directory: $PYTHON_PATH"
```
idodeclare, line 35: stderr maybe?
```suggestion
		>&2 echo "Not a directory: $PYTHON_PATH"
```

idodeclare (approved):
Nit about stderr but also OK as is

PR #3269, opened 2020-10-07
Update jQuery to 3.5.1

ahornace:
This change updates the jQuery by leveraging https://www.webjars.org. This is a recommended way how to include js/css in Java projects. We can try to include other dependencies in this way -> I just wanted to try it out. I've done some simple checks and the UI seemed to work properly but nothing extensive. @vladak could you please check?

Thanks.


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
```diff
@@ -86,27 +89,27 @@ public FileScript(String script, int priority) {
...
+
     /**
      * Aliases for the page scripts. The path in the FileScript is relatively to
      * the request's context path.
      *
      * @see HttpServletRequest#getContextPath()
      */
     static {
-        putjs("jquery", "js/jquery-3.4.1", 10);
+        putFromWebJar("jquery", "jquery.min.js", 10);
```
vladak, line 112: I wonder if this is going to work well w.r.t. caching. So far we relied on the version number to change I think.

ahornace: The final code looks like this so version is still there :)
```html
<script type="text/javascript" src="/opengrok_web_war/webjars/jquery/3.5.1/jquery.min.js" data-priority="10"</script>
```

tulinkry: Yes, caching.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
```diff
@@ -86,27 +89,27 @@ public FileScript(String script, int priority) {
...
-            builder.append(this.getScriptData());
-            builder.append("\" data-priority=\"");
-            builder.append(this.getPriority());
-            builder.append("\"></script>\n");
-            return builder.toString();
+            return "\t<script type=\"text/javascript\" src=\"" +
+                    this.getScriptData() +
+                    "\" data-priority=\"" +
+                    this.getPriority() +
+                    "\"></script>\n";
```
tulinkry, line 96: I didn't know + was better than a builder

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
ahornace: I'll probably need to modify `ExpiresOneDayFilter` to include `/webjars`.

tulinkry: Then it would be ok I guess.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Scripts.java
vladak: better = faster ?

ahornace: String concatenation uses StringBuilder in the backend so there is no performance gain but it's shorter this way.

vladak: In this case the builder approach was slightly more readable for me however that's probably just a matter of taste.

ahornace: This was suggested by IDEA and I thought removing the clutter is better but I can revert it back :)

tulinkry: Ok

tulinkry:
I don't know how it works exactly but can you configure it so it matches the path for the other scripts? I mean like so noone would recognized that it's loaded by different mechanism.

so instead of `webjars` it would be `js`

ahornace:
> so instead of webjars it would be js

What would be the main benefit? I would probably have to set up resource mapping (not sure how to do in EE) and I'm not sure if it would not interfere with the already present `/js/` directory.

tulinkry:
Consisntency

vladak:
If this can be done with some reasonable effort then go for it. Otherwise not such a big deal I think.

vladak:
The builds are currently failing with `java.lang.NoClassDefFoundError: Could not initialize class org.opengrok.indexer.web.Scripts`

ahornace:
> The builds are currently failing with java.lang.NoClassDefFoundError: Could not initialize class org.opengrok.indexer.web.Scripts

This is due to inherent problem that classes which should be part of the `web` module only are part of `indexer` module. Therefore, when running `ScriptsTest` there is no `jquery` on the classpath since it was added to the `web` module dependencies. One easy fix is to add it to indexer's `pom.xml` but this only promotes the problem. What do you think would be the best approach?

vladak:
Do they have to be really part of the index module ? If not how much work would it be to move them over to the web module ?

ahornace:
> Do they have to be really part of the index module ? If not how much work would it be to move them over to the web module ?

They don't. I don't think that indexer should care about HTTP requests etc. at all. I'll look into it.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/messages/Message.java
```diff
@@ -80,6 +80,7 @@ public String toString() {
             return messageLevelString;
         }
 
+        @SuppressWarnings("rawtypes")
```
vladak, line 83: there you go :-)

tulinkry:
> If this can be done with some reasonable effort then go for it. Otherwise not such a big deal I think.

As far as I understood, it wasn't possible.

ahornace:
> As far as I understood, it wasn't possible.

No, I did not have time to look into it properly and fixing the tests was easier. I will try over the weekend.

PR #3270, opened 2020-10-09
Move Message and related classes to web module

ahornace:
I've noticed the `TODO` messages in indexer's `pom.xml` and since I was already moving some classes in #3269 I decided to give messages a go as well.

I made `MessagesContainer` a singleton (we basically had only one instance in `RuntimeEnvironment`) so I had to change the tests around it a bit.

`MessagesContainer` was depending on configuration change so I introduced `OnConfigurationChangedListener` as a callback for any configuration dependent classes to use.

Thanks.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
@@ -1,0 +1,29 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.indexer.configuration;
+
+public interface OnConfigurationChangedListener {
+
+    void onConfigurationChanged();
+
+}
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
@@ -40,6 +40,7 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
@@ -73,9 +74,6 @@ import org.opengrok.indexer.util.PathUtils;
 import org.opengrok.indexer.util.ResourceLock;
 import org.opengrok.indexer.web.Prefix;
 import org.opengrok.indexer.web.Util;
-import org.opengrok.indexer.web.messages.Message;
-import org.opengrok.indexer.web.messages.MessagesContainer;
-import org.opengrok.indexer.web.messages.MessagesContainer.AcceptedMessage;
 
 /**
  * The RuntimeEnvironment class is used as a placeholder for the current
@@ -99,7 +97,6 @@ public final class RuntimeEnvironment {
 
     private String configURI;
     IncludeFiles includeFiles = new IncludeFiles();
-    private final MessagesContainer messagesContainer = new MessagesContainer();
 
     private static final IndexTimestamp indexTime = new IndexTimestamp();
 
@@ -122,6 +119,8 @@ public final class RuntimeEnvironment {
     private final transient Set<String> ctagsLanguages = new HashSet<>();
 
     public WatchDogService watchDog;
+
+    private final Set<OnConfigurationChangedListener> listeners = new CopyOnWriteArraySet<>();
 
     public List<String> getSubFiles() {
         return subFiles;
@@ -1677,7 +1676,9 @@ public final class RuntimeEnvironment {
         getAuthorizationFramework().setStack(getPluginStack());
         getAuthorizationFramework().reload();
 
-        messagesContainer.setMessageLimit(getMessageLimit());
+        for (OnConfigurationChangedListener l : listeners) {
+            l.onConfigurationChanged();
+        }
     }
 
     public void setIndexTimestamp() throws IOException {
@@ -1811,60 +1812,6 @@ public final class RuntimeEnvironment {
         return multiReader;
     }
 
-    public void startExpirationTimer() {
-        messagesContainer.setMessageLimit(getMessageLimit());
-        messagesContainer.startExpirationTimer();
-    }
-
-    public void stopExpirationTimer() {
-        messagesContainer.stopExpirationTimer();
-    }
-
-    /**
-     * Get the default set of messages for the main tag.
-     *
-     * @return set of messages
-     */
-    public SortedSet<AcceptedMessage> getMessages() {
-        return messagesContainer.getMessages();
-    }
-
-    /**
-     * Get the set of messages for the arbitrary tag.
-     *
-     * @param tag the message tag
-     * @return set of messages
-     */
-    public SortedSet<AcceptedMessage> getMessages(final String tag) {
-        return messagesContainer.getMessages(tag);
-    }
-
-    /**
-     * Add a message to the application.
-     * Also schedules a expiration timer to remove this message after its expiration.
-     *
-     * @param message the message
-     */
-    public void addMessage(final Message message) {
-        messagesContainer.addMessage(message);
-    }
-
-    /**
-     * Remove all messages containing at least one of the tags.
-     * @param tags set of tags
-     * @param text message text (can be null, empty)
-     */
-    public void removeAnyMessage(final Set<String> tags, final String text) {
-        messagesContainer.removeAnyMessage(tags, text);
-    }
-
-    /**
-     * @return all messages regardless their tag
-     */
-    public Set<AcceptedMessage> getAllMessages() {
-        return messagesContainer.getAllMessages();
-    }
-
     public Path getDtagsEftarPath() {
         return syncReadConfiguration(Configuration::getDtagsEftarPath);
     }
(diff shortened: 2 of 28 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -123,6 +120,8 @@
 
     public WatchDogService watchDog;
 
+    private final Set<OnConfigurationChangedListener> listeners = new CopyOnWriteArraySet<>();
```
vladak, line 123: why CopyOnWriteArraySet ?

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
```diff
@@ -222,6 +234,7 @@ public void setMessageLimit(int limit) {
     }
 
     public void startExpirationTimer() {
+        setMessageLimit(RuntimeEnvironment.getInstance().getMessageLimit());
```
vladak, line 237: getMessageLimit() ? i.e. why call into RuntimeEnvironment when there is the callback ?

vladak:
Makes sense to me. There seem to be some build failures due to Javadoc errors.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
ahornace: I'm iterating over the set when the configuration is changed. This could throw https://docs.oracle.com/javase/7/docs/api/java/util/ConcurrentModificationException.html if someone would try to register another listener at that time.

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
ahornace: To be honest, I was not thinking about this too much. This code was there before in `RuntimeEnvironment` so I wanted to preserve the functionality:
```java
    public void startExpirationTimer() {
         messagesContainer.setMessageLimit(getMessageLimit());
         messagesContainer.startExpirationTimer();
     }
```
I think I can set the limit in the constructor and leave this call here out..

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.indexer.configuration;
+
+public interface OnConfigurationChangedListener {
+
+    void onConfigurationChanged();
```
tulinkry, line 27: I'm thinking if it's a good idea to have the configuration as the parameter. Perhaps not as RE provides the locking mechanism for reads and writes.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.indexer.configuration;
+
+public interface OnConfigurationChangedListener {
```
tulinkry, line 25: I think the name should be `ConfigurationChangedListener`

opengrok-web/src/main/java/org/opengrok/web/api/constraints/PositiveDuration.java
```diff
@@ -37,7 +37,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface PositiveDuration {
 
-    String message() default "{org.opengrok.indexer.web.api.constraints.PositiveDuration.message}";
+    String message() default "{org.opengrok.web.api.constraints.PositiveDuration.message}";
```
tulinkry, line 40: I'm not familiar with this, can you in short explain what it does?

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
```diff
@@ -51,14 +53,16 @@
...
-public class MessagesContainer {
+public class MessagesContainer implements OnConfigurationChangedListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MessagesContainer.class);
 
     public static final String MESSAGES_MAIN_PAGE_TAG = "main";
 
     private static final int DEFAULT_MESSAGE_LIMIT = 100;
 
+    private static final MessagesContainer instance = new MessagesContainer();
```
tulinkry, line 64: capital letters

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/MessagesControllerTest.java
```diff
@@ -147,30 +146,29 @@ protected TestContainerFactory getTestContainerFactory() throws TestContainerExc
...
+
+    @SuppressWarnings("unchecked")
     @Before
     public void setupMessageListener() throws Exception {
-        setMessageContainer(env, new MessagesContainer());
-        env.startExpirationTimer();
+        Field messagesField = MessagesContainer.class.getDeclaredField("tagMessages");
+        messagesField.setAccessible(true);
+        ((Map<String, SortedSet<AcceptedMessage>>) messagesField.get(container)).clear();
+        container.startExpirationTimer();
```
tulinkry, line 157: wouldn't it be nices to introduce public constructor (without RE) and use different instance for each test?

tulinkry: like it was before (except the RE field hacking)

opengrok-web/src/test/java/org/opengrok/web/messages/MessagesUtilsTest.java
```diff
@@ -18,40 +18,50 @@
...
 
-    private MessagesContainer container;
+    private MessagesContainer container = MessagesContainer.getInstance();
 
     @After
-    public void tearDown() {
+    public void tearDown() throws Exception {
+        Constructor<MessagesContainer> constructor = MessagesContainer.class.getDeclaredConstructor();
+        constructor.setAccessible(true);
+        container = constructor.newInstance();
```
tulinkry, line 53: same as my previous comment, can you introduce public constructor without the RE parameter?

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
tulinkry: Also because this method is called when the webapp starts and it's not calling the applyConfig where the callbacks are executed

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
```diff
@@ -0,0 +1,29 @@
...
+
+    void onConfigurationChanged();
+
```
tulinkry, line 28: Although the thing is pretty self-descriptive, it is not self-evident how to obtain the new configuration values. Could you add a javadoc with brief information about that. I think this interface will get a lot of new use cases later.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
ahornace: If `Configuration` could be accessed then yes but the RE is there for the locking.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
ahornace: https://developer.android.com/reference/android/view/View.OnClickListener

It's a matter of taste but the naming like this is used quite often.

opengrok-web/src/main/java/org/opengrok/web/api/constraints/PositiveDuration.java
ahornace: https://www.baeldung.com/javax-validation

The default message if the validation of the bean does not pass.

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
ahornace: Actually no, this is not a constant in a matter of speaking and capital letters should not be used. I already posted the java language specification regarding this to @vladak once.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/MessagesControllerTest.java
ahornace: IMHO you should never modify the code for the ease of testing.

opengrok-web/src/main/java/org/opengrok/web/messages/MessagesContainer.java
tulinkry: 2 lines above the naming is consistently with capital letters

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OnConfigurationChangedListener.java
tulinkry:
```
opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
opengrok-indexer/src/main/java/org/opengrok/indexer/index/DefaultIndexChangedListener.java 
opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/NonSymbolMatchedListener.java
opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/SymbolMatchedListener.java
opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexChangedListener.java
```

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/MessagesControllerTest.java
tulinkry: On the other hand, classes should be easily testable which is not the case when the singleton is enforced like this

ahornace: We should use dependency injection but that is for another day.

tulinkry: Agreed

tulinkry: You can even use the existing consumer with the RE as parameter, IMHO still better than reflection without any pain

opengrok-web/src/main/java/org/opengrok/web/api/constraints/PositiveDuration.java
tulinkry: thanks

vladak:
needs rebase

PR #3278, opened 2020-10-14
Add tablesorter webjar dependency

ahornace:
This removes ca. 3k lines from our source code and uses minified version of tablesorter which might help a bit with the performance.

I've noticed that `npm` version of webjar artifacts has specified dependencies which was bringing a few artifacts that were not needed.

I've tested sorting the directory listing which was working fine. I've tried to include `theme.default.min.css` from the webjar but it broke the UI:
Screenshot 2020-10-14 at 14 33 00

Thanks.


tulinkry (approved):
Actually the table style looks nicer, shame it's not working.

ahornace:
> Actually the table style looks nicer, shame it's not working.

It is working but the width is 100% and looks different. If you like it, then we can use it :)

tulinkry:
Can you post a snippet how you included the `theme.default.min.css`. If I can find some time in the future to look at it

ahornace:
> Can you post a snippet how you included the theme.default.min.css? If I can find some time in the future to look at it

Sure thing:
```
diff --git a/opengrok-web/pom.xml b/opengrok-web/pom.xml
index 95c89d6e77..2df78c1afa 100644
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -115,6 +115,11 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
             <artifactId>activation</artifactId>
             <version>1.1</version>
         </dependency>
+        <dependency>
+            <groupId>org.webjars</groupId>
+            <artifactId>webjars-taglib</artifactId>
+            <version>0.3</version>
+        </dependency>
         <dependency>
             <groupId>org.webjars</groupId>
             <artifactId>jquery</artifactId>
diff --git a/opengrok-web/src/main/webapp/httpheader.jspf b/opengrok-web/src/main/webapp/httpheader.jspf
index 2479d95d62..672a3f1770 100644
--- a/opengrok-web/src/main/webapp/httpheader.jspf
+++ b/opengrok-web/src/main/webapp/httpheader.jspf
@@ -34,6 +34,7 @@ to set the title of the document before the include directive for this fragment:

 </pre>
 --%>
+<%@taglib uri="http://www.webjars.org/tags" prefix="wj"%>
 <%@page contentType="text/html; charset=UTF-8" pageEncoding="UTF-8" import="
 org.opengrok.indexer.Info,
 org.opengrok.web.PageConfig,
@@ -67,7 +68,9 @@ org.opengrok.web.Scripts"
 <link rel="stylesheet" type="text/css" href="<%=styleDir%>/jquery-ui-1.12.1-custom.structure.min.css" />
 <link rel="stylesheet" type="text/css" href="<%=styleDir%>/jquery-ui-1.12.1-custom.theme.min.css" />
 <link rel="stylesheet" type="text/css" href="<%=styleDir%>/jquery.tooltip.min.css" />
-<link rel="stylesheet" type="text/css" href="<%=styleDir%>/jquery.tablesorter.min.css" />
+<wj:locate path="theme.default.min.css" var="cssUrl" relativeTo="META-INF/resources">
+    <link rel="stylesheet" type="text/css" href="<%=ctxPath%>${cssUrl}">
+</wj:locate>
 <link rel="stylesheet" type="text/css" href="<%=styleDir%>/searchable-option-list-2.0.3.min.css" />

 <%
```
I'm not 100% what is the performance of `wj:locate`. I hope it caches the result so it would not have to search the classpath every page refresh.

vladak:
I don't mind the spacing of the table. The missing alternating line colors are deal breaker however I'd say.

opengrok-web/pom.xml
```diff
@@ -124,6 +124,23 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
             <groupId>org.webjars.npm</groupId>
             <artifactId>showdown</artifactId>
             <version>1.9.1</version>
+            <exclusions>
+                <exclusion>
```
vladak, line 128: why is this necessary ?

ahornace: https://www.npmjs.com/package/showdown has one dependency: `yargs` which in turn has 7 dependencies etc. and I haven't noticed that the dependency is required for our needs -> it has worked without it before and even now :D

tulinkry: I may be a good idea to write a comment there that we don't actually need any dependencies from it. After a year man can be surprised what was the intention here.

ahornace: done

vladak:
Thanks guys !

PR #3279, opened 2020-10-14
use statsd for indexer monitoring

vladak:
This change introduces remote monitoring of indexer via statsd registry in Micrometer. I wanted to use statsd for indexer because it does not require scraping and therefore the need for a web server. This would either conflict with already running web server on the same machine (assuming the indexer is usually run on the same machine as the web application) or would involve poking holes into firewall to allow the scraping to pass through to distinct TCP port.

Example read-only configuration:
```xml

  <void property="statsdConfig">
     <void property="port">
       <int>8125</int>
     </void>
     <void property="host">
       <string>localhost</string>
     </void>
     <void property="flavor">
       <string>etsy</string>
     </void>
  </void>

 </object>
</java>
```

This was used to generate the graphs in https://github.com/oracle/opengrok/pull/3248#issuecomment-706603506 (I used statsd-exporter for Prometheus and Grafana)

The metrics are not tagged unless the indexer is run per project.

During testing I noticed that some of the metrics recorded in the final stages of indexing are not sent out even with buffering set to false and big delay at the end of the Indexer#main. This is not such a big deal given that statsd traffic is usually not reliable anyway (assuming UDP by default). I will definitely investigate.

Once this is merged in I will update https://github.com/oracle/opengrok/wiki/Monitoring with example rig configuration.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
@@ -22,6 +22,9 @@
  */
 package org.opengrok.indexer;
 
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
@@ -29,12 +32,95 @@ import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
 import io.micrometer.core.instrument.binder.system.ProcessorMetrics;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
+import io.micrometer.statsd.StatsdConfig;
+import io.micrometer.statsd.StatsdMeterRegistry;
+import io.micrometer.statsd.StatsdFlavor;
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+import org.opengrok.indexer.index.Indexer;
+import org.opengrok.indexer.logger.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+/**
+ * Encapsulates logic of meter registry setup and handling.
+ * Generally, the web application publishes metrics to Prometheus and the Indexer to StatsD.
+ */
 public final class Metrics {
 
-    private static final PrometheusMeterRegistry registry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
+    private static final Logger LOGGER = LoggerFactory.getLogger(Metrics.class);
+
+    private static final StatsdConfig statsdConfig = new StatsdConfig() {
+        @Override
+        public String get(String k) {
+            return null;
+        }
+
+        @Override
+        public StatsdFlavor flavor() {
+            String flavor = RuntimeEnvironment.getInstance().getStatsdConfig().getFlavor().toLowerCase();
+            switch (flavor) {
+                case "etsy":
+                    return StatsdFlavor.ETSY;
+                case "datadog":
+                    return StatsdFlavor.DATADOG;
+                case "sysdig":
+                    return StatsdFlavor.SYSDIG;
+                case "telegraf":
+                    return StatsdFlavor.TELEGRAF;
+            }
+
+            return null;
+        }
+
+        @Override
+        public int port() {
+            return RuntimeEnvironment.getInstance().getStatsdConfig().getPort();
+        }
+
+        @Override
+        public String host() {
+            return RuntimeEnvironment.getInstance().getStatsdConfig().getHost();
+        }
+
+        @Override
+        public boolean buffered() {
+            return true;
+        }
+    };
+
+    private static PrometheusMeterRegistry prometheusRegistry;
+    private static StatsdMeterRegistry statsdRegistry;
 
     static {
+        MeterRegistry registry;
+
+        if (RuntimeEnvironment.getInstance().getStatsdConfig().isEnabled()) {
+            LOGGER.log(Level.INFO, "configuring StatsdRegistry");
+
+            statsdRegistry = new StatsdMeterRegistry(statsdConfig, Clock.SYSTEM);
+
+            // Add tag for per-project reindex.
+            ArrayList<String> subFiles = RuntimeEnvironment.getInstance().getSubFiles();
+            if (subFiles.size() > 0) {
+                List<String> sList = subFiles.stream().
+                        map(s -> s.startsWith(Indexer.PATH_SEPARATOR_STRING) ? s.substring(1) : s).
+                        collect(Collectors.toList());
+                statsdRegistry.config().commonTags(Collections.singleton(Tag.of("projects",
+                        String.join(",", sList))));
+            }
+
+            registry = statsdRegistry;
+        } else {
+            LOGGER.log(Level.INFO, "configuring PrometheusRegistry");
+            prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
+            registry = prometheusRegistry;
+        }
+
         new ClassLoaderMetrics().bindTo(registry);
         new JvmMemoryMetrics().bindTo(registry);
         new JvmGcMetrics().bindTo(registry);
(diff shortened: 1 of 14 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/StatsdConfig.java
```diff
@@ -0,0 +1,74 @@
...
+package org.opengrok.indexer.configuration;
+
+/**
+ * Configuration for Statsd metrics emitted by the Indexer via {@link org.opengrok.indexer.util.Statistics}.
+ */
+public class StatsdConfig {
+    private int port;
+    private String host;
+    private boolean enabled;
+    private String flavor;
```
ahornace, line 33: Cannot we use `io.micrometer.statsd.StatsdFlavor` enum? E.g. `RemoteSCM` is an enum and used in configuration so we should be good to go, right?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -123,6 +123,12 @@
 
     public WatchDogService watchDog;
 
+    public ArrayList<String> getSubFiles() {
```
ahornace, line 126:
```suggestion
    public List<String> getSubFiles() {
```

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -123,6 +123,12 @@
...
+        return subFiles;
+    }
+
+    private ArrayList<String> subFiles = new ArrayList<>();
```
ahornace, line 130:
```suggestion
    private List<String> subFiles = new ArrayList<>();
```

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -22,19 +22,105 @@
...
+        MeterRegistry registry;
+
+        if (RuntimeEnvironment.getInstance().getStatsdConfig().isEnabled()) {
+            LOGGER.log(Level.INFO, "configuring StatsdRegistry");
+
+            statsdRegistry = new StatsdMeterRegistry(statsdConfig, Clock.SYSTEM);
+
+            // Add tag for per-project reindex.
+            ArrayList<String> subFiles = RuntimeEnvironment.getInstance().getSubFiles();
+            if (subFiles.size() > 0) {
```
ahornace, line 109:
```suggestion
            if (!subFiles.isEmpty()) {
```

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -22,19 +22,105 @@
...
+                List<String> sList = subFiles.stream().
+                        map(s -> s.startsWith(Indexer.PATH_SEPARATOR_STRING) ? s.substring(1) : s).
+                        collect(Collectors.toList());
```
ahornace, line 112: You can create the joined String directly with `Collectors.joining()`

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -22,19 +22,105 @@
...
+        }
+    };
+
+    private static PrometheusMeterRegistry prometheusRegistry;
+    private static StatsdMeterRegistry statsdRegistry;
 
     static {
+        MeterRegistry registry;
+
+        if (RuntimeEnvironment.getInstance().getStatsdConfig().isEnabled()) {
```
ahornace, line 102: Why not use `RuntimeEnvironment.getInstance().isIndexer()` here as well?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -45,8 +131,23 @@
...
 
+    /**
+     * Get registry based on running context.
+     * @return MeterRegistry instance
+     */
+    public static MeterRegistry getRegistry() {
+        if (RuntimeEnvironment.getInstance().isIndexer()) {
+            return Metrics.getStatsdRegistry();
+        } else {
+            return Metrics.getPrometheusRegistry();
```
ahornace, line 150: Why use `Metrics.` prefix if the methods are in the same class?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/StatsdConfig.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getFlavor() {
+        return flavor;
+    }
+
+    public void setFlavor(String flavor) {
+        this.flavor = flavor;
+    }
+
+    public boolean isEnabled() {
+        return port != 0 && !host.isEmpty() && !flavor.isEmpty();
```
ahornace, line 60: This can cause NPE if `host` or `flavor` will be null or it should not happen?

ahornace (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: because statsd does not have to be configured. I don't want statsd export to be enabled by default.

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -22,19 +22,105 @@
...
+                        map(s -> s.startsWith(Indexer.PATH_SEPARATOR_STRING) ? s.substring(1) : s).
+                        collect(Collectors.toList());
+                statsdRegistry.config().commonTags(Collections.singleton(Tag.of("projects",
+                        String.join(",", sList))));
+            }
+
+            registry = statsdRegistry;
+        } else {
+            LOGGER.log(Level.INFO, "configuring PrometheusRegistry");
+            prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
```
ahornace, line 120: Let's say that statsd is disabled. What is the point of creating Prometheus registry if it cannot be accessed?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: fixed, thanks

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -22,19 +22,105 @@
...
     static {
+        MeterRegistry registry;
+
+        if (RuntimeEnvironment.getInstance().getStatsdConfig().isEnabled()) {
+            LOGGER.log(Level.INFO, "configuring StatsdRegistry");
+
+            statsdRegistry = new StatsdMeterRegistry(statsdConfig, Clock.SYSTEM);
+
+            // Add tag for per-project reindex.
+            ArrayList<String> subFiles = RuntimeEnvironment.getInstance().getSubFiles();
```
ahornace: This looks kind of iffy to me. We are filling `subFiles` in `main` method of `Indexer`. We are now basically counting on the fact that `Metrics` class won't be loaded by classloader before the list is filled.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/StatsdConfig.java
vladak: happens with no or incomplete StatsdConfig, fixed.

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: right, fixed.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/StatsdConfig.java
vladak: That certainly works however it will make the configuration slightly more ugly:
```xml

  <void property="statsdConfig">
     <void property="port">
       <int>8125</int>
     </void>
     <void property="host">
       <string>localhost</string>
     </void>
<!--
     <void property="flavor">
       <enum>etsy</enum>
     </void>
-->
     <void property="flavor">
       <object class="java.lang.Enum" method="valueOf">
         <class>io.micrometer.statsd.StatsdFlavor</class>
         <string>ETSY</string>
       </object>
    </void>
  </void>

 </object>
</java>
```

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: good catch. added a callback to main().

ahornace (approved):

vladak:
Thanks for the review !

PR #3281, opened 2020-10-15
upgrade Chronicle map

vladak:
This seems to fix the failing Docker build.

PR #3282, opened 2020-10-15
upgrade Chronicle map

vladak:
The previous fix in #3281 failed. It seems the chronicle-core needs to be more in line with chronicle-map.


ahornace:
Thanks! :)

PR #3284, opened 2020-10-16
add Graphite support

vladak:
This change introduces support for Graphite metering repository. Also it allows to mix and match metering repositories for the indexer and the webapp (the only exception is Prometheus registry for the indexer). There is always at most one repository for each.

The default is StatsD for the indexer and Prometheus for the web app.

Sample configuration:
```xml

  <void property="webAppMeterRegistryType">
     <object class="java.lang.Enum" method="valueOf">
       <class>org.opengrok.indexer.configuration.MeterRegistryType</class>
       <string>GRAPHITE</string>
     </object>
  </void>

  <void property="indexerMeterRegistryType">
     <object class="java.lang.Enum" method="valueOf">
       <class>org.opengrok.indexer.configuration.MeterRegistryType</class>
       <string>STATSD</string>
     </object>
  </void>

  <void property="graphiteConfig">
     <void property="port">
       <int>2005</int>
     </void>
     <void property="host">
       <string>localhost</string>
     </void>
    <void property="protocol">
       <object class="java.lang.Enum" method="valueOf">
         <class>io.micrometer.graphite.GraphiteProtocol</class>
         <string>UDP</string>
       </object>
    </void>
  </void>

  <void property="statsdConfig">
     <void property="port">
       <int>8125</int>
     </void>
     <void property="host">
       <string>localhost</string>
     </void>
     <void property="flavor">
       <object class="java.lang.Enum" method="valueOf">
         <class>io.micrometer.statsd.StatsdFlavor</class>
         <string>ETSY</string>
       </object>
    </void>
  </void>

 </object>
</java>
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
@@ -38,6 +38,8 @@ import io.micrometer.graphite.GraphiteMeterRegistry;
 import io.micrometer.statsd.StatsdConfig;
 import io.micrometer.statsd.StatsdMeterRegistry;
 import io.micrometer.statsd.StatsdFlavor;
+import org.opengrok.indexer.configuration.BaseGraphiteConfig;
+import org.opengrok.indexer.configuration.BaseStatsdConfig;
 import org.opengrok.indexer.configuration.Configuration;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.index.Indexer;
@@ -57,7 +59,7 @@ public final class Metrics {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(Metrics.class);
 
-    private StatsdConfig getStatsdConfig() {
+    private StatsdConfig getStatsdConfig(BaseStatsdConfig baseStatsdConfig) {
         return new StatsdConfig() {
             @Override
             public String get(String k) {
@@ -66,17 +68,17 @@ public final class Metrics {
 
             @Override
             public StatsdFlavor flavor() {
-                return RuntimeEnvironment.getInstance().getStatsdConfig().getFlavor();
+                return baseStatsdConfig.getFlavor();
             }
 
             @Override
             public int port() {
-                return RuntimeEnvironment.getInstance().getStatsdConfig().getPort();
+                return baseStatsdConfig.getPort();
             }
 
             @Override
             public String host() {
-                return RuntimeEnvironment.getInstance().getStatsdConfig().getHost();
+                return baseStatsdConfig.getHost();
             }
 
             @Override
@@ -86,7 +88,7 @@ public final class Metrics {
         };
     }
 
-    private GraphiteConfig getGraphiteConfig() {
+    private GraphiteConfig getGraphiteConfig(BaseGraphiteConfig baseGraphiteConfig) {
         return new GraphiteConfig() {
             @Override
             public String get(String k) {
@@ -95,17 +97,17 @@ public final class Metrics {
 
             @Override
             public int port() {
-                return RuntimeEnvironment.getInstance().getGraphiteConfig().getPort();
+                return baseGraphiteConfig.getPort();
             }
 
             @Override
             public String host() {
-                return RuntimeEnvironment.getInstance().getGraphiteConfig().getHost();
+                return baseGraphiteConfig.getHost();
             }
 
             @Override
             public GraphiteProtocol protocol() {
-                return RuntimeEnvironment.getInstance().getGraphiteConfig().getProtocol();
+                return baseGraphiteConfig.getProtocol();
             }
         };
     }
@@ -133,20 +135,22 @@ public final class Metrics {
     public void configure(Configuration.MeterRegistryType type) {
         switch (type) {
             case PROMETHEUS:
-                LOGGER.log(Level.INFO, "configuring Prometheus registry");
+                LOGGER.log(Level.INFO, "configuring Prometheus registry with default config");
                 prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
                 activeRegistry = prometheusRegistry;
                 break;
             case GRAPHITE:
-                if (RuntimeEnvironment.getInstance().getGraphiteConfig().isEnabled()) {
-                    LOGGER.log(Level.INFO, "configuring Graphite registry");
-                    activeRegistry = new GraphiteMeterRegistry(getGraphiteConfig(), Clock.SYSTEM);
+                BaseGraphiteConfig baseGraphiteConfig = RuntimeEnvironment.getInstance().getBaseGraphiteConfig();
+                if (baseGraphiteConfig.isEnabled()) {
+                    LOGGER.log(Level.INFO, "configuring Graphite registry with {0}", baseGraphiteConfig.toString());
+                    activeRegistry = new GraphiteMeterRegistry(getGraphiteConfig(baseGraphiteConfig), Clock.SYSTEM);
                 }
                 break;
             case STATSD:
-                if (RuntimeEnvironment.getInstance().getStatsdConfig().isEnabled()) {
-                    LOGGER.log(Level.INFO, "configuring Statsd registry");
-                    activeRegistry = new StatsdMeterRegistry(getStatsdConfig(), Clock.SYSTEM);
+                BaseStatsdConfig baseStatsdConfig = RuntimeEnvironment.getInstance().getBaseStatsdConfig();
+                if (baseStatsdConfig.isEnabled()) {
+                    LOGGER.log(Level.INFO, "configuring Statsd registry with {0}", baseStatsdConfig.toString());
+                    activeRegistry = new StatsdMeterRegistry(getStatsdConfig(baseStatsdConfig), Clock.SYSTEM);
                 }
                 break;
             case NONE:
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
@@ -154,6 +154,7 @@ public final class Metrics {
                 }
                 break;
             case NONE:
+                activeRegistry = null;
                 break;
             default:
                 throw new IllegalArgumentException("unsupported registry type");
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
@@ -149,6 +149,8 @@ public final class Metrics {
                     activeRegistry = new StatsdMeterRegistry(getStatsdConfig(), Clock.SYSTEM);
                 }
                 break;
+            case NONE:
+                break;
             default:
                 throw new IllegalArgumentException("unsupported registry type");
         }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
@@ -30,11 +30,15 @@ import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
 import io.micrometer.core.instrument.binder.system.ProcessorMetrics;
+import io.micrometer.graphite.GraphiteProtocol;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
+import io.micrometer.graphite.GraphiteConfig;
+import io.micrometer.graphite.GraphiteMeterRegistry;
 import io.micrometer.statsd.StatsdConfig;
 import io.micrometer.statsd.StatsdMeterRegistry;
 import io.micrometer.statsd.StatsdFlavor;
+import org.opengrok.indexer.configuration.Configuration;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.index.Indexer;
 import org.opengrok.indexer.logger.LoggerFactory;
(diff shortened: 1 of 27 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -315,6 +316,15 @@
         ON, OFF, DIRBASED, UIONLY
     }
 
+    public enum supportedMeterRegistryType {
```
tulinkry, line 319: I don't remember if I have ever seen a type with lowercase first letter

tulinkry: Anyway, I think MetricType is enough ... the "supported" part is kinda the purpose of the enum.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/GraphiteConfig.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.opengrok.indexer.configuration;
+
+import io.micrometer.graphite.GraphiteProtocol;
+
+public class GraphiteConfig {
+    private int port;
+    private String host;
+    private boolean enabled;
+    private GraphiteProtocol protocol;
```
tulinkry, line 32: Can this have a default value?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -53,90 +57,134 @@
...
-    private Metrics() {
+    /**
+     * Configure meter registry.
+     * @param type type of meter registry
+     */
+    public void configure(Configuration.supportedMeterRegistryType type) {
+        switch (type) {
+            case PROMETHEUS:
+                LOGGER.log(Level.INFO, "configuring Prometheus registry");
+                prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
```
tulinkry, line 137: prometheusRegistry does not need to be static if you set it inside an instance method. You still use this class as a singleton so you'll always have an instance from getInstance()

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
vladak: I will drop the 'supported' prefix. MetricType would be misleading - it should rather be MeterType however that would be imprecise since meter type is something like Timer, Gauge etc.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/GraphiteConfig.java
vladak: sure. which one do you think is reasonable ?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: right, good catch. This is leftover from before getInstance().

vladak:
Will update https://github.com/oracle/opengrok/wiki/Monitoring once this is in.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/GraphiteConfig.java
vladak: `isEnabled()` does not check the `protocol` field so it will remain set to `null`. The `GraphiteConfig#setProtocol()` will choose `GraphiteProtocol.PICKLED` in such case.

ahornace:
To be honest, I'm not sure I like this implementation. If we used only prometheus/statsd, then we could argue that we made a calculated decision to use those. However, there is ca. 20 micrometer plugins and support for each of those now would require code changes in OpenGrok. Why not add implementation for influx, wavefront, datadog or jmx which are more used than graphite (https://mvnrepository.com/search?q=micrometer-registry&p=1)? I think over time there will be feature requests for those as well.

I think we should make this a pluggable feature. Use a `ServiceLoader` or similar to load the `MetricsRegistry` since micrometer does not seem to support this at the moment (at least I had no luck finding anything). I've tried to check how they figured it out in Spring Boot but that seems too abstract: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java

vladak:
I think this change is a step forward, not the ultimate solution. To be honest the choice of StatsD+Prometheus was made purely on my personal preferences. If there are requests to add more meter registries we can always investigate a way how to inject them dynamically.

ahornace:
> I think this change is a step forward, not the ultimate solution. To be honest the choice of StatsD+Prometheus was made purely on my personal preferences. If there are requests to add more meter registries we can always investigate a way how to inject them dynamically.

Prometheus is the de facto standard at the moment and StatsD was chosen because of its push instead of pull model, right? I don't see anything wrong with that.

> If there are requests to add more meter registries we can always investigate a way how to inject them dynamically.

This is how it always starts. Now if I'd like to add support for anything else, then I would follow your example. "It's just one more argument." as they say :)

But okay, if you truly feel that this is just a step forwards, then I'm fine with it.

vladak:
The (minor) advantage compared to the current state is that you can have Statsd/Graphite registry for both webapp and the indexer. Not everyone wants to have separate monitoring data collection.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationHelp.java
```diff
@@ -174,6 +173,8 @@ private static Object getSampleValue(Method setter, Object defaultValue) {
             inm.add("f:user-specified-value");
             inm.add("d:user-specified-value");
             return inm;
+        } else if (paramType == Configuration.MeterRegistryType.class) {
```
idodeclare, line 176: Why doesn't the `paramType.isEnum()` handling of line 178 work for this?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/BaseGraphiteConfig.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Gets an instance version suitable for helper documentation by shifting
+     * most default properties slightly.
+     */
+    static BaseGraphiteConfig getForHelp() {
+        BaseGraphiteConfig res = new BaseGraphiteConfig();
+        res.setHost("foo.bar");
+        res.setPort(2004);
+        return res;
```
idodeclare, line 83: Missing a property shift for `protocol`?

opengrok-indexer/src/test/java/org/opengrok/indexer/configuration/MetricsConfigureTest.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package org.opengrok.indexer.configuration;
+
+import io.micrometer.graphite.GraphiteProtocol;
+import io.micrometer.statsd.StatsdFlavor;
+import org.junit.Test;
+import org.opengrok.indexer.Metrics;
+
+import static org.junit.jupiter.api.Assertions.*;
```
idodeclare, line 31: Doesn't `*` get disapproval?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -315,6 +316,16 @@
         ON, OFF, DIRBASED, UIONLY
     }
 
+    public enum MeterRegistryType {
```
idodeclare, line 319: Is there much value to having this be `configuration.Configuration.MeterRegistryType` instead of just `configuration.MeterRegistryType`? (Same question for that other, old enum.)

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationHelp.java
vladak: It chooses the value of `PROMETHEUS` which does not work with indexer, making tests fail:
```
java.lang.RuntimeException: error setting sample value PROMETHEUS for public void org.opengrok.indexer.configuration.Configuration.setIndexerMeterRegistryType(org.opengrok.indexer.configuration.Configuration$MeterRegistryType)

	at org.opengrok.indexer.configuration.ConfigurationHelp.getSamples(ConfigurationHelp.java:91)
	at org.opengrok.indexer.configuration.ConfigurationHelpTest.shouldCreateReadableUsage(ConfigurationHelpTest.java:36)
...
```

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/BaseGraphiteConfig.java
vladak: added.

opengrok-indexer/src/test/java/org/opengrok/indexer/configuration/MetricsConfigureTest.java
vladak: yes, I neglected that. We should really have that checked as part of the build (filed #3298).

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
vladak: Having this directly underneath `configuration` is certainly better. It seems I need to listen to my inner voices more carefully - had the same thought when introducing the enum and brushed it under carpet in the process of doing other changes.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -1317,17 +1323,47 @@ public void setSuggesterConfig(final SuggesterConfig config) {
...
+        this.webAppMeterRegistryType = registryType;
+    }
+
+    public MeterRegistryType getIndexerMeterRegistryType() {
+        return indexerMeterRegistryType;
+    }
+
+    public void setIndexerMeterRegistryType(MeterRegistryType registryType) {
+        if (registryType == MeterRegistryType.PROMETHEUS) {
+            throw new IllegalArgumentException("unsupported registry type");
```
idodeclare, line 1362: Can you elaborate on this? It smells off: a brand new enum yet partially illegal with this brand new API

opengrok-web/src/main/java/org/opengrok/web/servlet/MetricsServlet.java
```diff
@@ -39,7 +39,7 @@
     @Override
     protected void doGet(final HttpServletRequest req, final HttpServletResponse resp) throws IOException {
         try (PrintWriter pw = resp.getWriter()) {
-            pw.print(Metrics.getPrometheusRegistry().scrape());
+            pw.print(Metrics.getInstance().getPrometheusRegistry().scrape());
```
idodeclare, line 42: So unless `PROMETHEUS` then `NullPointerException` here?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -53,90 +59,145 @@
...
-    }
 
-    public static PrometheusMeterRegistry getPrometheusRegistry() {
-        return prometheusRegistry;
+        return commonTag;
     }
 
-    private static StatsdMeterRegistry getStatsdRegistry() {
-        return statsdRegistry;
+    public PrometheusMeterRegistry getPrometheusRegistry() {
```
idodeclare, line 192: Seemingly web-only...

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -53,90 +59,145 @@
...
+            new JvmThreadMetrics().bindTo(activeRegistry);
+        }
+    }
+
+    /**
+     * Set common tag according to list of files.
+     * @param subFiles list of files
+     * @return Tag object or null
+     */
+    public Tag updateSubFiles(List<String> subFiles) {
```
idodeclare, line 177: Seemingly indexer-only....

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -53,90 +59,145 @@
...
-    public static void updateSubFiles(List<String> subFiles) {
+    /**
+     * Configure meter registry.
+     * @param type type of meter registry
+     */
+    public void configure(MeterRegistryType type) {
+        switch (type) {
+            case PROMETHEUS:
+                LOGGER.log(Level.INFO, "configuring Prometheus registry with default config");
+                prometheusRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
```
idodeclare, line 139: If this method is called a second time with a different `type`, might not a previously existing, non-null `prometheusRegistry` be orphaned?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
```diff
@@ -53,90 +59,145 @@
...
+        return instance;
     }
 
-    public static void updateSubFiles(List<String> subFiles) {
+    /**
+     * Configure meter registry.
+     * @param type type of meter registry
+     */
+    public void configure(MeterRegistryType type) {
+        switch (type) {
```
idodeclare, line 136: If this method is called a second time, might not a previously existing, non-null `activeRegistry` be orphaned?

idodeclare (commented):
Wondering if `Metrics` is straining to cover both web and indexer

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
vladak: Well, the indexer does not have embedded web server so cannot support Prometheus registry. Do you think it would make sense to split `MeterRegistryType` ? Or even `Metrics` ?

opengrok-web/src/main/java/org/opengrok/web/servlet/MetricsServlet.java
vladak: This was handled before however got lost with rebasing/refactoring, reverting.

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: It is.

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: Yep.

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: I realize this is not a good design however makes testing easier in `MetricsConfigureTest`. Throw an exception as an alternative (or maybe silently return and document that) and split the test class ?

opengrok-indexer/src/main/java/org/opengrok/indexer/Metrics.java
vladak: ditto

PR #3286, opened 2020-10-18
increasing the max line length for python tools checkstyle

tulinkry:
Increase the line length in python tools to 120

### DIFF
--- a/opengrok-tools/pom.xml
+++ b/opengrok-tools/pom.xml
@@ -42,6 +42,7 @@ Portions Copyright (c) 2017-2018, 2020, Chris Fraire <cfraire@me.com>.
         <python.system.executable>python3</python.system.executable>
         <python.environment>${project.build.directory}/env/bin</python.environment>
         <project.python.package.version>${project.version}</project.python.package.version>
+        <python.checkstyle.line-length>119</python.checkstyle.line-length>
     </properties>
 
     <profiles>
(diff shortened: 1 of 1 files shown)


opengrok-tools/pom.xml
```diff
@@ -280,6 +280,8 @@ Portions Copyright (c) 2017-2018, 2020, Chris Fraire <cfraire@me.com>.
                             <arguments>
                                 <argument>-m</argument>
                                 <argument>pylint</argument>
+                                <argument>--max-line-length</argument>
+                                <argument>119</argument>
```
ahornace, line 284: Why 119? I thought that the rule is 120 inclusive.

opengrok-tools/pom.xml
```diff
@@ -299,6 +301,8 @@ Portions Copyright (c) 2017-2018, 2020, Chris Fraire <cfraire@me.com>.
                             <arguments>
                                 <argument>-m</argument>
                                 <argument>flake8</argument>
+                                <argument>--max-line-length</argument>
+                                <argument>119</argument>
```
vladak, line 305: introduce a variable ?

vladak:
I have no problem with this. This should match the style used for Java.

opengrok-tools/pom.xml
tulinkry: The default was:

```
  --max-line-length n   Maximum allowed line length for the entirety of this run. (Default: 79)
```

so I believe it is 120 including the newline.

opengrok-tools/pom.xml
tulinkry: sure

PR #3292, opened 2020-10-19
make Userwhitelist plugin generic

vladak:
This change allows to choose the field to check against in the `Userwhitelist` plugin.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2020, Chris Fraire <cfraire@me.com>.
  */
 package opengrok.auth.plugin;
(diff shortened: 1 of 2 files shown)


plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
```diff
@@ -77,7 +90,13 @@ private boolean checkWhitelist(HttpServletRequest request) {
             return false;
         }
 
-        return whitelist.contains(user.getUsername());
+        if (fieldName.equals(USERNAME_FIELD)) {
+            return user.getUsername() != null && whitelist.contains(user.getUsername());
+        } else if (fieldName.equals(ID_FIELD)) {
+            return user.getId() != null && whitelist.contains(user.getId());
+        }
```
tulinkry, line 97: I'd prefer enum, should be faster, could be arguably more readable

plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
```diff
@@ -44,18 +44,31 @@
...
+    // configuration parameters
     static final String FILE_PARAM = "file";
+    static final String FIELD_PARAM = "fieldName";
+
+    // valid values for the FIELD_NAME parameter
+    static final String USERNAME_FIELD = "username";
+    static final String ID_FIELD = "id";
 
     private final Set<String> whitelist = new TreeSet<>();
+    private String fieldName;
```
tulinkry, line 56: can be final

plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
```diff
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019-2020 Oracle and/or its affiliates. All rights reserved.
```
ahornace, line 21: Aren't we using `2019, 2020,` format?

vladak: yes, fixed. There should be a copyright format checker I guess. Filed #3300.

plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
vladak: It cannot. The value is assigned in `load()`.

plugins/src/main/java/opengrok/auth/plugin/UserWhiteListPlugin.java
vladak: I'd rather keep the configuration parameters of plugins restricted to strings. Or do you mean introduce the enum alongside the configuration only for internal handling in the plugin ?

vladak:
I have to say it was definitely pleasant to have the test already in place, @idodeclare. Increasing test coverage definitely makes sense.

vladak:
will update https://github.com/oracle/opengrok/wiki/Authorization-plugins#userwhitelistplugin

PR #3098, opened 2020-03-29
[WIP] Move to Java 11 as the minimum version to anticipate Lucene

idodeclare:
Upcoming Lucene 9.0.0 has updated its System Requirements with LUCENE-8738: Move to Java 11 as minimum Java version.

This is a patch to try a revised configuration to move OpenGrok to 11 too.

Issue #2528, opened 2018-11-21
TLS failures when running setup.py on newer macOS builds in Travis due to missing CA bundle


vladak:
Newer xcode environments in Travis fail with TLS failures due to missing CA bundles when running `setup.py`.

vladak:
This is blocking transition of the macOS builds to newer JDK.

vladak:
Example of the failure:
```
[INFO] --- exec-maven-plugin:1.6.0:exec (Install python package for tests/pylint) @ opengrok-tools ---

Download error on https://pypi.org/simple/pytest-runner/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1051) -- Some packages may not be found!

Couldn't find index page for 'pytest-runner' (maybe misspelled?)
```

vladak:
This failure is described in pypa/pip#4205

Installing `certifi` Python package (globally or in the venv) did not help.

The Linux builds succeed so the Python there has to access to a CA bundle.

vladak:
Looks like macOS also needs to run `Install Certificates.command` after the `certifi` installation: https://stackoverflow.com/a/56031239/11582827

vladak:
I think https://github.com/pypa/pip/issues/4205#issuecomment-290587840 is the closest to what is happening as it is the setuptools based installation that is failing.

vladak:
> Looks like macOS also needs to run `Install Certificates.command` after the `certifi` installation: https://stackoverflow.com/a/56031239/11582827

This only works if using official Python distribution, not the one from HomeBrew like the macOS build uses.

vladak:
Tried to reproduce the problem in https://github.com/vladak/setuptools_certs however failed so far.

tulinkry:
I am not able to reproduce the error. I took your setup, added maven, did some small changes to match it to the current situation in opengrok: https://github.com/tulinkry/setuptools_certs

```
mvn verify
```
executes just fine

tulinkry:
I think I got it, the commit which broke it is https://github.com/tulinkry/setuptools_certs/commit/a683be89d00966ccab52df19416bc4c4e10af65e.

tulinkry:
After series of tries this addon
```
  coverity_scan:
    project:
      name: OpenGrok/OpenGrok
      description: Build submitted via Travis CI
      branch_pattern: coverity_scan
[...]

PR #3294, opened 2020-10-20
macOS Xcode 10 vs. Python setuptools

vladak:
workaround for #2528 - in order to prevent setuptools from reaching out to PyPi, the needed packages are installed first via pip. This is done so that it is possible to merge #3098.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,10 @@ jobs:
     before_install: dev/before_install
     before_script: dev/before
   - stage: test
-    name: macOS + Oracle JDK 8
+    name: macOS + Oracle JDK 11
     os: osx
-    osx_image: xcode8.3
+    osx_image: xcode10.1
+    jdk: oraclejdk11
     install: true
     script: dev/main
     before_install: dev/before_install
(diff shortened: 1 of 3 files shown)


tulinkry (commented):
The only thing I don't like that you use the requirements.txt file which is there kept for reference and not really doing anything. Both of the requirements are now spread to two different places which is not ideal.

Let me try this evening to actually load the requirements in setup.py from the given files, so only one source of truth is available.

vladak:
> The only thing I don't like that you use the requirements.txt file which is there kept for reference only and not really doing anything. Both of the requirements are now spread to two different places which is not ideal.

Yeah, this is still confusing the heck out of me.
 
> I think the goal would be to actually load the requirements in setup.py from the given files, so only one source of truth is available. I can try tonight if you're busy.

Ok, will wait for that.

tulinkry:
I've come up with f00a8adec81620133893d550133c684936bd0d9e - loading the requirements from file. But if you find a different way - possibly just using the requirements in setup.py, that would be much better. Perhaps tweaking the macosx env?

Because the goal of requirements.txt is to replicate the dev environment used for developing the package (`pip freeze`) while the requirements in setup.py are requirements to actually run the package (without specific version locks).

vladak:
I had something like this in mind. As a workaround that would be acceptable I guess.

vladak:
closing in favor of #3295

PR #3304, opened 2020-10-22
change default path to dtags.eftar

vladak:
Keeping the `dtags.eftar` file underneath the `index` directory does not make sense to me. After all it does not hold index data. Also, when upgrading OpenGrok from scratch, I tend to clone the directory structure (yes, via ZFS) and recreate the subdirectories (er, datasets) such as `index,historycache,...`. With this procedure the file is lost.


vladak:
macOS build failed due to #3313

tulinkry (approved):

vladak:
Thanks @tulinkry !

PR #3311, opened 2020-10-22
Add copyright part to header check

### LINKED ISSUE #3300: copyright format checker
There should be copyright format checker run alongside the maven checkstyle checker that would enforce format of the copyright notes to be in the format of `{Portions ,}Copyright (c) <year>{, <year>}, comment`.

ahornace:
fixes #3300

Thanks.


idodeclare:
CDDL 3.3 "You may not remove or alter any copyright, patent or trademark notices contained within the Covered Software, or any notices of licensing or any descriptive text giving attribution to any Contributor or the Initial Developer."

tulinkry (approved):
Woah!

dev/checkstyle/fileheader.txt
```diff
@@ -1,19 +1,23 @@
...
+ \* fields enclosed by brackets "\[\]" replaced with your own identifying
+ \* information: Portions Copyright \[yyyy\] \[name of copyright owner\]
+ \*
+ \* CDDL HEADER END
+ \*/
 
+/\*
+ \* (Portions )?Copyright \(c\) \d\d\d\d(, \d\d\d\d)?, [^0-9].*\.
+ \*/
+package .*;
```
tulinkry, line 23: what is the purpose of this "package.*"?

ahornace: Basically it only prevents any blank lines between the end of copyright comment and the `package` definition.

vladak:
> CDDL 3.3 "You may not remove or alter any copyright, patent or trademark notices contained within the Covered Software, or any notices of licensing or any descriptive text giving attribution to any Contributor or the Initial Developer."

yep, this should be done incrementally for each change. Different model is needed. Github Checks ?

vladak (changes requested):
avoid changing the actual copyrights

ahornace:
> yep, this should be done incrementally for each change. Different model is needed. Github Checks ?

Let's be reasonable here. I'm not changing anything important (replacing `-` with `,` or adding `(c)`at most). This basically means that we have to carry the iron ball chained to our legs for the whole eternity.

The thing I can do is to add every file touched by non Oracle employees to suppressions and then every other file would force these restrictions. That means new files as well. But this is far from ideal.

tulinkry:
As most of the copyrights which are not connected to Oracle are from Chris, maybe we could ask him for a consent to change the format of the copyrights and put only the rest into suppressions.

vladak:
> > yep, this should be done incrementally for each change. Different model is needed. Github Checks ?
> 
> Let's be reasonable here. I'm not changing anything important (replacing `-` with `,` or adding `(c)`at most). This basically means that we have to carry the iron ball chained to our legs for the whole eternity.
> 
> The thing I can do is to add every file touched by non Oracle employees to suppressions and then every other file would force these restrictions. That means new files as well. But this is far from ideal.

I'd rather be cautious here. Changing the format (dashes to commas and similar) is probably fine. Adding new copyright headers I am not sure.

ahornace:
> Adding new copyright headers I am not sure.

Could you be please more specific? I've added a few instances where the copyright was missing and I think I removed Trond from one file since he was employee of Sun at that time and it seemed weird -> nor @tulinkry nor @tarzanek are adding their copyright to files they changed while being employees.

However, this would still be related to at most 10s of files which I'm willing to keep as they are and add them to suppressions.

vladak:
> > Adding new copyright headers I am not sure.
> 
> Could you be please more specific? I've added a few instances where the copyright was missing and I think I removed Trond from one file since he was employee of Sun at that time and it seemed weird -> nor @tulinkry nor @tarzanek are adding their copyright to files they changed while being employees.
> 
> However, this would still be related to at most 10s of files which I'm willing to keep as they are and add them to suppressions.

Sounds good.

vladak:
Thanks !

PR #3312, opened 2020-10-23
allow to access API using bearer token

vladak:
This change introduces bearer tokens to access the API over HTTPS. The configuration contains the tokens in verbatim form:
```xml
  <void property="authenticationTokens">
   <void method="add">
    <string>bar</string>
   </void>
   <void method="add">
    <string>foo</string>
   </void>
  </void>
```
One can access the API e.g. like so:
```
curl -v -X POST -H 'Authorization: Bearer foo' -H 'Content-Type:text/plain' -d @paths.tsv https://foo:8080/source/api/v1/system/pathdesc
```


vladak:
wikis to be updated:
  - https://github.com/oracle/opengrok/wiki/Web-services
  - https://github.com/oracle/opengrok/wiki/Webapp-configuration

tulinkry:
Will it work, for example, in your environment where you use apache to provide the https and then it proxy passes the request to the tomcat? Doesn't it set the forwarded header?

vladak:
> Will it work, for example, in your environment where you use apache to provide the https and then it proxy passes the request to the tomcat? Doesn't it set the forwarded header?

As long as it is forwarded to Tomcat via HTTPS. I did not want to make things too complicated as this is usually bad for security. Maybe reflecting the forwarded headers at all is bad.

vladak:
> > Will it work, for example, in your environment where you use apache to provide the https and then it proxy passes the request to the tomcat? Doesn't it set the forwarded header?
> 
> As long as it is forwarded to Tomcat via HTTPS. I did not want to make things too complicated as this is usually bad for security. Maybe reflecting the forwarded headers at all is bad.

I mean: denying based on the existence of the forwarded headers is done to handle specifically the reverse proxy environment where the request usually does not come from localhost (wanted to avoid parsing the contents of these headers). In such case it should be authenticated using a token. The responsibility of transfer security is up to whoever runs the proxy.

opengrok-web/src/main/java/org/opengrok/web/api/v1/filter/IncomingFilter.java
```diff
@@ -86,14 +97,38 @@ public void filter(final ContainerRequestContext context) {
...
         }
+
         String path = context.getUriInfo().getPath();
+
+        if (request.isSecure()) {
+            String authHeaderValue;
+            if ((authHeaderValue = request.getHeader(HttpHeaders.AUTHORIZATION)) != null &&
+                    authHeaderValue.startsWith(BEARER)) {
+                String tokenValue = authHeaderValue.substring(BEARER.length());
+                if (RuntimeEnvironment.getInstance().getAuthenticationTokens().contains(tokenValue)) {
```
tulinkry, line 108: could benefit from ConfiguratioChangeListener (https://github.com/oracle/opengrok/pull/3270) so you could cache the tokens as it requires a lock on each secure request which contains the header

opengrok-web/src/main/java/org/opengrok/web/api/v1/filter/IncomingFilter.java
```diff
@@ -86,14 +97,38 @@ public void filter(final ContainerRequestContext context) {
         if (request == null) { // happens in tests
             return;
         }
+
         String path = context.getUriInfo().getPath();
+
+        if (request.isSecure()) {
+            String authHeaderValue;
+            if ((authHeaderValue = request.getHeader(HttpHeaders.AUTHORIZATION)) != null &&
```
ahornace, line 105: I've noticed this construct a few times in OpenGrok and I think it makes things a bit unreadable. Why not rewrite it as:
```java
String authHeaderValue = request.getHeader(HttpHeaders.AUTHORIZATION);
if (authHeaderValue != null && authHeaderValue.startsWith(BEARER)) {
   ...
```
Then the definition is directly tied to declaration.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -1336,6 +1339,14 @@ public void setDisabledRepositories(Set<String> disabledRepositories) {
         this.disabledRepositories = disabledRepositories;
     }
 
+    public Set<String> getAuthenticationTokens() {
```
ahornace, line 1342: Should not we return a copy or at least return an unmodifiable collection? Theoretically, someone could change the content easily.

vladak: Good catch ! Definitely something to worry about given these are access tokens.

tulinkry: Similar to https://github.com/oracle/opengrok/pull/3312#discussion_r511657948 but perhaps better to solve it here also

opengrok-web/src/main/java/org/opengrok/web/api/v1/filter/IncomingFilter.java
vladak: yep, that looks nicer.

opengrok-web/src/main/java/org/opengrok/web/api/v1/filter/IncomingFilter.java
vladak: good idea. Will wait for the PR to land.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -564,6 +566,7 @@ public Configuration() {
         setUserPageSuffix("");
         setWebappLAF("default");
         // webappCtags is default(boolean)
+        setAuthenticationTokens(new HashSet<>());
```
tulinkry, line 569: // This list of calls is sorted alphabetically so please keep it.

vladak: fixed

opengrok-web/src/main/java/org/opengrok/web/api/v1/filter/IncomingFilter.java
vladak: @ahornace was so kind to allow me to take the `ConfiguratioChangedListener` from his PR.

tulinkry (approved):

vladak:
Thanks @tulinkry !

PR #3318, opened 2020-10-26
set user prior to commit

vladak:
Set user prior to commit() in the `test_incoming_retval` Python tools test to avoid failure.


tulinkry (approved):

vladak:
Thanks !

PR #3320, opened 2020-10-26
enable GitHub actions for building

vladak:
Travis backlog has started being unbearable recently (see https://www.traviscistatus.com/). This change enables basic Github actions for building. Relase related tasks will be done in later PR.

The macOS build fails due to #3296. For the Windows build to pass I had to exclude some of the tests. Will file issues for these.

Info about the build environments: https://github.com/actions/virtual-environments

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,5 @@ jobs:
         restore-keys: ${{ runner.os }}-m2
     - name: Install Universal ctags
       run: choco install universal-ctags
-    - name: Install Subversion
-      run: choco install svn
     - name: Build
       run: mvn -B -V verify
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,12 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
     - name: Install pre-requisites
       run: ./dev/before_install
     - name: Before build actions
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
+    - name: Install Universal ctags
+      run: choco install universal-ctags
+    - name: Build
+      run: mvn -B -V verify
```
ahornace, line 69:
```suggestion
      run: ./mvnw -B -V verify
```
Maybe use `mvnw`?

vladak: I tried that, does not work as the default shell is Powershell. Maven is installed by default the build runs fine.

.github/workflows/javadoc.yml
```diff
@@ -0,0 +1,22 @@
...
+      with:
+        java-version: 11
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
+    - name: Refresh Javadoc
+      run: ./dev/docker.sh
```
ahornace, line 22:
```suggestion
      run: ./dev/javadoc.sh
```
Maybe this script instead of `docker.sh`?

ahornace:
Why not remove javadoc and docker stages from travis as well?

vladak:
> Why not remove javadoc and docker stages from travis as well?

Not yet - they contain release specific pieces. At this point I'd like to preserve the ability to create releases via Travis, even though it is slow.

ahornace (approved):

vladak:
Thanks !

PR #3322, opened 2020-10-27
Extract and reuse pytest parameters

tulinkry:


vladak:
Pls rebase to get build results.

vladak:
Thanks !

PR #3326, opened 2020-10-27
Use annotations instead of declarations in web.xml

ahornace:
Using annotations is much more popular nowadays than declaring things in xml files. Therefore, I thought that this might be a good step forward.

I do not know how to deal with the jsp servlets properly so I left them be for the time being.

If you do not like this approach, then feel free to close :)

Thanks.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
@@ -39,6 +39,7 @@ import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 import javax.servlet.ServletRequestEvent;
 import javax.servlet.ServletRequestListener;
+import javax.servlet.annotation.WebListener;
 import java.io.File;
 import java.io.IOException;
 import java.time.Duration;
@@ -51,6 +52,7 @@ import java.util.logging.Logger;
  *
  * @author Trond Norbye
  */
+@WebListener
 public final class WebappListener
         implements ServletContextListener, ServletRequestListener {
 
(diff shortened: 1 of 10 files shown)


vladak:
Not thrilled. To me, being able to configure the order and the parameters in one place is making the administration more transparent.

tulinkry:
I agree but this is the future old man!

vladak:
Fair enough :-) How does one configure the parameters in the new world ? Looks like https://github.com/oracle/opengrok/wiki/Webapp-configuration#changing-webapp-parameters will need to be updated.

vladak (changes requested):
I'd like to see a test case that checks the ordering of the filters.

ahornace:
> How does one configure the parameters in the new world ?

I think the same way as before. `@WebListener` does not have any attributes for that but even if it did, it would still pose a problem for our purposes since recompilation would be needed.

> I'd like to see a test case that checks the ordering of the filters.

It seems that Servlet 3.x API does not allow this. Maybe 4.x fixed this issue which is available in Tomcat 9.x but I don't know. 
https://stackoverflow.com/questions/6560969/how-to-define-servlet-filter-order-of-execution-using-annotations-in-war

vladak:
> > I'd like to see a test case that checks the ordering of the filters.
> 
> It seems that Servlet 3.x API does not allow this. Maybe 4.x fixed this issue which is available in Tomcat 9.x but I don't know.
> https://stackoverflow.com/questions/6560969/how-to-define-servlet-filter-order-of-execution-using-annotations-in-war

what about https://docs.oracle.com/javaee/7/api/javax/ws/rs/Priorities.html ?

ahornace:
> what about https://docs.oracle.com/javaee/7/api/javax/ws/rs/Priorities.html ?

That is from jax-rs project (REST) instead of `javax.servlet` so I don't think they will work together :/

Issue #3307, opened 2020-10-22
GitRepository.isWorking access to `this.working` not thread-safe and unnecessarily expensive


azakkerman:
**Describe the bug**
https://github.com/oracle/opengrok/blob/09e5ac1aead3f586713c0ade3d6e27b9c174c7e7/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java#L522-L542

The code mutates `this.working`, so if isWorking is called from multiple threads, there is no guarantee other threads see the outcome of an already made isWorking call, resulting in unnecessary work.

In addition, `GitRepository.isWorking` only performs a git version check, independent of what repository it is for, so the check doesn't have to be done for each Repository object. Instead, it can be done once for all GitRepository instance (see more below)

I have not looked at other implementations of Repository, but it is likely they have the same issue.

**Expected behavior**
Whether git command is working or not, can be done
* once for the GitRepository class via static isWorking (suggest using memoizing supplier for it)
* or periodically via expiring memoizing supplier, see [guava](https://guava.dev/releases/19.0/api/docs/com/google/common/base/Suppliers.html#memoizeWithExpiration(com.google.common.base.Supplier,%20long,%20java.util.concurrent.TimeUnit))

tulinkry:
I always thought this isWorking is per directory but it is actually per repository type - it's checking if git works...good catch

PR #3333, opened 2020-10-28
Fix for #3307: Make GitRepository#isWorking synchronized to avoid read/write proble…

azakkerman:
…ms with mutable working field. Check whether git is working once per JVM run and don't recheck every time.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -49,6 +49,7 @@ import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.BufferSink;
 import org.opengrok.indexer.util.Executor;
 import org.opengrok.indexer.util.HeadHandler;
+import org.opengrok.indexer.util.LazilyInstantiate;
 import org.opengrok.indexer.util.StringUtils;
 import org.opengrok.indexer.util.Version;
 
@@ -95,6 +96,12 @@ public class GitRepository extends Repository {
      */
     private static final Version MINIMUM_VERSION = new Version(2, 1, 2);
 
+    /**
+     * This is a static replacement for 'working' field. Effectively, check if git is working once in a JVM
+     * instead of calling it for every GitRepository instance.
+     */
+    private static final LazilyInstantiate<Boolean> GIT_IS_WORKING = LazilyInstantiate.using(GitRepository::isGitWorking);
+
     public GitRepository() {
         type = "git";
         /*
@@ -107,6 +114,21 @@ public class GitRepository extends Repository {
 
         ignoredDirs.add(".git");
         ignoredFiles.add(".git");
+    }
+
+    private static boolean isGitWorking() {
+        String repoCommand = getCommand(GitRepository.class, CMD_PROPERTY_KEY, CMD_FALLBACK);
+        Executor exec = new Executor(new String[] {repoCommand, "--version"});
+        if (exec.exec(false) == 0) {
+            final String outputVersion = exec.getOutputString();
+            final String version = outputVersion.replaceAll(".*? version (\\d+(\\.\\d+)*).*", "$1");
+            try {
+                return Version.from(version).compareTo(MINIMUM_VERSION) >= 0;
+            } catch (NumberFormatException ex) {
+                LOGGER.log(Level.WARNING, String.format("Unable to detect git version from %s", outputVersion), ex);
+            }
+        }
+        return false;
     }
 
     /**
@@ -521,21 +543,7 @@ public class GitRepository extends Repository {
     @Override
     public boolean isWorking() {
         if (working == null) {
-            ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);
-            Executor exec = new Executor(new String[]{RepoCommand, "--version"});
-
-            if (exec.exec(false) == 0) {
-                final String outputVersion = exec.getOutputString();
-                final String version = outputVersion.replaceAll(".*? version (\\d+(\\.\\d+)*).*", "$1");
-                try {
-                    working = Version.from(version).compareTo(MINIMUM_VERSION) >= 0;
-                } catch (NumberFormatException ex) {
-                    LOGGER.log(Level.WARNING, String.format("Unable to detect git version from %s", outputVersion), ex);
-                    working = false;
-                }
-            } else {
-                working = false;
-            }
+            working = GIT_IS_WORKING.get();
         }
 
         return working;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
@@ -117,7 +117,7 @@ public abstract class Repository extends RepositoryInfo {
 
     /**
      * Gets the instance's repository command, primarily for testing purposes.
-     * @return null if not {@link isWorking}, or otherwise a defined command
+     * @return null if not {@link #isWorking()}, or otherwise a defined command
      */
     public String getRepoCommand() {
         isWorking();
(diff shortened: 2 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -518,24 +528,32 @@ boolean isNestable() {
...
+            } catch (NumberFormatException ex) {
+                LOGGER.log(Level.WARNING, String.format("Unable to detect git version from %s", outputVersion), ex);
+            }
+        }
+        return false;
+    }
+
     @Override
-    public boolean isWorking() {
+    public synchronized boolean isWorking() {
```
idodeclare, line 546: How about double-locking inside the method (with proper `volatile working`) as the lowest-cost synchronization vs always synchronizing?

(I would think this affects all `RepositoryInfo` sub-classes but as just a mostly-harmless, occasional redundant-run. Or am I misunderstanding?)

azakkerman: It seems that double-locking is an overkill here. Once GIT_IS_WORKING is initialized by one instance of GitRepository, the method will be very quick and lock contention is not likely to be an issue. Also, there are no other synchronized methods just yet. In general, inheritance from RepositoryInfo does not look like a good choice here, as it is mutable and not properly synchronized. Repository should be an interface, with, maybe a base abstract implementation class, allowing for more flexibility in how to make it thread-safe (immutability, etc)

I have not looked closely at other Repository subclasses, iirc, some of them actually run commands pertaining to the actual repository in order to determine whether a given Repository instance is working. I am trying to contribute back to the community some of the optimizations I ended up doing for my employer where we use opengrok with git, hence, only fixing GitRepository. This fix is useful to improve indexing performance of installations with many (e.g. 10k) repositories, so it is harmless from correctness perspective but not from indexing speed perspective.

idodeclare: Ah OK I'm clear now that you're striving for once-per-run initialization of Git is-working. The pattern in OpenGrok would be to add a `static LazilyInstantiate<Boolean>` that synchronizes until initialized and thereafter doesn't bother to synchronize at all.

idodeclare: @azakkerman, apologies if already answered, but you write the following:

> ... trying to contribute back to the community some of the optimizations I ended up doing for my employer ....

and on [another issue](https://github.com/oracle/opengrok/issues/3306):

> it requires some work to get past our corporate process for submitting OSS patches.

Would these require your company copyright to be asserted in contributions?

azakkerman: > Ah OK I'm clear now that you're striving for once-per-run initialization of Git is-working. The pattern in OpenGrok would be to add a static LazilyInstantiate<Boolean> that synchronizes until initialized and thereafter doesn't bother to synchronize at all.

Updated the code

> Would these require your company copyright to be asserted in contributions?

I received approval. They ask to include work email in addition to personal email. Will double-check if they want to add copyright as well. The issue is mostly about exporting patches from the internal network, but I chose to simply re-implement the code.

idodeclare (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
azakkerman: LazilyInstantiate only synchronizes the static GIT_IS_WORKING, but not this.working. Java Memory Model does not guarantee any visibility for read after write across threads w/o use of synchronizing primitives to force memory barriers. My assertion is that synchronized here is needed for correctness, unless you want to make working field of RepositoryInfo volatile. If you insist, I will remove synchronized, as, in the worst case, this would result in an extra GIT_IS_WORKING.get(), so it should be idempotent.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -43,6 +43,8 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.concurrent.ConcurrentException;
+import org.apache.commons.lang3.concurrent.LazyInitializer;
```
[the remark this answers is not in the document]
azakkerman: btw, looking at LazilyInstantiate, I don't think it is truly safe, the 'current' member is not swapped in a thread-safe way. It should be declared volatile (like the active field) or one should use an AtomicReference for current to be truly thread-safe.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
idodeclare: Indeed. I mean since the reliance now is on `LazilyInstantiate` for synchronization, and `GIT_IS_WORKING.get()` can only ever return the same value, there is no need to synchronize here (nor to make `working` volatile) because redundant `get()` or assignment to `working` are harmless.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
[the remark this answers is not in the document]
idodeclare: Please see how `swapper()` is `synchronized`, and analyze what happens if redundant calls occur

vladak:
Needs Javadoc fix:
```
 /home/runner/work/opengrok/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java:120: error: no tag name after @
Error:       * @return null if not {@see isWorking}, or otherwise a defined command
Error:                             ^
```
Otherwise this is good to go I think.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
```diff
@@ -548,6 +548,16 @@ static Boolean checkCmd(String... args) {
         return exec.exec(false) == 0;
     }
 
+    protected static String getCommand(String className, String propertyKey, String fallbackCommand) {
```
tulinkry, line 551: I wouldn't mind a Class parameter instead of string. @vladak ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -95,6 +96,28 @@
...
     private static final Version MINIMUM_VERSION = new Version(2, 1, 2);
 
+    /**
+     * This is a static replacement for 'working' field. Effectively, check if git is working once in a JVM
+     * instead of calling it for every GitRepository instance.
+     */
+    private static final LazilyInstantiate<Boolean> GIT_IS_WORKING = LazilyInstantiate.using(
+            () -> isGitWorking(getCommand(GitRepository.class.getCanonicalName(), CMD_PROPERTY_KEY, CMD_FALLBACK)));
+
+    private static boolean isGitWorking(String repoCommand) {
```
tulinkry, line 106: As this is not used anywhere else it does not need to be generic (wrt to the repository command). You can get the command inside this function and do just 

```
LazilyInstantiate.using(GitRepository::isGitWorking);
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -95,6 +96,28 @@
...
+            final String outputVersion = exec.getOutputString();
+            final String version = outputVersion.replaceAll(".*? version (\\d+(\\.\\d+)*).*", "$1");
+            try {
+                return Version.from(version).compareTo(MINIMUM_VERSION) >= 0;
+            } catch (NumberFormatException ex) {
+                LOGGER.log(Level.WARNING, String.format("Unable to detect git version from %s", outputVersion), ex);
+            }
+        }
+        return false;
+    }
```
tulinkry, line 118: Would you mind moving this private static method below the constructor at least?

azakkerman:
Tests passed on my ubuntu machine, not sure what went wrong with the automated build.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
vladak: I concur.

vladak:
Checked that OCA is in place, merging. Thanks !

PR #3334, opened 2020-10-28
Speed up IndexDatabase.accept() by avoiding history lookup for files …

azakkerman:
…if unversioned files are allowed by the config.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -843,19 +843,20 @@ public class IndexDatabase {
             return true;
         }
 
-        if (HistoryGuru.getInstance().hasHistory(file)) {
-            // versioned files should always be accepted
-            return true;
-        }
 
-        // this is an unversioned file, check if it should be indexed
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -843,19 +843,21 @@ private boolean accept(File file, AcceptSymlinkRet ret) {
...
-        if (!res) {
-            LOGGER.log(Level.FINER, "not accepting unversioned {0}",
-                absolutePath);
+        // Lookup history if indexing versioned files only.
+        // Skip the lookup entirely (which is expensive) if unversioned files are allowed
+        if (env.isIndexVersionedFilesOnly()) {
+            if (HistoryGuru.getInstance().hasHistory(file)) {
+                // versioned files should always be accepted
+                return true;
+            } else {
```
idodeclare, line 854: superfluous `else`?

idodeclare (approved):
nit but LGTM

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
azakkerman: updated.

ahornace:
Thanks!

PR #3336, opened 2020-10-29
Build mvn modules in parallel fashion

ahornace:
I was inspired by `--parallel` option in Gradle. This should create a DAG from modules and build indexer, tools, and suggester modules in a parallel fashion which could save 1-2 minutes from the build times. Paid GitHub actions use 2 core VMs but I was not able to find the specs for free ones -> hopefully it's the same.

`1C` option means that one thread will be created for each core available.

The drawback is that it might be more difficult to find the failure in the logs.

What do you think?

Thanks.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,4 +74,4 @@ jobs:
     - name: Install Universal ctags
       run: choco install universal-ctags
     - name: Build
-      run: mvn -B -V verify
+      run: mvn -T 1C -B -V verify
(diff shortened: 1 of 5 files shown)


vladak:
I am all for parallelism. However the builds currently fail.

ahornace:
It seems that running `compile`, `test` or `package` works; however, it fails with `verify` or `site`. I can't see anything significant from the logs :/ 

Maybe some plugin is doing some harm.

PR #3339, opened 2020-10-30
Update junit to 5.7.0

ahornace:
While trying to make https://github.com/oracle/opengrok/pull/3336 work I've tried to update the junit and had to fix a few files.

`assertThat` and `ExpectedException.none()` became deprecated in junit 4.13.

Thanks.


vladak:
Nice, I wanted to do this earlier so that we can use `@ParameterizedTest` with `@ValueSource` (https://www.baeldung.com/parameterized-tests-junit-5).

Issue #3347, opened 2020-11-02
suggester rebuild should use single ExecutorService


vladak:
Another post-mortem induced issue: when chasing really long latencies (minutes) of requests to web application in our deployment I found out this is often happening when the indexer is running, namely when the suggester is being rebuilt. This deployment uses per project workflow so in fact several indexers (in our case 16) are running in parallel. This easily hits the default maximum thread limit of Tomcat (200), leading to connections being accepted however not handled for a long time.

When observing the situation I noticed that there were bunch of indexer processes waiting on the RESTful API call to `/{project}/indexed`, concretely in `suggester.rebuild()`, waiting for the lock to be acquired. Firstly, the handling of the `/indexed` request should be asynchronous, secondly the rebuild should use single thread pool and the requests to rebuild suggester(s) should be submitted there. This will limit the number of threads consumed by the indexers and also speed up the rebuilds in general. Like this is done now the rebuilds of individual projects are processed in linerar fashion.

I already updated the wikis with tuning advice.

vladak:
For the record here are the threads:
```
"AsyncFileHandlerWriter-1837760739" #25 daemon prio=5 os_prio=64 cpu=41221.39ms elapsed=612562.48s tid=0x0000000000d8f800 nid=0x14 waiting on condition  [0x00007fff752fe000]
"Attach Listener" #34166 daemon prio=9 os_prio=64 cpu=5.20ms elapsed=316845.65s tid=0x000000000561b000 nid=0x89ab waiting on condition  [0x0000000000000000]
"C1 CompilerThread0" #15 daemon prio=9 os_prio=64 cpu=60306.86ms elapsed=612562.73s tid=0x0000000000a94800 nid=0xd waiting on condition  [0x0000000000000000]
"C2 CompilerThread0" #5 daemon prio=9 os_prio=64 cpu=912135.03ms elapsed=612562.73s tid=0x0000000000a90000 nid=0xc waiting on condition  [0x0000000000000000]
"Common-Cleaner" #21 daemon prio=8 os_prio=64 cpu=792.92ms elapsed=612562.65s tid=0x0000000000c8a000 nid=0x11 in Object.wait()  [0x00007fff759f8000]
"ContainerBackgroundProcessor[StandardEngine[Catalina]]" #5404 daemon prio=5 os_prio=64 cpu=129110.54ms elapsed=612519.20s tid=0x00000000036b1800 nid=0x155c waiting on condition  [0x00007fff6d5fe000]
[...]

PR #3349, opened 2020-11-03
make ProjectsController API calls using suggester async

vladak:
This change removes unnecessary Counter and making the API calls to spawn a thread when calling into suggester.

The change merely changes the behavior of the API calls. In terms of #3347 it actually makes the situation worse since there will be more threads created.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
@@ -186,9 +186,9 @@ public class ProjectsController {
 
     @DELETE
     @Path("/{project}/data")
-    public void deleteProjectData(@PathParam("project") String projectName) throws HistoryException {
+    public void deleteProjectData(@PathParam("project") String projectNameParam) {
         // Avoid classification as a taint bug.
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -206,7 +206,7 @@ public void deleteProjectData(@PathParam("project") String projectName) throws H
         deleteHistoryCache(projectName);
 
         // Delete suggester data.
-        suggester.delete(projectName);
+        new Thread(() -> suggester.delete(projectName)).start();
```
tulinkry, line 209: could you CompletableFuture.runAsync to possibly reuse resources

tulinkry (approved):

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
vladak: done

ktulinger (approved):

tulinkry (approved):

vladak:
Thanks @tulinkry ! (2 reviews !)

PR #3350, opened 2020-11-03
support OPTIONAL control flag in authorization framework

vladak:
This introduces the `OPTIONAL` authorization control flag.

Tested with stack that contained this at the top level:
```
LdapUserPlugin (OPTIONAL)
  - substack (REQUIRED)
    UserWhiteListPlugin (SUFFICIENT)
    LdapFilterPlugin (REQUIRED)
```
Works fine except when `LdapUserPlugin` fails the check because the user does not exist in LDAP it will cause a LDAP query each time the user sends a request as there is no negative caching.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
@@ -55,6 +55,7 @@ public enum AuthControlFlag {
      */
     SUFFICIENT("sufficient"),
     /**
+     * A success of a optional plugin is ignored and processing of the stack continues unaffected.
      * The failure is returned only if stack traversal finished and
      * there was no prior result recorded.
      */
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
@@ -48,13 +48,17 @@ public enum AuthControlFlag {
      */
     REQUISITE("requisite"),
     /**
-     * If such a plugin succeeds and no prior required plugin has failed the
-     * authorization framework returns success to the application immediately
-     * without calling any further plugins in the stack. A failure of a
-     * sufficient plugin is ignored and processing of the plugin list continues
-     * unaffected.
+     * If such a plugin succeeds and no prior required plugin has failed,
+     * the authorization framework returns success immediately
+     * without calling any further plugins in the stack.
+     * A failure of a sufficient plugin is ignored and processing of the stack continues unaffected.
      */
-    SUFFICIENT("sufficient");
+    SUFFICIENT("sufficient"),
+    /**
+     * The failure is returned only if stack traversal finished and
+     * there was no prior result recorded.
+     */
+    OPTIONAL("optional");
 
     private final String flag;
 
@@ -77,6 +81,10 @@ public enum AuthControlFlag {
 
     public boolean isSufficient() {
         return SUFFICIENT.equals(this);
+    }
+
+    public boolean isOptional() {
+        return OPTIONAL.equals(this);
     }
 
     /**
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.authorization;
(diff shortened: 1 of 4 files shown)


vladak:
https://github.com/oracle/opengrok/wiki/Authorization#a-stack-of-plugins will need to be updated.

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationStack.java
```diff
@@ -277,6 +292,14 @@ protected boolean processStack(Nameable entity,
...
         }
 
+        if (overallDecision == null && optionalFailure) {
+            return false;
+        }
+
+        if (overallDecision == null) {
+            return true;
+        }
+
```
tulinkry, line 302: So the idea is that it fails the decision only when the OPTIONAL are the only plugins in the stack?

If so, the original structure may stay it was (except the OPTIONAL else-if) and here you could do

```
if(optionalFailure && getStack().stream().filter(Predicate::not(AuthorizationEntity::isOptional)).findAny().isEmpty()) {
  return optionalFailure;
}
```

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
```diff
@@ -48,13 +48,17 @@
...
+     * If such a plugin succeeds and no prior required plugin has failed,
+     * the authorization framework returns success immediately
+     * without calling any further plugins in the stack.
+     * A failure of a sufficient plugin is ignored and processing of the stack continues unaffected.
      */
-    SUFFICIENT("sufficient");
+    SUFFICIENT("sufficient"),
+    /**
+     * The failure is returned only if stack traversal finished and
+     * there was no prior result recorded.
```
tulinkry, line 59: maybe worth noting that: A success of a sufficient plugin is ignored and processing of the stack continues unaffected.

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
```diff
@@ -48,13 +48,17 @@
      */
     REQUISITE("requisite"),
     /**
```
tulinkry, line 50: copyright

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationStack.java
```diff
@@ -219,7 +219,13 @@ protected boolean processStack(Nameable entity,
             PluginDecisionPredicate pluginPredicate,
             PluginSkippingPredicate skippingPredicate) {
 
-        boolean overallDecision = true;
+        Boolean overallDecision = null;
+        boolean optionalFailure = false;
+
+        if (getStack().isEmpty()) {
+            return true;
+        }
```
tulinkry, line 227: "if" not necessary here, but i don't mind it

opengrok-indexer/src/test/java/org/opengrok/indexer/authorization/AuthorizationFrameworkTest.java
```diff
@@ -654,6 +654,57 @@ public AuthorizationFrameworkTest(StackSetup setup) {
...
+                new StackSetup(
+                    NewStack(AuthControlFlag.REQUIRED,
+                            new AuthorizationPlugin(AuthControlFlag.OPTIONAL, createAllowedPrefixPlugin())
+                    ),
+                    // optional plugin returns false => false
+                    NewTest(false, createUnallowedProject()),
+                    NewTest(false, createUnallowedGroup()),
+                    // optional plugin returns true => true
+                    NewTest(true, createAllowedProject()),
+                    NewTest(true, createAllowedGroup()))
```
tulinkry, line 706: a test with two optionals, both returning different result, would be nice

tulinkry: a test including sufficient and optional would be nice too

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationStack.java
vladak: I think it makes the code a little more easier to follow.

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
vladak: will fix

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthControlFlag.java
vladak: added

opengrok-indexer/src/test/java/org/opengrok/indexer/authorization/AuthorizationFrameworkTest.java
vladak: added

opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationStack.java
vladak: Used the original pattern according to the Unix PAM implementation (it is actually more complicated there). The stream interface makes the code arguably a bit more readable at the expense of traversing the plugins.

PR #3352, opened 2020-11-03
log more info during init() in plugins using whitelist

vladak:
This change adds a bit more logging to the init()/load() functions of the plugins using whitelists.

Sample output:
```
03-Nov-2020 16:16:06.058 FINE [http-nio-8080-exec-29] opengrok.auth.plugin.UserWhiteListPlugin.load UserWhiteList plugin loaded with filePath=/opengrok/auth/config/whitelists/foo.txt (404 entries), fieldName=id
03-Nov-2020 16:16:07.770 FINE [http-nio-8080-exec-29] opengrok.auth.plugin.LdapAttrPlugin.init LdapAttrPlugin plugin loaded with attr=gidNumber, whitelist=/opengrok/auth/config/whitelists/bar.txt (2 entries), instance=33
```


tulinkry (approved):

vladak:
Thanks @tulinkry !

PR #3353, opened 2020-11-04
use common Maven build for all OSes

### LINKED ISSUE #3330: make Coveralls+Sonarcloud work on Windows again
During the conversion from Travis to Github actions we lost Coveralls for Windows. This is because by default the command runs in Poweshell so `dev/main` (which adjusts the command line options to include extra arguments) cannot be used as it is a bash script.

vladak:
Changes all Maven builds to start from `bash`.  This should resolve  #3330.


tulinkry (approved):

vladak:
Thanks @tulinkry !

PR #3355, opened 2020-11-04
lucene 8.7.0

tarzanek:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
     </scm>
 
     <properties>
-        <lucene.version>8.6.2</lucene.version>
+        <lucene.version>8.7.0</lucene.version>
         <mavenjavadocplugin.version>3.0.1</mavenjavadocplugin.version>
         <!-- The following changed syntax from e.g. 1.8 to 11. -->
         <compileSource>11</compileSource>


tarzanek:
don't merge, I just submitted it to see if tests won't break on API change https://issues.apache.org/jira/browse/LUCENE-9449

vladak:
Seems like they did :-)

vladak:
Is reindex from scratch needed ? Are there any notable changes/bug fixes ?

tarzanek:
https://lucene.apache.org/core/8_7_0/changes/Changes.html#v8.7.0
there are optimizations
LUCENE-9386: RegExpQuery added case insensitive matching option. 
is interesting
also there is bugfix:
LUCENE-9427: Fix a regression where the unified highlighter didn't produce highlights on fuzzy queries that correspond to exact matches. 

I still can't believe 
LUCENE-9449: Field comparators for numeric fields and _doc were moved to their own package. TopFieldCollector sets TotalHits.relation to GREATER_THAN_OR_EQUAL_TO, as soon as the requested total hits threshold is reached, even though in some cases no skipping optimization is applied and all hits are collected. 
didn't cause any test to fail ... well ... we need to keep it in mind and see if users won't find bugs with it (e.g. on 10th page of results)

tarzanek:
reindex is needed, version was bumped (and is transparent now, hence no code change)
they optimized storage a bit

tarzanek:
so since tests pass I don't have any objection to merge this ...

vladak:
> so since tests pass I don't have any objection to merge this ...

Okay, I have couple of simple PRs that I'd like to merge in 1.5.x and then will merge this which will lead to 1.6.x.

idodeclare:
> > so since tests pass I don't have any objection to merge this ...
> 
> Okay, I have couple of PRs that I'd like to merge in 1.5.x and then will merge this which will lead to 1.6.x.

Almost all Lucene minor releases (and generally even a release from Lucene `X.Y` to `++X.0`) would not require a bump in OpenGrok minor version — i.e. `y` in `x.y.z` — since a bump in `y` per OpenGrok README advises that a "full clean reindex of your repositories is needed (e. g. index format has changed)."

I believe it would have to be a relevant Lucene bug fix to require an OpenGrok full reindex for a minor release.

tarzanek:
I didn't check deep, but e.g. previous release of lucene didn't change the version of index storage,
this one however did, hence I suggested reindex
fwiw 8.8 is planned in mid/late January

tarzanek:
closing in favour of https://github.com/oracle/opengrok/pull/3415

PR #3358, opened 2020-11-04
raise exception from do_api_call()

vladak:


opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
```diff
@@ -38,14 +38,24 @@ def do_api_call(verb, uri, params=None, headers=None, data=None):
...
+    r = handler(
         uri,
         data=data,
         params=params,
         headers=headers,
         proxies=get_proxies(uri)
     )
 
+    if r is not None:
+        logger.error("API call failed: {}".format(r))
```
tulinkry, line 54: you don't know the call failed at this point

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
```diff
@@ -38,14 +38,24 @@ def do_api_call(verb, uri, params=None, headers=None, data=None):
...
+        r.raise_for_status()
```
tulinkry, line 55: I'd interpret this behavior to the doc string (exception on failed requests)

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
tulinkry: is the logging necessary here? if you throw the exception you kinda expect someone else to handle it (and log it perhaps)

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
```diff
@@ -38,14 +38,24 @@ def do_api_call(verb, uri, params=None, headers=None, data=None):
     if handler is None or not callable(handler):
         raise Exception('Unknown HTTP verb: {}'.format(verb))
 
-    return handler(
+    logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
vladak: will fix

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
vladak: true, removed

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
```diff
@@ -34,18 +34,27 @@
...
+                 format(verb, uri, data, headers))
+    r = handler(
         uri,
         data=data,
         params=params,
         headers=headers,
         proxies=get_proxies(uri)
     )
 
+    if r is not None:
```
tulinkry, line 53: thinking about it, i don't remember if this can occur with requests, but even if it could occur (`r is None`) then you should probably raise an exception here (like when handler is none)

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
vladak: added a doc string

opengrok-tools/src/main/python/opengrok_tools/utils/restful.py
vladak: makes sense

tulinkry (approved):

vladak:
Thanks @tulinkry ! Those were all helpful comments.

Issue #3308, opened 2020-10-22
Allow indexer's search paths (-s flag) to be specified via a file or specially named files in the search dir itself


azakkerman:
**Is your feature request related to a problem? Please describe.**
Specifying many search paths (e.g. 10k repositories to index) can blow through command line limits. It would be better if one could specify all search paths via a file. Another possibility is to allow special files in the search tree (e.g. `.opengrok.path`) to specify which subdirectories are repositories to be indexed.

Yet another suggestion is to enhance discovery of repositories inside a git repository to automatically discover [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) as valid Repositories.

vladak:
Specifying the input via a file (new option or perhaps follow the `curl` style and use `-s @inputfile`) is fine. We try to avoid implicit config files.

Please file separate issues for the Git submodule detection.

PR #3359, opened 2020-11-04
Fix for #3308: Allow @path syntax for both search paths and repositor…

azakkerman:
…y paths to be able to load values from a given file.


vladak:
Thanks !

PR #3364, opened 2020-11-06
run the build once a week

vladak:
also adds some ignore paths to the `push` event.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,13 @@
 name: Build
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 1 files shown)


ahornace:
Could you please elaborate why run it in a scheduled manner every week? I think it's safe to assume that only code changes can break the code and this is covered by running builds on PRs, right?

vladak:
Some of the build components do not fix the versions, primarily the Python tools. Running the builds periodically will cover quiet periods (with no PRs) and uncover any problems due to changing modules that the tools depend on.

PR #3365, opened 2020-11-06
another suggester.rebuild() to become async

vladak:
Forgot make one of the `suggester.rebuild()` calls async when working on #3349. Noticed this in Grafana as it skews the overall request max latency graphs.


vladak:
For the record, the call to the `/system/refresh` API endpoint happens at the end of per project reindex in `signalTorefreshSearcherManagers()` called from `refreshSearcherManagers()` near the very end of Indexer's `main()`.

PR #3366, opened 2020-11-06
Catch all IOExceptions thrown by toRealPath and fallback to normalize…

azakkerman:
…().toAbsolutePath()

This is necessary, because getRelativeToCanonical is called for paths that are not real paths
(e.g. in the case of reading git history and resolving historical paths relative to current)


vladak:
Thanks !

PR #3372, opened 2020-11-19
generate xref for non empty compressed files

vladak:
This change makes it possible to produce xref's for compressed files. The most complex part is trying to make sure that empty xref files are not lying around.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.index;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -33,7 +33,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
-import java.net.ConnectException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
+import java.net.ConnectException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -486,17 +487,17 @@ public void update() throws IOException {
                     reader.close();
                 }
             }
-
-            try {
```
idodeclare, line 490: Please don't relocate this. That breaks known failure handling where we would actually not want to finish writing. I.e. `finishWriting()` in advance of a `RuntimeException` catch is deliberate.

Recommend you swallow web exceptions as per your alternative idea. Web exceptions didn't used to abort an index run; you used to be able to index without Opengrok web running. Not sure what changed or when to break that.

vladak: Oh, okay. Also added a comment there. The API failure is really a soft one.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -725,8 +739,28 @@ private void addFile(File file, String path, Ctags ctags)
...
             analyzerGuru.populateDocument(doc, file, path, fa, xrefOut);
+
+            // Avoid producing empty xref files.
+            if (xrefOut != null && xrefOut.getCount() > 0) {
+                PendingFileRenaming ren = new PendingFileRenaming(xrefAbs,
+                        transientXref.getAbsolutePath());
+                completer.add(ren);
+            } else if (xrefOut != null) {
+                LOGGER.log(Level.FINER, "xref for {0} would be empty, removing", path);
+                transientXref.delete();
```
idodeclare, line 762: I worry on NTFS that a deletion of an open file wouldn't succeed as it would do on *nix. Perhaps instead schedule a `PendingFileDeletion` of the `transientXref`?

vladak: Good point. I had the same idea however thought it is unnecessary.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -353,27 +354,38 @@ private void markProjectIndexed(Project project) {
...
+                    .target(env.getConfigURI())
+                    .path("api")
+                    .path("v1")
+                    .path("projects")
+                    .path(Util.URIEncode(project.getName()))
+                    .path("indexed")
+                    .request()
+                    .put(Entity.text(""));
+        } catch (RuntimeException e) {
+            LOGGER.log(Level.WARNING, "Couldn''t notify the webapp that project {0} was indexed: {1}",
```
idodeclare, line 381: Not sure the ramification, but usually we would see used the `log(Level, String, Throwable)` method when an entire `Throwable` is reported (versus e.g. just the `getMessage()`)

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -1629,46 +1666,63 @@ public int hashCode() {
...
-        AbstractAnalyzer.Genre g = fa.getFactory().getGenre();
-        return (g == AbstractAnalyzer.Genre.PLAIN || g == AbstractAnalyzer.Genre.XREFABLE);
+    private class CountingWriter extends Writer {
+        private long count;
+        private Writer out;
+
+        CountingWriter(Writer out) {
+            super(out);
+            this.out = out;
+            this.count = 0;
```
idodeclare, line 1676: redundant?

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
vladak: True, changed.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
vladak: Looks like it. I also did bunch of other changes suggester by IDEA.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java
```diff
@@ -353,27 +354,38 @@ private void markProjectIndexed(Project project) {
...
+                    .target(env.getConfigURI())
+                    .path("api")
+                    .path("v1")
+                    .path("projects")
+                    .path(Util.URIEncode(project.getName()))
+                    .path("indexed")
+                    .request()
+                    .put(Entity.text(""));
+        } catch (RuntimeException e) {
+            LOGGER.log(Level.WARNING, String.format("Couldn''t notify the webapp that project {0} was indexed",
```
idodeclare, line 381: Does `{0}` work with `String.format`?

vladak: Good catch, this will likely not work.

vladak, line 381:
```suggestion
            LOGGER.log(Level.WARNING, String.format("Couldn''t notify the webapp that project %s was indexed",
```

idodeclare (approved):
LGTM

vladak:
Thanks @idodeclare , I appreciate the thorough review of my sloppy changes.

PR #3375, opened 2020-11-23
change /system/pathdesc API endpoint to accept JSON

vladak:


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SystemController.java
```diff
@@ -60,9 +68,10 @@ public void reloadIncludes() {
 
     @POST
     @Path("/pathdesc")
-    @Consumes(MediaType.TEXT_PLAIN)
-    public void loadPathDescriptions(final String input) throws IOException {
+    @Consumes(MediaType.APPLICATION_JSON)
```
tulinkry, line 71: I'm all for this change but wasn't the original reason that the file is somewhat standard format wrt to the eftar reader? Then I'd be for keeping both formats perhaps.

vladak: I think the whitespace/tab separated format is error prone so it's no loss. Would like to keep things simple.

PR #3392, opened 2020-12-04
control thread pool size in invalidateRepositories()

vladak:
This change allows to set thread pool size in `HistoryGuru#invalidateRepositories()` via `-T` indexer option.


opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -757,7 +757,8 @@ public static void main(String[] argv) {
                     cfg.setWebappLAF((String) stylePath));
 
             parser.on("-T", "--threads", "=number", Integer.class,
-                    "The number of threads to use for index generation and repository scan.",
+                    "The number of threads to use for index generation, repository scan",
```
idodeclare, line 760: Nit: not reformatting the paragraph fully

idodeclare (approved):
Looks fine to me

vladak:
Thanks @idodeclare !

PR #3393, opened 2020-12-04
do not hardcode exact versions in requirements.txt

vladak:
I don't think capturing the exact developer environment via `requirements.txt` for the Python tools is necessary. This change will make `pip install -r requirements.txt` to install whatever is in the package dependencies in `setup.py`.

### DIFF
--- a/opengrok-tools/README.md
+++ b/opengrok-tools/README.md
@@ -39,12 +39,6 @@ python3 -m venv env
 . env/bin/activate
 ```
 
-Install development dependencies
-
-```bash
-python -m pip install -r requirements.txt
-```
-
 ## Developing
 
 When you start developing, install the package in a development mode.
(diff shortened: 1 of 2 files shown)


vladak:
Any comments @tulinkry ?

tulinkry:
The only purpose of this file is to replicate the developer environment so there is a source of thruth when someone else is developing/debugging the code.

If you don't think capturing the env is a good idea, then this files looses its purpose and can be deleted entirely.

vladak:
I think that's not really needed for the opengrok tools development.

tulinkry:
Another thing which is touching this is that the versions in the setup.py should have lower and upper bounds to prevent update surprises.

tulinkry (approved):

vladak:
Thanks @tulinkry !

PR #3394, opened 2020-12-04
close inactive servers in prepareServers()

vladak:
This change makes sure that inactive LDAP servers are closed in `LdapFacade#prepareServers()` after the working one is determined. The main goal is resource efficiency.

I had contemplated this change for a while. While it is nice to have connections/contexts established for these servers for quicker failover, it is not guaranteed that they will be in a working state by the time search request is made through them. Also, usually the 3-way TCP hadshake and LDAP bind operation do not take too long compared to LDAP query latency so the benefit of keeping them around is arguable.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/ldap/LdapFacade.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/ldap/LdapFacade.java
@@ -197,12 +197,24 @@ public class LdapFacade extends AbstractLdapProvider {
      * Go through all servers in the pool and record the first working.
      */
     void prepareServers() {
+        LOGGER.log(Level.FINER, "checking servers for {0}", this);
         for (int i = 0; i < servers.size(); i++) {
             LdapServer server = servers.get(i);
             if (server.isWorking() && actualServer == -1) {
                 actualServer = i;
             }
         }
+
+        // Close the connections to the inactive servers.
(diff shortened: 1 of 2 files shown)


plugins/src/main/java/opengrok/auth/plugin/ldap/LdapFacade.java
```diff
@@ -197,12 +197,24 @@ private void setWebHooks(WebHooks webHooks) {
...
+
+        // Close the connections to the inactive servers.
+        LOGGER.log(Level.FINER, "closing unused servers");
+        for (int i = 0; i < servers.size(); i++) {
+            if (i != actualServer) {
+                servers.get(i).close();
+            }
+        }
+
+        LOGGER.log(Level.FINER, String.format("server check done (current server: %s)",
```
idodeclare, line 216: Nit: not wrapping with `if (LOGGER.isLoggable(Level.FINER)) { }` since the argument is a composed string

idodeclare (approved):
Looks fine to me

plugins/src/main/java/opengrok/auth/plugin/ldap/LdapFacade.java
vladak: ack

vladak:
Thanks @idodeclare !

PR #3396, opened 2020-12-05
Suggester rebuild terminate on undeploy

vladak:
This should take care of the long wait on undeploy when there is a number of outstanding suggester rebuild jobs.

It does so by setting a flag so if there is a long running rebuild in progress already it will still wait for that. That would be the case also when shutting down an executor (interrupting the threads by shutting the executor down immediately would not be a good practice).

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/Suggester.java
+++ b/suggester/src/main/java/org/opengrok/suggest/Suggester.java
@@ -71,7 +71,7 @@ public final class Suggester implements Closeable {
 
     private static final String PROJECTS_DISABLED_KEY = "";
 
-    private static final Logger logger = Logger.getLogger(Suggester.class.getName());
+    private static final Logger LOGGER = Logger.getLogger(Suggester.class.getName());
 
     private final Map<String, SuggesterProjectData> projectData = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -278,8 +279,12 @@ public void rebuild(final Collection<NamedIndexDir> indexDirs) {
         rebuildLock.unlock();
 
         synchronized (lock) {
+            if (terminating) {
```
idodeclare, line 282: What would you think about also short-circuiting the init- and rebuild-runnables to check `terminating`?

vladak: Inside the runnables the benefit is small I think however checking it in `init()` makes sense.

idodeclare: The runnables are the items potentially in queue. It makes sense for them to be `terminating`-aware so that when they're awakened they do no unnecessary work. If you have a for-loop that is just submitting to an asynchronous executor (as this code does), you have a minuscule amount of time where being `terminating`-aware has a chance of being effective.

Moreover, `shutdown()` for a work-stealing pool only prevent new tasks from being submitted; it is an "orderly shutdown " that allows all submitted tasks to be executed. So again, the runnables should be `terminating`-aware for there to be any realistic chance of avoiding work.

I recommend reverting to your commit 62b8e45 and just addressing the runnables.

vladak: You're right, addressing this in the runnables is better and actually more readable.

idodeclare (approved):
LGTM

vladak:
Thanks @idodeclare !

PR #3399, opened 2020-12-09
add Indexer option to set API token

vladak:
This change adds `--token` Indexer option to allow setting API tokens in the requests made to the web application.

Out of the occurrences where requests are made to the web app `plugins/src/main/java/opengrok/auth/plugin/util/RestfulClient.java#postIt()` was left untouched because it is effectively running in the web app context.

Tested with the `isSecure()` check commented out in the `IncomingFilter` and making sure the token matched.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -35,7 +35,6 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -756,6 +756,23 @@ public final class Indexer {
                     "stylesheet. The factory-setting is: \"default\".").execute(stylePath ->
                     cfg.setWebappLAF((String) stylePath));
 
+            parser.on("--token", "=string|@file_with_string",
+                    "Authorization bearer API token to use when making API calls",
+                    "to the web application").
+                    execute(optarg -> {
+                        String value = ((String) optarg).trim();
+                        if (value.startsWith("@")) {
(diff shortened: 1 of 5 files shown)


vladak:
> Do you plan on adding IndexerAuthenticationToken to the set of "normal" tokens as well? Or is your plan that one should configure both?

Good question. I'd like to keep these separate for simplicity.

The bootstrap step would look like this (assuming the source/data root is shared by other means):
```
opengrok-indexer -R config-with-tokens.xml -W config.xml ... # no -U
scp config.xml webapp-host:/the/location/of/config
ssh webapp-host "opengrok-deploy -c /the/location/of/config ..." # or use the /configuration API endpoint locally
```
and then run the indexer normally:
```
opengrok-indexer --token @file_with_token -U https://webapp-host/source ...
```

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -756,6 +756,23 @@ public static void main(String[] argv) {
...
                     cfg.setWebappLAF((String) stylePath));
 
+            parser.on("--token", "=string|@file_with_string",
+                    "Authorization bearer API token to use when making API calls",
+                    "to the web application").
+                    execute(optarg -> {
+                        String value = ((String) optarg).trim();
+                        if (value.startsWith("@")) {
+                            try {
+                                String token = new String(Files.readAllBytes(Path.of(value))).trim();
```
idodeclare, line 766: Is this correct handling w.r.t. the `@` sigil?

idodeclare (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -756,6 +756,23 @@ public static void main(String[] argv) {
                     "stylesheet. The factory-setting is: \"default\".").execute(stylePath ->
                     cfg.setWebappLAF((String) stylePath));
 
+            parser.on("--token", "=string|@file_with_string",
```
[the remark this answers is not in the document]
vladak: fixed

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
vladak: Maybe only single line from the file should be read ?

vladak:
thanks @idodeclare !

PR #3401, opened 2020-12-11
move opengrok-tools -> tools

vladak:
This change renames the `opengrok-tools` directory to just `tools`.


vladak:
@tulinkry can you take a look ?

Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY suggester/pom.xml /mvn/suggester/
 
 # distribution and opengrok-tools do not have dependencies to cache
```
tulinkry, line 17: also only tools

distribution/assembly.xml
```diff
@@ -20,17 +20,17 @@
         <destName>source.war</destName>
       </file>
       <file>
-        <source>${project.basedir}/../opengrok-tools/${project.build.directory}/dist/opengrok-tools-${project.python.package.version}.tar.gz</source>
+        <source>${project.basedir}/../tools/${project.build.directory}/dist/opengrok-tools-${project.python.package.version}.tar.gz</source>
```
tulinkry, line 23: so the target archive is still opengrok-tools? I'm all for, just asking.

tulinkry (approved):

distribution/assembly.xml
vladak: Yes, I think it should be recognizable among all the other Python packages. 'tools' would be too generic.

Dockerfile
vladak, line 17:
```suggestion
# distribution and tools do not have dependencies to cache
```

vladak:
Tanks @tulinkry !

Issue #3221, opened 2020-09-11
use opengrok-sync to mirror+reindex in Docker image


vladak:
The Docker image can use `opengrok-sync` to perform repository sync and indexing instead of running `opengrok-mirror` and `opengrok-indexer` separately. The `opengrok-mirror` that will be run via `opengrok-sync` will use `-I` to avoid unnecessary indexer run.

vladak:
Assuming this will just replace `index.sh` with execution of `opengrok-sync` that will in turn run `opengrok-mirror` and indexer.

Some things for considerations:
  - do we want to run `opengrok-mirror` with `-I` / `--check-changes` by default ?
  - make it possible to avoid running the mirroring ?

I'd say the answers should be yes and no. respectively.

vladak:
The extra indexer flags passed via `INDEXER_OPT` environment variable need to be honored. Same for `NOMIRROR`.

vladak:
Also, with the conversion to `opengrok-sync` the `-P` indexer option can no longer be overridden as the program assumes projects.

PR #3402, opened 2020-12-14
Docker script refactoring and rewrite to Python

vladak:
This change rafactors the scripts used for Docker. The initial motivation was to avoid the non-existent-configuration exception thrown by the web app during the initial startup of a container based on the opengrok image.

Some notable changes:
  - no special casing between initial and non-initial indexing
  - the sequence goes like this:
    1. deploy
    1. bare config is created (to avoid above mentioned exception) 
    1. indexer is started in the background (waits for the Tomcat to come up before it enters the infinite loop)
    1. Tomcat is started on foreground
  - to preserve the locking I added filelock to the `indexer` program

~This is a step towards #3221.~

~Eventually the `start.sh` script can be rewritten to Python and use native interfaces to `opengrok-sync`, `opengrok-deploy`.~

I used the opportunity to overhaul everything - the main Docker program is now based on the `opengrok-sync` and rewritten into Python.


ahornace (approved):

vladak:
I used the opportunity and changed start.sh to use `opengrok-sync` and then rewritten it to Python. The way it works now is that the `sync` runs the mirroring/indexing/... in parallel for each projects and once this is all done, it will print the logs for each project to the console.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -118,9 +118,11 @@ public Response addProject(String projectName) {
                     }
                 }
                 // deleted repository
-                for (RepositoryInfo repo : map.get(project)) {
-                    if (!repos.contains(repo)) {
-                        allrepos.remove(repo);
+                if (map.get(project) != null) {
```
ahornace, line 121:
```suggestion
                if (map.containsKey(project)) {
```
I think this is nicer :)

vladak:
Incidentally, #3421 describes a use case when project's repositories set is changed. This change does not support such use case even though it can be easily added.

tools/src/main/python/opengrok_tools/reindex_project.py
```diff
@@ -75,50 +75,68 @@ def get_config_file(logger, uri):
...
                       java=args.java, java_opts=java_opts,
                       env_vars=args.environment, doprint=args.doprint)
     indexer.execute()
     ret = indexer.getretcode()
     os.remove(conf_file)
-    os.remove(logprop_file)
+    if logprop_file:
+        os.remove(logprop_file)
+
+    # TODO
```
ahornace, line 137: What about this `#TODO`?

ahornace (approved):

tools/src/main/python/opengrok_tools/reindex_project.py
vladak: That still needs addressing. Like it is now it would print the output twice on failure.

vladak:
thanks @ahornace !

Issue #3418, opened 2021-02-11
Index database(s) not found in OpenGrok for a particular repository.


avnish-sharma:
Facing an issue 'Index database(s) not found.' in OpenGrok for a particular repository. Issue persists post index recreation.

vladak:
How do you run the indexer exactly ? Could you post the configuration here ? (ideally extracted from the web app using http://localhost:8080/source/api/v1/configuration - assuming the name of the application is `source`) For which project does this happen ?

vladak:
I think this error usually happens if the source root is mis-configured (which is probably not your case as this happens only for some projects) or might happen if the project is not marked as indexed.

vladak:
Maybe the web app should emit more detailed error message. After all, there is a commented out line in the code: https://github.com/oracle/opengrok/blob/876391fae90ef7239a29b98c01205b3ab9a89fcc/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java#L346-L348

avnish-sharma:
> How do you run the indexer exactly ? Could you post the configuration here ? (ideally extracted from the web app using http://localhost:8080/source/api/v1/configuration - assuming the name of the application is `source`) For which project does this happen ?

we are executing the indexer like : 
/scratch/ofssobp/jdk-11.0.3/bin/java -Xms2048m -Xmx4096m -XX:-UseCompressedOops -Dorg.opensolaris.opengrok.history.cvs=cvs -Dorg.opensolaris.opengrok.history.Subversion=svn -Djava.util.logging.config.file=/scratch/opengrok_setup/opengrok/data/logging.properties -jar /scratch/opengrok_setup/opengrok/lib/opengrok.jar -c ctags -W /scratch/opengrok_setup/opengrok/data/configuration.xml -P -S -r on -v -U localhost:8080 -s /scratch/opengrok_setup/opengrok/src -d /scratch/opengrok_setup/opengrok/data

vladak:
Okay, that looks normal. If you go to the `/scratch/opengrok_setup/opengrok/data/index` directory, is there a subdirectory matching the name of the project exhibiting the failure ? If yes, what is inside this directory ? What are the permissions of this directory and the files therein ?

vladak:
Is there any related log messages in Tomcat logs ?

vladak:
[...]

PR #3424, opened 2021-02-16
more verbose error for index problems

vladak:
Should help in cases such as #3418.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
@@ -350,7 +350,7 @@ public class SearchHelper {
             if (projects.size() > 0) {
                 errorMsg = errorMsg + " for projects: " + String.join(", ", projects);
             }
-            errorMsg = errorMsg + " " + e.getMessage();
+            errorMsg = errorMsg + "; " + e.getMessage();
         } catch (IOException e) {
             errorMsg = e.getMessage();
         }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
@@ -319,7 +319,7 @@ public class SearchHelper {
                     searcher = new IndexSearcher(reader);
                 } else {
                     errorMsg = "Failed to initialize search. Check the index";
-                    if (projects.size() > 0) {
+                    if (!projects.isEmpty()) {
                         errorMsg += " for projects: " + String.join(", ", projects);
                     }
                     return this;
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
```diff
@@ -314,12 +314,14 @@ public SearchHelper prepareExec(SortedSet<String> projects) {
...
                 // around set of IndexReader objects.
-                reader = RuntimeEnvironment.getInstance().getMultiReader(
-                    projects, searcherList);
+                reader = RuntimeEnvironment.getInstance().getMultiReader(projects, searcherList);
                 if (reader != null) {
                     searcher = new IndexSearcher(reader);
                 } else {
-                    errorMsg = "Failed to initialize search. Check the index.";
+                    errorMsg = "Failed to initialize search. Check the index";
+                    if (projects.size() > 0) {
```
ahornace, line 322: Just a nitpick, I think it is generally preferred to use `isEmpty()` instead of comparing `size()` to `0`.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
```diff
@@ -344,8 +346,11 @@ public SearchHelper prepareExec(SortedSet<String> projects) {
         } catch (ParseException e) {
             errorMsg = PARSE_ERROR_MSG + e.getMessage();
         } catch (FileNotFoundException e) {
-//          errorMsg = "Index database(s) not found: " + e.getMessage();
-            errorMsg = "Index database(s) not found.";
+            errorMsg = "Index database not found. Check the index";
+            if (projects.size() > 0) {
+                errorMsg = errorMsg + " for projects: " + String.join(", ", projects);
+            }
+            errorMsg = errorMsg + " " + e.getMessage();
```
ahornace, line 353: Maybe use different separator instead of space? E.g. `.`, `;`. I think it will improve the readability a bit.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
```diff
@@ -314,12 +314,14 @@ public SearchHelper prepareExec(SortedSet<String> projects) {
...
-                reader = RuntimeEnvironment.getInstance().getMultiReader(
-                    projects, searcherList);
+                reader = RuntimeEnvironment.getInstance().getMultiReader(projects, searcherList);
                 if (reader != null) {
                     searcher = new IndexSearcher(reader);
                 } else {
-                    errorMsg = "Failed to initialize search. Check the index.";
+                    errorMsg = "Failed to initialize search. Check the index";
+                    if (projects.size() > 0) {
+                        errorMsg = errorMsg + " for projects: " + String.join(", ", projects);
```
ahornace, line 323: Another nitpick, why not use `+=`? I personally like it more.

ahornace (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
vladak: good idea

opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
vladak: good idea

PR #3425, opened 2021-02-16
add API to get last index time

vladak:
Works like this:
```
$ curl -s http://localhost:8080/source/api/v1/system/indextime | jq
"2021-02-15T16:39:16.409+00:00"
```

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SystemController.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SystemController.java
@@ -18,14 +18,14 @@
  */
 
 /*
- * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opengrok.web.api.v1.controller;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.util.StdDateFormat;
+import org.opengrok.indexer.configuration.IndexTimestamp;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.web.EftarFile;
 import org.opengrok.indexer.logger.LoggerFactory;
@@ -41,9 +41,7 @@ import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
-import java.io.File;
 import java.io.IOException;
-import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SystemController.java
```diff
@@ -74,4 +81,16 @@ public void loadPathDescriptions(@Valid final PathDescription[] descriptions) th
...
+
+    @GET
+    @Path("/indextime")
+    @Produces(MediaType.APPLICATION_JSON)
+    public String getIndexTime() throws JsonProcessingException {
+        Date date = new IndexTimestamp().getDateForLastIndexRun();
+        ObjectMapper mapper = new ObjectMapper();
+        // StdDateFormat is ISO8601 since jackson 2.9
+        mapper.setDateFormat(new StdDateFormat().withColonInTimeZone(true));
+        String result = mapper.writeValueAsString(date);
```
ahornace, line 93: Why not return directly by `return mapper.writeValueAsString(date);` ? I think `result` variable name is too common to validate its use here.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
```diff
@@ -130,4 +136,26 @@ public void testDtagsEftarReload() throws IOException {
         // Cleanup
         IOUtils.removeRecursive(dataRoot);
     }
+
+    @Test
+    public void testIndexTime() throws IOException, ParseException {
+        Path dataRoot = Files.createTempDirectory("indexTimetest");
+        env.setDataRoot(dataRoot.toString());
+        File indexTimeFile = Paths.get(dataRoot.toString(), "timestamp").toFile();
```
ahornace, line 144: I recommend using:
`dataRoot.resolve("timestamp")`
and then methods from
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
```diff
@@ -130,4 +136,26 @@ public void testDtagsEftarReload() throws IOException {
...
+        SimpleDateFormat f = new SimpleDateFormat("yyyy-MM-dd_hh:mm:ss+ZZ");
+        Date date = f.parse("2021-02-16_11:18:01+UTC");
+        indexTimeFile.setLastModified(date.getTime());
+
+        Response r = target("system")
+                .path("indextime")
+                .request().get();
+        String result = r.readEntity(String.class);
+
+        assertThat(result, containsString("2021-02-16T11:18:01.000+00:00"));
```
ahornace, line 156: Should not we check for equality?

ahornace (approved):

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
vladak: That's because JSON returns the string with double quotes and I thought checking the whole input is too strong. Thinking about this again, it should not hurt.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SystemController.java
vladak: Separate variable could be handy for debugging however in this case it is probably unwarranted.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SystemControllerTest.java
vladak: The `resolve()` is beneficial. Not so sure about the `Files` methods.

ahornace (approved):

vladak:
thanks @ahornace !

PR #3436, opened 2021-02-25
Update README.md: "git pull --ff-only"

45hvnL:
OpenGrok mirroring by default does a "git pull --ff-only" for Git repos, not a regular "git pull". This avoids merges occurring.


vladak:
Thanks !

Issue #3434, opened 2021-02-24
Indexer option to run historycache only


ChristopheBordieu:
**Is your feature request related to a problem? Please describe.**
Let's write yes.
I am reindexing all my company Git repos (30k) with latest OG 1.5.11 / Tomcat 9 / Java 11 / RHEL 8.3.
Historycache generation is extremely long...

**Describe the solution you'd like**
I would like when reindexing:
1a run indexer without historycache
1b run another indexer process in parallel with historycache only that's to say skipping file indexing since this is done in a short time by process of step 1a
2 each night run indexer without historycache
3 when historycache started in step 1b finishes ok, then and only then, update job run nightly in step 2 adding historycache too

**Describe alternatives you've considered**
Optimizing the historycache generation? Issue to be opened further to your answer to this one.
But in all case, because of the size of our code sources, historycache process has always been very long.

**Additional context**
Some of our repos have very long history, with many tags and also many files. This does not help generating quickly the history.

vladak:
The history cache update for Git is incremental so the long historycache run has to be caused by either the sheer amount of repositories and/or the processing overhead. I wonder if replacing the executions of the `git` binary in the `GitRepository` class with pure Java implementation (say Eclipse JGit) would help things.

louie0817:
> The history cache update for Git is incremental so the long historycache run has to be caused by either the sheer amount of repositories and/or the processing overhead. I wonder if replacing the executions of the `git` binary in the `GitRepository` class with pure Java implementation (say Eclipse JGit) would help things (making it hopefully more I/O bound).

Vlad, can you elaborate on "incremental" please. We have a huge GHE repo that has over 500k commits. at the top level, the uncompressed OpenGrokDirHist.gz file is over 4 GB. yesterday just one file was updated in the top level dir and about 8 files change in various subdirs. The history cache generation for just the top level dir, took 1 hour and 40 minutes and this happens everynight as they update the file every night. Incremental to me would to run git log and stop when I hit the sha1 saved in the OpenGroklatestRev file.

vladak:
[...]

PR #3438, opened 2021-02-27
report duration for file history cache updates

vladak:
This is to help identify the long history index times reported in #3434.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -138,6 +141,10 @@ private void doFileHistory(String filename, List<HistoryEntry> historyEntries,
         }
 
         storeFile(hist, file, repository, !renamed);
+
+        statRepoHist.report(LOGGER, Level.FINER,
+                String.format("Done storing history cache for '%s'", filename),
```
vladak, line 146: Here, the `String.format()` will be evaluated every time. Thinking of introducing static method `Statistics.eligible(LogLevel level)` that would do something like:
```java
return (LOGGER.isLoggable(level) || Metrics.getRegistry());
```

PR #3439, opened 2021-02-27
Suggester rebuild does not reflect fields change

ahornace:
Fixes issue when suggester did not work in docker image due to changes in https://github.com/oracle/opengrok/pull/3402

Thanks.

### DIFF
--- a/suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java
+++ b/suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java
@@ -316,7 +316,7 @@ public class SuggesterProjectDataTest {
         }
         init(false);
 
-        // add new field after suggester data were initialized
+        // add new field after suggester data was initialized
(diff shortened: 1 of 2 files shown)


suggester/src/test/java/org/opengrok/suggest/SuggesterProjectDataTest.java
```diff
@@ -308,4 +308,22 @@ public void testGetSearchCountMapSorted() throws IOException {
...
 
+    @Test
+    public void testRebuildPicksUpNewFields() throws IOException {
+        // create dummy index
+        try (IndexWriter iw = new IndexWriter(dir, new IndexWriterConfig())) {
+            iw.isOpen(); // dummy operation
+        }
+        init(false);
+
+        // add new field after suggester data were initialized
```
vladak, line 319:
```suggestion
        // add new field after suggester data was initialized
```

vladak:
Thanks @ahornace !

PR #3440, opened 2021-03-01
Update Configuration.java

herkura:
Converted into HashMap


vladak (changes requested):
I don't see how this could ever work. You need to modify the setter/getter and other functions accessing the `groups`.

PR #3446, opened 2021-03-02
add option/config for ignoring projects

vladak:
Needed for local deployment where one of the projects has extensive pre/post mirror hooks so I'd like to run the sync for it separately, with different schedule. With this change I'd just run the main sync as `opengrok-sync --ignore_project foo --indexed ...`.

### DIFF
--- a/tools/src/main/python/opengrok_tools/sync.py
+++ b/tools/src/main/python/opengrok_tools/sync.py
@@ -131,13 +131,15 @@ def main():
     group1 = parser.add_mutually_exclusive_group()
     group1.add_argument('-d', '--directory',
                         help='Directory to process')
-    group1.add_argument('-P', '--projects', nargs='*',
-                        help='List of projects to process')
+    group1.add_argument('-P', '--project', nargs='*',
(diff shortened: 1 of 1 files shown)


ahornace (approved):

tools/src/main/python/opengrok_tools/sync.py
```diff
@@ -131,13 +131,15 @@ def main():
...
-    group1.add_argument('-P', '--projects', nargs='*',
-                        help='List of projects to process')
+    group1.add_argument('-P', '--project', nargs='*',
+                        help='project(s) to process')
 
     parser.add_argument('-I', '--indexed', action='store_true',
                         help='Sync indexed projects only')
     parser.add_argument('-i', '--ignore_errors', nargs='*',
                         help='ignore errors from these projects')
+    parser.add_argument('--ignore_project', nargs='+',
```
ahornace, line 141: Should not we use `*` instead of `+`?

ahornace: Nevermind :D

Issue #3453, opened 2021-03-08
convert command list to dictionary


vladak:
A list is currently used to set various API call properties in command sequence (URI, HTTP verb, data, headers). This is inconvenient when one wants to specify per command proxies or timeout value. Dictionary would be more flexible.

vladak:
From:
```yml
- command:
  - http://foo:8080/source/api/v1/messages
  - POST
  - messageLevel: warning
    duration: PT1H
    tags: ['%PROJECT%']
    text: resync + reindex in progress
```
to something like:
```yml
- command:
  uri: http://foo:8080/source/api/v1/messages
  verb: POST
  data:
    messageLevel: warning
    duration: PT1H
    tags: ['%PROJECT%']
    text: resync + reindex in progress
  headers:
    one: two
    foo: bar
  timeout: 42
```

vladak:
This will change the config file format for both `opengrok-mirror` and `opengrok-sync`.

PR #3454, opened 2021-03-08
requests timeout for tools

vladak:
This change introduces the request timeout to the tools. The default is set to 60 seconds in `requests#do_api_call()`. The command line tools received the `--api_timeout` option.

So far it is not possible to specify per command (as in command sequence) timeout, tracked by #3453.


ahornace (approved):

vladak:
thanks @ahornace !

Issue #3455, opened 2021-03-08
does REINDEX=0 have any effect in new start.py?


louie0817:
**Describe the bug**

Setting  REINDEX=0 to prevent reindexing in docker image no longer has any effect.

Docker 1.6.1 image

in the syncer function, the reindex arg is only used to calculate number of seconds to sleep, there is no check for zero value

jnovak-praetorian:
I have also noticed this in the newest release - with `REINDEX=0` my opengrok instance is continuously re-indexing, not disabling it.

jnovak-praetorian:
For a temporary, not great fix -- set `REINDEX=-1`. This will throw an exception in Python's `time.sleep()` function, thereby breaking you out of the sync loop.

jnovak-praetorian:
By my reading this commit doesn't seem to allow you to set the `SYNC_PERIOD_MINUTES` equal to zero (thereby bypassing the sync loop)

https://github.com/vladak/OpenGrok/commit/5b81e69972e5a4e3788db9f69ae9588aceb4e55f#diff-d0e0d1fe01e659a4debf8bf6d66912cfafb1aa1d88537f03686f789af4c94c92R309-R318

PR #3456, opened 2021-03-09
breaking sync loop when REINDEX <= 0

jnovak-praetorian:


jnovak-praetorian:
Closes #3455

vladak:
Thanks, closing in favor of #3458 to fix this in more robust way.

PR #3458, opened 2021-03-09
rename REINDEX to SYNC_PERIOD_MINUTES and allow disable

vladak:
This change renames the `REINDEX` environment variable used in the Docker image to `SYNC_PERIOD_MINUTES` so that it is more telling. Also, I made the checking of any integer arguments more strict.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,9 @@ COPY suggester/pom.xml /mvn/suggester/
 RUN sed -i 's:<module>distribution</module>::g' /mvn/pom.xml
 RUN sed -i 's:<module>tools</module>::g' /mvn/pom.xml
 
+# get clean state
+RUN mvn clean
+
 RUN mkdir -p /mvn/opengrok-indexer/target/jflex-sources
 RUN mkdir -p /mvn/opengrok-web/src/main/webapp
 RUN mkdir -p /mvn/opengrok-web/src/main/webapp/WEB-INF/ && touch /mvn/opengrok-web/src/main/webapp/WEB-INF/web.xml
--- a/docker/start.py
+++ b/docker/start.py
@@ -218,7 +218,7 @@ def merge_commands_env(commands, env):
     return commands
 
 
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -18,6 +18,9 @@ COPY suggester/pom.xml /mvn/suggester/
 RUN sed -i 's:<module>distribution</module>::g' /mvn/pom.xml
 RUN sed -i 's:<module>tools</module>::g' /mvn/pom.xml
 
+# get clean state
+RUN mvn clean
```
vladak, line 22: sometimes the docker image build barfs in tools build due to leftover `target/env` from local system build. This takes care of that.

docker/start.py
```diff
@@ -305,13 +305,19 @@ def main():
...
-    else:
-        reindex_min = 10
-    logger.debug("reindex period = {} minutes".format(reindex_min))
+    # default period for syncing (in minutes)
+    sync_period = 10
+    sync_env = os.environ.get('SYNC_TIME_MINUTES')
+    if sync_env:
+        try:
+            n = int(sync_env)
+            if n > 0:
```
jnovak-praetorian, line 314: This should be `n >= 0` to allow for a zero value (and bypassing the sync loop)

Dockerfile
ahornace: This does not help, does it? At this point we only copied `pom.xml` files to the docker image. I think we should modify `RUN mvn -DskipTests=true -Dmaven.javadoc.skip=true -B -V package` to include `clean` target as well, right?

docker/start.py
vladak: will fix, thanks !

ahornace (approved):

jwehler:
fyi, it looks like the documentation is showing SYNC_PERIOD_MINUTES, but the code is using SYNC_TIME_MINUTES.
`    sync_env = os.environ.get('SYNC_TIME_MINUTES')`

vladak:
fixed, thanks !

PR #3466, opened 2021-03-10
enable projects separately in Docker

vladak:
This brings back repositories in Docker in setup with projects.


vladak:
Arguably, -P with --noIndex should not perform project scan however at this point I decided to fix this in Docker only.

ahornace (approved):

PR #3472, opened 2021-03-11
changed Set to HashMap - fix #1587

### LINKED ISSUE #1587: use hashmap for groups too
projects are using hashmap (#1543) for fast lookup, the same should be implemented for groups

adityamathur97:
I tried testing the test cases available in the project after making the changes and this is the result I got,

image

Please review the changes and let me know how I can improve. Thank you!


vladak:
The tests have to pass. Looks like this does not even compile at the moment. Also, get rid of these unnecessary whitespace changes, they make the changes hard to comprehend.

ahornace:
Thank you for your effort! Closing due to inactivity. Please reopen once the tests pass.

PR #3484, opened 2021-03-17
get last revision date from the index first

vladak:
This change stores last revision of given file in the index. Web app will then take advantage of this when getting the revision number in order to perform redirection. If the newly introduced `LASTREV` field is not found, it will fall back to the old behavior.

In general this should be quicker than getting the history from the history cache since the index is usually memory mapped, i.e. avoids reading the history cache file, deserializing potentially huge `History` object and picking just one `HistoryEntry`.


ahornace (approved):

opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
```diff
@@ -1291,11 +1294,26 @@ public File findDataFile() {
...
+            return lastRev;
+        }
+
+        // fallback
+        return getLastRevFromHistory();
+    }
+
+    @Nullable
+    private String getLastRevFromHistory() {
+        // fallback
```
vladak, line 1316:
```suggestion
```

vladak:
thanks @ahornace !

PR #3485, opened 2021-03-17
add more search.latency categories

vladak:
When investigating big latencies I came across the `search.latency` meter and realized that it is only used when searching the index via part of the UI. This change adds more categories. Because of Prometheus limitation the set of tags for the common meter name has to be the same.

Tested by performing search in the main form and via RESTful API and listing a directory.


ahornace (approved):

vladak:
thanks @ahornace !

PR #3486, opened 2021-03-17
Projects controller vs history

vladak:
This change should take care of the project addition/removal via the RESTful API w.r.t. repositories and disabled history.


ahornace (approved):

vladak:
thanks @ahornace !

PR #3489, opened 2021-03-21
set default value to avoid nullptr at start.py:391

xundaoxd:


ahornace:
I could not find your name in https://www.oracle.com/technical-resources/oracle-contributor-agreement.html
Could you please sign it? We cannot accept your change without it.

vladak:
For starters, please state that you are okay with the OCA conditions.

xundaoxd:
OK, I have signed an Individual OCA.

PR #3490, opened 2021-03-21
First set of chages for JGit

vladak:
This is the initial phase of converting `GitRepository` to use JGit instead of executing the `git` binary. In [this round](https://github.com/oracle/opengrok/issues/1113#issuecomment-803470717), all  functions necessary for `RepositoryFactory.getRepository()` are converted. This should speed up repository scan and invalidation.

Repository scan (-S) of 10k Git repositories (dotfiles) takes some 6 and half minutes with the original code and some 45 seconds with the code from this change. That is on my i5 based mac mini with SSD.


opengrok-indexer/pom.xml
```diff
@@ -233,6 +233,11 @@ Portions Copyright (c) 2020-2020, Lubos Kosco <tarzanek@gmail.com>.
             <artifactId>annotations</artifactId>
             <version>20.1.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jgit</groupId>
+            <artifactId>org.eclipse.jgit</artifactId>
+            <version>5.10.0.202012080955-r</version>
```
ahornace, line 239: Just a note, 5.11 version was released last week: https://mvnrepository.com/artifact/org.eclipse.jgit/org.eclipse.jgit/5.11.0.202103091610-r

However, if it has some breaking changes, then we can create a separate PR for that.

vladak: should go with the latest. The changes use official APIs only.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
```diff
@@ -86,13 +97,133 @@ public void tearDown() {
...
+        GitRepository gitrepo = (GitRepository) RepositoryFactory.getRepository(root);
+        assertNotNull(gitrepo);
+        String ver = gitrepo.determineCurrentVersion();
+        assertNotNull(ver);
+        Date date = new Date((long) (timestamp) * 1000);
+        assertEquals(Repository.format(date) + " " + commitId + " " + shortComment, ver);
+    }
+
     @Test
     public void testDetermineCurrentVersion() throws Exception {
```
ahornace, line 111: I feel the tests are overly long and are testing multiple things at once. I'd personally break them into multiple tests, e.g.
- testDetermineCurrentVersionForSingleRepository
- testBuildTagListForSingleHeadTag
- testBuildTagListForMultipleHeadTags
- testBuildTagListForTagOnSpecificCommit

If such a test fails and is short, then I find it easier to determine what is actually wrong.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -578,6 +595,44 @@ boolean hasFileBasedTags() {
...
+                for (Ref ref : refList) {
+                    String tagName = ref.getName().replace("refs/tags/", "");
+                    String existingTags;
+                    RevCommit commit = getCommit(repository, ref);
+                    if ((existingTags = commit2Tags.get(commit)) != null) {
+                        existingTags = existingTags + TAGS_SEPARATOR + tagName;
+                        commit2Tags.put(commit, existingTags);
+                    } else {
+                        commit2Tags.put(commit, tagName);
+                    }
```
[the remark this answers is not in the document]
vladak: good idea !

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -578,6 +595,42 @@ boolean hasFileBasedTags() {
...
+
+    private void rebuildTagList(File directory) {
+        this.tagList = new TreeSet<>();
+        try (org.eclipse.jgit.lib.Repository repository = getJGitRepository(directory.getAbsolutePath())) {
+            try (Git git = new Git(repository)) {
+                List<Ref> refList = git.tagList().call(); // refs sorted according to tag names
+                Map<RevCommit, String> commit2Tags = new HashMap<>();
+                for (Ref ref : refList) {
+                    String tagName = ref.getName().replace("refs/tags/", "");
+                    String existingTags;
```
ahornace, line 610: `String existingTags;` line is no longer needed.

ahornace (approved):
Great job! :)

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
vladak: make sense, refactored.

vladak:
thanks @ahornace !

PR #3491, opened 2021-03-22
run hg binary only once per isWorking()

vladak:
This mimics the changes done in `GitRepository` to avoid checking Mercurial command for some of repository operations repeatedly. Should help repository scan/invalidation performance and thus web app startup.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java
@@ -32,6 +32,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.TreeSet;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java
```diff
@@ -104,6 +105,12 @@
     private static final Pattern LOG_COPIES_PATTERN
             = Pattern.compile("^(\\d+):(.*)");
 
+    /**
+     * This is a static replacement for 'working' field. Effectively, check if hg is working once in a JVM
+     * instead of calling it for every MercurialRepository instance.
+     */
+    private static final LazilyInstantiate<Boolean> HG_IS_WORKING = LazilyInstantiate.using(MercurialRepository::isHgWorking);
```
ahornace, line 112: Just a note for future, I'd personally change the name of `LazilyInstantiate` to `MemoizingSupplier` which follows more common naming or even better use one from Guava https://guava.dev/releases/19.0/api/docs/com/google/common/base/Suppliers.html#memoize(com.google.common.base.Supplier)

ahornace (approved):
Good idea!

opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java
ahornace: Also, I think using supplier type is nicer:
 `Supplier<Boolean> HG_IS_WORKING = ...`

vladak: ok, used Supplier.

vladak:
thanks @ahornace !

PR #3495, opened 2021-03-24
print list of analyzers with -h guru

vladak:
This change adds table of analyzer -> name to the output of `-h guru`:

```
List of analyzers:
The names of the analyzers (left column) can be used for the -A indexer option:

Ada        : Ada
Asm        : Asm
BZip2      : Bzip(2)
C          : C
CSharp     : C#
Cxx        : C++
Clojure    : Clojure
Eiffel     : Eiffel
ELF        : ELF
Erlang     : Erlang
File       : N/A
Fortran    : Fortran
Golang     : Golang
GZIP       : GZIP
Haskell    : Haskell
HCL        : HCL
Ignorant   : N/A
Image      : Image file
Jar        : Jar
Java       : Java
JavaClass  : Java class
JavaScript : JavaScript
Json       : Json
Kotlin     : Kotlin
Lisp       : Lisp
Lua        : Lua
Mandoc     : Manual pages
Pascal     : Pascal
Perl       : Perl
Php        : PHP
PLSQL      : PL/SQL
Plain      : Plain Text
Powershell : PowerShell script
Python     : Python
R          : R
Ruby       : Ruby
Rust       : Rust
Scala      : Scala
Sh         : Shell script
SQL        : SQL
Swift      : Swift
Tar        : Tar
Tcl        : Tcl
Terraform  : Terraform
Troff      : Troff
TypeScript : TypeScript
Uuencode   : UUEncoded
Verilog    : Verilog
VB         : Visual Basic
XML        : XML
Zip        : Zip

AnalyzerGuru prefixes:
MAKEFILE*  : Shell script
...
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuruHelp.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuruHelp.java
@@ -48,6 +48,18 @@ public class AnalyzerGuruHelp {
      */
     public static String getUsage() {
         StringBuilder b = new StringBuilder();
+
+        b.append("List of analyzers:\n");
+        b.append("The names of the analyzers (left column) can be used for the -A indexer option:\n\n");
+        byFactory(AnalyzerGuru.getAnalyzerFactories().stream().
+                collect(Collectors.toMap(f -> f.getClass().getSimpleName(), f -> f))).
+                forEach((factory) -> {
+            b.append(String.format("%-10s : %s\n",
+                    factory.fac.getClass().getSimpleName().replace("AnalyzerFactory", ""),
+                    factory.fac.getName() != null ? factory.fac.getName() : "N/A"));
+        });
+        b.append("\n");
+
         b.append("AnalyzerGuru prefixes:\n");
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuruHelp.java
```diff
@@ -48,7 +49,18 @@ private AnalyzerGuruHelp() {
...
     public static String getUsage() {
         StringBuilder b = new StringBuilder();
-        b.append("AnalyzerGuru prefixes:\n");
+
+        b.append("List of analyzers:\n");
+        b.append("The names of the analyzers (left column) can be used for the -A indexer option:\n\n");
+        byFactory(AnalyzerGuru.getAnalyzerFactories().stream().
+                collect(Collectors.toMap(f -> f.getClass().getSimpleName(), f -> f))).
+                forEach((factory) -> {
+            b.append(String.format("%-10s : %s\n",
```
ahornace, line 58: Just a nitpick, even though it should work fine on Windows, maybe it would be better to use `System.lineSeparator()`?

Btw, I think that with with `String.format` you can use `%n`.

ahornace (approved):

PR #3496, opened 2021-03-24
Merge commits tunable

vladak:
This change allows to control whether the merge commits should be stored in history. Should work as workaround with indexing large repositories that have rich history. This is relevant only to Git so far.

I did not want to introduce yet another command line option so this needs to be used via read-only configuration.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -232,6 +232,8 @@ public final class Configuration {
      */
     private boolean handleHistoryOfRenamedFiles;
 
+    private boolean mergeCommitsEnabled;
+
     public static final double defaultRamBufferSize = 16;
 
     /**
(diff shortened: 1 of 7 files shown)


ahornace (approved):

vladak:
wikis that will need an update:
  - https://github.com/oracle/opengrok/wiki/Indexer-configuration
  - https://github.com/oracle/opengrok/wiki/Indexer-configuration

vladak:
Tested with read-only configuration:
```xml

    <void method="addGroup">
        <object class="org.opengrok.indexer.configuration.Group">
            <void property="name">
                <string>OpenGrok</string>
            </void>
            <void property="pattern">
                <string>opengrok</string>
            </void>
        </object>
    </void>

<!--
    <boolean>false</boolean>
  </void>
-->
   <boolean>true</boolean>
  </void>

   <void property="projects">
     <void method="put">
      <string>one</string>
        <boolean>true</boolean>
       </void>
        <boolean>true</boolean>
       </void>
       <void property="indexed">
        <boolean>true</boolean>
       </void>
       <void property="name">
        <string>one</string>
       </void>
       <void property="path">
        <string>/one</string>
       </void>
        <boolean>false</boolean>
       </void>
      </object>
     </void>
   </void>

 </object>
</java>
```

PR #3499, opened 2021-03-24
change GitRepository#GetHistoryGet() to use JGit

vladak:
Another step in adapting `GitRepository` to use JGit.


ahornace (approved):
Nice job!

PR #3504, opened 2021-03-25
skip over false tags

vladak:
With this change the tag rebuild for the linux-mainline repository finishes fine, with some 687 `tagList` entries currently.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -649,7 +654,7 @@ private void rebuildTagList(File directory) {
             // In case of partial success, do not null-out tagList here.
         }
 
-        if (LOGGER.isLoggable(Level.FINEST)) {
+        if (LOGGER.isLoggable(Level.FINER)) {
```
ahornace, line 657: Why this change? It does not make much sense to check for higher log level, right?

vladak: I think it should be higher than the `FINEST` log above.

PR #3505, opened 2021-03-25
use JGit for annotations

vladak:
Another set of JGit changes, this time for getting annotations.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -488,61 +465,71 @@ private String getFirstRevision(String fullpath) {
...
+            ObjectId commitId = repository.resolve(revision);
+            blameCommand.setStartCommit(commitId);
+            blameCommand.setFollowFileRenames(isHandleRenamedFiles());
+            final BlameResult result = blameCommand.setTextComparator(RawTextComparator.WS_IGNORE_ALL).call();
+            if (result != null) {
+                final RawText rawText = result.getResultContents();
+                for (int i = 0; i < rawText.size(); i++) {
+                    final PersonIdent sourceAuthor = result.getSourceAuthor(i);
+                    final RevCommit sourceCommit = result.getSourceCommit(i);
+                    annotation.addLine(sourceCommit.getId().abbreviate(8).name(), sourceAuthor.getName(), true);
```
ahornace, line 524: Maybe create a constant for `8`?

ahornace (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: sure. I thought there is one however could not find it.

PR #3511, opened 2021-03-30
add FileHistoryCache.get() counters for hits/misses

vladak:
No project info to avoid leakage.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
@@ -55,6 +55,8 @@ import java.util.logging.Logger;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
+import io.micrometer.core.instrument.Counter;
+import org.opengrok.indexer.Metrics;
 import org.opengrok.indexer.configuration.PathAccepter;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.logger.LoggerFactory;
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexerMainTest.java
```diff
@@ -34,7 +34,7 @@
 
 import static org.junit.Assert.assertFalse;
 
-public class IndexerMainTest {
+public class  IndexerMainTest {
```
ahornace, line 37: Why the additional space? Maybe a mistake?

vladak: yep, inadvertent modification.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -174,7 +180,17 @@ protected Expression instantiate(Object oldInstance, Encoder out) {
...
-        // nothing to do
+        MeterRegistry meterRegistry = Metrics.getRegistry();
+        if (meterRegistry != null) {
+            fileHistoryCacheHits = Counter.builder("filehistorycache.history.get").
+                    description("file history cache hits").
+                    tag("what", "hits").
+                    register(meterRegistry);
+            fileHistoryCacheMisses = Counter.builder("filehistorycache.history.get").
+                    description("file history cache misses").
+                    tag("what", "hits").
```
ahornace, line 191: Should not this be a "miss"?

vladak: oops, thanks !

vladak:
thanks @ahornace !

PR #3514, opened 2021-03-31
Code cleanup

ahornace:
I've applied automatic cleanup in IDEA plus ran the analysis and applied IDEA suggestions when it made sense to me.

This is not a complete list, I left out some suggestions as the diff grew big. 

Thanks :)


vladak:
whoa, thanks !

Issue #3517, opened 2021-03-31
indexing with https


chkp-baselz:
Hey, how could I index with https website?
I though to change it into this, looks like it's doing some issues
`java     -Djava.util.logging.config.file=/opengrok/etc/logging.properties     -jar /opengrok/dist/lib/opengrok.jar     --ctags  /usr/local/bin/ctags     -s /opengrok/src -d /opengrok/data -H -P -S -G     -W /opengrok/etc/configuration.xml -U https://opengrok02:8443/source/`

the change in server.xml is 
`<Connector port="8443" protocol="HTTP/1.1" SSLEnabled="true"
        keystoreFile="jks" keystorePass="tomcat9"
               maxThreads="150" scheme="https" secure="true"
               clientAuth="false" sslProtocol="TLS" keyalias="tomca"/>`

vladak:
Looks like you are trying to run the indexer on different machine than the web app is running on. If that's the case, see https://github.com/oracle/opengrok/wiki/Web-services#authenticationauthorization

chkp-baselz:
@vladak I don't think this is the case, indexing a specific version works good
`java -Djava.util.logging.config.file=/opengrok/etc/logging.properties -jar /opengrok/dist/lib/opengrok.jar --ctags /usr/local/bin/ctags -s /opengrok/src -d /opengrok/data -H -P -G -R /opengrok/etc/configuration.xml -U  https://opengrok02:8443/source/ VERSION`

chkp-baselz:
indexing with localhost gives this error
![image](https://user-images.githubusercontent.com/77238449/113168344-3ee0f780-924d-11eb-9d18-0d1a09edac3e.png)

also the opengrok on the same machine

vladak:
Yes, the -U option has to have a URL where the web app is reachable.

chkp-baselz:
@vladak so I didn't understand how to fix this?

vladak:
Basically, the web app has to be running on the URI you are supplying with the `-U` indexer option. For example, if you run the indexer with `-U http://localhost:8080/source`, the web app has to be accepting TCP connections on port 8080 on `localhost` via plain HTTP.

If you specify non-localhost URI then you need to use the Bearer token (see the link above) so that the API requests issued by the indexer can go through.

chkp-baselz:
@vladak okay thank you, on a diffrent topic, should src folder contain gz files or only data? the indexer looks like he left gz files in src folder is that fine or should I remove them?

[...]

PR #3520, opened 2021-03-31
fail early if web app URI is not reachable

vladak:
#3517 inspired me to add this check. Wanted to do this outside of argument parsing - if there is any problem with options this should fail first. These methods come from authnz plugins so this is code duplication however plugins needs to be independent.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -173,7 +173,7 @@ public final class Indexer {
                 try {
                     for (InetAddress addr : InetAddress.getAllByName(HostUtil.urlToHostname(webappURI))) {
                         int port = HostUtil.urlToPort(webappURI);
-                        if (port < 0) {
+                        if (port <= 0) {
                             LOGGER.log(Level.SEVERE, "invalid port number for " + webappURI);
                             break;
                         }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
@@ -33,7 +33,6 @@ import java.io.PrintStream;
 import java.io.UncheckedIOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
-import java.net.InetAddress;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -161,9 +163,36 @@ public static void main(String[] argv) {
 
         boolean createDict = false;
 
+        int CONNECT_TIMEOUT = 1000;  // in milliseconds
+
         try {
             argv = parseOptions(argv);
 
+            if (webappURI != null) {
```
ahornace, line 171: I'd move this check inside a separate method inside the `HostUtil` class. I think the `main` method is already too long.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -161,9 +163,36 @@ public static void main(String[] argv) {
...
+                            connectWorks = true;
+                            break;
+                        }
+                    }
+                } catch (URISyntaxException e) {
+                    LOGGER.log(Level.WARNING, String.format("URI not valid: %s", webappURI), e);
+                }
+
+                if (!connectWorks) {
+                    LOGGER.log(Level.SEVERE, webappURI + " is not reachable.");
```
ahornace, line 191: It seems we're printing the message to `System.err` right before the exit in other places. Will the log be always displayed?

opengrok-indexer/src/main/java/org/opengrok/indexer/util/HostUtil.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return true;
+    }
+
+    public static boolean isReachable(String webappURI, int timeOutMillis) {
+        boolean connectWorks = false;
+        try {
+            for (InetAddress addr : InetAddress.getAllByName(HostUtil.urlToHostname(webappURI))) {
+                int port = HostUtil.urlToPort(webappURI);
```
ahornace, line 88: The port can actually be checked only once before the `for`, right?

vladak: true !

ahornace (approved):

vladak:
thanks @ahornace !

PR #3524, opened 2021-04-05
Move jflex sources to src/main/jflex

ahornace:
The first step for removing ant from build as this removes `-collect-lex-lexh` target.

Thanks :)


vladak:
Whoa, thanks !

Issue #3523, opened 2021-04-05
Failed to get xref file


chkp-baselz:
Hey,
![image](https://user-images.githubusercontent.com/77238449/113573407-c601e700-9622-11eb-9a80-f29247e056ff.png)

 I had this issue before([3444](https://github.com/oracle/opengrok/issues/3444)) and fixed it by removing the gz files from the data folder and re-indexing the version, now I have only gz.org_opengrok extension file, no gz file, **removing it and touching the src file re-index won't solve the problem**
![image](https://user-images.githubusercontent.com/77238449/113573423-d0bc7c00-9622-11eb-892c-85859838c40f.png)

vladak:
This is indeed different than #3444 where the xref file was present but stale. Here the indexer did not finish the move from temporary location to final destination.

As a workaround you can try moving the `globals.cc.gz.org_opengrok` to `globals.cc.gz` and touching it.

chkp-baselz:
@vladak it worked, thank you

PR #3525, opened 2021-04-06
generate xref on the fly if xref is stale or non existent

vladak:
This should help #3523. It hides the original problem with indexing, though.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/FileExtraZipper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/FileExtraZipper.java
@@ -45,8 +45,7 @@ public class FileExtraZipper {
      * @param extras some OpenGrok-analyzed extra metadata
      * @return a list of the same size as {@code files}
      */
-    public List<DirectoryEntry> zip(File dir, List<String> files,
-        List<FileExtra> extras) {
+    public List<DirectoryEntry> zip(File dir, List<String> files, List<FileExtra> extras) {
 
         if (extras == null) {
             return files.stream().map(f ->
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/util/FileExtraZipper.java
```diff
@@ -75,9 +74,11 @@ private FileExtra findExtra(Map<String, FileExtra> byName, File fileobj) {
     private Map<String, FileExtra> indexExtraByName(List<FileExtra> extras) {
         Map<String, FileExtra> byPath = new HashMap<>();
         for (FileExtra extra : extras) {
-            File f = new File(extra.getFilepath());
-            String filename = f.getName();
-            byPath.put(filename, extra);
+            if (extra.getFilepath() != null) {
```
vladak, line 77: this I added because of interaction with index created with the #3138 changes - was getting NPEs due to `filepath = null`.

opengrok-web/src/main/webapp/xref.jspf
```diff
@@ -0,0 +1,193 @@
+<%--
+$Id$
```
ahornace, line 2: Just curious, what does this actually do? Is it necessary or some leftover? Some `.jspf` files in OpenGrok have it and some don't.

ahornace (approved):

opengrok-web/src/main/webapp/xref.jspf
vladak: That must somehow come from history. The `$Id$` is RCS (CVS) keyword that gets expanded into file metadata (https://web.mit.edu/gnu/doc/html/cvs_17.html).

Issue #2329, opened 2018-08-31
choose different serialization scheme for storing configuration


vladak:
The XML encoder used for configuration serialization is not very robust (e.g. in the face of changing class hierarchy and removing configuration options) and has some quirks (#2002). We should consider using something else (YAML/JSON ?).

Also, this serialization is used not only for configuration but also elsewhere (`AnalyzerGuru` settings).

tulinkry:
Yes, finally.

tulinkry:
Looks like yaml would be the way to go.

vladak:
Also, the configuration should be treated as data, not serialized objects.

vladak:
The other reason for using something else is performance. Lately, I realized that `XMLEncoder` does not scale when retrieving configuration using the RESTful API. When running a multithreaded program where each thread just retrieves the configuration in a loop, where the number of threads matches the number of CPUs, the times shoot up to almost 2 seconds, compared to single threaded program with 0.4 seconds. The XML file with the configuration has some 1.38 MB. When I got a `jstack` snapshot, it revealed that lots of the XMLEncoder processing threads (like 25 out of the 32 threads I was using) are waiting on internal synchronization object, with top of the stack looking like this:
```
"http-nio-8080-exec-1427" #29360 daemon prio=5 os_prio=64 cpu=38052.59ms elapsed=2859934.54s tid=0x000000000531c000 nid=0x7981 waiting for monitor entry  [0x00007fff808fa000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at com.sun.beans.util.Cache.get(java.desktop@11.0.7-internal/Cache.java:119)
        - waiting to lock <0x00007ff387d7f320> (a java.lang.ref.ReferenceQueue)
        at com.sun.beans.finder.MethodFinder.findMethod(java.desktop@11.0.7-internal/MethodFinder.java:81)
        at java.beans.Statement.getMethod(java.desktop@11.0.7-internal/Statement.java:369)
        at java.beans.Statement.invokeInternal(java.desktop@11.0.7-internal/Statement.java:273)
        at java.beans.Statement$2.run(java.desktop@11.0.7-internal/Statement.java:187)
        at java.security.AccessController.doPrivileged(java.base@11.0.7-internal/Native Method)
        at java.beans.Statement.invoke(java.desktop@11.0.7-internal/Statement.java:184)
        at java.beans.Expression.getValue(java.desktop@11.0.7-internal/Expression.java:155)
[...]

PR #3526, opened 2021-04-07
Prevent XMLDecoder from loading other than whitelisted classes

ahornace:
Creating this as a draft because this fix is only a workaround. I'm not really sure I whitelisted all the possible classes and I do not know how to achieve this more nicely. I think we can use this as an interim solution until we get the yaml config working.

Thanks to Bobby Rauch (Accenture) for reporting the issue.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationClassLoader.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationClassLoader.java
@@ -35,6 +35,7 @@ import java.beans.XMLDecoder;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


vladak:
The #2329 will be the final fix (YAML or some custom format).

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationClassLoader.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Temporary hack to prevent {@link XMLDecoder} to deserialize other than allowed classes. This tries to prevent
+ * calling of methods on {@link ProcessBuilder} or {@link Runtime} (or similar) which could be used for code execution.
+ */
+public class ConfigurationClassLoader extends ClassLoader {
+
+    private static final Set<String> allowedClasses = Set.of(
+            ArrayList.class,
```
vladak, line 49: needs `Enum.class`

vladak: with that it appears to work fine on my staging server that has semi-complex configuration (authnz stacks)

opengrok-indexer/src/test/java/org/opengrok/indexer/configuration/ConfigurationTest.java
```diff
@@ -168,4 +175,62 @@ public void serializationOrderTest() throws IOException {
...
+                    "</java>"
+    })
+    void testDeserializationOfNotWhiteListedClassThrowsError(final String exploit) {
+        assertThrows(IllegalAccessError.class, () -> Configuration.makeXMLStringAsConfiguration(exploit));
+    }
+
+    @Test
+    void testLoadingValidConfiguration() throws IOException {
+        try (var br = new BufferedReader(new InputStreamReader(ConfigurationTest.class.getClassLoader()
+                .getResourceAsStream("configuration/valid_configuration.xml")))) {
```
vladak, line 229: would be more robust to encode a `Configuration` instance and store it in the file however let's leave it like it is for now.

PR #3527, opened 2021-04-07
restrict XML decoding of History object

vladak:
Copies the approach from #3526.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryClassLoader.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.HashMap;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+/**
+ * Temporary hack to prevent {@link XMLDecoder} to deserialize other than allowed classes. This tries to prevent
+ * calling of methods on {@link ProcessBuilder} or {@link Runtime} (or similar) which could be used for code execution.
+ */
+public class HistoryClassLoader extends ClassLoader {
```
ahornace, line 38: Maybe we could create something like `WhiteListClassLoader` which would accept the whitelisted classes in the constructor to avoid the duplication? Even though it's minimal. I was thinking about that approach yesterday but forgot about the history use case. However, I think it's good to go and we can do that later as it is really simple to do.

ahornace (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryClassLoader.java
vladak: I think this is acceptable; it's a small class.

PR #3530, opened 2021-04-07
Use set for renamed files in history

### LINKED ISSUE #3529: faster lookup of renamed files in History
As mentioned in the code https://github.com/oracle/opengrok/blob/0e510f5b1c7ef0653125e789f0618ba0ef23861a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java#L119-L125 the renamed file list should use something other than `List` to perform better.

ahornace:
fixes #3529

Thanks.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
@@ -24,7 +24,11 @@
 package org.opengrok.indexer.history;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
```diff
@@ -107,24 +108,19 @@ public boolean hasFileList() {
      * tag list, {@code false} otherwise
      */
     public boolean hasTags() {
-        // TODO Use a private variable instead of for loop?
```
ahornace, line 110: So far this is only used once in `history.jsp` so I do not think memoizing or precomputing the result in the constructor is justified.

vladak:
thanks !

PR #3533, opened 2021-04-08
Add serial filtering when deserializing classes

ahornace:
I was debating whether to put some default allowed classes to the whitelist filter (e.g. `Integer.class`, `Double.class`) but since it's used only in a couple of places, I decided against it.

Thanks.


vladak:
Interesting how much serialization happens in the code. Thanks !

PR #3534, opened 2021-04-09
use automatic port allocation in LdapServerTest

vladak:
~Copies the `PortChecker` from the web app tests to the plugin tests and uses that for the `LdapServerTest`.~ This should avoid spurious failures due to the 6336 TCP port being occupied.


plugins/src/test/java/opengrok/auth/plugin/LdapServerTest.java
```diff
@@ -79,9 +81,24 @@ public void testSetGetUsername() {
...
+            }
+            if (++triesCount > MAX_PORT_TRIES) {
+                throw new RuntimeException("Could not find an available port after " +
+                        MAX_PORT_TRIES + " tries");
+            }
+        }
+
+        // Start simple TCP server on test port.
         InetAddress localhostAddr = InetAddress.getLocalHost();
         ServerSocket serverSocket = new ServerSocket(testPort, 1, localhostAddr);
```
ahornace, line 103: Why not just specify `0` for port here?
From https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/ServerSocket.html#%3Cinit%3E(int,int,java.net.InetAddress)
> A port number of 0 means that the port number is automatically allocated, typically from an ephemeral port range. This port number can then be retrieved by calling getLocalPort.

vladak: That's not a bad idea. Let me try that.

vladak: Looks like it works. This is probably better solution because there is no time-of-use-time-of-check gap.

ahornace (approved):

PR #3540, opened 2021-04-13
disable merge commits by default

vladak:
After trying to create history for linux-mainline Git repository from scratch (https://github.com/oracle/opengrok/issues/3539#issuecomment-818752840) and seeing the difference when merge commits are disabled I think it is reasonable to disable merge commits by default.

This is the difference on Java 11 VM with 16 GB heap:

linux-merge-on_vs_off

The indexer with merge commits disabled was started at 16:00 and very quickly progressed to create the file history cache while the first actually ended with OOM and did not even get to storing a single file history cache entry on disk.


vladak:
This should be revisited once #3243 is fixed.

Issue #1113, opened 2016-06-15
Move GitRepository to JGit


fzakaria:
It would be great to move as many of the external tools to pure java implementations.
a) It would simplify the testing
- JGit for instance provides JUnit test helpers that setup a test repository that gets deleted after test execution. This would remove the need to hanve `ant` scripts create and bundle test repositories.

b) speedup the execution
- No need to call out to external process

c) make setup easier
- No need to figure out the path to Git and more repositories can be by default supported

I've explored swapping GitRepository to JGit and it's definitely feasible. The tough point at the moment is that there are few tests for some of the functions.

tarzanek:
well I would always want a fallback method to not java implementations, but native ones
I have a very bad experience with jgit in jenkins, where I always have because of various bugs to switch back to native git (which I can control the version of)
But obviously your experience might vary, I'd just be cautious about moving everything to java ...
(I guess better would be to have a set of recommended external tool versions ... )

tarzanek:
regarding versions - there is a bug #970 - where basically since hg 3.4 mercurial is not working properly with opengrok ...
so it's really hard to decide to go just for one way (since both seem to have probs long term :( )

fzakaria:
Interesting, I would have thought the inprocess solutions (java/JNI implementations) the way to go.

tarzanek:
Yes, jni would be good too, better than pure java, since the original libs from OS are used ...
Jgit is jni?

fzakaria:
Looks like there are other projects that provide JNI bindings for [libgit](https://github.com/ethomson/jagged)  but nothing worthwhile.

vladak:
With the renamed file handling in Git (issue #1345) we need to make sure that whatever library is chosen supports this functionality.

vladak:
Also, we need to carefully evaluate this in terms of performance and stability. Surely we don't want to switch to library which is not being actively developed and has uncertain future.

marcelhuberfoo:
As far as I checked today for [jgit](https://github.com/eclipse/jgit), it seems to be actively maintained and developed.
jgit also supports [renamed file handling](https://github.com/eclipse/jgit/blob/f3ec7cf3f0436a79e252251a31dbc62694555897/org.eclipse.jgit/src/org/eclipse/jgit/revwalk/FollowFilter.java)
[...]

PR #3544, opened 2021-04-15
JGit history bulk

vladak:
This change replaces the Git history executors and associated parsing of `git` command output with JGit, next step for #1113. It is a trade-off: for more robust Git operations handling the performance will be slightly degraded and memory consumption will increase. For the testing data set I used the  `master` branch of the OpenSSL Git repository.

For functional testing, I generated the history cache first using the `master` branch, moved it aside, then generated the cache using this branch and compared the directory trees by checking they contain the same set of files and that the files' SHA1 digest matches. There were some differences due to the missing leading whitespace in some of the commit messages. Looks like nothing can be done about it.

I tested from scratch history cache generation with various indexer tunables to see what the performance impact will be (times in seconds if there is freestanding number, otherwise in minutes:seconds):

**version** | **renamed off + merge off** | **renamed on + merge off** | **renamed on + merge on**
------------ | ------------- | ---------- | ----------
`master` | 60 | 8:52 | 9:13
`jgit_history_bulk` | 1:09 | 12:56 | 13:17

The indexer was running with `-Xmx16g` and the options were `-s /var/opengrok/src.openssl -d /var/opengrok/data.openssl -P -H -S -G` plus read-only configuration to set the renames/merges.

Of course, the memory demands have grown as well given that now everything happens inside single JVM instance. The sort of left side of the graph are the 3 indexing variants (as in the table above, left to right) done with `master`, on the right is done with the JGit changes:

jgit-indexer

For instance, the reindex with jgit and renamed=on,merge=on begins at 14:36. The previous reindex with jgit and renamed=on,merge=off begins cca at 14:22. Mostly I ran them back to back (after clobbering the history cache which takes a second or two on my system with SSD).


opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
```diff
@@ -123,4 +123,27 @@ public boolean isRenamed(String file) {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        History that = (History) o;
+        return this.getHistoryEntries().equals(that.getHistoryEntries()) &&
```
ahornace, line 136: I'd advise using the newer comparison as it can handle theoretical NPEs even in the future: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)

You can let IDEA generate it as well if you choose Java7+ equals/hashcode option.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
```diff
@@ -123,4 +123,27 @@ public boolean isRenamed(String file) {
...
+                this.getRenamedFiles().equals(that.getRenamedFiles());
+    }
+
+    @Override
+    public int hashCode() {
+        return getHistoryEntries().hashCode() * getRenamedFiles().hashCode();
```
ahornace, line 142: I'd use: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#hash(java.lang.Object...)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
```diff
@@ -207,4 +216,44 @@ public void stripFiles() {
     public void stripTags() {
         tags = null;
     }
+
+    private boolean checkEquals(Object o1, Object o2) {
```
ahornace, line 220: Please use methods mentioned above.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#hash(java.lang.Object...)

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
```diff
@@ -207,6 +206,41 @@ public void testDetermineBranchAfterChange() throws Exception {
         }
     }
 
+    @Test
+    public void testGetHistoryInBranch() throws Exception {
```
ahornace, line 210: Just a note, with JUnit5 the methods no longer have to be public and using "package private"/default access is recommended (at least by Sonar :D ).

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
```diff
@@ -61,12 +62,10 @@
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.opengrok.indexer.condition.RepositoryInstalled.Type.GIT;
 
 /**
  * @author austvik
  */
-@EnabledForRepository(GIT)
```
ahornace, line 69: Do we still have some other places which are requiring git command or can we remove the GIT check altogether?

ahornace (commented):
Awesome job! :)

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
vladak: Yes, there are quite a few. I plan to revisit them separately. In fact, I added a item to https://github.com/oracle/opengrok/issues/1113#issuecomment-803470717

opengrok-indexer/src/test/java/org/opengrok/indexer/history/GitRepositoryTest.java
vladak: Good to know, thanks !

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -586,19 +523,129 @@ History getHistory(File file) throws HistoryException {
...
+                    }
+                } else {
+                    getFiles(repository, commit.getParent(0), commit, files, renamedFiles);
+                }
+
+                historyEntry.setFiles(files);
+                entries.add(historyEntry);
+            }
+        } catch (IOException | ForbiddenSymlinkException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
vladak: actually, I think we want to rethrow the exception to properly signal the failure to `createCache()`.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
vladak: Exactly what I needed ! Thanks !

opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
vladak: Thanks !

opengrok-indexer/src/main/java/org/opengrok/indexer/history/History.java
```diff
@@ -123,4 +123,27 @@ public boolean isRenamed(String file) {
...
+            return false;
+        }
+        History that = (History) o;
+        return Objects.equals(this.getHistoryEntries(), that.getHistoryEntries()) &&
+                Objects.equals(this.getRenamedFiles(), that.getRenamedFiles());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getHistoryEntries().hashCode(), getRenamedFiles().hashCode());
```
ahornace, line 142: You are not supposed to call `hashCode()` on the objects, the implementation should handle that :) 
```suggestion
        return Objects.hash(getHistoryEntries(), getRenamedFiles());
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
```diff
@@ -207,4 +217,36 @@ public void stripFiles() {
...
+                Objects.equals(this.getRevision(), that.getRevision()) &&
+                Objects.equals(this.getDate(), that.getDate()) &&
+                Objects.equals(this.getMessage(), that.getMessage()) &&
+                Objects.equals(this.getFiles(), that.getFiles()) &&
+                Objects.equals(this.getTags(), that.getTags());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getHashCode(getAuthor()), getHashCode(getRevision()), getHashCode(getDate()),
```
ahornace, line 249: Same here, you do not have to call `hashCode` yourself.

ahornace (approved):

PR #3545, opened 2021-04-15
LdapServerTest redux

vladak:
The LdapServerTest is still failing due to

```
Error:  Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.225 s <<< FAILURE! - in opengrok.auth.plugin.LdapServerTest
Error:  testIsReachable  Time elapsed: 0.019 s  <<< ERROR!
java.net.BindException: Cannot assign requested address (Bind failed)
	at java.base/java.net.PlainSocketImpl.socketBind(Native Method)
	at java.base/java.net.AbstractPlainSocketImpl.bind(AbstractPlainSocketImpl.java:436)
	at java.base/java.net.ServerSocket.bind(ServerSocket.java:395)
	at java.base/java.net.ServerSocket.<init>(ServerSocket.java:257)
	at opengrok.auth.plugin.LdapServerTest.testIsReachable(LdapServerTest.java:84)
```

The fix done in 476443edddbb606ccf738095d424034494b8dd71 was not sufficient it seems so reintroducing the original approach.

### DIFF
--- a/plugins/src/test/java/opengrok/auth/plugin/LdapServerTest.java
+++ b/plugins/src/test/java/opengrok/auth/plugin/LdapServerTest.java
@@ -81,7 +81,7 @@ public class LdapServerTest {
     public void testIsReachable() throws IOException, InterruptedException, URISyntaxException {
         // Start simple TCP server on test port.
         InetAddress localhostAddr = InetAddress.getLoopbackAddress();
-        try (ServerSocket serverSocket = new ServerSocket(0, 1, localhostAddr)) {
+        try (ServerSocket serverSocket = new ServerSocket(0, 1)) {
             Thread thread = new Thread(() -> {
                 try {
                     while (true) {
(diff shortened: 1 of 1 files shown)


vladak:
Seems like the problem is elsewhere.

ahornace:
> Seems like the problem is elsewhere.

Yes, I think so too. I believe 0 port should be handled by Java properly. Maybe we can omit the localhost address? I'm not sure if it's necessary. Furthermore, maybe we should also properly close the resources with try-with-resources structure.

vladak:
Looks like the combination of loopback + wildcard works.

plugins/src/test/java/opengrok/auth/plugin/LdapServerTest.java
```diff
@@ -80,49 +80,50 @@ public void testSetGetUsername() {
...
+                    e.printStackTrace();
+                }
+            });
+
+            int testPort = serverSocket.getLocalPort();
+            thread.start();
+            Socket socket = null;
+            for (int i = 0; i < 3; i++) {
+                try {
+                    socket = new Socket(localhostAddr, testPort);
```
ahornace, line 101: There is a resource leak as we are not closing the first 2 sockets and also the waits make the test flaky and execute for too long. I believe the purpose is to make it work for now and we can revisit later.

ahornace (approved):

plugins/src/test/java/opengrok/auth/plugin/LdapServerTest.java
vladak: filed #3546

PR #3551, opened 2021-04-19
Use font instead of png for logo

ahornace:
The logo was not looking good on scaled displays (e.g. MacBook Pro with Retina display).

I also removed the Java logo. I don't think it's relevant to users in which programming language is OpenGrok written. Furthermore, the hype about Java seems to be over :D

I removed the combined image and the script for combining images. I'm not sure what was the reason to do this. I know that games are using texture atlases but I have not seen that in web development. Theoretically, it might be good to save http requests but I'm not sure it's the way things should be done. Do you have any links that would support this? I just googled very quickly and did not find much. But there are only 2 images left so it does not make much sense anyway.

I also wanted to use `svg` for the favicon but Chrome does not support external resources in favicon svg (e.g. font). Maybe I could copy the shapes of the font but I'm not that skilled in graphics :D I'll try to look into it more. I'd like to avoid generating multiple png files with different sizes.

Thanks!

(tested on Safari and Chrome)

Screen Shot 2021-04-19 at 20 13 17


opengrok-web/src/main/webapp/foot.jspf
```diff
@@ -42,10 +42,12 @@ org.opengrok.indexer.web.Prefix"
     
         %>
     <div id="footer">
-<p><a href="https://oracle.github.io/opengrok/"
- title="Served by OpenGrok (<%= Info.getVersion() %> - <%= Info.getShortRevision() %>)"><span id="fti"></span></a></p>
+<p><a href="https://oracle.github.io/opengrok/" class="link-no-decoration"
+      title="Served by OpenGrok (<%= Info.getVersion() %> - <%= Info.getShortRevision() %>)">
+    <span id="served-by">served&nbsp;by&nbsp;</span><span class="logo logo-gray-color">{</span><span class="logo logo-blue-color">Open</span><span class="logo logo-gray-color">Grok</span>
```
ahornace, line 47: The spans need to be on the same line due to https://stackoverflow.com/questions/2519923/get-rid-of-spaces-between-spans -> I did not like the other solutions as they are mostly hacks.

vladak:
Nice !

PR #3556, opened 2021-04-22
add port and REST endpoint to trigger reindex

vladak:
This change allows to trigger reindex from outside of the Docker container via `GET` request to `/reindex` on certain port (5000 by default). This works only if the periodic reindex is disabled, i.e. when `SYNC_TIME_MINUTES` environment variable is set to 0.

### DIFF
--- a/docker/start.py
+++ b/docker/start.py
@@ -30,6 +30,9 @@ import threading
 import time
 from pathlib import Path
 from requests import get, ConnectionError
+from flask import Flask
+from flask_httpauth import HTTPTokenAuth
+from waitress import serve
 
 from opengrok_tools.utils.log import get_console_logger, \
     get_log_level, get_class_basename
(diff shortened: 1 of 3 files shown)


docker/start.py
```diff
@@ -230,7 +263,14 @@ def indexer_no_projects(logger, uri, config_path, sync_period,
 
     wait_for_tomcat(logger, uri)
 
+    periodic_sync = True
+    if sync_period is None or sync_period == 0:
+        periodic_sync = False
+
     while True:
+        if not periodic_sync:
```
ahornace, line 271: Should not this check be at the end of `while` loop? It seems more consistent to me or are we trying to prevent even the 1st invocation?

vladak: I have not made my mind yet whether zero period should disable the reindex altogether. This preserves current behavior.

ahornace (approved):

vladak:
Thanks @ahornace !

PR #3557, opened 2021-04-22
Use label for minisearch's "current directory" checkbox

ahornace:
The look is the same but now clicking on the text also checks the checkbox.

Thanks.


vladak:
Cool. Does this work across different browsers ?

ahornace:
> Cool. Does this work across different browsers ?

Works in Chrome, Safari and Firefox :)

PR #3561, opened 2021-04-25
Update Java dependencies

ahornace:
I've done some simple manual tests and everything seemed to work fine.

Thanks.

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -18,7 +18,7 @@ information: Portions Copyright [yyyy] [name of copyright owner]
 
 CDDL HEADER END
 
-Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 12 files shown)


plugins/pom.xml
```diff
@@ -84,11 +84,16 @@ Portions Copyright (c) 2020, Chris Fraire <cfraire@me.com>.
             <artifactId>jersey-test-framework-provider-grizzly2</artifactId>
             <version>${jersey.version}</version>
             <scope>test</scope>
+            <exclusions>
```
vladak, line 87: why this exclusion ? Does grizzly supply its own junit dependency ?

ahornace: Yes, I think so. The `JerseyTest` contains `@Before` and `@After` imports from JUnit4 and without this exclusion we could use the JUnit4 asserts. Hence, some of the changes in the test files.

PR #3562, opened 2021-04-26
allow to use read-only configuration in Docker

vladak:
This change allow to specify read-only configuration in Docker.


ahornace (approved):

vladak:
thanks @ahornace !

PR #3567, opened 2021-04-30
attempt at graceful exit in Docker

vladak:
This change is sort of low effort approach to addressing termination of the Docker container via Ctrl-C when running as `docker run`. Setting the threads with `daemon=True` allows them to be terminated (https://stackoverflow.com/a/2564282/11582827) along with the main thread. Otherwise Tomcat would exit and the threads would be still running with the root process/thread waiting in `sys.exit()`. Perhaps more proper way to address that would be to wake them up and let them finish on their own, using a global termination flag. For the REST thread that would probably mean adding a separate API endpoint solely for the purpose of terminating the Flask server (use something like https://code-examples.net/en/q/ed76ce) or muck around with `waitress` internals  (https://stackoverflow.com/a/59923433/11582827).

This is probably not 100% robust (https://exceptionshub.com/why-cant-i-handle-a-keyboardinterrupt-in-python-3.html) and will not be graceful if sync is in progress.

Lastly, this will not help with `docker stop` since it uses different signal (on Unix) to terminate the root process (https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/). For that it would be prudent to install proper signal handlers (also, `docker kill` can specify arbitrary signal number such as `SIGQUIT`).

The code is `flake8` clean.

### DIFF
--- a/docker/start.py
+++ b/docker/start.py
@@ -487,7 +487,6 @@ def main():
                        extra_indexer_options)
 
     logger.debug("Starting sync thread")
-    global sync_thread
     sync_thread = threading.Thread(target=worker_function, name="Sync thread",
                                    args=syncer_args, daemon=True)
     sync_thread.start()
@@ -500,7 +499,6 @@ def main():
                      "on port {} to '{}'".format(rest_port, token))
         expected_token = token
     logger.debug("Starting REST thread")
-    global rest_thread
     rest_thread = threading.Thread(target=rest_function, name="REST thread",
                                    args=(logger, rest_port), daemon=True)
     rest_thread.start()
--- a/docker/start.py
+++ b/docker/start.py
@@ -26,6 +26,7 @@ import logging
 import multiprocessing
 import shutil
 import subprocess
+import sys
 import tempfile
 import threading
 import time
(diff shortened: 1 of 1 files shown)


docker/start.py
```diff
@@ -207,6 +208,9 @@ def refresh_projects(logger, uri):
     Ensure each immediate source root subdirectory is a project.
     """
     webapp_projects = list_projects(logger, uri)
+    if not webapp_projects:
```
vladak, line 211: for cases like when Tomcat exits first and the thread is still running

docker/start.py
```diff
@@ -321,28 +325,29 @@ def project_syncer(logger, loglevel, uri, config_path, sync_period,
...
-        do_sync(loglevel, commands, config.get('cleanup'),
-                projects, config.get("ignore_errors"), uri,
-                numworkers, driveon=True, logger=logger, print_output=True)
-        logger.info("Sync done")
-
-        # Workaround for https://github.com/oracle/opengrok/issues/1670
-        Path(os.path.join(OPENGROK_DATA_ROOT, 'timestamp')).touch()
-
-        save_config(logger, uri, config_path)
+        if projects:
```
vladak, line 328: ditto as in `refresh_projects()`

docker/start.py
```diff
@@ -494,14 +499,25 @@ def main():
...
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except KeyboardInterrupt:
+        global tomcat_popen
+        print("Terminating Tomcat {}".format(tomcat_popen))
+        tomcat_popen.terminate()
+
+        sys.exit(1)
```
vladak, line 523: might be better to exit with 128 + signal number on Unix systems ?

ahornace:
Cool! This was driving me crazy while testing :D

PR #3572, opened 2021-05-03
publish Docker image for non release builds as well

vladak:
This change should build and push `opengrok/docker:master` for each non release build in the main repo.

### DIFF
--- a/dev/docker.sh
+++ b/dev/docker.sh
@@ -3,7 +3,7 @@
 #
 # Build and optionally push new image to Docker hub.
 #
-# When pushing, this script uses the following Travis secure variables:
+# When pushing, this script uses the following secure variables:
 #  - DOCKER_USERNAME
 #  - DOCKER_PASSWORD
 #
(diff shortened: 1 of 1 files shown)


vladak:
For testing, I already pushed `:master` to Docker hub by running the script with `DOCKER_USERNAME`, `DOCKER_PASSWORD` and `OPENGROK_REPO_SLUG` environment variables.

dev/docker.sh
```diff
@@ -84,12 +81,12 @@ if [[ -z $DOCKER_PASSWORD ]]; then
 fi
 
 # Publish the image to Docker hub.
-if [ -n "$DOCKER_PASSWORD" -a -n "$DOCKER_USERNAME" -a -n "$VERSION" ]; then
+if [ -n "$DOCKER_PASSWORD" -a -n "$DOCKER_USERNAME" -a -n "$TAGS" ]; then
```
ahornace, line 84: Isn't `-n "$TAGS"` always true?

ahornace:
Just a nitpick I noticed, maybe we could remove the `Travis` reference as that is no longer true?
```
When pushing, this script uses the following Travis secure variables:
```

ahornace (approved):

dev/docker.sh
vladak: it is, however a bit of defensive programming does not hurt in shell scripts :-)

vladak:
> Just a nitpick I noticed, maybe we could remove the `Travis` reference as that is no longer true?
> 
> ```
> When pushing, this script uses the following Travis secure variables:
> ```

True, removed.

PR #3574, opened 2021-05-04
check index and wipe out data root on mismatch

vladak:
Another piece for the OpenGrok demo. This change introduces the `CHECK_INDEX` environment variable that triggers the run of the indexer with `--checkIndex`. If this fails, the data root is wiped out. This is necessary for upgrades across major versions.


ahornace:
Looks good :)

PR #3577, opened 2021-05-05
handle SIG{TERM,INT} in Docker

vladak:
This takes care of the Ctrl+C case in `docker-compose`, at least in the basic sense. This event leads to `docker stop` which sends `SIGTERM` to PID 1 in the container (https://github.com/docker/compose/issues/7135#issuecomment-573621185).

Still, the sync and REST thread could be dealt with in a better way.

### DIFF
--- a/docker/start.py
+++ b/docker/start.py
@@ -24,6 +24,7 @@
 import os
 import logging
 import multiprocessing
+import signal
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


vladak:
> We could also provide some simple `docker-compose` file with grafana, prometheus, etc.
> 
> Similarly as in https://github.com/grafana/loki/blob/main/production/docker-compose.yaml

It's already on https://github.com/OpenGrok/opengrok-monitoring-docker

Issue #3578, opened 2021-05-06
Calling `/reindex` does not appear to trigger a reindex


dhose:
This is related to https://github.com/oracle/opengrok/issues/3535

**Describe the bug**
Running OpenGrok in a container and trying to manually trigger a reindex with the `/reindex` endpoint. I have checked that my repositories is updated on disk, but that is not reflected in the _Current version_ timestamp in OpenGroks webui. As a sidenote I don't see any increase in CPU usage on the host when calling `/reindex`.

I am running the container (from image **opengrok/docker:1.7.2**) with the following environment variables:

```
SYNC_PERIOD_MINUTES=0
NOMIRROR=true
```

**To Reproduce**

- Calling `/bin/curl http://127.0.0.1:5000/reindex` from outside of my container
- Response from container is `Reindex triggered`

Thank's!

vladak:
Just tried that with the latest image and seems to be working fine. The thing with the reindex trigger by the API endpoint is that there needs to be something listening for the event. If the reindex thread has not reached the `wait()` call, the event will be lost because the `index` function clears the event right after `set()`.

Will add a log message to signal that the thread is waiting for the event.

dhose:
Ok, I see. Does this mean I have to implement some changes on my side?

Thank's for a quick response!

vladak:
What you can try is to do the API call repeatedly. As explained above, it depends in which stage of the container runtime the call is made, i.e. you need to wait for the reindex/sync thread to reach a point where it is waiting for the reindex event.

For reference here is the `docker-compose.yml` I am using:
```yml
version: "3"

# More info at https://github.com/oracle/opengrok/docker/
services:
  opengrok:
    container_name: opengrok-test
    image: opengrok/docker:latest
    ports:
      - "8080:8080/tcp"
      - "5000:5000/tcp"
    environment:
      SYNC_PERIOD_MINUTES: '0'
      NOMIRROR: 1
    # Volumes store your data between container upgrades
    volumes:
       - '~/opengrok/src/:/opengrok/src/'  # source code
       - '~/opengrok/etc/:/opengrok/etc/'  # folder contains configuration.xml
       - '~/opengrok/data/:/opengrok/data/'  # index and other things for source code
```

vladak:
I know what is going on - the `SYNC_PERIOD_MINUTES` is the name of the environment variable that is in the `master`, 1.7.2 is still using `SYNC_TIME_MINUTES`.

vladak:
[...]

PR #3579, opened 2021-05-06
rework reindex/sync event triggers in Docker

vladak:
This change allows to the reindex/sync to be triggered based on both timeout and RESTful API call. This has some limitiations that stem from the underlying implementation, e.g. if reindex is already running and the `/reindex` API endpoint is called, the event will be lost.

Inspired by #3578.

Also added some signal handling related changes that I think prevent `watchtower` from restarting the container on image update.


ahornace:
Looks good.

PR #3586, opened 2021-05-17
Fix subrepository display on Windows

ahornace:
On Windows the repository can be misinterpreted as subrepository which can cause display similar to below:
repo_win

The problem seems to be that `Project`s are using `/` as path separator  whereas `RepositoryInfo` can have it as `\` which is causing the inequality:
http://demo.opengrok.org/xref/OpenGrok/opengrok-indexer/src/main/java/org/opengrok/indexer/history/RepositoryInfo.java?r=93915c65#192

I believe the best way would be to use `/` in `RepositoryInfo` as well but I'm not sure what else such change could affect. @vladak what do you think?

Thanks!


vladak:
> I believe the best way would be to use `/` in `RepositoryInfo` as well but I'm not sure what else such change could affect. @vladak what do you think?

I think this inconsistency was mentioned in some issue, should be fixed separately.

PR #3592, opened 2021-05-19
convert findOriginalName() to JGit

vladak:
One of the last steps in the JGit conversion (#1113). Executing the `git` command is no longer needed, at least for the indexer/webapp.


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -305,61 +284,55 @@ String findOriginalName(String fullpath, String changeset)
...
                 }
             }
         }
 
-        if (status != 0 || originalFile == null) {
-            LOGGER.log(Level.WARNING,
-                    "Failed to get original name in revision {2} for: \"{0}\" Exit code: {1}",
-                    new Object[]{fullpath, String.valueOf(status), changeset});
+        if (originalFile == null) {
+            LOGGER.log(Level.WARNING, "Failed to get original name in revision {2} for: \"{0}\"",
```
ahornace:
```suggestion
            LOGGER.log(Level.WARNING, "Failed to get original name in revision {1} for: \"{0}\"",
```

ahornace (approved):
Great! It's almost done :)

PR #3594, opened 2021-05-20
JGit related test/build cleanup

vladak:
Last piece of the changes for #1113. Cleans the tests + removes the need for the `git` command during build. The javadoc workflow still needs it (installed by default) however that is separate from the regular build.

The Docker image still needs the `git` command to perform repository synchronization (until #2852 is implemented).


ahornace:
Awesome!

Issue #3542, opened 2021-04-14
parallelize doFileHistory() for regular files in FileHistoryCache#store()


vladak:
Playing with a proof of concept fix for #3243 I realized that regular files could be parallelized in the same way as renamed files, i.e. create the directories first and then use thread pool to perform doFileHistory() for each file. This will cost more memory - same assumption applies as for the proof of concept fix - history of individual files is reasonably big.

vladak:
Also, the current way how the directories are created: https://github.com/oracle/opengrok/blob/610d9085dcb52fc16b3f50d16e64cc03a62b7ea1/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java#L537-L555
is sub-optimal: it should really assemble the directories to be created in a set first and then go through the set and call `mkdirs()` for each item in the set. Like this is done now it calls `isDirectory()` more than is needed. Of course, more intelligent algorithm can be used only to call `mkdirs()` on the longest paths first.

vladak:
One observation made when working on the fix for #3243 - when creating history cache for single repository with large history (e.g. Linux) the CPU is only lightly utilized.

Also, this change is almost necessary given that the XML serialization filtering seems to impose additional non trivial workload.

PR #3601, opened 2021-05-21
Mercurial history per partes

vladak:
Converts Mercurial history handling to the per partes style done in #3589 for Git.

In Mercurial I happen to be using (5.3.1) the revision filtering for file history log is still broken, i.e. `hg log -r X:Y --follow <file>` does not display expected list of changesets. This is only relevant to renamed file handling. I strip the unwanted `HistoryEntry` objects as a workaround. Even if it worked, it would be probably unusable - thinking the revision filtering for files works, I tried it first and it resulted in some of the `hg` processes to consume over 12 GB RSS and the system got into unhappy place (heavy swapping, the OOM killer striking).

The history cache was much slower to generate with the repository I tried this on (OpenSolaris ON gate with some 16k changesets, the majority of files added in the first changeset), like 4 minutes vs. 10 minutes (renamed handling on, 8 GB heap), however I think this is partly due to the fact that the history cache for renamed files is generated in parallel (there are some 2800 renamed files there) and for the regular files still sequentially (#3542).

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
@@ -492,6 +492,8 @@ class FileHistoryCache implements HistoryCache {
          * hash map entry for the file) in a file. Skip renamed files
          * which will be handled separately below.
          */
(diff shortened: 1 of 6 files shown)


ahornace:
Looks good! Awesome job! :)

PR #3602, opened 2021-05-23
Fix @EnabledForRepository usages in webapp

ahornace:
test-jar is not being propagated with dependencies which caused `NoClassDefFoundError` when `@EnabledForRepository` was used in webapp. This consequently caused the tests to be silently ignored. There are 3 failing tests in `PageConfigTest` caused by the recent jgit changes. 

I've tried to fix the tests by doing something like this (as jgit does not seem to have any nice API for this scenario) which fixed the tests but seems to add a lot of unnecessary work and tests for renamed files started to fail:
```java
SortedSet<String> files = new TreeSet<>();
SortedSet<String> renamedFilesHolder = new TreeSet<>();
getFilesForCommit(renamedFilesHolder, files, commit, repository);

boolean inFiles = files.stream().anyMatch(f -> f.startsWith(relativePath));
if (!inFiles) {
    continue;
}
if (isDirectory) {
    renamedFiles.addAll(renamedFilesHolder);
    historyEntry.setFiles(files);
}
``` 

@vladak you have far more experience with jgit, do you think it can be solved any way without getting the files for every commit?

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/condition/RepositoryInstalled.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/condition/RepositoryInstalled.java
@@ -22,8 +22,6 @@
  */
 package org.opengrok.indexer.condition;
 
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
 import org.opengrok.indexer.history.BazaarRepository;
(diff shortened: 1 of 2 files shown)


vladak:
Thanks for noticing this ! Will take a look at the failing tests.

PR #3604, opened 2021-05-24
fix path filtering for Git

vladak:
As noticed in #3602, path filtering for Git is broken. The tree filter needs to be used in different way as noted on https://stackoverflow.com/a/12697369/11582827


ahornace:
Cool! I was expecting a fix like this. Your stackoverflow search skill is better than mine :D

PR #3607, opened 2021-05-24
Remove stale links on gh-pages

ahornace:
Just a couple of stale links I noticed while searching for javadoc.

Thanks!


vladak:
thanks !

PR #3609, opened 2021-05-24
Refactored IndexerTest to use conditional test execution annotation

eas5:
Problem:
Conditions within the test method may produce a false test outcome. Once a condition is not met, and the code block inside the condition is not executed, a false 'passed' outcome can be produced without the test ever being executed.

Solution:
Instead of using a conditional to control the test execution, we propose using the Conditional Test Execution Annotation @EnabledIf, which results in a test being aborted before a previous test. Conditional execution is typically used whenever it does not make sense to continue executing a given test method.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexerTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexerTest.java
@@ -57,6 +57,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIf;
 import org.opengrok.indexer.analysis.AnalyzerFactory;
 import org.opengrok.indexer.analysis.AnalyzerGuru;
 import org.opengrok.indexer.condition.EnabledForRepository;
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexerTest.java
```diff
@@ -451,39 +452,32 @@ public void testIncrementalIndexAddRemoveFile() throws Exception {
...
+        idb.addIndexChangedListener(listener);
+        System.out.println("Trying to index a special file - FIFO in this case.");
+        idb.update();
+        assertEquals(0, listener.files.size());
+    }
 
-        } else {
-            System.out.println("Skipping test for bug 11896. Could not find a mkfifo in path.");
-        }
+    public boolean mkinfoInPath() {
```
ahornace, line 479: This should be named `mkfifoInPath`

ahornace, line 479: No need for `public` modifier, can be even `private` it seems

ahornace (changes requested):

opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexerTest.java
ahornace: https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-conditional-execution-custom has an example with package private access so I think we should go with that :)

eas5:
Thank you for your revision @ahornace. I have made the suggested changes :)

ahornace:
> Thank you for your revision @ahornace. I have made the suggested changes :)

Cool! Thanks for the PR :) I'll merge as soon as the checks pass

PR #3610, opened 2021-05-24
 cleanup of renamed file handling in FileHistoryCache

vladak:
This set of changes should save same memory and make the code in `FileHistoryCache#store()` more readable.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialHistoryParser.java
@@ -170,7 +170,7 @@ class MercurialHistoryParser implements Executor.StreamHandler {
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialHistoryParser.java
```diff
@@ -185,8 +185,8 @@ public void processStream(InputStream input) throws IOException {
                      String[] move = part.split(" \\(");
                      File f = new File(mydir + move[0]);
                      if (!move[0].isEmpty() && f.exists() &&
-                         !renamedFiles.contains(move[0])) {
-                             renamedFiles.add(move[0]);
+                         !renamedFiles.contains(repository.getDirectoryNameRelative() + File.separator + move[0])) {
```
ahornace, line 188: What about changing the `renamedFiles` to `Set`? In that case we would not need the `contains` check and it should be also faster.

ahornace:
Nice work! :)

PR #3612, opened 2021-05-25
Leverage util methods from apache commons and java 11

ahornace:
Small cleanup of either unused methods or methods that could be leveraged from libraries.

Thanks!


vladak:
Nice cleanup, thanks !

Issue #3500, opened 2021-03-25
Docker sync stops project index if the repository contains changes that prevent from pull


ahornace:
I think we should go with the best effort and still index the project.

Excerpt from the log:
```
Adding project opengrok
Sync starting
command 'opengrok-mirror -I -U http://localhost:8080/ opengrok' for project opengrok failed with code 1, breaking
processing of project 'opengrok' failed
  failed commands:
    'opengrok-mirror -I -U http://localhost:8080/ opengrok': 1
      Project opengrok has not been indexed yet
      Synchronizing repository /opengrok/src/opengrok
      failed to perform command: command ['/usr/bin/git', 'pull', '--ff-only'] in directory /opengrok/src/opengrok exited with 1
      output of '['/usr/bin/git', 'pull', '--ff-only']':
      From https://github.com/oracle/opengrok
         9bb1a6638d..cd560f8676  master     -> origin/master
         74646ce8cb..51af015f08  gh-pages   -> origin/gh-pages
       * [new tag]               1.6.0      -> 1.6.0
       * [new tag]               1.6.1      -> 1.6.1
       * [new tag]               1.6.2      -> 1.6.2
       * [new tag]               1.6.3      -> 1.6.3
       * [new tag]               1.6.4      -> 1.6.4
      error: Your local changes to the following files would be overwritten by merge:
      	README.md
      Please commit your changes or stash them before you merge.
      Aborting
      Updating 893629cc30..cd560f8676
      failed to synchronize repository /opengrok/src/opengrok


Output from 'opengrok-mirror -I -U http://localhost:8080/ opengrok':
Project opengrok has not been indexed yet
Synchronizing repository /opengrok/src/opengrok
failed to perform command: command ['/usr/bin/git', 'pull', '--ff-only'] in directory /opengrok/src/opengrok exited with 1
output of '['/usr/bin/git', 'pull', '--ff-only']':
From https://github.com/oracle/opengrok
   9bb1a6638d..cd560f8676  master     -> origin/master
   74646ce8cb..51af015f08  gh-pages   -> origin/gh-pages
 * [new tag]               1.6.0      -> 1.6.0
 * [new tag]               1.6.1      -> 1.6.1
 * [new tag]               1.6.2      -> 1.6.2
 * [new tag]               1.6.3      -> 1.6.3
 * [new tag]               1.6.4      -> 1.6.4
error: Your local changes to the following files would be overwritten by merge:
	README.md
Please commit your changes or stash them before you merge.
[...]

PR #3614, opened 2021-05-26
incoming check override

vladak:
This change allows to override the incoming check in `opengrok-mirror`. It also adds mirror configuration to Docker.

Sort of paves the way for #3500.


ahornace:
Looks good

vladak:
Thanks. The test cases are insufficient, will add more later I hope.

vladak:
updated the documentation: https://github.com/oracle/opengrok/wiki/Repository-synchronization#overriding-incoming-check

PR #3619, opened 2021-06-03
deal with read_config() for empty YAML file properly

vladak:
This fixes a problem in the recent PR #3614 which assumed that `read_config()` returns empty dictionary on a valid yet empty YAML config however `None` is returned.

### DIFF
--- a/tools/src/test/python/test_readconfig.py
+++ b/tools/src/test/python/test_readconfig.py
@@ -22,7 +22,6 @@
 #
 # Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
 #
-import pytest
 import tempfile
 from mockito import mock
 import logging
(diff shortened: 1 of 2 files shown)


ahornace (approved):

tools/src/test/python/test_readconfig.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import tempfile
+from mockito import mock
+import logging
+
+from opengrok_tools.utils.readconfig import read_config
+
+
+def test_read_config_empty_yaml():
+    tmpf = tempfile.NamedTemporaryFile(mode='w+t', delete=False)
```
vladak, line 34: cannot use context manager because of Windows does not allow the file to be opened twice.

PR #3621, opened 2021-06-04
Fix "on the fly" xref generation for empty revisions

### LINKED ISSUE #3620: "Error Reading File" page while updating
**Describe the bug**
(Opengrok V1.7.2) While downloading new files and reindexing, the Opengrok webpage displays "Error reading file" and the webserver logs:
```
org.opengrok.web.PageConfig.checkFileInner file '/var/opengrok/src/project/file.c' is newer than '/var/opengrok/data/xref/project/file.c.gz'
```

Is there a way for opengrok to just serve the old page, or an unindexed version of the update? Right now my reindex cycles are pretty long, this means updated files are unavailable for any hour or two when the reindexing job goes off.

ahornace:
fixes #3620

I've reproduced the error by removing the xref file. The implementation tried to get the file contents from history with empty revision which then produced an error from jgit.

At first I wanted to use `PageConfig#isLatestRevision` method but we map `rev` to empty string whereas `isLatestRevision` can use a `null` value and it did not work properly.

Thanks!

### DIFF
--- a/opengrok-web/src/main/webapp/xref.jspf
+++ b/opengrok-web/src/main/webapp/xref.jspf
@@ -56,7 +56,7 @@ org.opengrok.indexer.web.QueryParameters"
         InputStream in = null;
         File tempf = null;
         try {
-            if (rev.equals(DUMMY_REVISION)) {
+            if (rev.isBlank() || rev.equals(DUMMY_REVISION)) {
                 in = new BufferedInputStream(new FileInputStream(resourceFile));
             } else {
                 tempf = File.createTempFile("ogtags", basename);


vladak:
thanks @ahornace !

liamwarfield:
Does opengrok have a donation link?

ahornace:
> Does opengrok have a donation/buy the dev a beer link?

We still don't know if it's really fixed :D joking aside, we really appreciate the thought :)

Me and @vladak are Oracle employees and even though OpenGrok is only a side project for us, it still would feel wrong and probably against company policy to accept donations.

Just thanks is more than enough for us :)

vladak:
> Just thanks is more than enough for us :)

Saying thanks is actually pretty important I think, esp. for a project where the contributor/user ratio is very low. Coincidentally, https://news.ycombinator.com/item?id=27420554 is making rounds today and it contains some interesting thoughts.

Issue #3622, opened 2021-06-04
Error viewing any file in a specific project with Docker + version 1.7.7/1.7.8


jwehler:
**Describe the bug**
Running OG 1.7.7 in Docker (and same in 1.7.8 verified just now), one of my git repos under a project fails to render any file that I click on with the following dump:
```
There was an error!
Walk failure.

org.eclipse.jgit.errors.RevWalkException: Walk failure.
	at org.eclipse.jgit.revwalk.RevWalk.nextForIterator(RevWalk.java:1380)
	at org.eclipse.jgit.revwalk.RevWalk.iterator(RevWalk.java:1402)
	at org.opengrok.indexer.history.GitRepository.getHistory(GitRepository.java:520)
	at org.opengrok.indexer.history.GitRepository.getHistory(GitRepository.java:461)
	at org.opengrok.indexer.history.GitRepository.getHistory(GitRepository.java:456)
	at org.opengrok.indexer.history.FileHistoryCache.get(FileHistoryCache.java:608)
	at org.opengrok.indexer.history.HistoryGuru.getHistory(HistoryGuru.java:264)
	at org.opengrok.web.PageConfig.getLastRevFromHistory(PageConfig.java:1296)
	at org.opengrok.web.PageConfig.getLatestRevision(PageConfig.java:1288)
	at org.apache.jsp.list_jsp._jspService(list_jsp.java:286)
	at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:71)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:770)
	at org.apache.jasper.servlet.JspServletWrapper.service(JspServletWrapper.java:467)
	at org.apache.jasper.servlet.JspServlet.serviceJspFile(JspServlet.java:379)
	at org.apache.jasper.servlet.JspServlet.service(JspServlet.java:327)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:770)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:223)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:158)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:185)
[...]

PR #3623, opened 2021-06-07
PageConfig#getLastRevFromHistory() should work with history efficiently

vladak:
When looking at #3622 I realized that `PageConfig#getLastRevFromHistory()` retrieves complete history for given file only to throw most of it away. This change fixes that for Git. This is covered by tests in `PageConfigTest`.

The `ui` parameter is ignored in `GitRepository#getLastHistoryEntry()`, assuming this sort of operation will be quick.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -497,6 +497,11 @@ public class GitRepository extends RepositoryWithPerPartesHistory {
 
     public History getHistory(File file, String sinceRevision, String tillRevision,
                               Integer numCommits) throws HistoryException {
+
+        if (numCommits == null || numCommits <= 0) {
+            return null;
+        }
+
         final List<HistoryEntry> entries = new ArrayList<>();
         final Set<String> renamedFiles = new HashSet<>();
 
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -483,7 +484,19 @@ public void accept(String sinceRevision, Consumer<String> visitor) throws Histor
...
+        History hist = getHistory(file, null, null, 1);
+        return getLastHistoryEntry(hist);
+    }
+
     public History getHistory(File file, String sinceRevision, String tillRevision) throws HistoryException {
+        return getHistory(file, sinceRevision, tillRevision, null);
+    }
+
+    public History getHistory(File file, String sinceRevision, String tillRevision,
+                              Integer numCommits) throws HistoryException {
```
ahornace, line 499: Maybe add check to the start of the method that if `numCommits` is `null` or `<= 0`, then return `null` or empty? Now it seems that it would return 1 entry even when 0 would be passed.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
```diff
@@ -111,6 +111,42 @@
...
+     * @param file file
+     * @return last history entry or null
+     * @throws HistoryException on error
+     */
+    public HistoryEntry getLastHistoryEntry(File file, boolean ui) throws HistoryException {
+        History hist;
+        try {
+            hist = HistoryGuru.getInstance().getHistory(file, false, ui);
+        } catch (HistoryException ex) {
+            return null;
```
ahornace, line 144: Should not we log something?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
vladak: good catch, yes.

vladak:
added changes for Mercurial while I was at it.

ahornace (approved):
Looks good!

vladak:
thanks @ahornace !

PR #3626, opened 2021-06-07
use try-with-resources on the Reader

vladak:
should fix https://lgtm.com/projects/g/oracle/opengrok/snapshot/36022920d6d1efa52fd40b81a19b27d95c0ea3b3/files/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java?sort=name&dir=ASC&mode=heatmap#x503edddf86372eff:1

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
@@ -1196,31 +1196,31 @@ public final class Util {
      * {@link IOException}s get caught and logged, but not re-thrown.
      * @param out dump destination
      * @param file file to dump
-     * @param compressed if {@code true} the denoted file is assumed to be
-     * gzipped
+     * @param compressed if {@code true} the denoted file is assumed to be gzipped
      * @param contextPath an optional override of "/source/" as the context path
      * @return {@code true} on success (everything read and written)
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -1196,31 +1196,31 @@ public static boolean dump(Writer out, File file, boolean compressed) {
...
          * For backward compatibility, read the OpenGrok-produced document
          * using the system default charset.
          */
-        try (InputStream iss = new BufferedInputStream(
-                new FileInputStream(file))) {
-            Reader in = compressed ? new InputStreamReader(new GZIPInputStream(
-                iss)) : new InputStreamReader(iss);
-            dumpXref(out, in, contextPath);
-            return true;
+        try (InputStream iss = new BufferedInputStream(new FileInputStream(file))) {
```
ahornace, line 1214: You can actually combine multiple resources in try-with-resources and then they will be closed in the reverse order, e.g.:
```java
try (var iss = ...; var in = ...(iss)) {
}
```
This should help with indentation and might be a little clearer.

vladak: yep, was too lazy to do it.

ahornace (approved):

PR #3629, opened 2021-06-08
deal with symlinks when sorting directory entries

### LINKED ISSUE #3628: Error Comparison method violates its general contract!
Hi team

On OG 1.6.6 / Tomcat 10.0.4 / Java JDK 11.0.9.
I have a project PROJ with several Git repositories. When browsing xref paths, for some of the repos, the root access generates an error:
image

Examples of URL accesses:

- https://.../xref/PROJ is OK
- https://.../xref/PROJ/repo1 is OK
- https://.../xref/PROJ/repo2 is **KO**
- https://.../xref/PROJ/repo2/file1 is OK

Nothing special found into the indexing logfiles.

Any idea?

vladak:
This should address #3628.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
@@ -477,9 +477,9 @@ public final class PageConfig {
         return dirFileList;
     }
 
-    List<String> getSortedFiles(File[] files) {
+    private Comparator<File> getFileComparator() {
         if (getEnv().getListDirsFirst()) {
-            Arrays.sort(files, (f1, f2) -> {
+            return (f1, f2) -> {
                 if (f1.isDirectory() && !f2.isDirectory()) {
                     return -1;
                 } else if (!f1.isDirectory() && f2.isDirectory()) {
@@ -487,12 +487,14 @@ public final class PageConfig {
                 } else {
                     return f1.getName().compareTo(f2.getName());
                 }
-            });
+            };
         } else {
-            Arrays.sort(files, Comparator.comparing(File::getName));
-
+            return Comparator.comparing(File::getName);
         }
-        return Arrays.stream(files).map(File::getName).collect(Collectors.toList());
+    }
+
+    List<String> getSortedFiles(File[] files) {
+        return Arrays.stream(files).sorted(getFileComparator()).map(File::getName).collect(Collectors.toList());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
```diff
@@ -446,28 +446,11 @@ public String canProcess() {
             if (isDir() && getResourcePath().length() > 1) {
                 files = getResourceFile().listFiles();
             }
+
             if (files == null) {
                 dirFileList = Collections.emptyList();
             } else {
-                List<String> listOfFiles;
-                if (env.getListDirsFirst()) {
-                    Arrays.sort(files, (f1, f2) -> {
```
ahornace, line 454: Wouldn't it make more sense to fix this comparator? E.g. something like this:
```java
if (f1.isDirectory() && !f2.isDirectory()) {
    return -1;
} else if (!f1.isDirectory() && f2.isDirectory()) {
    return 1;
} else {
    return f1.getName().compareTo(f2.getName());
}
```

I think file can be neither `isFile()` and `isDirectory()` as mentioned in the issue with the symbolic link which can cause inconsistency in the current implementation - even symmetry is not upheld.

vladak: Good point ! I assume we want to sort the links with files.

opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
```diff
@@ -494,6 +477,24 @@ public String canProcess() {
...
+                    return 1;
+                } else {
+                    return f1.getName().compareTo(f2.getName());
+                }
+            });
+        } else {
+            Arrays.sort(files, Comparator.comparing(File::getName));
+
+        }
+        return Arrays.stream(files).map(File::getName).collect(Collectors.toList());
```
ahornace, line 495: What about going one step further? In this stream we could use something like `sorted(getFileComparator())`.

vladak: nice idea !

opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
```diff
@@ -494,6 +477,26 @@ public String canProcess() {
...
+                } else {
+                    return f1.getName().compareTo(f2.getName());
+                }
+            };
+        } else {
+            return Comparator.comparing(File::getName);
+        }
+    }
+
+    List<String> getSortedFiles(File[] files) {
```
ahornace, line 496: I think this is the good candidate for the `@VisibleForTesting` annotation we talked about :) 

https://javadoc.io/doc/org.jetbrains/annotations/latest/org/jetbrains/annotations/VisibleForTesting.html

ahornace (approved):
Good work! :)

PR #3633, opened 2021-06-11
Use Google Closure JS compiler

### LINKED ISSUE #3627: yuicompressor should be replaced with different js/css minifier
yuicompressor is not maintained anymore and does not support newer ES6 features: https://github.com/yui/yuicompressor/issues/262 

The build fails when using even simple things like `let`.

I've tried to use https://github.com/blutorange/closure-compiler-maven-plugin but the produced minified js did not seem to work. I did not find any up-to-date maven plugin for uglifyJS.

ahornace:
fixes #3627

Generated js is in `strict` mode. The only problem was the following code which was missing `p` declaration:
```js
p = inner.getLinesParts(urlPart);
```
So I rewrote `var` usages to newer `let`/`const` which have proper scope.

Thanks! :)

Tested the basic functionality (search, diffs, annotate, nav, scope, suggester) in Chrome and did not see any warnings/errors.

### DIFF
--- a/opengrok-web/pom.xml
+++ b/opengrok-web/pom.xml
@@ -257,13 +257,39 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
                 </configuration>
             </plugin>
             <plugin>
+                <groupId>com.github.blutorange</groupId>
+                <artifactId>closure-compiler-maven-plugin</artifactId>
+                <version>2.21.0</version>
+                <configuration>
+                    <baseSourceDir>${project.basedir}/src/main/webapp</baseSourceDir>
+                    <baseTargetDir>${project.build.directory}/source</baseTargetDir>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>default-minify</id>
+                        <configuration>
+                            <encoding>UTF-8</encoding>
+                            <sourceDir>js</sourceDir>
+                            <targetDir>js</targetDir>
+                            <includes>**/*.js</includes>
+                            <excludes>**/*.min.js</excludes>
+                            <skipMerge>true</skipMerge>
+                            <closureLanguageOut>ECMASCRIPT5</closureLanguageOut>
+                        </configuration>
+                        <goals>
+                            <goal>minify</goal>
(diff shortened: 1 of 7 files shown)


opengrok-web/pom.xml
```diff
@@ -256,21 +256,48 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
...
+            </plugin>
             <plugin>
                 <groupId>net.alchim31.maven</groupId>
                 <artifactId>yuicompressor-maven-plugin</artifactId>
                 <version>1.5.1</version>
                 <executions>
                     <execution>
                         <goals>
-                            <goal>jslint</goal>
                             <goal>compress</goal>
```
vladak, line 293: this is still needed for CSS compression ?

opengrok-web/pom.xml
```diff
@@ -256,21 +256,48 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
...
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>net.alchim31.maven</groupId>
                 <artifactId>yuicompressor-maven-plugin</artifactId>
                 <version>1.5.1</version>
                 <executions>
                     <execution>
                         <goals>
-                            <goal>jslint</goal>
```
vladak, line 266: does the closure compressor provide JS lint functionality or are we going to rely on the Github's checks ?

opengrok-web/pom.xml
ahornace: Yes, I left it there for that reason, we might want to investigate https://github.com/cssnano/cssnano or https://github.com/css/csso

opengrok-web/pom.xml
ahornace: I've tried to set https://blutorange.github.io/closure-compiler-maven-plugin/minify-mojo.html#closureWarningLevel to `VERBOSE` but it failed the build and I did not immediately see an option which would set treating errors as warnings. It would make sense to try to actually fix the errors. I'll try to look into it.

PR #3641, opened 2021-06-18
reintroduce setTags() in HistoryEntry

vladak:
This is another follow up change to #3636. The removal of tags in `HistoryEntry` does not cause a functional problem with old history cache however produces warning messages when deserializing these objects as visible in  `FileHistoryCacheTest#testReadCacheValid()`:
```
java.lang.NoSuchMethodException: <unbound>=HistoryEntry.setTags("1.6.4, 1.6.3, 1.6.2, 1.6.1, 1.6.0");
Continuing ...
...
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
@@ -181,7 +181,9 @@ public class HistoryEntry implements Serializable {
     /**
      * The method is kept only for backward compatibility to avoid warnings when deserializing objects
      * from the previous format. The tags were moved to the {@link History} class.
+     * Will be removed sometime after the OpenGrok 1.8.0 version.
      */
+    @Deprecated
     public void setTags(String tags) {
     }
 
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
```diff
@@ -178,6 +178,13 @@ public void setFiles(SortedSet<String> files) {
         this.files = files;
     }
 
+    /**
+     * The method is kept only for backward compatibility to avoid warnings when deserializing objects
+     * from the previous format. The tags were moved to the {@link History} class.
+     */
+    public void setTags(String tags) {
```
ahornace, line 185: Maybe use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Deprecated.html with `since` and explicitly mention that it should be removed in the next (1.8.*) minor version?

ahornace (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryEntry.java
vladak: I suspected there is some better way, thanks.

PR #3642, opened 2021-06-19
Add user-specific settings page

ahornace:
Screen Shot 2021-06-19 at 12 31 45

This is an attempt to introduce user settings page which would allow every user to slightly modify their OpenGrok experience in a nicer way (i.e. not using cookies). It uses local storage to store the values which should be persisted for a long time unless user explicitly clears them.

This change just introduces suggester enabled/disabled checkbox and is meant to be a founding stone for more things like:
- suggester configuration (show projects, speed time, number of results, min chars, ...)
- font type or size
- line spacing
- color themes
- default project
- ...

https://github.com/oracle/opengrok/wiki/Cookies will need to be modified.

Tested in Chrome and Safari.

Thanks!

### DIFF
--- a/opengrok-web/src/main/webapp/js/utils-0.0.41.js
+++ b/opengrok-web/src/main/webapp/js/utils-0.0.41.js
@@ -2426,10 +2426,19 @@ function resetAllSettings() {
  * @param el settings {@code input} element
  */
 function setDefaultSettingsValue(el) {
+    setSettingsInputValue(el, el.dataset.defaultValue);
+}
+
+/**
+ * Sets the settings {@code input} element value to the provided one.
+ * @param el element to change
+ * @param value value to set
+ */
+function setSettingsInputValue(el, value) {
     if (el.type === 'checkbox') {
-        el.checked = el.dataset.defaultValue === el.dataset.checkedValue;
+        el.checked = el.dataset.checkedValue === value;
     } else {
-        el.value = el.dataset.defaultValue;
+        el.value = value;
     }
 }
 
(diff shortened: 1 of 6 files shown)


opengrok-web/src/main/webapp/js/utils-0.0.41.js
```diff
@@ -1700,8 +1700,7 @@ function pageReadyMast() {
 }
 
 function domReadyMenu(minisearch) {
-    if (getCookie('OpenGrokSuggester.enabled') === 'false') {
-        console.log('Suggester disabled by a cookie');
```
vladak, line 1704: do we want to retain some sort of `console.log()` ?

opengrok-web/src/main/webapp/js/utils-0.0.41.js
```diff
@@ -2378,22 +2377,76 @@ function escapeHtml(string) { // taken from https://stackoverflow.com/questions/
...
+}
+
+/**
+ * Initializes all settings elements with {@code local-setting} class.
+ */
+function initSettings() {
+    for (const el of document.getElementsByClassName('local-setting')) {
+        const value = getSettingsValue(el.name);
+        if (value) {
+            if (el.type === 'checkbox') {
```
vladak, line 2443: could be refactored w.r.t. setDefaultSettingsValue() ?

vladak (approved):

opengrok-web/src/main/webapp/js/utils-0.0.41.js
ahornace: I think it's a good javascript practice to remove `console.log()` calls from production code. I quickly checked if google closure could not be configured to do that but it does not seem so. Although this hack could be used: https://stackoverflow.com/questions/2934509/exclude-debug-javascript-code-during-minification#answer-5969752

But I can add it back if you think the logs are beneficial as I do not have a strong opinion about it.

vladak: I am fine with whatever the current best practice is.

PR #3647, opened 2021-06-23
bump history cache chunk counts and make it tunable

vladak:
This change increases the default counts of changesets in history cache chunks. In general this is hard to get right because this is inherently dependent on the input data (i.e. the resitories being indexed) and because of nested parallelism - both the history cache for repositories and individual files for each repository are processed in parallel. Also, by default the thread pool sizes for these parallelisms depend on the number of CPUs; the more CPUs, the more memory will be consumed. (maybe there is a way to flatten this using similar concept as `IndexerParallelizer` for the 2nd phase of indexing ? That would require some pre-processing such as creating the history cache directories.)

Lastly, this depends on the heap size available. The baseline I used is 8 GB however some users might want to push this down at the cost of slower history cache creation/refresh, hence the new `historyChunkCount` tunable (that overrides all repositories - I did not want to add per project tunable as this would be too big hammer and introducing per repository type tunable would take longer as the concept is not there yet). So, this is sort of a band aid until someone comes with some cool formula to compute the chunk size that would be based on the number of CPUs, number of indexed projects and available heap (still making assumptions about the input data) and/or flatten the history cache parallelism.

For the measurement I used AMD based OCI shape with 8 OCPUs (16 processors reported in `/proc/cpuinfo`) and 48 GB RAM, setting the indexer to use 8 GB heap:
```
/usr/bin/time -o /dev/stdout --format '%e' java -Xmx8g -classpath 'distribution/target/dist/*' \
    org.opengrok.indexer.index.Indexer -s $HOME/opengrok/src -d $HOME/opengrok/data -c /usr/local/bin/ctags \
    -H -S -P -c /usr/bin/ctags -R ~/opengrok/etc/ro.xml
```
The read-only configuration contained just StatsD export variables. The source root contained fresh checkout of the Linux Git repository. Since this is using otherwise default configuration, merge commits and renamed files are disabled.

The times in seconds to create history cache (I modified `Indexer.java` to exit after `prepareIndexer()`):

```
8192	1814.90
16384	1456.70
32768	1086.02
65536	813.16
131072	633.39
262144	518.22
524288	445.40
```

This is how it looked like:

linux-chunks

The blue annotation triangles at the bottom of the JVM graph correspond to the chunk sizes above.

Obviously, this is just one repository. Indexing multiple Linux repositories at once would probably need more heap or smaller chunk size or both. That's why I decided to leave the defaults low.

I left the Mercurial chunk count higher (the number being arbitrary) because the processing is done by external command (as opposed to JGit), assuming it consumes less heap.

The new tunable and the 8 GB heap assumption should be documented in respective wiki pages after this goes in.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
@@ -25,18 +25,22 @@ package org.opengrok.indexer.history;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.Mockito;
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.util.TestRepository;
 
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Random;
 import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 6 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
```diff
@@ -74,6 +78,16 @@ void testInvalidMaxCount(int maxCount) {
         assertThrows(RuntimeException.class, () -> new BoundaryChangesets(gitSpyRepository));
     }
 
+    @Test
+    void testMaxCountConfiguration() {
+        int maxCount = 42 * new Random().nextInt(100) + 1;
```
ahornace, line 83: Why random? I think it's better to avoid random stuff in tests.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
```diff
@@ -74,6 +78,16 @@ void testInvalidMaxCount(int maxCount) {
...
+        assertNotEquals(0, maxCount);
+        RuntimeEnvironment.getInstance().setHistoryChunkCount(maxCount);
+        GitRepository gitSpyRepository = Mockito.spy(gitRepository);
```
ahornace, line 86: Why use of `spy` here? Won't it just delegate all the calls to `gitRepository`?

ahornace (approved):

opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
vladak: good catch, fixed.

opengrok-indexer/src/test/java/org/opengrok/indexer/history/BoundaryChangesetsTest.java
vladak: just wanted to have non default value. Agreed that randomness in tests should be avoided, unless it is seeded to make it reproducible.

PR #3652, opened 2021-06-24
log the actual HTTP error from requests

vladak:
As visible in #3651 the API errors lack the information about what went wrong. This was bugging me for quite some time so let's fix that. Works like this now:
```
>>> import logging
>>> logger = logging.getLogger()
>>> from opengrok_tools.utils.opengrok import get_repos
>>> get_repos(logger, 'fooa', 'http://localhost:8080/surce/')
could not get repositories for project 'fooa': 404 Client Error:  for url: http://localhost:8080/surce/api/v1/projects/fooa/repositories
>>> 
```


ahornace:
Looks good.

PR #3654, opened 2021-06-26
Fix suggester display of hiragana terms

### LINKED ISSUE #3603: Suggester and search do not work properly for words written in hiragana (and probably other CJK alphabets)
While testing https://github.com/oracle/opengrok/pull/1391 I've found out that OpenGrok has some issues with hiragana letters.

I've added three words in hiragana to the index:
```
あめ
ありがと
あした
```

Typing `あ` does not make suggester to suggest anything meaningful. Searching for the whole word splits it into multiple characters (e.g. `+full:あ +full:し +full:た`) and does not provide any results. Adding `*` seems to be a workaround for search and suggester but distorts the suggester output:
Screen Shot 2021-05-23 at 18 42 43

ahornace:
fixes #3603

Now it seems that hiragana terms work properly. I'm not sure if I made a mistake during my testing of #3603 or the new Lucene version made a difference.

This is just a quick hack to fix the suggester display. The problem was caused by hiragana characters being slightly taller than latin letters.

Screen Shot 2021-06-26 at 16 40 26

Thanks!


vladak:
Thanks !

PR #3663, opened 2021-07-15
add option to turn off per partes history cache

vladak:
Introduces the `historyCachePerPartesEnabled` tunable to disable per partes history cache creation. This is handy for those who have repositories with sizeable history (either in terms of number of changesets or changed files or both; thinking hundreds thousands of changesets) and can throw lots of memory on the JVM (think tens of gigabytes) for the sake of speeding the indexing.


vladak:
will need to add the tunable to https://github.com/oracle/opengrok/wiki/Indexer-configuration

ahornace:
Looks good.

PR #3669, opened 2021-07-19
Fix testRemoveFileOnFileChange test

### LINKED ISSUE #3276: IndexerTest.testRemoveFileOnFileChange fails intermittently
From time to time I'm getting the following error when running `./mvnw clean test` on my MacBook with JDK 1.8: `IndexerTest.testRemoveFileOnFileChange:347 history cache file should be preserved`

Mercurial version 5.5.1 -> not sure if related.

ahornace:
This was failing almost all the time on my mac and I remembered the problem to be elusive so I was ignoring it. The root cause seems to be different than #3276 and this only fixes the parallel execution problem.

Thanks.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
@@ -335,9 +335,7 @@ class FileHistoryCache implements HistoryCache {
      * @param mergeHistory whether to merge the history with existing or
      *                     store the histNew as is
      */
-    private void storeFile(History histNew, File file, Repository repo,
-            boolean mergeHistory) throws HistoryException {
-
+    private void storeFile(History histNew, File file, Repository repo, boolean mergeHistory) throws HistoryException {
         File cacheFile;
         try {
             cacheFile = getCachedFile(file);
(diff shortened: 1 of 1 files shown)


vladak:
Could you elaborate on what the race is / how it happens ?

ahornace:
You have 2 threads executing this condition:
`if (!dir.isDirectory() && !dir.mkdirs()) {`
Now imagine that both pass the first part `!dir.isDirectory()` with `false`.

Now both threads will try to execute the `!dir.mkdirs()` -> the first will return `true` because the dir was not created; however, the second one will return `false` as the dir is already created by the previous one and there is nothing to do (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#mkdirs()). Hence for the 2nd thread the condition will evaluate to `true` and throw `HistoryException`.

vladak:
I wonder in which cases it is necessary to create the directories - during reindex `createDirectoriesForFiles()` should create the directories before the `storeFile()` is called.

ahornace:
> I wonder in which cases it is necessary to create the directories - during reindex createDirectoriesForFiles() should create the directories before the storeFile() is called.

On my Mac which has 6 cores the `testRemoveFileOnFileChange` test fails almost all the time. Not sure if there are any other real world consequences though.

vladak:
This probably has to do with the test calling `IndexDatabase#update()` rather than the standard indexer path, probably `storeFile()` is called from `FileHistoryCache#get()` instead of via `FileHistoryCache#doFileHistory()` which gets called from `store()` that utiilizes `createDirectoriesForFiles()`. As a workaround for faiiing test it's probably fine however I'd like to see a solution that would remove the check from `storeFile()` altogether.

vladak:
Interestingly I have just hit very similar race in `store()`, getting the `WARNING` log message even though the directory was created successfully: https://github.com/oracle/opengrok/blob/4bc99e10af7f2a559277c9380a4d62281495d1fe/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java#L462-L465
when indexing a repository and its sub-repository (`/var/opengrok/src.gitsub/main_repo/submd_repo` and `/var/opengrok/src.gitsub/main_repo`) from scratch. This is to do with history cache being generated for these in parallel.

PR #3672, opened 2021-07-20
Fix missing email addresses in go xref

### LINKED ISSUE #3659: 1.6.3 email addresses are lost in Golang source files
**Describe the bug**

We have added go code to our grok index, source files that contain email addresses seem to have them stripped out when viewing go source files.

Email addresses remain visible in other languages and in searches and in the raw files when downloaded from grok.

**To Reproduce**
On 1.6.3, add a Go source file containing an email address in a variable such as a string.

**Expected behavior**
Golang strings containing email addresses should be viewable 

**Screenshots**

Email addresses missing from go code:

image

Email addresses found correctly in go code by search:

image

ahornace:
fixes #3659

Thanks.


vladak:
Nice, thanks !

PR #3673, opened 2021-07-21
add check for commands section of the mirror configuration

vladak:
As discovered in https://github.com/oracle/opengrok/issues/3664#issuecomment-883564423 , the `commands` part of the `opengrok-mirror` configuration has to be checked.


vladak:
This is how it looks when run with the invalid configuration referenced above:
```
$ opengrok-mirror -c mirror-invalid-commands.yml --all
value of 'repo' under 'commands' is not string
```

ahornace:
Looks good 👍

PR #3682, opened 2021-07-27
revamp forbidden check

vladak:
This change performs the check for forbidden access right in `mast.jsp` in order to produce better looking forbidden page. Unlike the `enoent` page it does not add the full menu (a practice which I find confusing).


ahornace:
Looks good!

PR #3690, opened 2021-08-03
Fix clicking on revision in annotation mode returns 400

### LINKED ISSUE #3689: Clicking on revision in annotation mode returns 400 bad request
**Describe the bug**
Annotation seems to be broken on 1.7.15, clicking on a specific commit in annotation mode results in 400.

**To Reproduce**

1. Go to http://demo.opengrok.org/xref/OpenGrok/opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginClassLoader.java?a=true&r=8ae5e262
2. click on any commit

**Expected behavior**
Load the file in the desired revision.

Results in this uri:
```
http://demo.opengrok.org/xref/OpenGrok/opengrok-indexer/src/main/java/org/opengrok/indexer/framework/opengrok-indexer%2Fsrc%2Fmain%2Fjava%2Forg%2Fopengrok%2Findexer%2Fframework%2FPluginClassLoader.java?a=true&r=0cdb674f
```
 which returns 400 bad request.

ahornace:
fixes #3689

The problem was that the new git history implementation was passing the relative path to the source directory instead of the file name. I also improved the url encoding while I was at it.

Thanks.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -28,6 +28,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Date;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


tulinkry (approved):

tulinkry (approved):
Makes sense

ahornace:
The changes break the renamed file handling between directories.

ahornace:
> The changes break the renamed file handling between directories.

It's still not ideal and some things were not working properly because
```java
for (DiffEntry diff : diffs) {
    if (diff.getChangeType() == DiffEntry.ChangeType.RENAME &&
            originalFile.equals(diff.getNewPath())) {
        originalFile = diff.getOldPath();
    }
}
```
did not recognize a rename by `git mv` but as `ADD` with `R100` and was therefore ignored.

vladak:
> > The changes break the renamed file handling between directories.
> 
> It's still not ideal and some things were not working properly because
> 
> ```java
> for (DiffEntry diff : diffs) {
>     if (diff.getChangeType() == DiffEntry.ChangeType.RENAME &&
>             originalFile.equals(diff.getNewPath())) {
>         originalFile = diff.getOldPath();
>     }
> }
> ```
> 
> did not recognize a rename by `git mv` but as `ADD` with `R100` and was therefore ignored.

file a new issue to track this ?

vladak:
Also, a test would be nice.

PR #3696, opened 2021-08-12
allow directory in commands

vladak:
The recently added check for commands in `opengrok-mirror` configuration is too strict. Specifically, Teamware needs to accept a directory: https://github.com/oracle/opengrok/blob/7f430e467c592672b420a552530c6e091e5b87f3/tools/src/main/python/opengrok_tools/scm/teamware.py#L34-L47

While I could just remove the check for file, this still captures (rare) cases such as FIFOs and such. Possibly better approach would be to reach out to the `scm` module and perform per repository type check however I need a quick fix.

### DIFF
--- a/tools/src/main/python/opengrok_tools/utils/mirror.py
+++ b/tools/src/main/python/opengrok_tools/utils/mirror.py
@@ -607,8 +607,8 @@ def check_commands(commands):
                          format(name, COMMANDS_PROPERTY, value))
             return False
 
-        if not os.path.isfile(value):
-            logger.error("path for '{}' under '{}' is not a file: {}".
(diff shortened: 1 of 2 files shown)


tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -607,8 +607,8 @@ def check_commands(commands):
                          format(name, COMMANDS_PROPERTY, value))
             return False
 
-        if not os.path.isfile(value):
-            logger.error("path for '{}' under '{}' is not a file: {}".
+        if not os.path.isfile(value) and not os.path.isdir(value):
```
ahornace, line 610: Is not `os.path.exists` enough?

PR #3699, opened 2021-08-12
introduce timeout for generating xrefs

vladak:
This change introduces time out for xref processing/generation. Also, I took the opportunity to join some short wrapped lines.

The default timeout will be 30 seconds (default ctags timeout is 10 seconds).

In the indexer log this will look like this:
```
2021-08-12 13:44:00.802+0000 WARNING t143 IndexDatabase.addFile: File '/solaris-userland/components/desktop/firefox/firefox-78.11.0/accessible/windows/msaa/HyperTextAccessibleWrap.cpp' interrupted--failed to gen
erate xref :java.util.concurrent.ExecutionException: java.util.concurrent.TimeoutException
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -425,8 +425,8 @@ public class Ctags implements Resettable {
      * @throws IOException I/O exception
      * @throws InterruptedException interrupted command
      */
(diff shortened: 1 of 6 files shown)


vladak:
triggers update of https://github.com/oracle/opengrok/wiki/Indexer-configuration

ahornace (approved):

ChristopheBordieu:
Shouldn't this be reported as an error instead of  warning?

PR #3700, opened 2021-08-12
add test for xref timeout

vladak:
This change adds basic functional test for the changes in #3699.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/PlainAnalyzerTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/PlainAnalyzerTest.java
@@ -1,0 +1,87 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/analysis/PlainAnalyzerTest.java
```diff
@@ -0,0 +1,87 @@
...
+                return new ByteArrayInputStream(bytes);
+            }
+        };
+    }
+
+    @Test
+    void testXrefTimeout() {
+        RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+        long timeoutOriginal = env.getXrefTimeout();
+        int timeout = 3;
```
ahornace, line 56: Is not `1` enough? Saving 2 seconds for each build can be good.

vladak: good point, fixed.

PR #3702, opened 2021-08-18
Fix project selection by space

### LINKED ISSUE #3701: project list does not allow checkbox selection by space
Until recently it was possible to type part of project name in the project selection box, and then use cursors and space to select a project. The first 2 steps still work, but pressing space does not result in marking the checkbox - it's necessary to click it. I assume this is inadvertent change, since there does not seem to be any benefit to disabling this feature - i.e. space does not seem to do anything else instead.

I've tested this behavior in Firefox, Vivaldi and Chrome.

ahornace:
fixes #3701

Thanks.


vladak:
Thanks !

Issue #3709, opened 2021-09-06
NOMIRROR env variable seems to have no effect


mishka81:
when running latest Docker image (1.7) and specifying "NOMIRROR=1" environment variable mirroring is not disabled.
In the logs I see "opengrok-mirror" working.

With latest 1.6 image we do not have this behavior.

vladak:
Trying with 1.7.16, this works as expected. Note that `opengrok-mirror` actually runs however it exits right after finding out that the environment variable is set.

For reference, here's the `docker-compose` configuration:
```yml
version: "3"

# More info at https://github.com/oracle/opengrok/docker/
services:
  opengrok:
    container_name: opengrok-test
    image: opengrok/docker:master
    ports:
      - "8080:8080/tcp"
      - "5000:5000/tcp"
    environment:
      OPENGROK_LOG_LEVEL: 'debug'
      SYNC_PERIOD_MINUTES: '0'
      NOMIRROR: 1
    # Volumes store your data between container upgrades
    volumes:
      - '~/opengrok/src/:/opengrok/src/'  # source code
      - '~/opengrok/etc/:/opengrok/etc/'  # folder contains configuration.xml
      - '~/opengrok/data/:/opengrok/data/'  # index and other things for source code
```
Setting the `OPENGROK_LOG_LEVEL` to `debug` reveals this during container startup:
```
opengrok-dev | 2021-09-13 10:24:26,574    DEBUG opengrok_tools | Extra environment: {'OPENGROK_NO_MIRROR': '1'}
```
When I modified `sync.yml` to make `opengrok-mirror` run with debug logging (by using `-l DEBUG`) I am seeing this:
```
opengrok-dev | 2021-09-13 10:24:35,908    DEBUG opengrok_tools | skipping mirror based on the OPENGROK_NO_MIRROR environment variable
```

PR #3713, opened 2021-09-13
share the OPENGROK_NO_MIRROR env var definition

vladak:
When looking at #3709 I noticed that the `OPENGROK_NO_MIRROR` environment variable name is not shared properly between the main Docker program and the tooling. This change fixes that.


ahornace:
Looks good.

PR #3715, opened 2021-09-15
replace HashSet with concurrent set in PendingFileCompleter

vladak:
This change should take care of the spurious failures when indexing data with large set of changed files.


ahornace:
Looks good :)

PR #3717, opened 2021-09-18
Fix header overflow

ahornace:
When opening pages with a mast the header was covering part of the content. Example:
http://demo.opengrok.org/xref/OpenGrok/pom.xml?r=5ec8ffa7
http://demo.opengrok.org/xref/OpenGrok/
Screen Shot 2021-09-17 at 21 36 45

Based on https://github.com/oracle/opengrok/pull/1194 it seems the change was done to allow scrolling with space or arrow keys. It's still possible even without the explicit focus (tested in Chrome and Safari). And the scrollbar stays on the same position when page is refreshed (which is in my opinion an added bonus).

Thanks!


vladak:
Thanks !

PR #3721, opened 2021-09-21
track logged chunks in Progress

vladak:
Rather than checking for current count to be dividable by given count in `Progress#logLoop()`, track the last logged chunk.

There are still problems with `Progress` - I cannot get any progress logged when tracking history cache creation process (by file) even with this change and default `INFO` log level. When I attach debugger and step through breakpoints in `Progress` and `FileHistoryCache`, there is sometimes log entry emitted. Possibly related to thread scheduling so not related to the problem addressed by this change.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Progress.java
@@ -25,6 +25,8 @@ package org.opengrok.indexer.util;
 
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


vladak:
> There are still problems with `Progress` - I cannot get any progress logged when tracking history cache creation process (by file) even with this change and default `INFO` log level. When I attach debugger and step through breakpoints in `Progress` and `FileHistoryCache`, there is sometimes log entry emitted. Possibly related to thread scheduling so not related to the problem addressed by this change.

This was actually caused by erroneous programming - the try-with-resources block needs to span the `await()` too, otherwise the task submissions can be done before the progress thread starts running, the `Progress` instance will be closed and this will make the thread go away. Done properly in #3735.

PR #3724, opened 2021-09-24
always create mirror.conf

vladak:
The sync configuration runs `opengrok-mirror` always with the mirror config file so the file needs to be always present.


ahornace:
Looks good.

PR #3729, opened 2021-09-30
Cleanup file locks in opengrok python tools

ohm314:
This patch ensures that lock files created by FileLock are always
eventually cleaned up. FileLock does not delete lock files itself, which
usually isn't an issue, but it can lead to permission problems if a tool
finds a file lock left over from a previous (finished) run and can't overwrite it.

### DIFF
--- a/tools/src/main/python/opengrok_tools/indexer.py
+++ b/tools/src/main/python/opengrok_tools/indexer.py
@@ -27,6 +27,7 @@ import argparse
 import os
 import sys
 import tempfile
+from os import path
 
 from .utils.indexer import FindCtags, Indexer
 from .utils.log import get_console_logger, get_class_basename, fatal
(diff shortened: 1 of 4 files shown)


vladak:
> This patch ensures that lock files created by FileLock are always eventually cleaned up. FileLock does not delete lock files itself, which usually isn't an issue, but it can lead to permission problems if a tool finds a file lock left over from a previous (finished) run and can't overwrite it.

do you mean permission problem in case the tools run under different users ?

vladak:
Anyway, this changed would lead to wrong locking as noted in https://github.com/tox-dev/py-filelock/blob/c9b6d90b8d0d6f3971b97ce7bfd1b3402d58220e/src/filelock/_unix.py#L27-L30 , hence closing. If there is problem with permissions, then set the umask and group permissions accordingly.

ohm314:
Ah, I did not think that through! sorry for the bad patch...

PR #3731, opened 2021-10-01
Projects controller vs repositories

vladak:
Fixes a problem when setting per project property that is not shadowed by repository property.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
@@ -315,17 +315,10 @@ public class ProjectsController {
             List<RepositoryInfo> riList = env.getProjectRepositoriesMap().get(project);
             if (riList != null) {
                 for (RepositoryInfo ri : riList) {
-                    Repository repo = getRepository(ri, CommandTimeoutType.RESTFUL);
-
                     // Set the property if there is one.
-                    try {
-                        ClassUtil.getFieldValue(repo, field);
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -311,13 +311,20 @@ public void set(
...
+            // Refresh field values for project's repositories for this project as well.
             List<RepositoryInfo> riList = env.getProjectRepositoriesMap().get(project);
             if (riList != null) {
                 for (RepositoryInfo ri : riList) {
                     Repository repo = getRepository(ri, CommandTimeoutType.RESTFUL);
 
-                    // set the property
+                    // Set the property if there is one.
+                    try {
+                        ClassUtil.getFieldValue(repo, field);
```
ahornace, line 322: Shouldn't the `setFieldValue` be called here and the invocation below removed?

ahornace: Or, if we want to propagate errors, then maybe introduce a new method `hasField` or similar?

vladak: Definitely want to propagate the errors. While adding `hasField()` I realized the controller actually sets the values on wrong repository object set. Setting the values on `Repository` objects constructed from `RepositoryFactory` will not have any effect on the configuration. The test was wrong also.

ahornace:
Looking good.

PR #3735, opened 2021-10-04
report file history cache progress

vladak:
This change adds progress reporting for file history cache. For repositories that support per partes history cache creation an annotation of the terminal changeset is added.


vladak:
The build failures seems to be all related to the use of `filelock` in tools.

ahornace (approved):
Looks good to me.

vladak:
Thanks @ahornace !

Issue #3764, opened 2021-11-18
Many ctags warnings due to kind name redefinition


ohm314:
**Describe the bug**

While looking through indexer logs I keep stumbling over warning lines like this:

```
2021-11-18 03:20:35.063+0000 WARNING t27 Ctags.lambda$run$0: Error from ctags: ctags: Warning: Don't reuse the kind letter `c' in a language Haskell (old: "constructor", new: "classes")
```

Those are just warnings, so could be ignored, but I was wondering whether they could be fixed before they lead to errors down the line (similar to #1137). For above example, the Haskell definition is overridden by opengrok here: https://github.com/oracle/opengrok/blob/00e4c74bf30151dafe908105e79986013b8671c1/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java#L357
More precisely, line 362 is defining a new regex to tag a constructor (or class). The problem is with the kind definition, the code uses the extended form, but changed the long name from the one defined in ctags:

```
/usr/local/bin/ctags --list-kinds-full | grep Haskell
Haskell              c      constructor           yes     no      0      NONE   type constructors
Haskell              f      function              yes     no      0      NONE   functions
Haskell              m      module                yes     no      0      NONE   modules
Haskell              t      type                  yes     no      0      NONE   types
```
Looking at ctags help I see that the full kind name is optional and its use deprecated, but I am not sure if actually the redefined name is necessary for Opengrok.

(From ctags documentation)
> A kind specifier (<kind-spec>) for tags matching regexp may follow `<name_pattern>`, which will determine what kind of tag is reported in the kind extension field (see tags(5)).
>
> `<kind-spec>` has two forms: one-letter form and full form.
>
> The one-letter form in the form of `<letter>`. It just refers a kind `<letter>` defined with --kinddef-<LANG>. This form is recommended in Universal Ctags.
>
> The full form of `<kind-spec>` is in the form of `<letter>,<name>,<description>`. Either the kind `<name>` and/or the `<description>` can be omitted. See the description of `--kinddef-<LANG>=<letter>,<name>,<description>` option about the elements.
>
> The full form is supported only for keeping the compatibility with Exuberant Ctags which does not have `--kinddef-<LANG> `option. Supporting the form will be removed from Universal Ctags in the future.

**Versions:**
Opengrok: 1.7.18
ctags: 5.9.0

**Possible Solution**

If the long names is not necessary, they should be just removed from all `--regex` definitions:

So for Haskell this would be:

[...]

Issue #1672, opened 2017-07-18
put the ctags language defs into a file instead of lengthy command line


tarzanek:
the ctags command line is TOO LONG:
https://github.com/OpenGrok/OpenGrok/blob/master/src/org/opensolaris/opengrok/analysis/Ctags.java#L88

we should write this into a file and use the file and --options param of ctags:
http://docs.ctags.io/en/latest/optlib.html

vladak:
This would clash with user configurable options already present in `Ctags#initialize()`:

```
242              /* Add extra command line options for ctags. */
243              if (CTagsExtraOptionsFile != null) {
244                  LOGGER.log(Level.INFO, "Adding extra options to ctags");
245                  command.add("--options=" + CTagsExtraOptionsFile);
246              }
```

I.e. the file needs to be generated by merging user config.

tarzanek:
I was hoping 2 --options can be used ;)

vladak:
As long as the double option works for both Exuberant and Universal ctags fine with me.

tarzanek:
(I don't know if 2 --options can be used, I was just joking, anyways, this should be investigated, since I am not sure if all definitions are being used right now)

cross:
Is there any status on this?  Old open bug.  I'm trying to track down some warnings I'm getting from ctags, so how these are all specified may come into play...

masatake:
You can specify multiple option files like: https://github.com/sourcegraph/sourcegraph/tree/df2c4da895f6fb7935e8858f7a87ff403fdd11c1/cmd/symbols/.ctags.d .

PR #3770, opened 2021-11-19
Update ctag language regexp

ohm314:
- Use only short form for kinds, the long form is deprecated
- Define new kinds with the --kinddef-<LANG> flag
- Harmonize kind names and descriptions

This is an attempt to address  #3764

Note:
The number of flags to ctags is huge. Maybe it would be worth writing all these options to an option file passed via --options=PATHNAME (as described [here](https://docs.ctags.io/en/latest/option-file.html#options-pathname-option))
In this case the file can be part of the repository.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -368,7 +368,7 @@ public class Ctags implements Resettable {
         command.add("--kinddef-clojure=m,macro,Macros");
         command.add("--kinddef-clojure=i,inline,Inlines");
         command.add("--kinddef-clojure=a,multimethodDefinition,Multimethod\\ definitions");
-        command.add("--kinddef-clojure=b,multimethodInstance,Multimethod\\ instnances");
+        command.add("--kinddef-clojure=b,multimethodInstance,Multimethod\\ instances");
         command.add("--kinddef-clojure=c,definitionOnce,Definition\\ once");
         command.add("--kinddef-clojure=s,struct,Structs");
         command.add("--kinddef-clojure=v,intern,Interns");
@@ -404,7 +404,7 @@ public class Ctags implements Resettable {
         if (!env.getCtagsLanguages().contains("Scala")) { // Built-in would be capitalized.
             command.add("--langdef=scala"); // below is bug 61 to get full scala support. Lower-case
         }
-        command.add("--kinddef-scala=c,class,Classesw");
+        command.add("--kinddef-scala=c,class,Classes");
         command.add("--kinddef-scala=o,object,Objects");
         command.add("--kinddef-scala=C,caseClass,Case\\ classes");
         command.add("--kinddef-scala=O,caseObject,Case\\ objects");
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -248,10 +248,10 @@ public class Ctags implements Resettable {
         }
 
         // The following are not supported yet in Universal Ctags b13cb551
-        command.add("--kinddef-rust=\"C,const,Static constants\"");
-        command.add("--kinddef-rust=\"I,impl,Trait implementation\"");
-        command.add("--kinddef-rust=\"r,trait,Traits\"");
-        command.add("--kinddef-rust=\"V,variable,Variables\"");
+        command.add("--kinddef-rust=C,const,Static\\ constants");
+        command.add("--kinddef-rust=I,impl,Trait\\ implementation");
+        command.add("--kinddef-rust=r,trait,Traits");
+        command.add("--kinddef-rust=V,variable,Variables");
         command.add("--regex-rust=/^[[:space:]]*(pub[[:space:]]+)?(static|const)[[:space:]]+(mut[[:space:]]+)?" +
                 "([[:alnum:]_]+)/\\4/C/");
         command.add("--regex-rust=/^[[:space:]]*(pub[[:space:]]+)?(unsafe[[:space:]]+)?impl([[:space:]\n]*<[^>]*>)?" +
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -341,57 +363,76 @@ private void addClojureSupport(List<String> command) {
...
+        command.add("--regex-haskell=/[[:space:]]+([a-zA-Z0-9_]+).*[[:space:]]+={1}[[:space:]]+/\\1/f/");
+        command.add("--regex-haskell=/^(let|where)[[:space:]]+([a-zA-Z0-9_]+).*[[:space:]]+={1}[[:space:]]+/\\2/f/");
+        command.add("--regex-haskell=/[[:space:]]+(let|where)[[:space:]]+([a-zA-Z0-9_]+).*[[:space:]]+={1}[[:space:]]+/\\2/f/");
     }
 
     private void addScalaSupport(List<String> command) {
         if (!env.getCtagsLanguages().contains("Scala")) { // Built-in would be capitalized.
             command.add("--langdef=scala"); // below is bug 61 to get full scala support. Lower-case
         }
+        command.add("--kinddef-scala=c,class,Classesw");
```
ahornace, line 407: `Classesw` -> `Classes`

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -341,57 +363,76 @@ private void addClojureSupport(List<String> command) {
...
-        command.add("--regex-clojure=/\\([[:space:]]*defmethod[[:space:]]+([-[:alnum:]*+!_:\\/.?]+)/\\1/b,multimethodInstance/");
-        command.add("--regex-clojure=/\\([[:space:]]*defonce[[:space:]]+([-[:alnum:]*+!_:\\/.?]+)/\\1/c,definitionOnce/");
-        command.add("--regex-clojure=/\\([[:space:]]*defstruct[[:space:]]+([-[:alnum:]*+!_:\\/.?]+)/\\1/s,struct/");
-        command.add("--regex-clojure=/\\([[:space:]]*intern[[:space:]]+([-[:alnum:]*+!_:\\/.?]+)/\\1/v,intern/");
+        command.add("--kinddef-clojure=d,definition,Definitions");
+        command.add("--kinddef-clojure=p,privateFunction,Private\\ functions");
+        command.add("--kinddef-clojure=m,macro,Macros");
+        command.add("--kinddef-clojure=i,inline,Inlines");
+        command.add("--kinddef-clojure=a,multimethodDefinition,Multimethod\\ definitions");
+        command.add("--kinddef-clojure=b,multimethodInstance,Multimethod\\ instnances");
```
ahornace, line 371: `instnances` -> `instances`

ahornace:
Could you please fix the tests? Otherwise looks good to me.

vladak:
> Note: The number of flags to ctags is huge. Maybe it would be worth writing all these options to an option file passed via 

This is tracked by #1672

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -248,22 +248,26 @@ private void addRustSupport(List<String> command) {
         }
 
         // The following are not supported yet in Universal Ctags b13cb551
+        command.add("--kinddef-rust=C,const,Static\\ constants");
+        command.add("--kinddef-rust=I,impl,Trait\\ implementation");
+        command.add("--kinddef-rust=r,trait,Traits");
+        command.add("--kinddef-rust=V,variable,Variables");
```
ahornace, line 254: I'm seeing another warning:
```
Error from ctags: ctags: Warning: the kind for name `variable' specified in "--kinddef-rust" option is already defined.
```

Maybe we will have to pick another name or maybe just skip the kind definition?

ohm314: I must have missed this one. Looking at the existing definitions I get:

```
Rust                 M      macro                 yes     no      0      NONE   Macro Definition
Rust                 P      method                yes     no      0      NONE   A method
Rust                 c      implementation        yes     no      0      NONE   implementation
Rust                 e      enumerator            yes     no      0      NONE   An enum variant
Rust                 f      function              yes     no      0      NONE   Function
Rust                 g      enum                  yes     no      0      NONE   Enum
Rust                 i      interface             yes     no      0      NONE   trait interface
Rust                 m      field                 yes     no      0      NONE   A struct field
Rust                 n      module                yes     no      0      NONE   module
Rust                 s      struct                yes     no      0      NONE   structural type
Rust                 t      typedef               yes     no      0      NONE   Type Alias
Rust                 v      variable              yes     no      0      NONE   Global variable
```
You see there is already a `v` defined, but I'm not sure if the newly defined `V` is another type of variable? in this case we should just use another shortname in the kinddef? Otherwise the whole kinddef can be dropped and we use the the regex options `v` instead of `V`.

ohm314:
@vladak: thanks, maybe the option file stuff should be tackled in another PR anyway. Reading that issue I was wondering if there is still a requirement to support Exuberant ctags? It does not support multiple `--options` flags afaik and so one would have to merge the two files (which will be annoying to do 😄 )

vladak:
> @vladak: thanks, maybe the option file stuff should be tackled in another PR anyway. Reading that issue I was wondering if there is still a requirement to support Exuberant ctags? It does not support multiple `--options` flags afaik and so one would have to merge the two files (which will be annoying to do smile )

Yeah, definitely a topic for another PR. As for Exuberant ctags, this is long dead so the references are purely historical I'd say. If we need some OpenGrok specific changes in Universal ctags, it should not be a problem raising PR there.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
ahornace: I've tested with this snippet found here: https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html
```rust
fn main() {
    let x = 5;
    println!("The value of x is: {}", x);
    x = 6;
    println!("The value of x is: {}", x);
}
```
and ctags did not recognize the variable `x` by default. Funny thing is that OpenGrok has a bug in the regex as well as it should probably be:
`/^[[:space:]]*let([[:space:]]+mut)?[[:space:]]+([[:alnum:]_]+)/\\2/V/`

Either way, I'd personally go with a new name like `localVariable`. What do you think?

ohm314: Makes sense. Sorry didn't get around to try this out myself. I applied the fix.

ahornace:
Thanks!

PR #3772, opened 2021-11-20
Compress configuration when sending it to webapp

### LINKED ISSUE #3771: setting new configuration should be asynchronous
Hi team,

Looking at:
https://github.com/oracle/opengrok/blob/0e4c55544f8ea0a68e8bae37b0e502097e008ec1/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java#L1433
it seems that at the end of the indexing process, when new configuration file is ready and has been written, it is sent to webapp without being compressed.

If I am right, would it be possible the PUT request compresses the file (as much as possible) when sending it?
I guess that could be done updating the request header with the directive: _Content-Encoding: gzip_. 

My current configuration file is ~20M and the configuration sending request takes long time. This could help decrease this time (and so limit the risk of a timeout or a connection loss).

Regards,
Christophe

ahornace:
fixes #3771

Thanks.


vladak:
I'd like to understand the original problem first. This will not generally hurt anything (besides debugging the request by snooping the network) however my suspicion is that it will not address the actual problem.

vladak:
Based on the information received in the issue, this would not help.

Issue #2147, opened 2018-06-08
projects lost when merging per-project setting in read-only configuration with active configuration


vladak:
https://github.com/oracle/opengrok/wiki/Per-project-configuration advises to put per-project configuration into read-only configuration file. However when merged with a configuration retrieved from webapp that already contains some projects via:

```shell
/opengrok/dist/bin/projadm.py -b /opengrok -R /opengrok/etc/readonly_configuration.xml -r
```

the resulting `configuration.xml` will only have the projects specified in read-only configuration.

vladak:
One way how to deal with this would be to change the workflow - augment the `Project` `Message` to be able to set per-project properties and then retrieve the configuration so that it is persistent.

vladak:
I think this is a fall-out of how XML serialization is done. As a result, merge is not really a merge.

tulinkry:
I think some customizations in this code

vladak:
This has a workaround so stripping the stopper tag for now.

vladak:
Also, this needs to work with full global (not per project) indexing - the per-project setting in read-only configuration needs to be applied after the projects are discovered.

vladak:
Thel workaround needs a tweak for full global (not per project) indexing - I believe the per-project setting in read-only configuration is applied after the projects are discovered however once the initial index is done, the per-project settings have to be removed from read-only configuration and set in the webapp.

PR #3774, opened 2021-11-22
allow to configure bug/review properties per project

vladak:
This change makes it possible to configure bug/review links on per project level. Due to the way de-serialization of the `Project` objects works (setters are not called for unspecified properties), the getters have to be equipped with null checks and fall back to the global configuration. The code in `completeWithDefaults()` is used only when adding new projects in the indexer.

https://github.com/oracle/opengrok/wiki/Per-project-configuration will have to be updated.

Sample configuration:
```xml

   <boolean>true</boolean>
  </void>

   <void property="projects">
     <void method="put">
      <string>one</string>
        <boolean>true</boolean>
       </void>
        <boolean>true</boolean>
       </void>
       <void property="indexed">
        <boolean>true</boolean>
       </void>
       <void property="name">
        <string>one</string>
       </void>
       <void property="path">
        <string>/one</string>
       </void>
       <void property="bugPage">
        <string>http://example.com/foobar</string>
       </void>
        <boolean>false</boolean>
       </void>
      </object>
     </void>
   </void>

 </object>
</java>
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/PatternUtil.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/PatternUtil.java
@@ -34,7 +34,7 @@ public class PatternUtil {
      * A check if a pattern contains at least one pair of parentheses meaning
      * that there is at least one capture group. This group must not be empty.
      */
-    static final String PATTERN_SINGLE_GROUP = ".*\\([^\\)]+\\).*";
+    private static final String PATTERN_SINGLE_GROUP = ".*\\([^\\)]+\\).*";
     /**
      * Error string for invalid patterns without a single group. This is passed
      * as a first argument to the constructor of PatternSyntaxException and in
@@ -43,7 +43,7 @@ public class PatternUtil {
      * @see PatternSyntaxException
      * @see #PATTERN_SINGLE_GROUP
      */
-    static final String PATTERN_MUST_CONTAIN_GROUP = "The pattern must contain at least one non-empty group -";
+    private static final String PATTERN_MUST_CONTAIN_GROUP = "The pattern must contain at least one non-empty group -";
 
     /**
      * Check and compile the bug pattern.
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/PatternUtil.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/PatternUtil.java
@@ -1,0 +1,62 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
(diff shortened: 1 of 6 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/PatternUtil.java
```diff
@@ -0,0 +1,62 @@
...
+public class PatternUtil {
+    private PatternUtil() {
+        // private to enforce static
+    }
+
+    /**
+     * A check if a pattern contains at least one pair of parentheses meaning
+     * that there is at least one capture group. This group must not be empty.
+     */
+    static final String PATTERN_SINGLE_GROUP = ".*\\([^\\)]+\\).*";
```
ahornace, line 37: Make these `private`?

opengrok-indexer/src/test/java/org/opengrok/indexer/configuration/ProjectTest.java
```diff
@@ -143,6 +147,10 @@ public void testMergeProjects1() {
 
         assertEquals(env.getTabSize(), p1.getTabSize());
         assertEquals(env.isNavigateWindowEnabled(), p1.isNavigateWindowEnabled());
+        assertEquals(env.getBugPage(), p1.getBugPage());
```
ahornace, line 150: Just FYI, recently found out about https://stackoverflow.com/questions/40796756/assertall-vs-multiple-assertions-in-junit5 which might be useful here? No need to change though

ahornace (approved):

opengrok-indexer/src/test/java/org/opengrok/indexer/configuration/ProjectTest.java
vladak: Nice tip !

vladak:
When updating a configuration of a staging machine with the changes in order to set JDK bug links, I ran into an old issue with per project workflow (#2147). First thing I added the XML snippet with `bugPage`/`bugPattern` into read-only configuration and ran `opengrok-projadm` to merge it with the current configuration in the web app. This resulted in the `Project` objects disappearing from the configuration.

Luckily, thanks to the easily accessible ZFS snapshots (in the hidden by default `.zfs` directory) and the script wrapping `opengrok-projadm` that creates the snapshots whenever it is run, I was able to recover valid configuration quickly - otherwise I'd have to run the indexer for all projects.

When working with the per project workflow, the correct way is to modify the properties (once the project is added) like so:
```
curl -X PUT -d 'https://bugs.openjdk.java.net/browse/JDK-' http://localhost:8080/source/api/v1/projects/jdk8u/property/bugPage
curl -X PUT -d '([1-9][0-9]{6,7})' http://localhost:8080/source/api/v1/projects/jdk8u/property/bugPattern
```
and then grab the configuration from the web app and make it persistent.

PR #3775, opened 2021-11-22
add tunable for concurrency level of repository invalidation 

vladak:
This change lowers the pressure on the system imposed by `HistoryGuru#invalidateRepositories()` by changing the concurrency level (the web app uses _parallelism_ which is not technically correct, I think) for the web app and avoiding repository tag construction in that code path. This is mainly targeting setting the configuration in the web app.

https://github.com/oracle/opengrok/wiki/Webapp-configuration will have to be updated.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2021, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.history;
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -815,7 +815,14 @@ public void invalidateRepositories(Collection<? extends RepositoryInfo> repos, C
          * run in parallel to speed up the process.
          */
         final CountDownLatch latch = new CountDownLatch(repos.size());
-        final ExecutorService executor = Executors.newFixedThreadPool(env.getIndexingParallelism(),
+        int parallelismLevel;
+        // Both indexer and web app startup should be as quick as possible.
+        if (cmdType.equals(CommandTimeoutType.INDEXER) || cmdType.equals(CommandTimeoutType.WEBAPP_START)) {
```
ahornace, line 820: I think it's generally preferable to use `==` for enum comparison.

vladak: I wondered about that when I was writing the comparisons.

PR #3781, opened 2021-11-23
sanitize class paths

vladak:
This change should fix the "zip slip" issue in plugin framework reported by Sonar. Normally the plugin archives are considered trusted however a bit of sanitization does not hurt.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
@@ -278,14 +278,14 @@ public abstract class PluginFramework<PluginType> {
      * delegates the loading to the custom class loader
      * {@link #loadClass(String)}.
      *
-     * @param jarfiles list of jar files containing java classes
+     * @param fileList list of jar files containing java classes
      * @see #handleLoadClass(String)
      * @see #loadClass(String)
      */
-    private void loadJarFiles(List<File> jarfiles) {
+    private void loadJarFiles(List<File> fileList) {
         PluginType pf;
 
-        for (File file : jarfiles) {
+        for (File file : fileList) {
             try (JarFile jar = new JarFile(file)) {
                 Enumeration<JarEntry> entries = jar.entries();
                 while (entries.hasMoreElements()) {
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -313,9 +319,24 @@ private String getClassName(File f) {
...
+    @Nullable
     private String getClassName(JarEntry f) {
+        final String filePath = f.getName();
+        if (!filePath.endsWith(CLASS_SUFFIX)) {
+            return null;
+        }
+
+        File file = new File(pluginDirectory.getAbsolutePath(), filePath);
+        try {
+            if (!file.getCanonicalPath().startsWith(pluginDirectory.getCanonicalPath())) {
```
ahornace, line 331: Shouldn't the `File.separator` be appended?
```java
if (!file.getCanonicalPath().startsWith(pluginDirectory.getCanonicalPath() + File.separator)) {
```
At least the example on https://snyk.io/research/zip-slip-vulnerability has it.

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -313,9 +319,24 @@ private String getClassName(File f) {
...
+                return null;
```
ahornace, line 332: Maybe add a warning log?

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -313,9 +319,24 @@ private String getClassName(File f) {
...
+            }
+        } catch (IOException e) {
+            return null;
```
ahornace, line 335: I think this shouldn't happen often, maybe add a log?

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
vladak: Good point.

PR #3782, opened 2021-11-23
use try-with-resources on DefinitionsTokenStream

vladak:
This change is basically a no-op due to empty close() however it silences Sonar warnings.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/plain/PlainAnalyzer.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/plain/PlainAnalyzer.java
@@ -203,27 +203,28 @@ public class PlainAnalyzer extends TextAnalyzer {
     private void tryAddingDefs(Document doc, Definitions defs, StreamSource src)
             throws IOException {
 
-        DefinitionsTokenStream defstream = new DefinitionsTokenStream();
-        defstream.initialize(defs, src, this::wrapReader);
+        try (DefinitionsTokenStream defstream = new DefinitionsTokenStream()) {
+            defstream.initialize(defs, src, this::wrapReader);
 
-        /*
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/plain/PlainAnalyzer.java
```diff
@@ -203,27 +203,28 @@ public void analyze(Document doc, StreamSource src, Writer xrefOut) throws IOExc
...
+             * expensive, particularly when using offsets in postings."
+             *     DEFS re-analysis will not be correct, however, as the
+             * PlainAnalyzer which UnifiedHighlighter will use on-the-fly will
+             * not correctly integrate ctags Definitions.
+             *     Storing term vectors, however, allows UnifiedHighlighter to
+             * avoid re-analysis at the cost of a larger index. As DEFS are a
+             * small subset of source text, it seems worth the cost to get
+             * accurate highlighting for DEFS MTQs.
+             */
+            doc.add(new OGKTextVecField(QueryBuilder.DEFS, defstream));
```
ahornace, line 226: I don't think this is going to work.

`doc.add()` only adds the field to its internal list but does not interact with it in any other way. The `OGKTextVecField` delegates to `Field`'s constructor which again only stores the `defstream` locally but does not consume it yet. So I think the stream will be closed before being read.

ahornace (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/plain/PlainAnalyzer.java
vladak: I have been trying to wrap my head around this very thing. You are right that closing is risky. Will need to find a way how to suppress the Sonar issue.

PR #3787, opened 2021-11-24
another round of fixes for Sonar warnings

vladak:
Trying to address more Sonar blockers.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
@@ -113,7 +113,7 @@ public class SSCMRepository extends Repository {
                 LOGGER.log(Level.WARNING,
                         "Failed to work with {0} file of {1}: {2}", new Object[]{
                             MYSCMSERVERINFO_FILE,
-                            getDirectoryName(), ex.getClass().toString()});
+                            getDirectoryName(), ex.getClass()});
             }
         }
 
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
```diff
@@ -140,8 +140,12 @@ Executor getHistoryLogExecutor(final File file, String sinceRevision) throws IOE
         } else {
             argv.add(file.getName());
         }
-        if (sinceRevision != null && new Scanner(sinceRevision).hasNextInt()) {
-            argv.add("-v" + (Integer.parseInt(sinceRevision) + 1) + ":" + Integer.MAX_VALUE);
+        if (sinceRevision != null) {
+            try (Scanner scanner = new Scanner(sinceRevision)) {
+                if (scanner.hasNextInt()) {
+                    argv.add("-v" + (Integer.parseInt(sinceRevision) + 1) + ":" + Integer.MAX_VALUE);
```
ahornace, line 146: I'd use here Scanner's `nextInt` method:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Scanner.html#nextInt()

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
```diff
@@ -249,8 +252,12 @@ boolean fileHasAnnotation(File file) {
                     if (parts[0].equals(file.getName())) {
                         // Check if the version field is greater than 1
                         //  which indicates that annotate will work
-                        if (parts.length > 2 && new Scanner(parts[2]).hasNextInt()) {
-                            return Integer.parseInt(parts[2]) > 1;
+                        if (parts.length > 2) {
+                            try (Scanner scanner = new Scanner(parts[2])) {
+                                if (scanner.hasNextInt()) {
+                                    return Integer.parseInt(parts[2]) > 1;
```
ahornace, line 258: Again, I'd go with `nextInt`

ahornace (approved):
Good job with the cleanups! :)

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SSCMRepository.java
vladak: Then I'd have to handle the exceptions which would make the code a bit more difficult to read, I think.

ahornace: Oh, I see, forgot about the exception, then let's leave it as it is

PR #3789, opened 2021-11-24
suppress warning about lack of sanitization

vladak:
Looks like the change to sanitize paths from the archive did not help w.r.t. Sonar warnings so this change adds error suppression.


ahornace:
I think the vulnerability needs to be reviewed manually, or not? Can't you login and mark it as "not applicable"?

https://sonarcloud.io/project/security_hotspots?id=org.opengrok%3Aopengrok-top#

vladak:
I marked this as a false positive in the UI and sent feedback to Sonar team however this should probably be marked in the code as well.

PR #3792, opened 2021-11-25
cleanup Sonar issues in repositories

vladak:
Sonar cleanup in history/repositories.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -156,11 +156,8 @@ public class GitRepository extends RepositoryWithPerPartesHistory {
                     treeWalk.setRecursive(true);
                     treeWalk.setFilter(PathFilter.create(filename));
                     if (!treeWalk.next()) {
-                        if (LOGGER.isLoggable(Level.FINEST)) {
-                            LOGGER.log(Level.FINEST,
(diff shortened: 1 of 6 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -144,21 +144,23 @@ private HistoryRevResult getHistoryRev(OutputStream out, String fullpath, String
...
+                // Now try to find a specific file.
                 try (TreeWalk treeWalk = new TreeWalk(repository)) {
                     treeWalk.addTree(tree);
                     treeWalk.setRecursive(true);
                     treeWalk.setFilter(PathFilter.create(filename));
                     if (!treeWalk.next()) {
-                        LOGGER.log(Level.FINEST,
-                                String.format("Did not find expected file '%s' in revision %s for repository '%s'",
-                                        filename, rev, directory));
+                        if (LOGGER.isLoggable(Level.FINEST)) {
```
ahornace, line 159: Why not just use:
```java
LOGGER.log(Level.FINEST, "Did not find expected file ''{0}'' in revision {1} for repository ''{2}''", new Object[] {filename, rev, directory}); 
```
In that case the check is hidden and no message formatting is happening. Can be also applied to the warning log below.

vladak: Thanks !

PR #3798, opened 2021-11-25
Fix terraform definitions regex

ahornace:
There is a warning during indexing:
```
WARNING: Error from ctags: ctags: Warning: regcomp [[:<:]]resource[[:space:]]*"([[:alpha:]][-_[:alpha:]]*)"[[:space:]]*"([[:alpha:]][-_[:alpha:]]*)"[[:space:]]*\{: Invalid character class name
```
which causes no tf definitions to be recognized.

Thanks!

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -455,8 +455,8 @@ public class Ctags implements Resettable {
          */
         command.add("--kinddef-terraform=s,struct,Resource\\ names");
         command.add("--regex-terraform=" +
-                "/[[:<:]]resource[[:space:]]*\"([[:alpha:]][-_[:alpha:]]*)\"[[:space:]]*" +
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
```diff
@@ -413,7 +413,7 @@ private void addTerraformSupport(List<String> command) {
          * the possibility of leaving out some matches.
          */
         command.add("--regex-terraform=" +
-                "/[[:<:]]resource[[:space:]]*\"([[:alpha:]][-_[:alpha:]]*)\"[[:space:]]*" +
+                "/[[:space:]]*resource[[:space:]]*\"([[:alpha:]][-_[:alpha:]]*)\"[[:space:]]*" +
```
ahornace, line 416: It seems that ctags on windows have some problem with this regex, maybe it doesn't support POSIX? However, there is no warning, only the line is not matched. I'll investigate tomorrow.

PR #3799, opened 2021-11-29
Add support for additional terraform definitions

ahornace:
I found a problem that many of the custom definitions (including the ones being added in this PR) are not displayed in xref due to missing types in:
https://github.com/oracle/opengrok/blob/610d9085dcb52fc16b3f50d16e64cc03a62b7ea1/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/XrefStyle.java#L52

Should we match similar constructs together? E.g. `resource`, `provider`, etc. could have the `struct` style.

Thanks.


vladak:
> Should we match similar constructs together? E.g. `resource`, `provider`, etc. could have the `struct` style.

Either that or make the `XrefStyle` pluggable.

PR #3801, opened 2021-11-30
Sonar fixes in AnalyzerGuru

vladak:
Another round of Sonar fixes.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -870,19 +870,14 @@ public class AnalyzerGuru {
      *
      * Use if you just want to find file type.
      *
-     *
      * @param in The input stream containing the data
      * @param file The file name to get the analyzer for
      * @return the analyzer factory to use
-     * @throws java.io.IOException If a problem occurs while reading the data
+     * @throws java.io.IOException If a problem occurred while reading the data
      */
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1158,10 +1156,8 @@ private static String readOpening(InputStream in, byte[] sig)
 
             // If the opening starts with "#!", then track so that any
             // trailing whitespace after the hashbang is ignored.
-            if (opening.length() == 2) {
-                if (opening.charAt(0) == '#' && opening.charAt(1) == '!') {
-                    postHashbang = true;
-                }
+            if (opening.length() == 2 && opening.charAt(0) == '#' && opening.charAt(1) == '!') {
```
ahornace, line 1159: Wouldn't `opening.equals("#!")` be nicer?

vladak: Definitely nicer, thanks !

ahornace: My bad, I didn't notice that `opening` is a `StringBuilder`, in that case `toString` operation might be expensive if we execute it for every character.

vladak: ok, reverting

PR #3802, opened 2021-11-30
make kwd protected in Consts

vladak:
Seems like the `kwd` can be made protected in the various `Consts` since they are used in the lexers that belong to the same package. This addresses bunch of issues reported by Sonar.


ahornace (commented):
It seems that `Consts` classes have `private` constructor so they cannot be extended. In that case we can make the `kwd` fields `package-private` (no/default modifier).

vladak:
> It seems that `Consts` classes have `private` constructor so they cannot be extended. In that case we can make the `kwd` fields `package-private` (no/default modifier).

Makes sense.

PR #3805, opened 2021-11-30
fix Sonar issues in CtagsReader

vladak:
Fix for another issue identified by Sonar.


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/CtagsReader.java
```diff
@@ -142,7 +142,7 @@
             boolean match;
             for (tagFields x : tagFields.values()) {
                 match = true;
-                for (i = 0; i <= charCmpEndOffset; i++) {
+                for (i = 0; i <= CHAR_CMP_END_OFFSET; i++) {
```
ahornace, line 145: Since you are touching the code, then I'd move `int i` declaration into the `for` as is the industry standard. E.g.
`for (int i = 0; i <= CHAR_CMP_END_OFFSET; i++) {`

PR #3806, opened 2021-11-30
Suggester Sonar fixes

vladak:
This change fixes bunch of minor issues in Suggester.

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/Suggester.java
+++ b/suggester/src/main/java/org/opengrok/suggest/Suggester.java
@@ -238,7 +238,7 @@ public final class Suggester implements Closeable {
                 Duration d = Duration.between(start, Instant.now());
                 LOGGER.log(Level.FINE, "Finished initialization of {0}, took {1}", new Object[] {indexDir, d});
             } catch (Exception e) {
-                LOGGER.log(Level.SEVERE, "Could not initialize suggester data for " + indexDir, e);
+                LOGGER.log(Level.SEVERE, String.format("Could not initialize suggester data for %s", indexDir), e);
             }
         };
     }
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -199,7 +199,9 @@ public void init(final Collection<NamedIndexDir> luceneIndexes) {
      * @throws InterruptedException
      */
     public void waitForInit(long timeout, TimeUnit unit) throws InterruptedException {
-        initDone.await(timeout, unit);
+        if (!initDone.await(timeout, unit)) {
+            LOGGER.log(Level.WARNING, "await timed out due to timeout");
```
ahornace, line 203: `timed out due to timeout` sounds a bit funny :D maybe something like this would be better?
`Initialization did not finish in {0} {1}, new Object[] {timeout, unit})`

suggester/src/main/java/org/opengrok/suggest/Suggester.java
```diff
@@ -328,7 +330,9 @@ public void waitForRebuild(long timeout, TimeUnit unit) throws InterruptedExcept
         rebuildLock.lock();
         try {
             while (rebuilding) {
-                rebuildDone.await(timeout, unit);
+                if (!rebuildDone.await(timeout, unit)) {
+                    LOGGER.log(Level.WARNING, "await timed out due to timeout");
```
ahornace, line 334: Again, maybe something like:
`Rebuild did not finish in {0} {1}, new Object[] {timeout, unit})`

vladak: Thanks !

PR #3807, opened 2021-11-30
Sonar fixes in Definitions

vladak:
This change addresses bunch of issues in `Definitions` found by Sonar and IDEA.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
@@ -264,9 +264,9 @@ public class Definitions implements Serializable {
 
     public void addTag(int line, String symbol, String type, String text,
             String namespace, String signature, int lineStart, int lineEnd) {
-        Tag new_tag = new Tag(line, symbol, type, text, namespace, signature,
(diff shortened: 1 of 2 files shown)


vladak:
I wanted to change the `addTag()` to use `computeIfAbsent()`, however there seems to be some different semantics (likely related to the `null` map value handling) that makes `PascalXrefTest` fail so I will save it for another day.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
```diff
@@ -60,27 +60,27 @@
     public static class LineTagMap implements Serializable {
 
         private static final long serialVersionUID = 1191703801007779481L;
-        private final Map<String, Set<Tag>> sym_tags; //NOPMD
+        private final Map<String, Set<Tag>> symTags; //NOPMD
```
ahornace, line 63: Is this going to work? I think this breaks backwards compatibility as serialized classes contain the names of the fields (at least based on the simple test I did).

vladak: Good catch, reverting.

PR #3822, opened 2021-12-02
reduce field visibility in PendingToken

vladak:
This change makes the `nonpos` field package-private.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/PendingToken.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/PendingToken.java
@@ -18,6 +18,7 @@
  */
 
 /*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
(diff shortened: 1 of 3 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/PendingToken.java
```diff
@@ -34,11 +35,12 @@
     public final String str;
     public final int start;
     public final int end;
+
     /**
      * When tokenizers allow overlapping tokens, the following field is set to
      * {@code true} for tokens that should not increment the position attribute.
      */
-    public boolean nonpos;
+    boolean nonpos;
```
ahornace, line 43: It seems this is not used, even 3 years after it was introduced. I'd personally follow YAGNI and remove it.

ahornace (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/PendingToken.java
vladak: okay, removed.

Issue #1114, opened 2016-06-17
incremental history sometimes doesn't add proper records


tarzanek:
repos indexed from scratch and then incrementally indexed few times after gets history records missing for certain files
we've seen this usually on mercurial repos (might happen on others too)
it needs to be investigated how/why can this happen with incremental index and fixed, marking as stopper for now ...

tarzanek:
defer, we don't have proper steps to repro, nor root cause

vladak:
I can still see this problem on our production instance running 1.7.23 in bigger Mercurial repository that has renamed file handling enabled. The sync happens every 4 hours and usually the repository receives only a handful of changesets.

vladak:
Looking at the code in `get()`: https://github.com/oracle/opengrok/blob/110674a804dc9faa9a665dbac513689aaad3eca2/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java#L674-L686
makes me wonder whether the following situation could cause the problem (assuming Git or Mercurial repository):
1. repository is updated on disk (assuming it was indexed previously). This will cause a file to be updated.
1. someone triggers `get()` for the file. Given the cache for the file is now stale, this will make it updated. The `storeFile()` is called with `mergeHistory = false` so the cache file will be overwritten completely.
1. indexer is run. It will grab the latest indexed changeset (from the `OpenGroklatestRev` file), the incoming changesets will be extracted, converted to the inverted map and the incremental changesets will be added to the history cache of the file.

This sort of scenario would produce duplicate entries in the history cache files, however this is not what I am seeing - it is rather truncated history. Anyhow, the above code block is suspicious and I'd rather remove it.

vladak:
Comparing the number of history entries for each cached file in a repo with some 42k files: a checkout that was recently reindexed from scratch had generally more history entries per file than a checkout of the same repo whose history cache has been there for much longer:
```
find usr/src -type f -name '*.gz' | while read f; do n=$( gzcat $f | grep HistoryEntry |wc -l ); m=$( gzcat ../from-scatch-repo/$f | grep HistoryEntry |wc -l); if [[ $n != $m ]]; then echo "$f $n $m"; fi; done
```
The number of cache files that had different number of history entries was around 5k which given the arguably light traffic the service receives combined with the fact that the more fresh history cache has always more history entries per file does not support the hypothesis that the problem is in the forced cache refresh done by the `get()` method.

[...]

PR #3826, opened 2021-12-02
log merged/written revisions in FileHistoryCache

vladak:
This change is aimed at aiding debugging of #1114. I suspect there is either something fishy in the history merging process or the forced history cache refresh in `get()`. Having the debug log capability in the release allows me to use in on our production instance.

I fixed some minor issues in the (neighboring) code and augmented the `Repositoryinfo` copy constructor (which might be actually relevant to the issue), otherwise the semantics should be retained.

Both the indexer and webapp would have to run with `FINEST` log level in order to get as much data about the problem as possible.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
@@ -676,15 +676,20 @@ class FileHistoryCache implements HistoryCache {
             return null;
         }
 
-        // Either the cache is stale or retrieving the history took too long, cache it!
-        if (!file.isDirectory() && (cacheFile.exists() || (time > env.getHistoryReaderTimeLimit()))) {
-            // Also, don't cache history-information for directories, since the
-            // history information on the directory may change if a file in
-            // a sub-directory change. This will cause us to present a stale
-            // history log until the current directory is updated and
-            // invalidates the cache entry.
-            LOGGER.log(Level.FINEST, "getting history for ''{0}'' took longer than {1} ms, caching it [{2}]",
-                    new Object[]{file, env.getHistoryReaderTimeLimit(), history.getRevisionList()});
+        // Don't cache history-information for directories, since the
+        // history information on the directory may change if a file in
+        // a sub-directory change. This will cause us to present a stale
+        // history log until the current directory is updated and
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -245,13 +245,17 @@ static History readCache(File file) throws IOException {
      * @param cacheFile the file to store the history to
      */
     private void writeHistoryToFile(File dir, History history, File cacheFile) throws HistoryException {
+
+        LOGGER.log(Level.FINEST, "writing history entries to ''{0}'': {1}",
+                new Object[]{cacheFile, history.getRevisionList()});
```
ahornace, line 250: It seems a bit weird calling `history.getRevisionList()` which creates a new list for a `FINEST` log but is executed every time. Maybe using the supplier like `LOGGER.log(Level.FINEST, () -> ...)` would be better?

However, maybe I'm just overthinking this, the actual performance impact should be negligible.

vladak: That's a good point. I wrapped the added `FINEST` logs inside `isLoggable()`.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileHistoryCache.java
```diff
@@ -666,23 +672,25 @@ public History get(File file, Repository repository, boolean withFiles)
...
-                storeFile(history, file, repository);
+            // Either the cache is stale or retrieving the history took too long, cache it!
+            if (cacheFile.exists()) {
+                LOGGER.log(Level.FINEST, "refreshing history for ''{0}'': {1}",
+                        new Object[]{file, history.getRevisionList()});
+            } else if (time > env.getHistoryReaderTimeLimit()) {
+                LOGGER.log(Level.FINEST, "getting history for ''{0}'' took longer than {1} ms, caching it: {2}",
+                        new Object[]{file, env.getHistoryReaderTimeLimit(), history.getRevisionList()});
             }
+            storeFile(history, file, repository);
```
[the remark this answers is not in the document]
vladak: D'oh! definitely. Thanks. Should not do changes late at night.

PR #3827, opened 2021-12-02
Add additional printable chars to full index

ahornace:
This change adds additional chars to the list of printable characters that is being recognized by full tokenizer.

This change will be especially useful for `""` queries that could now return proper results even for the following:
- `!= null`
- `new int[0]`
- `#!/bin/`
- `/**`

Thanks to @Aldarrion for the suggestion.


ahornace:
It seems that some other additional chars like `<>(),;~` are being ignored, will investigate those. Converting to draft for the time being.

ahornace:
> It seems that some other additional chars like `<>(),;~` are being ignored, will investigate those. Converting to draft for the time being.

My bad, didn't recompile the lexer :D

PR #3828, opened 2021-12-03
FileHistoryCache#get() should not modify history cache

vladak:
This change removes the ability of `FileHistoryCache#get()` to modify the history cache as this can lead to cache corruption. I find this approach more robust than trying to add detection. This solves the problem described in https://github.com/oracle/opengrok/issues/1114#issuecomment-985059373

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -108,11 +108,6 @@ public final class Configuration {
      */
     private boolean historyCache;
     /**
-     * The maximum time in milliseconds {@code HistoryCache.get()} can take
-     * before its result is cached.
-     */
-    private int historyCacheTime;
-    /**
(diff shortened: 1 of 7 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -807,28 +806,6 @@ public void setHistoryCache(boolean historyCache) {
...
-    /**
-     * How long can a history request take before it's cached? If the time is
-     * exceeded, the result is cached. This setting only affects
-     * {@code FileHistoryCache}.
-     *
-     * @return the maximum time in milliseconds a history request can take
-     * before it's cached
-     */
-    public int getHistoryCacheTime() {
-        return historyCacheTime;
```
ahornace, line 819: Maybe also remove the `historyCacheTime` field?

Isn't this going to break the deserialization if the field is present in XML?

vladak: yes and yes, will add a note to the release.

PR #3830, opened 2021-12-06
Suggester rebuild vs undeploy

vladak:
This change fixes suggester rebuild termination when web app undeploy is in progress.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
@@ -105,7 +105,7 @@ public final class WebappListener
         }
 
         // Check index(es).
-        check_index(env);
+        checkIndex(env);
 
         env.startExpirationTimer();
         startupTimer.record(Duration.between(start, Instant.now()));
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
```diff
@@ -116,21 +116,21 @@ public void contextInitialized(final ServletContextEvent servletContextEvent) {
...
      */
-    private void check_index(RuntimeEnvironment env) {
+    private void checkIndex(RuntimeEnvironment env) {
         if (env.isProjectsEnabled()) {
             LOGGER.log(Level.FINE, "Checking indexes for all projects");
             Map<String, Project> projects = env.getProjects();
             File indexRoot = new File(env.getDataRootPath(), IndexDatabase.INDEX_DIR);
             if (indexRoot.exists()) {
-                for (String projectName : projects.keySet()) {
+                projects.keySet().forEach(projectName -> {
```
ahornace, line 125: I think that it doesn't make sense rewriting all `enhanced-for` loops to `forEach`. 

https://stackoverflow.com/questions/16635398/java-8-iterable-foreach-vs-foreach-loop

ahornace (approved):

opengrok-web/src/main/java/org/opengrok/web/WebappListener.java
vladak: Thanks for the pointer. Actually I wanted to use entrySet() and get lured into using `forEach` by the Sonar plugin in IDEA.

PR #3831, opened 2021-12-07
Diff page html cleanup

vladak:
This change replaces the `b` and `i` elements in the diff page with `strong` and `em` elements, respectively.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/Scripts.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/Scripts.java
@@ -34,7 +34,7 @@ import java.util.TreeSet;
 import org.webjars.WebJarAssetLocator;
 
 /**
- * A list-like container for javascripts in the page.
+ * A list-like container for JavasScript sources in JSP/HTML pages.
  *
  * @author Krystof Tulinger
  */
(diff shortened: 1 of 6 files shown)


ahornace (approved):
I don't think that `<b>` or `<i>` are deprecated and should not be used. It depends on the context. So for example I don't see a problem using `<i>` to style the line numbers (all of them look the same and there is no emphasis on them).

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#b_vs._strong
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#i_vs._em

vladak:
> I don't think that `<b>` or `<i>` are deprecated and should not be used. It depends on the context. So for example I don't see a problem using `<i>` to style the line numbers (all of them look the same and there is no emphasis on them).
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#b_vs._strong https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#i_vs._em

Thanks for the links. The change deserves more thought, I think. While for `help.jsp` the `<strong>` is warranted I think, the change of `<i>` to `<em>` in `diff.jsp` should probably be replaced by CSS as we do not really put an emphasis but to make the font italics.

vladak:
> Thanks for the links. The change deserves more thought, I think. While for `help.jsp` the `\<strong\>` is warranted I think, the change of `<i>` to `\<em\>` in `diff.jsp` should probably be replaced by CSS as we do not really put an emphasis but to make the font italics.

Changed the `<em>` to use `<span>` with class `it` that sets the font style to italics. This will avoid semantic tags and gives greater flexibility w.r.t. the style.

PR #3833, opened 2021-12-09
lucene 9.0.0

tarzanek:
this changes format of lucene index

@ahornace can you check the visitors if they should be like that?
(seems it depends on subqueries, but I don't understand if those wrappers have any , other samples how they implement it:
https://github.com/apache/lucene/search?p=2&q=visit%28QueryVisitor+queryVisitor%29 )


tarzanek:
ah ... asi to bude chciet aj update toho testu 
```
Error:  Failures: 
Error:    IndexCheckTest.testIndexVersionOldIndex:140 Unexpected exception type thrown ==> expected: <org.opengrok.indexer.index.IndexCheck.IndexVersionException> but was: <org.apache.lucene.index.IndexFormatTooOldException>
Error:    SearchHelperTest.testSearchAfterReindex:160 expected: <5> but was: <3>
```

PR #3835, opened 2021-12-10
add alt attribute to images in xref

vladak:
This change should silence Sonar warning about missing the `alt` attribute in images in `xref.jspf`.

### DIFF
--- a/opengrok-web/src/main/webapp/xref.jspf
+++ b/opengrok-web/src/main/webapp/xref.jspf
@@ -129,7 +129,7 @@ org.opengrok.indexer.web.QueryParameters"
                             defs, annotation, project);
                     } else if (g == AbstractAnalyzer.Genre.IMAGE) {
         %></pre>
-    <img src="<%= rawPath %>?<%= QueryParameters.REVISION_PARAM_EQ %><%= Util.URIEncode(rev) %>" alt="<%= path %>"/>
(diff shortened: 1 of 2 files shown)


opengrok-web/src/main/webapp/xref.jspf
```diff
@@ -129,7 +129,7 @@ org.opengrok.indexer.web.QueryParameters"
                             defs, annotation, project);
                     } else if (g == AbstractAnalyzer.Genre.IMAGE) {
         %></pre>
-    <img src="<%= rawPath %>?<%= QueryParameters.REVISION_PARAM_EQ %><%= Util.URIEncode(rev) %>"/>
+    <img src="<%= rawPath %>?<%= QueryParameters.REVISION_PARAM_EQ %><%= Util.URIEncode(rev) %>" alt="<%= path %>"/>
```
ahornace, line 132: Maybe use `"Image from Source Repository"` to be consistent with https://github.com/oracle/opengrok/blob/539b5afffee12c5504fba86b2e8efbfe9f22839b/opengrok-web/src/main/webapp/xref.jspf#L180 ?

ahornace (approved):

opengrok-web/src/main/webapp/xref.jspf
vladak: That might do it, thanks.

PR #3845, opened 2021-12-21
fix Sonar issues in PluginFramework

vladak:
This change fixes 2 issues in `PluginFramework` found by Sonar.


opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -108,7 +108,7 @@ public synchronized void setPluginDirectory(File pluginDirectory) {
      *
      * @param directory the directory path
      */
-    public void setPluginDirectory(String directory) {
+    public synchronized void setPluginDirectory(String directory) {
```
vladak, line 111: Technically this is not needed. It silences Sonar and works because the lock used is reentrant.

PR #3847, opened 2021-12-22
Move repositories to test resources

### LINKED ISSUE #2244: get rid of dependency on Ant
Even after the changes in #2032, Maven still calls Ant to perform certain tasks (in the `opengrok-indexer` module, for the `-collect-lex-lexh`, `-update-build-info` and `-post-compile-test` Ant targets) which is strange to say the least.

ahornace:
fixes #2244

This removes the ant dependency and should make the testing a bit more straightforward. Thanks!


opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestRepository.java
```diff
@@ -53,6 +54,23 @@
...
 
+    private static final Map<Path, Path> renameMappings = new LinkedHashMap<>(Map.of(
+            Path.of("bazaar", "bzr"), Path.of("bazaar", ".bzr"),
+
+            Path.of("bitkeeper", "bk", "BitKeeper", "etc", "SCCS_dir"),
+            Path.of("bitkeeper", ".bk", "BitKeeper", "etc", "SCCS"),
+            Path.of("bitkeeper", "bk", "SCCS_dir"),
+            Path.of("bitkeeper", ".bk", "SCCS"),
+
+            Path.of("bitkeeper", "bk"), Path.of("bitkeeper", ".bk"),
```
vladak, line 65: perhaps keep each `Path.of()` on separate line ?

ahornace: Sure, wanted to keep the key -> value mapping on the same line but some were just too long :D probably better to make it consistent.

pom.xml
```diff
@@ -412,6 +412,14 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-resources-plugin</artifactId>
+                    <version>3.2.0</version>
+                    <configuration>
+                        <addDefaultExcludes>false</addDefaultExcludes>
```
vladak, line 420: I assume this is because of the various `git` paths etc.

ahornace: https://maven.apache.org/plugins/maven-resources-plugin/resources-mojo.html#addDefaultExcludes

I've added it mostly because of `BitKeeper ` as the other directories were already renamed somehow. I was just surprised that they were ignored in the first place. If needed, I can just create more mappings and remove this setting.

opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestRepository.java
vladak: or insert empty line between each `Path.of()`

opengrok-indexer/src/test/java/org/opengrok/indexer/util/TestRepository.java
```diff
@@ -84,19 +105,37 @@ public void create(@NotNull final URL url) throws IOException, URISyntaxExceptio
...
         try (Stream<Path> stream = Files.walk(src)) {
-            stream.forEach(source -> {
-                if (source.equals(src)) {
+            stream.forEach(sourceFile -> {
+                if (sourceFile.equals(src)) {
                     return;
                 }
                 try {
-                    Files.copy(source, dest.resolve(src.relativize(source).toString()), REPLACE_EXISTING);
+                    Path destRelativePath = getDestinationRelativePath(src, sourceFile);
```
vladak, line 113: I was about to say this might be refactored :-)

pom.xml
vladak: I am confused. So even with the `addDefaultExcludes=false` the plugin actually moves/renames some of these entries but not all ?

ahornace: >I am confused. So even with the addDefaultExcludes=false the plugin actually moves/renames some of these entries but not all ?

No, without `addDefaultExcludes=false` the files that are matching these patterns are completely ignored and are not being copied to `target` directory:
 ```
Misc: **/*~, **/#*#, **/.#*, **/%*%, **/._*
CVS: **/CVS, **/CVS/**, **/.cvsignore
RCS: **/RCS, **/RCS/**
SCCS: **/SCCS, **/SCCS/**
VSSercer: **/vssver.scc
MKS: **/project.pj
SVN: **/.svn, **/.svn/**
GNU: **/.arch-ids, **/.arch-ids/**
Bazaar: **/.bzr, **/.bzr/**
SurroundSCM: **/.MySCMServerInfo
Mac: **/.DS_Store
Serena Dimension: **/.metadata, **/.metadata/**
Mercurial: **/.hg, **/.hg/**
GIT: **/.git, **/.gitignore, **/.gitattributes, **/.git/**
Bitkeeper: **/BitKeeper, **/BitKeeper/**, **/ChangeSet, **/ChangeSet/**
Darcs: **/_darcs, **/_darcs/**, **/.darcsrepo, **/.darcsrepo/****/-darcs-backup*, **/.darcs-temp-mail
 ```

vladak: Okay, I understand now. With `addDefaultExcludes=false` the plugin will copy everything from the resources directory to the output directory.

PR #3852, opened 2021-12-23
Add MethodName CheckStyle check

ahornace:
Just another simple check.

Thanks.


vladak:
I bet this will clean some of the Sonar reported issues.

PR #3855, opened 2021-12-30
deal with exceptions in asynchronous API tasks

vladak:
This change is follow up to the asynchronous API introduction in #3853. It adds the possibility to set payload in the final response. Either to the string representation of an exception (if one was thrown) or the string representation of object returned from the `Callable`.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
@@ -185,15 +185,13 @@ public class ApiTask {
                 throw new IllegalStateException(String.format("API task %s already submitted", this));
             }
 
-            return new Callable<>() {
-                public Object call() throws Exception {
-                    LOGGER.log(Level.FINE, "API task {0} started", this);
-                    setSubmitted();
-                    Object ret = callable.call();
-                    setCompleted();
-                    LOGGER.log(Level.FINE, "API task {0} done", this);
-                    return ret;
-                }
+            return () -> {
+                LOGGER.log(Level.FINE, "API task {0} started", this);
+                setSubmitted();
+                Object ret = callable.call();
+                setCompleted();
+                LOGGER.log(Level.FINE, "API task {0} done", this);
+                return ret;
             };
         }
     }
--- a/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
@@ -166,10 +166,10 @@ public class ApiTask {
                 public Object call() throws Exception {
                     LOGGER.log(Level.FINE, "API task {0} started", this);
                     setSubmitted();
-                    callable.call();
+                    Object ret = callable.call();
                     setCompleted();
                     LOGGER.log(Level.FINE, "API task {0} done", this);
-                    return null;
+                    return ret;
                 }
             };
         }
--- a/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
@@ -145,6 +145,12 @@ public class ApiTask {
         }
     }
 
+    /**
+     * Provides simple Exception to status code mapping. The Exception match is exact, i.e. exception class hierarchy
+     * is not considered.
+     * @param exception Exception
+     * @return Response status
+     */
     private Response.Status mapExceptionToStatus(ExecutionException exception) {
         return exceptionStatusMap.getOrDefault(exception.getCause().getClass(), Response.Status.INTERNAL_SERVER_ERROR);
     }
--- a/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
@@ -96,7 +96,7 @@ public class ApiTask {
     /**
      * @return response status
      */
-    public Response.Status getResponseStatus() {
+    Response.Status getResponseStatus() {
         return responseStatus;
     }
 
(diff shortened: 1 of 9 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
```diff
@@ -50,68 +55,145 @@
...
                 throw new IllegalStateException(String.format("API task %s already submitted", this));
             }
 
-            return () -> {
-                LOGGER.log(Level.FINE, "API task {0} started", this);
-                setSubmitted();
-                runnable.run();
-                setCompleted();
-                LOGGER.log(Level.FINE, "API task {0} done", this);
+            return new Callable<>() {
```
ahornace, line 188: I think you can still use lambda.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
```diff
@@ -50,68 +55,145 @@
...
+    public boolean isDone() {
+        if (future != null) {
+            return future.isDone();
+        } else {
+            return false;
+        }
+    }
+
+    private Response.Status mapExceptionToStatus(ExecutionException exception) {
+        return exceptionStatusMap.getOrDefault(exception.getCause().getClass(), Response.Status.INTERNAL_SERVER_ERROR);
```
ahornace, line 149: This will work only if there is an exact match, what if there are subtypes? E.g. `IOException` will not catch `FileNotFoundException`. Is it okay?

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
```diff
@@ -50,68 +55,145 @@
...
+     */
+    public Response getResponse() {
+        // Avoid thread being blocked in future.get() below.
+        if (!isDone()) {
+            throw new IllegalStateException(String.format("task %s not yet done", this));
+        }
+
+        Object obj;
+        try {
+            obj = future.get();
```
ahornace, line 165: `future` can be `null` here.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
```diff
@@ -50,68 +55,145 @@
...
 
-    public void setCompleted() {
+    void setCompleted() {
         state = ApiTaskState.COMPLETED;
     }
 
+    /**
+     * @param future Future object used for tracking the progress of the API task
+     */
+    public void setFuture(Future<Object> future) {
```
ahornace, line 133: Maybe not `public`?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
```diff
@@ -68,6 +68,7 @@ public String get() {
 
     @PUT
     @Consumes(MediaType.APPLICATION_XML)
+    @Produces("text/plain")
```
ahornace, line 71: Why these? Does it actually matter when there is no output?

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/StatusController.java
```diff
@@ -55,8 +55,8 @@ public Response getStatus(@PathParam("uuid") String uuid) {
             return Response.status(Response.Status.NOT_FOUND).build();
         }
 
-        if (apiTask.isCompleted()) {
-            return Response.status(apiTask.getResponseStatus()).build();
+        if (apiTask.isDone()) {
```
ahornace, line 58: I think this should be still using `isCompleted`, the same below

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
```diff
@@ -50,68 +55,145 @@
...
+        this.future = future;
+    }
+
+    /**
+     * @return whether the task is finished (normally or with exception)
+     */
+    public boolean isDone() {
```
ahornace, line 140: Why this? Isn't `isCompleted` an indicator already?

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
vladak: True, fixed.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
vladak: Hmm, the exception hierarchy is something I did not consider. I'd like to keep it simple (and a bit stupid) so I will just document the limitation.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
vladak: I think it is safe given the `isDone()` check above.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
vladak: Definitely !

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ConfigurationController.java
vladak: At one point Grizzly complained about not having the media type however this does not seem to be the case anymore. Reverting.

ahornace (approved):
Looks good!

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/StatusController.java
vladak: `getResponse()` relies on `isDone()` and handles exceptions. `isCompleted()` would indicate only successful result of the task.

opengrok-web/src/main/java/org/opengrok/web/api/ApiTask.java
vladak: I was a bit torn whether not to turn `isCompleted()` to `isDone()` however decided to leave it as it is quite handy for testing.

vladak:
> Looks good!

Thanks for the review !

PR #3860, opened 2022-01-04
negative cache for LdapUserPlugin 

vladak:
I decided to store the negative "flag" in the `LdapUser` object itself as an attribute (as opposed to e.g. having the `attributes` field being `null` - this brings all kinds of trouble along the way). This also means that the negative cache is per session. In Tomcat the default session duration is 30 minutes however the authnz changes on the server/identity side are usually not fast either.

### DIFF
--- a/plugins/src/main/java/opengrok/auth/plugin/LdapUserPlugin.java
+++ b/plugins/src/main/java/opengrok/auth/plugin/LdapUserPlugin.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2016, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2022, Oracle and/or its affiliates. All rights reserved.
  */
 package opengrok.auth.plugin;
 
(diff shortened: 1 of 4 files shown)


vladak:
@tulinkry can you can a look ?

ahornace (approved):
Looks good to me!

vladak:
Thanks !

plugins/src/main/java/opengrok/auth/plugin/LdapUserPlugin.java
```diff
@@ -157,8 +158,7 @@ public void fillSession(HttpServletRequest req, User user) {
         String dn = null;
```
ktulinger, line 158: the user is still set to `null` (if that is something you tried to avoid based on your description)

ktulinger: line 151

tulinkry, line 158: the user is still set to null (if that is something you tried to avoid based on your description) liine 151

tulinkry (commented):
I am late to the party. Looks good, I only worry when someone configures tomcat to have longer sessions (like day long). There is no mechanism to retry the ldap lookup in a shorter period.

I'd rather much see storing like a timestamp of last fetch and then retrying only after 30 minutes (can be configurable in the plugin) internally in the plugin, as opposed to expecting tomcat to have sane session expiration.

plugins/src/main/java/opengrok/auth/plugin/LdapUserPlugin.java
vladak: As far as I remember, this was a problem when I tried to use the `null` `LdapUser` session attribute as a indication of negative caching. Otherwise both `LdapAttrPlugin` and `LdapFilterPlugin` deal with it just fine.

PR #3866, opened 2022-01-14
cleanup timeout handling in tools

vladak:
This change addresses a problem with checking response status:
```
could not delete project 'foo' in web application: 'Response' object has no attribute 'status'
```
and also moves the timeout defaults to the programs.


kulikjak (approved):

vladak:
Thanks !

PR #3869, opened 2022-01-17
check commands structure in CommandSequenceBase()

vladak:
This change should catch more invalid configurations of `opengrok-sync`. It introduces the `CommandConfigurationException`.

Looks like this:
```
$ opengrok-sync -c sync-invalid.yml 
Assuming directory: /var/opengrok/src
Invalid configuration: command 'foo' is not a dictionary
```


ahornace (approved):
Looks good.

vladak:
Thanks !

PR #3873, opened 2022-01-23
changing set to map for groups, fixing issue #1587

JonathanGranado:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -47,7 +47,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeSet;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -130,7 +130,7 @@ public final class Configuration {
     private boolean authorizationWatchdogEnabled;
     private AuthorizationStack pluginStack;
     private Map<String, Project> projects; // project name -> Project
-    private Set<Group> groups;
+    private Map<String, Group> groups; // group name -> group
     private String sourceRoot;
     private String dataRoot;
     /**
@@ -537,7 +537,8 @@ public final class Configuration {
         setFetchHistoryWhenNotInCache(true);
         setFoldingEnabled(true);
         setGenerateHtml(true);
-        setGroups(new TreeSet<>());
+        setGroups(new TreeMap<>());
+    
         setGroupsCollapseThreshold(4);
         setHandleHistoryOfRenamedFiles(false);
         setHistoryCache(true);
@@ -847,25 +848,31 @@ public final class Configuration {
     }
 
     /**
-     * Adds a group to the set. This is performed upon configuration parsing
+     * Adds a group to the map. This is performed upon configuration parsing
      *
      * @param group group
      * @throws IOException when group is not unique across the set
      */
     public void addGroup(Group group) throws IOException {
-        if (!groups.add(group)) {
+        if(!groups.containsKey(group.getName())){
             throw new IOException(
                     String.format("Duplicate group name '%s' in configuration.",
                             group.getName()));
-        }
+            }else{
+                groups.put(group.getName(), group);
+            }
     }
 
-    public Set<Group> getGroups() {
+    public Map<String, Group> getGroups() {
         return groups;
     }
 
-    public void setGroups(Set<Group> groups) {
-        this.groups = groups;
+    public void setGroups(Map<String, Group> groups) {
+        Map<String, Group> settedGroup = new HashMap<String, Group>();
+        for (Group entry : groups.values()){
+            settedGroup.put(entry.getName(), entry);
+        }
+        this.groups = settedGroup;
     }
 
     public String getSourceRoot() {
@@ -1480,22 +1487,25 @@ public final class Configuration {
         // This ensures that when the configuration is reloaded then the set
         // contains only root groups. Subgroups are discovered again
         // as follows below
-        conf.groups.removeIf(g -> g.getParent() != null);
+        conf.groups.entrySet().removeIf( entries -> entries.getValue().getParent() != null);
 
         // Traversing subgroups and checking for duplicates,
         // effectively transforms the group tree to a structure (Set)
         // supporting an iterator.
-        TreeSet<Group> copy = new TreeSet<>();
-        LinkedList<Group> stack = new LinkedList<>(conf.groups);
+        Map<String, Group> copy = new TreeMap<>();
+        LinkedList<Group> stack = new LinkedList<>(conf.groups.values());
         while (!stack.isEmpty()) {
             Group group = stack.pollFirst();
             stack.addAll(group.getSubgroups());
 
-            if (!copy.add(group)) {
+            if (!copy.containsKey(group.getName())) {
                 throw new IOException(
                         String.format("Duplicate group name '%s' in configuration.",
                                 group.getName()));
+            }else{
+                copy.put(group.getName(), group);
             }
+            
 
             // populate groups where the current group in in their subtree
             Group tmp = group.getParent();
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
@@ -23,8 +23,11 @@
  */
 package org.opengrok.indexer.configuration;
 
+import java.util.Collection;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
@@ -287,7 +290,7 @@ public class Group implements Comparable<Group>, Nameable {
         Group ret = null;
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         if (env.hasGroups()) {
-            for (Group grp : env.getGroups()) {
+            for (Group grp : env.getGroups().values()) {
                 if (name.equals(grp.getName())) {
                     ret = grp;
                 }
@@ -301,12 +304,12 @@ public class Group implements Comparable<Group>, Nameable {
      * the project's description.
      *
      * @param project the project
-     * @param groups set of groups
+     * @param map set of groups
      * @return set of groups matching the project
      */
-    public static Set<Group> matching(Project project, Set<Group> groups) {
-        Set<Group> copy = new TreeSet<>(groups);
-        copy.removeIf(g -> !g.match(project));
+    public static Map<String, Group> matching(Project project, Map<String, Group> map) {
+        Map<String, Group> copy = new TreeMap<>(map);
+        ((Collection<Group>) copy).removeIf(g -> !g.match(project));
         return copy;
     }
 }
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
@@ -28,8 +28,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Locale;
-import java.util.Set;
-import java.util.TreeSet;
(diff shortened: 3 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -847,25 +848,31 @@ public void setProjects(Map<String, Project> projects) {
...
     /**
-     * Adds a group to the set. This is performed upon configuration parsing
+     * Adds a group to the map. This is performed upon configuration parsing
      *
      * @param group group
      * @throws IOException when group is not unique across the set
      */
     public void addGroup(Group group) throws IOException {
-        if (!groups.add(group)) {
+        if(!groups.containsKey(group.getName())){
```
ahornace, line 857: Wrong formatting, checkstyle will probably fail. Please try to compile the code locally by `./mvnw compile`.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -847,25 +848,31 @@ public void setProjects(Map<String, Project> projects) {
...
 
-    public Set<Group> getGroups() {
+    public Map<String, Group> getGroups() {
         return groups;
     }
 
-    public void setGroups(Set<Group> groups) {
-        this.groups = groups;
+    public void setGroups(Map<String, Group> groups) {
+        Map<String, Group> settedGroup = new HashMap<String, Group>();
```
ahornace, line 871: In constructor, there is a `TreeMap` so I'd assume it was used for a reason. If tests pass, then using `HashMap` is okay; however, please use "copy constructor" like: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/TreeMap.html#%3Cinit%3E(java.util.Map)

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -1480,22 +1487,25 @@ private static Configuration decodeObject(InputStream in) throws IOException {
         // This ensures that when the configuration is reloaded then the set
         // contains only root groups. Subgroups are discovered again
         // as follows below
-        conf.groups.removeIf(g -> g.getParent() != null);
+        conf.groups.entrySet().removeIf( entries -> entries.getValue().getParent() != null);
```
ahornace, line 1490: It's just one `entry`

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -1480,22 +1487,25 @@ private static Configuration decodeObject(InputStream in) throws IOException {
         // This ensures that when the configuration is reloaded then the set
```
ahornace, line 1487: set -> map ?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
```diff
@@ -301,12 +304,12 @@ public static Group getByName(String name) {
      * the project's description.
      *
      * @param project the project
-     * @param groups set of groups
+     * @param map set of groups
```
ahornace, line 307: map of groups

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
```diff
@@ -301,12 +304,12 @@ public static Group getByName(String name) {
...
      * @return set of groups matching the project
      */
-    public static Set<Group> matching(Project project, Set<Group> groups) {
-        Set<Group> copy = new TreeSet<>(groups);
-        copy.removeIf(g -> !g.match(project));
+    public static Map<String, Group> matching(Project project, Map<String, Group> map) {
+        Map<String, Group> copy = new TreeMap<>(map);
+        ((Collection<Group>) copy).removeIf(g -> !g.match(project));
```
ahornace, line 312: This will not work. `Map` is not a `Collection`:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
```diff
@@ -301,12 +304,12 @@ public static Group getByName(String name) {
...
      *
      * @param project the project
-     * @param groups set of groups
+     * @param map set of groups
      * @return set of groups matching the project
      */
-    public static Set<Group> matching(Project project, Set<Group> groups) {
-        Set<Group> copy = new TreeSet<>(groups);
-        copy.removeIf(g -> !g.match(project));
+    public static Map<String, Group> matching(Project project, Map<String, Group> map) {
```
ahornace, line 310: Rename `map` to `groups`

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1545,30 +1546,30 @@ private void generateProjectRepositoriesMap() throws IOException {
      * If any of the groups contain some projects or repositories already,
      * these get discarded.
      *
-     * @param groups   set of groups to be filled with matching projects
+     * @param map   set of groups to be filled with matching projects
      * @param projects projects to classify
      */
-    public void populateGroups(Set<Group> groups, Set<Project> projects) {
-        if (projects == null || groups == null) {
+    public void populateGroups(Map<String, Group> map, Map<String, Project> projects) {
```
ahornace, line 1552: Rename `map` to `groups`

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
ahornace, line 310: Do you need to change the return value to map? It seems that the only usage just calls `.values()` immediately.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -1636,11 +1637,12 @@ public synchronized void setConfiguration(Configuration configuration, List<Stri
         }
 
         // populate groups is dependent on repositories map
-        populateGroups(getGroups(), new TreeSet<>(getProjects().values()));
+        populateGroups(getGroups(), new TreeMap<>(getProjects()));
 
         includeFiles.reloadIncludeFiles();
     }
 
+
```
ahornace, line 1645: Revert this blank line.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
ahornace, line 1552: I don't think the change from `Set` to `Map` for `projects` is guaranteed here.

ahornace (commented):
In order for your contribution to be accepted, you have to sign https://oca.opensource.oracle.com/?ojr=home -> I wasn't able to find your name there. Or at the very least, please state that you accept the terms in the PR.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
JonathanGranado: Thanks for the all the comments, I'll go ahead and fix everything

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
tulinkry: Originally a treeset was used so that the groups are ordered on the main page. This needs be respected.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
tulinkry: also the condition is wrong

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
tulinkry: the original code of this method does not do a copy, why do you need it now?

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -1480,22 +1487,25 @@ private static Configuration decodeObject(InputStream in) throws IOException {
...
-        TreeSet<Group> copy = new TreeSet<>();
-        LinkedList<Group> stack = new LinkedList<>(conf.groups);
+        Map<String, Group> copy = new TreeMap<>();
+        LinkedList<Group> stack = new LinkedList<>(conf.groups.values());
         while (!stack.isEmpty()) {
             Group group = stack.pollFirst();
             stack.addAll(group.getSubgroups());
 
-            if (!copy.add(group)) {
+            if (!copy.containsKey(group.getName())) {
```
tulinkry, line 1501: the condition is wrong

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: yeah projects can stay Set

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
```diff
@@ -287,7 +290,7 @@ public static Group getByName(String name) {
         Group ret = null;
         RuntimeEnvironment env = RuntimeEnvironment.getInstance();
         if (env.hasGroups()) {
-            for (Group grp : env.getGroups()) {
+            for (Group grp : env.getGroups().values()) {
```
tulinkry, line 293: The entire goal of the issue and therefore of this change was to avoid this loop

tulinkry (changes requested):

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
```diff
@@ -102,7 +102,7 @@
     /**
      * Set of groups which match this project.
      */
-    private transient Set<Group> groups = new TreeSet<>();
+    private transient Map<String, Group> groups = new TreeMap<>();
```
tulinkry, line 105: project doesnt need map of groups, set is enough

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Group.java
tulinkry: you dont need to return a Map, Set is enough

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
```diff
@@ -516,10 +517,10 @@ public boolean hasGroups() {
      *
      * @param groups the set of groups to use
      */
-    public void setGroups(Set<Group> groups) {
+    public void setGroups(Map<String, Group> groups) {
         syncWriteConfiguration(groups, (c, g) -> {
-            populateGroups(g, new TreeSet<>(getProjects().values()));
-            c.setGroups(g);
+            populateGroups(g, new TreeMap<>(getProjects()));
+            c.setGroups(groups);
```
tulinkry, line 523: you're not using the lambda parameter

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/RuntimeEnvironment.java
tulinkry: you dont even need map of groups for this method, just the Set. perhaps using the same trick as with getProjects() would do the trick

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
```diff
@@ -294,22 +294,18 @@ public void setMergeCommitsEnabled(boolean flag) {
      *
      * @return set of groups|empty if none
      */
-    public Set<Group> getGroups() {
+    public Map<String,Group> getGroups() {
```
tulinkry, line 297: wrong formatting

tulinkry (changes requested):

vladak:
This is more than 1 year old, without a response, hence closing.

PR #3875, opened 2022-01-28
set default for API response timeout in opengrok-sync

vladak:
Forgot to set the default for `opengrok-sync` in PR #3866.


ahornace (approved):

vladak:
Thanks !

PR #3876, opened 2022-01-28
rename displayRepositories getter to match the return type

vladak:
The methods for checking whether to display repositories/projects on the index page should be called `isDisplayRepositories()` to match the boolean.


ahornace (approved):

vladak:
Thanks !

Issue #3872, opened 2022-01-22
"opengrok-mirror"without effect，cannot update source code and index


chenchuanliang:
When I use "opengrok-mirror",source code and index cannot be updated.
I found some level 2 directory( such as E60/ADSP )'s source code updated and reindex
But If level 2 directory not have git( such as E60/android , E60/android/package/apps/Settings have git) can not update source code and index.
![2022-01-20 10-36-22屏幕截图](https://user-images.githubusercontent.com/20430797/150623035-6295a73c-b12b-41d5-9a5a-5f81b33a5e8b.png)
You can see:SQ53 no update,but the actual source code have changes.source code and index no changes when run "opengrok-mirror".
So I checked the `tools/env/lib/python3.8/site-packages/opengrok_tools/scm/git.py` and `repo.py`,changed the `incoming_check` command to `['repo','forall','-c','git','pull']`, source code can be updated,but index not.

chenchuanliang:
Or change "def incoming_check(self)" return True,check every time,Is it feasible？

vladak:
I am not familiar with the E60/ADSP structure. I assume this is something similar to the AOSP structure, i.e. using the [repo] tool (https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md) to manage bunch of Git repositories.

The output from `opengrok-mirror` looks as if the initial `repo sync` was done directly to the OpenGrok source root. Anyhow, the problem might be that `opengrok-mirror` tries to descend into the sub-repositories for the `repo` managed tree instead of using the `repo` directly for the project top level directory.

vladak:
> Or change "def incoming_check(self)" return True,check every time,Is it feasible？

Technically, this is actually what happens currently. `repo sync -n` will perform fetch only (i.e. no working tree update) so it boils down to it returning 0 exit value and producing non empty output which it normally does, hence `repo.py#incoming_check()` will normally return true which means if `opengrok-mirror` is run with the `-I` option it will perform the actual synchronization and will indicate via its return value that the indexer can be run (if running it via `opengrok-sync`).

vladak:
In general, `opengrok-mirror` should stop when it encounters a `repo` type repository, use its `incoming_check()`/`reposync()` and do not descend into the Git sub-repositories. The web app should still provide treat the sub-repositories separately in order to get the correct history.

PR #3880, opened 2022-01-28
fix repo repository mirroring

vladak:
While playing with `repo` for #3872, I discovered the support for it is broken in `opengrok-mirror`. Firstly, the `RepoRepository`  init function has wrong order or arguments, secondly the `repo` synchronization should not descend into sub-repositories.


tulinkry (approved):

vladak:
thanks @tulinkry !

PR #3884, opened 2022-01-31
fix getAuthors() in SCCSRepository

vladak:
Looking at the output from the `SCCSRepositoryTest#testAnnotation()` I can see a warning:
```
WARNING: Non-zero exit status 1 from command [sccs prs -e -d $':I: :P:' /tmp/source11618603591071727860/teamware/main.c] in directory /tmp/source11618603591071727860/teamware: prs: Cannot open SCCS file SCCS/s.:I: :P: for reading : No such file or directory
```
and also looking at the test it does not assert the authors in the `Annotation` object which is why this went uncovered.

The problem is that the 'cssc' implementation of SCCS requires the `-d` and its optarg to be single argument (as opposed to the SCCS implementation on Solaris):
```
$ sccs -V
sccs from GNU CSSC 1.4.1
$Id: sccs.c,v 1.44 2007/12/19 00:21:14 jay Exp $
SccsPath = 'SCCS'
SccsDir = ''
Default prefix for SCCS subcommands is '/usr/lib/x86_64-linux-gnu/cssc/'
$ sccs prs -e -d ':I: :P:' main.c
prs: Cannot open SCCS file SCCS/s.:I: :P: for reading : No such file or directory



$ sccs prs -e -d':I: :P:' main.c
1.2 trond
1.1 trond
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
@@ -91,6 +91,14 @@ public class Annotation {
         return ret;
     }
 
+    Set<String> getAuthors() {
+        Set<String> ret = new HashSet<>();
+        for (Line ln : this.lines) {
+            ret.add(ln.author);
+        }
+        return ret;
+    }
+
     /**
      * Gets the author who last modified the specified line.
      *
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
@@ -94,11 +94,7 @@ public class Annotation {
 
     @TestOnly
     Set<String> getAuthors() {
-        Set<String> ret = new HashSet<>();
-        for (Line ln : this.lines) {
-            ret.add(ln.author);
-        }
-        return ret;
+        return lines.stream().map(ln -> ln.author).collect(Collectors.toSet());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -91,6 +92,15 @@ public String getRevision(int line) {
         return ret;
     }
 
+    @VisibleForTesting
```
ahornace, line 95: If the only purpose is to use it only in tests, then maybe `@TestOnly` is a better choice.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
```diff
@@ -91,6 +92,15 @@ public String getRevision(int line) {
...
+    Set<String> getAuthors() {
+        Set<String> ret = new HashSet<>();
```
ahornace, line 97: I personally like the concise:
```java
this.lines.stream().map(ln -> ln.author).collect(Collectors.toSet());
```

ahornace (approved):
Good catch.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Annotation.java
vladak: Yeah, I was thinking about using stream() and then forgot so thanks. It definitely looks better.

vladak:
> Good catch.

Yeah, I should have written sound test in the first place.

Thanks for the reviews !

PR #3888, opened 2022-02-02
return response from call_finished()

vladak:
Fixes a logic problem introduced in 959b34e9. This could lead to false warnings produced by the Python tools such as:
```
$ opengrok-projadm --async_api_timeout 300 -a openssh-1
Adding project openssh-1
API request still not completed after 1 seconds: <Response [202]>
could not add project 'openssh-1' in web application
```


vladak:
Thanks !

PR #3894, opened 2022-02-11
--strip-outgoing option for opengrok-mirror

vladak:
This change adds the `--strip-outgoing` CLI option and `strip_outgoing` config option (both global and per project) to `opengrok-mirror`. When set to true, the outgoing changesets (in repositories where this option is in effect and where supported) will be stripped and project data will be deleted (not the source code, just index/xrefs/cache) so that the project can be reindexed from scratch. This is handy in situations when mirroring a repository that has its history frequently rewritten and not caring about the reindex time (the project will disappear from the UI after its data is deleted. It will reappear after reindex automatically). So far only Git is supported, although this should be possible to implement also for Mercurial and other changeset based SCMs.

### DIFF
--- a/tools/src/main/python/opengrok_tools/mirror.py
+++ b/tools/src/main/python/opengrok_tools/mirror.py
@@ -108,7 +108,7 @@ def main():
                              ' repositories,'
                              ' terminate the processing'
                              ' if no change is found.')
-    parser.add_argument('--strip_outgoing', type=bool, default=False,
+    parser.add_argument('--strip_outgoing', action='store_true', default=False,
                         help='check outgoing changes for each repository of a project,'
                         'strip any such changes and remove all project data so that'
                         'it can be reindexed from scratch')
--- a/tools/src/main/python/opengrok_tools/mirror.py
+++ b/tools/src/main/python/opengrok_tools/mirror.py
@@ -108,7 +108,7 @@ def main():
                              ' repositories,'
                              ' terminate the processing'
                              ' if no change is found.')
-    parser.add_argument('--strip_outgoing', action='store_true', default=False,
+    parser.add_argument('--strip-outgoing', action='store_true', default=False,
                         help='check outgoing changes for each repository of a project,'
                         'strip any such changes and remove all project data so that'
                         'it can be reindexed from scratch')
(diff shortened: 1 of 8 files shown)


tools/src/main/python/opengrok_tools/mirror.py
```diff
@@ -107,12 +108,18 @@ def main():
                              ' repositories,'
                              ' terminate the processing'
                              ' if no change is found.')
+    parser.add_argument('--strip-outgoing', action='store_true', default=False,
```
ahornace, line 111: Maybe add a note that it's only supported by some repositories (e.g. git)? It seems to me that otherwise it's just ignored.

tools/src/main/python/opengrok_tools/utils/mirror.py
```diff
@@ -393,18 +401,43 @@ def get_mirror_retcode(ignore_errors, value):
...
+
+def mirror_project(config, project_name, check_changes, check_outgoing, uri,
+                   source_root, headers=None, timeout=None, api_timeout=None):
     """
     Mirror the repositories of single project.
     :param config global configuration dictionary
     :param project_name: name of the project
     :param check_changes: check for changes in the project or its repositories
      and terminate if no change is found
+    :param check_outgoing: check for outgoing changes in the repositories of the project,
```
ahornace, line 434: Why the change to `check_outgoing`? To me `check` has neutral meaning (it's just a check - whatever :D ) whereas `strip` is something I'd be more aware of.

ahornace (approved):

tools/src/main/python/opengrok_tools/mirror.py
vladak: Makes sense, added.

tools/src/main/python/opengrok_tools/utils/mirror.py
vladak: yep, check is not fitting; renamed.

vladak:
Thanks for the review !

vladak:
documented the feature on https://github.com/oracle/opengrok/wiki/Repository-synchronization

PR #3897, opened 2022-02-14
terminate the outgoing check for top level repository

vladak:
When introducing the changes for outgoing check+strip I forgot to add handling for top level repositories, breaking AOSP mirroring. This change fixes that.


kulikjak (approved):
LGTM

vladak:
Thanks !

PR #3898, opened 2022-02-14
outgoing check needs to take current branch into account

vladak:
When testing the outgoing changes handling with internal Linux UEK repository, a changeset that had no parent was wrongly detected as outgoing. This change fixes that.

Also, I realized that `git fetch` is needed before checking the outgoing changes.


kulikjak (approved):
LGTM

vladak:
Thanks !

PR #3920, opened 2022-03-29
print default values of options

vladak:
This change adds formatter to some of the Python tools so that default values of options are printed in help, e.g.:
```
$ opengrok-mirror -h

...

  --api_timeout API_TIMEOUT
                        Set response timeout in seconds for RESTful API calls (default: 3)
  --async_api_timeout ASYNC_API_TIMEOUT
                        Set timeout in seconds for asynchronous RESTful API calls (default: 300)
```


vladak:
Tried adding the formatter to the base parser, however that does not work.

kulikjak (approved):

vladak:
Thanks !

PR #3927, opened 2022-04-07
sort revisions for CVS repository semantically

vladak:
This change should fix sorting of CVS revision numbers.


vladak:
thanks @ahornace !

PR #3930, opened 2022-04-08
make sure Suggester#init() does not receive duplicate entries

vladak:
When writing https://github.com/oracle/opengrok/discussions/3928#discussioncomment-2530471 I realized it would not hurt to be more careful with the collection passed to `Suggester#init()`.


ahornace (approved):

vladak:
thanks @ahornace

PR #3940, opened 2022-04-19
remove /system/refresh API endpoint

vladak:
The API endpoint is redundant and causes suggester to be rebuilt twice in case of per project handling.


ahornace (approved):
Looks good!

vladak:
Thanks @ahornace !

PR #3941, opened 2022-04-22
add --check_config to opengrok-sync

vladak:
After releasing 1.7.31 with the incompatible change to the API call configuration (#3924) used by both `opengrok-mirror` and `opengrok-sync`, I realized that only `opengrok-mirror` has an option to check the configuration. This change adds `--check_config` to `opengrok-sync` so they are on par.


ahornace (approved):

kulikjak (approved):

vladak:
Thanks guys !

PR #3942, opened 2022-04-26
make sure getHistoryGet() implementations for CVSRepository are aligned

vladak:
This change transforms the implementation of `CVSRepository#getHistoryGet(String parent, String basename, String rev)` into `CVSRepository#getHistoryGet(OutputStream out, String parent, String basename, String rev)` so that getting contents of older revisions actually works.

This is done because `HistoryGuru#getRevision()` (used by `xref.jsp`) calls the 4-arg version and `CVSRepository` extends `RCSRepository` which implements it.

I did a bit of cleanup in other files as well.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/history/CVSRepositoryTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/history/CVSRepositoryTest.java
@@ -154,9 +154,9 @@ public class CVSRepositoryTest {
         try (FileChannel outChan = new FileOutputStream(mainC, true).getChannel()) {
             outChan.truncate(0);
         }
-        FileWriter fw = new FileWriter(mainC);
-        fw.write("#include <foo.h>\n");
-        fw.close();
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/test/java/org/opengrok/indexer/history/CVSRepositoryTest.java
```diff
@@ -135,24 +137,23 @@ void testGetBranchNoBranch() throws Exception {
...
 
         // Change the content and commit.
         File mainC = new File(root, "main.c");
-        FileChannel outChan = new FileOutputStream(mainC, true).getChannel();
-        outChan.truncate(0);
-        outChan.close();
+        try (FileChannel outChan = new FileOutputStream(mainC, true).getChannel()) {
+            outChan.truncate(0);
+        }
         FileWriter fw = new FileWriter(mainC);
```
ahornace, line 157: nit: maybe use try-with-resources on `fw` as well?

ahornace (approved):

vladak:
Thanks @ahornace !

PR #3983, opened 2022-07-05
make index database "optimization" explicit operation

vladak:
This change removes the automatic index "optimization" (a.k.a. reduction of segment count for given index to 1) at the end of the indexing. The outcome is reduced indexing time - indexing [linux kernel repo](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git) with history off takes 06:49 minutes, out of which 50 seconds is the "optimization" step; with the step off, the indexing time is 6:03 minutes, which means 12 % improvement. On the other hand, it might theoretically slow down the search a bit and also the index size might increase - in multi segment index, each segment typically contains deleted documents. These are expunged on segment merge which happens during indexing. Comparing the size of the index of the linux kernel repository without and with the changes, there is no significant difference - in both cases the index directory had 1.5 GB (the real difference was some 45 MB) and there were 16 segments at the end of the indexing in the changed case. The number of the segments is not the same for each reindex of the same data due to the parallel nature of the indexing.

While this might seem as a overturn to the less optimal handling, it just normalizes the situation - Lucene documents consider the reduction to single segment to be extreme and useful only for long term index archival. The other alternative would be to reduce the segments to some other number however that would be highly dependent on given index. To address the search latency, I plan to submit a PR that will create [`IndexSearcher`](https://lucene.apache.org/core/8_8_1/core/org/apache/lucene/search/IndexSearcher.html#IndexSearcher-org.apache.lucene.index.IndexReader-java.util.concurrent.Executor-) objects with a dedicated executors so that index segments can be searched in parallel.

I chose a bit of a more aggressive approach - instead of setting the default to false, this is now explicit operation invoked via the `--reduceSegmentCount` indexer option. Still softer approach than ripping this functionality out altogether. The limitation of the programming of this option's handling is that when used together with index database update, it reduces the segment count for all index databases, not just those that changed. It can be run as a standalone operation when combined with the `-n` / `--noIndex` option.

That said, I think the `reduceSegmentCount` functionality should be eventually replaced by tunables to control the Lucene segment merge policy type and its parameters.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -167,7 +167,6 @@ public final class Configuration {
     private String reviewPattern;
     private String webappLAF;
     private RemoteSCM remoteScmSupported;
-    private boolean optimizeDatabase;
     private boolean quickContextScan;
 
     private LuceneLockName luceneLocking = LuceneLockName.OFF;
@@ -557,7 +556,6 @@ public final class Configuration {
         setMessageLimit(500);
         setNavigateWindowEnabled(false);
         setNestingMaximum(1);
-        setOptimizeDatabase(true);
         setPluginDirectory(null);
         setPluginStack(new AuthorizationStack(AuthControlFlag.REQUIRED, "default stack"));
         setPrintProgress(false);
@@ -1086,14 +1084,6 @@ public final class Configuration {
 
     public void setRemoteScmSupported(RemoteSCM remoteScmSupported) {
         this.remoteScmSupported = remoteScmSupported;
-    }
-
-    public boolean isOptimizeDatabase() {
-        return optimizeDatabase;
-    }
-
-    public void setOptimizeDatabase(boolean optimizeDatabase) {
-        this.optimizeDatabase = optimizeDatabase;
     }
 
     public LuceneLockName getLuceneLocking() {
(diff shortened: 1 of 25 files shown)


vladak:
After running the tests without the segment count reduction, `ProjectsControllerTest#testListFiles()` started failing. The cause was `IndexDatabase#getFiles()` used a weird way how to interact with iterator, in this case `TermsEnum` object. Normally, one uses iterator in Java like this:
```java
Iterator<T> iterator = SomeType.iterator();
 
while (iterator.hasNext()) {
        //  Moving cursor to next element
        T data = iterator.next();
        ... // process data
}
```
or something like:
```java
Iterator<T> iterator = SomeType.iterator();
 
T data;
while ((data = iterator.next()) != null) {
        ...  // process data
}
```
however in this case the pattern was:
https://github.com/oracle/opengrok/blob/9bbba82610629a748452a72656648a9c23e33aa3/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java#L1791-L1808

For some reason, when an index has multiple segments, the first call to `iter.term()` is `null`, even though `iter.next()` returns valid object. Rewriting the `getFiles()` method to use the iterator properly fixes the test case failure. Same goes for `IndexDatabase#listTokens()`.

Now, `IndexDatabase` contains various methods used via `update()` that follow this pattern, notably  `processFile()`, `processFileIncremental` and `processTrailingTerms()`. It works there because of this code: https://github.com/oracle/opengrok/blob/9bbba82610629a748452a72656648a9c23e33aa3/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexDatabase.java#L634-L635 

After the `seekCeil()` method is called, `uidIter.term()` becomes non-null. It seems that all the above listed indexer methods rely on this fact. This is the reason the indexing still works after this change. Chaning the `uidIter` interactions blindly to call `uidIter.next()` first thing would break indexing I think, because the `seekCeil()` likely leads to calling `next()`, so the first term would be skipped.

PR #4002, opened 2022-07-20
make username/password configurable on project/repository level

vladak:
This change introduces project/repository properties for setting username/password. Currently this is used only for Subversion, where it replaces the environment variables used so far (`OPENGROK_SUBVERSION_USERNAME`, `OPENGROK_SUBVERSION_PASSWORD`). This impacts both indexer and the webapp.

This is a departure from global username/password towards per project username/password. While technically the username/password can be defined on repository level, the code is not ready for this just yet (`RepositoryInfo#fillFromProject()`) and also it is not clear whether this level of granularity is actually needed.

Example of per project [read-only configuration](https://github.com/oracle/opengrok/wiki/Read-only-configuration) with username/password:
```xml

  <void property="projects">
   <void method="put">
    <string>some-project-with-subversion-repositories</string>
     <void property="username">
      <string>foo</string>
     </void>
     <void property="password">
      <string>bar</string>
     </void>
    </object>
   </void>
  </void>

 </object>
</java>
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
@@ -92,7 +92,13 @@ public class Project implements Comparable<Project>, Nameable, Serializable {
      */
     private Boolean mergeCommitsEnabled = null;
 
+    /**
+     * Username to used for repository authentication. This is propagated to all repositories of this project.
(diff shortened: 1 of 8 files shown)


vladak:
https://github.com/oracle/opengrok/wiki/Indexer-configuration#environment-variables will need to be replaced.

PR #4004, opened 2022-07-22
docker: add openssh-client to the docker image

rgroux:
add ssh cli to be abble to fetch/pull git repository over ssh protocol.


rgroux:
Done with  Issue #4005

PR #4017, opened 2022-07-29
make Project/Group comparison case sensitive

vladak:
This change removes the upper case conversion of Project/Group names. It is not clear to me why it was there in the first place.


ahornace (approved):

vladak:
Thanks @ahornace !

PR #4018, opened 2022-07-29
use docker pushrm to update README on Docker Hub

vladak:
This change fixes Docker hub README update. Before this change it failed because the payload was not actually formed JSON even though the Content-type header claimed so. Originally https://github.com/christian-korneck/docker-pushrm based Github action was used, however this was not possible.

As mentioned on https://stackoverflow.com/a/58478262/11582827 , it would be better if the conditional in the Github workflow was based on the `release` build type however the workflow needs to work for other types as well.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,4 +1,4 @@
-name: Build docker image
+name: Build Docker image
 
 on: [push, pull_request]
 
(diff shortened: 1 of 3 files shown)


vladak:
The build failed with:
```
christian-korneck/update-container-description-action@v1 is not allowed to be used in oracle/opengrok
```
so I need to sort this out with the Oracle Github maintainers.

vladak:
Because of internal regulations, third party Github actions cannot be used. Rewriting the program to Python.

PR #4020, opened 2022-07-29
fix NPE in SuggesterProjectData

vladak:
This change makes the lookup operation for a term more resilient.

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
+++ b/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
@@ -446,8 +446,8 @@ class SuggesterProjectData implements Closeable {
         }
 
         try {
-            WFSTCompletionLookup lookup;
-            if ((lookup = lookups.get(term.field())) == null || lookup.get(term.text()) == null) {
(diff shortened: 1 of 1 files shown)


suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
```diff
@@ -446,7 +446,8 @@ boolean incrementSearchCount(final Term term, final int value, boolean waitForLo
         }
 
         try {
-            if (lookups.get(term.field()).get(term.text()) == null) {
+            WFSTCompletionLookup lookup;
```
ahornace, line 449: nit: this looks more readable to me
```java
WFSTCompletionLookup lookup = lookups.get(term.field());
if (lookup == null || lookup.get(term.text()) == null) {
```

ahornace (approved):

suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
vladak: makes sense

vladak:
Thanks @ahornace !

PR #4061, opened 2022-10-19
add opengrok-mirror tunable to ignore a project

vladak:
Sometimes it makes sense to take `opengrok-mirror` out of the loop (`opengrok-sync`) completely and just make it passthru. This change adds the `ignore` per project tunable that does that.

Sample config:
```yml
projects:
  foo:
    ignore: true
```


kulikjak (approved):
LGTM

PR #4067, opened 2022-10-22
fix event handling in Docker (redux)

vladak:
This change fixes how the Docker startup program works with the recently introduced `PeriodicTimer` class. Firstly, the instance need to be referenced with `global`. Also, it should be created before the sync threads and the Flask app are started. Also, it should be possible to trigger reindex even though the periodic timer is set to 0.

### DIFF
--- a/docker/periodic_timer.py
+++ b/docker/periodic_timer.py
@@ -8,13 +8,13 @@ import time
 
 class PeriodicTimer:
     """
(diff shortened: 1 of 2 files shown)


kulikjak (approved):
LGTM

vladak:
Thanks Jakub !

PR #4098, opened 2022-11-09
use Smile to serialize AnnotationData

vladak:
This is another tradeoff I am going to do (trading space for speed). To change the serialization scheme for annotation cache, I had multiple requirements for the implementation:
  - speedy enough not to be noticeable in the UI (tens of ms)
  - robust and well maintained library
  - easy to use from Java (i.e. without running external tools or defining serialization scheme by hand)
  - ideally extendable with Java annotations and such
  - space efficient without compression => potentially being usable in streaming mode

I did not have many requirements for the format itself, assumed it will be some binary format.

Looked at bunch of alternatives to discover Jackson (that is already used in the webapp) offers some binary serialization formats so I went with exploring these.

To measure the impact, I used the OpenSolaris ON Mercurial repository (yeah). The de-serialization latency test used `usr/src/cmd/svc/configd/rc_node.c` from the UI.

Type | indexing time | cache size | deserialization
--- | --- | --- | ---
compressed XML (current) | 30:58 | 209M | >600ms
CBOR | 24:47 | 893M | <30ms
Smile | 25:36 | 562M | <20ms

So, Smile looks like a viable format. I do realize that currently older version of Jackson is used that most probably does not have all the performance improvements for CBOR.

Another advantage is that this avoids maintaining class white/black-lists.

This will require reindexing projects that have annotation cache enabled from scratch. There is currently no way how annotation cache can be recreated from scratch. That said, one can leave xrefs and history cache and just wipe out the index and annotation cache (there are REST APIs).

Also, at least conceptually, this is paving the way for other serialization format changes, notably for history cache.

### DIFF
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -196,7 +196,7 @@ Portions Copyright (c) 2020-2020, Lubos Kosco <tarzanek@gmail.com>.
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
-            <artifactId>jackson-dataformat-cbor</artifactId>
+            <artifactId>jackson-dataformat-smile</artifactId>
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -195,6 +195,11 @@ Portions Copyright (c) 2020-2020, Lubos Kosco <tarzanek@gmail.com>.
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


vladak:
On upgrade from the previous annotation cache format, the error to get a cache entry will be ENOENT e.g.:
```
11-Nov-2022 10:36:34.836 WARNING [http-nio-8080-exec-385] org.opengrok.indexer.history.HistoryGuru.getAnnotation org.opengrok.indexer.history.CacheException:
 java.io.FileNotFoundException: /opengrok/data/annotationcache/solaris-userland/components/aalib/Makefile (No such file or directory)
```
This is because the file name of the cache file changed - it no longer has the ".gz" suffix.

PR #4109, opened 2022-11-22
add support for Tomcat 10.1

vladak:
This change addresses the lack of support for Tomcat 10.1. The biggest challenge were the Jackson-Jersey dependencies. I casually tested the UI when deployed on Tomcat 10.1.0 and also checked the search API, plus built a Docker image and let it run for a bit to see if the API calls done during resync do not blow up.


vladak:
https://github.com/oracle/opengrok/wiki/How-to-setup-OpenGrok updated.

PR #4172, opened 2023-02-08
Fix svn history for files which were in a renamed directory

### LINKED ISSUE #3443: Subversion history is incomplete when a folder has been renamed
**Describe the bug**
For a subversion repository where a folder has been renamed and then a file within that folder has subsequently been amended, the cached history created by the indexer is incomplete. This is similar to #761.
If the cached history is deleted or somehow ignored then the history gathered via SubversionHistoryParser for the single file is correct.

This is reproducible in the latest version of OpenGrok.

**To Reproduce**
This can be seen by replacing the existing svnlog.dump file in the testdata with the one attached which has 4 additional commits.
[svnlog.zip](https://github.com/oracle/opengrok/files/6068435/svnlog.zip)

Using a test over the subversion repository in FileHistoryCacheTest you can then verify the history of the file which should have 3 history entries:
```
        File testFileInRenamedFolder = new File(reposRoot.toString() +
          File.separatorChar + "renamedFolder" + File.separatorChar +  "FileInRenamedFolder.txt");
        History cachedHistory = cache.get(testFileInRenamedFolder, repo, false);
        History fullHistory = repo.getHistory(testFileInRenamedFolder);
        assertEquals(fullHistory.getHistoryEntries(), cachedHistory.getHistoryEntries());
```

**Expected behavior**
The history produced by the initial indexing of a subversion repository should contain the complete history for a file which is in a folder which has been renamed.
The correct history in this case can be seen by bypassing the cache (as shown in the test snippet), which would be achieved if the initial repository History was able to correctly identify the file as _renamed_

**Additional context**
The subversion history contains renames of folders in the following format:
```
<path
   text-mods="false"
   kind="dir"
   copyfrom-path="/repo/originalfoldername/src"
   copyfrom-rev="564779"
   action="A"
   prop-mods="false">/repo/newfoldername/src</path>
```

I don't think that there is an easy way to determine the contents of the folder at the point of discovering such a directive, and therefore it doesn't seem immediately obvious that this could be turned into a list of _renamedFiles_ by the history parser.
The approach that I have considered would be to simply extend History to track renamed directories as well. Then History.isRenamed could check both the file and the directory. 
The downside here is that this will catch too many files causing any file in a directory that was ever renamed as a renamed file, this may lead to some performance degredation of initial indexing. For extreme cases where much of the repository has been renamed it might even lead to the consumption of the complete history being largely redundant and instead an alternate approach of simply generating the history for each file independently being prefered. Possibly some overlap with comments on #3243 about chunking the history consumption.

harrisric:
Fixes #3443

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
@@ -20,7 +20,7 @@
 /*
  * Copyright (c) 2006, 2020, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
- * Portions Copyright (c) 2020, Ric Harris <harrisric@users.noreply.github.com>.
+ * Portions Copyright (c) 2020, 2023, Ric Harris <harrisric@users.noreply.github.com>.
  */
 package org.opengrok.indexer.history;
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
@@ -32,8 +32,11 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -64,6 +67,8 @@ class SubversionHistoryParser implements Executor.StreamHandler {
 
     private static class Handler extends DefaultHandler2 {
 
+        private static final String COPYFROM_PATH = "copyfrom-path";
+
         /**
          * Example of the longest date format that we should accept - SimpleDateFormat cannot cope with micro/nano seconds.
          */
@@ -74,10 +79,12 @@ class SubversionHistoryParser implements Executor.StreamHandler {
         final int length;
         final List<HistoryEntry> entries = new ArrayList<>();
         final Set<String> renamedFiles = new HashSet<>();
+        final Map<String, String> renamedToDirectoryRevisions = new HashMap<>();
         final SubversionRepository repository;
         HistoryEntry entry;
         StringBuilder sb;
-        boolean isRenamed;
+        boolean isRenamedFile;
+        boolean isRenamedDir;
 
         Handler(String home, String prefix, int length, SubversionRepository repository) {
             this.home = home;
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionRepository.java
```diff
@@ -282,6 +291,73 @@
...
+    private Executor getDirectoryListExecutor(final String dir, String atRevision, CommandTimeoutType cmdType) {
+
+        List<String> cmd = new ArrayList<>();
+        ensureCommand(CMD_PROPERTY_KEY, CMD_FALLBACK);
+        cmd.add(RepoCommand);
+        cmd.add("ls");
+        cmd.add(escapeFileName(dir));
+        cmd.add("-r" + atRevision);
+
+        return new Executor(cmd, new File(getDirectoryName()),
```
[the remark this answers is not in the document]
harrisric: Using the same pattern as L135, L257 and L509.

vladak:
Have you signed OCA before ? If not, please start the OCA process so the PR can be merged in.

harrisric:
@vladak my company [alfa  ](https://github.com/alfasoftware) has just signed the OCA.

vladak:
> @vladak my company [Alfa ](https://github.com/alfasoftware) has just signed the OCA.

Cool. One more thing is needed:
1. Contact us and we will add you to the list of contributors for the existing agreement.
Send an email to [opensource_ww_grp@oracle.com ](mailto:opensource_ww_grp@oracle.com) including your first name, last name, email, GitHub username (optional) and GitHub project/s that you would intend to contribute to.
2. After you are added to the list of members, go back the the homepage and check the list of signatories from the top menu, your name should be displayed on the list.

harrisric:
@vladak should all be done now thanks.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
```diff
@@ -87,19 +94,31 @@ private static class Handler extends DefaultHandler2 {
...
+            isRenamedFile = false;
+            isRenamedDir = false;
             if ("logentry".equals(qname)) {
                 entry = new HistoryEntry();
                 entry.setActive(true);
                 entry.setRevision(attr.getValue("revision"));
             } else if ("path".equals(qname)) {
-                isRenamed = attr.getIndex("copyfrom-path") != -1;
+                if (attr.getIndex(COPYFROM_PATH) != -1) {
+                  LOGGER.info("rename for:" + attr.getValue(COPYFROM_PATH) );
```
vladak, line 115: could you reduce the level to e.g. `FINE` ? This would be too noisy in the logs, I think.

vladak:
> @vladak should all be done now thanks.

Cool. Besides the logging nit this is good to go.

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SubversionHistoryParser.java
vladak: actually, I will fix it after that fact.

vladak:
Thanks !

Issue #3487, opened 2021-03-20
address isWorking() for other repository types


vladak:
The fix for #3307 was done for `GitRepository` only. This bug tracks the rest of the repositories.

vladak:
Explore the related issues and history. Specifically, for Git this was fixed in c47b7e716a31dc48f006bc51fc330c0b946affe3 (only to be removed by b289da3c) and for Mercurial in de458d8d. There are other types of SCMs under `opengrok-indexer/src/main/java/org/opengrok/indexer/history/` that should receive the same treatment.

gaurangkudale:
Hi @vladak,
Thank you for your response Can you please tell me more about this issue I'm trying to understand

Akash3121:
Hi @vladak,
I'm willing to work on this issue, 
https://github.com/oracle/opengrok/tree/master/opengrok-indexer/src/main/java/org/opengrok/indexer/history 
In this tab, for which file is the problem araising for  isWorking() function? can you specify the file, please.

Akash3121:
Hey @vladak @gaurangkudale 
https://github.com/oracle/opengrok/blob/master/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java
In the above file for line no 21: ( * Copyright (c) 2006, 2022, Oracle and/or its affiliates. All rights reserved.)
I think the year 2022 need's to be changed to 2023, please confirm.
Thank you!

Akash3121:
Hey @vladak,
https://github.com/oracle/opengrok/blob/03246776315da611de8587e903c16d2d6854a89c/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java#L21
In this line, I think the year needs to be updated to 2023. please confirm.
Thank you!

vladak:
`MercurialRepository` is already addressed, see [above.](https://github.com/oracle/opengrok/issues/3487#issuecomment-1011982888)

Akash3121:
In which file does this problem arises, could you please specify one?
Thank you!

vladak:
> In which file does this problem arises, could you please specify one? Thank you!

Basically in all `Repository` descendants that do not employ lazy loading of the working attribute.
 
> > Hey @vladak,
> > https://github.com/oracle/opengrok/blob/03246776315da611de8587e903c16d2d6854a89c/opengrok-indexer/src/main/java/org/opengrok/indexer/history/MercurialRepository.java#L21
> > 
[...]

PR #4199, opened 2023-03-05
[Fix] Lazily Instantiate works for SCCSRepository

Akash3121:
Lazily Instantiate works for SCCSRepository.java

helps [#3487]().
Hey @vladak! Thank you for sharing the [Mercuial Reference](https://github.com/oracle/opengrok/commit/de458d8daebcb30b8a5600f3eb6835da3ee228ae)

Understood the problem from there and fixed it here. please take a look and verify.
Thank you!

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SCCSRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/SCCSRepository.java
@@ -41,6 +41,7 @@ import org.opengrok.indexer.configuration.CommandTimeoutType;
 import org.opengrok.indexer.configuration.RuntimeEnvironment;
 import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.util.Executor;
+import org.opengrok.indexer.util.LazilyInstantiate;
 
 /**
  * This class gives access to repositories built on top of SCCS (including TeamWare).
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/SCCSRepository.java
```diff
@@ -62,6 +63,12 @@ public class SCCSRepository extends Repository {
     @VisibleForTesting
     static final String CODEMGR_WSDATA = "Codemgr_wsdata";
 
+     /**
+     * This is a static replacement for 'working' field. Effectively, check if hg is working once in a JVM
+     * instead of calling it for every SCCSRepository instance.
+     */
+    private static final Supplier<Boolean> HG_IS_WORKING = LazilyInstantiate.using(SCCSRepository::isHgWorking);
```
vladak, line 70: this variable should probably by named differently to match the repository implementation (fix copy-n-paste programming).

opengrok-indexer/src/main/java/org/opengrok/indexer/history/SCCSRepository.java
```diff
@@ -171,15 +178,22 @@ boolean isRepositoryFor(File file, CommandTimeoutType cmdType) {
         }
         return false;
     }
+    
+    private static boolean isHgWorking() {
+        String repoCommand = getCommand(SCCSRepository.class, CMD_PROPERTY_KEY, CMD_FALLBACK);
+        boolean works = checkCmd(repoCommand);
```
vladak, line 184: This does not match the original behavior of `isWorking()`.

PR #4235, opened 2023-03-22
tools: utils: do not error during directory creation if it exists

kubiko:
Do not error out when creating a directory and directory already exists.

I got an error while running the sync job. There was an error during git fetch, and while reporting error sync job failed when creating the logs directory.

```
processing of project 'xxxxx' failed
  failed commands:
    'opengrok-mirror -U http://localhost:8080/source -c /etc/mirror-config.yml xxxxx': 1
      cannot create /tmp/logs directory
```

### DIFF
--- a/tools/src/main/python/opengrok_tools/utils/utils.py
+++ b/tools/src/main/python/opengrok_tools/utils/utils.py
@@ -43,7 +43,7 @@ def check_create_dir(logger, path):
     """
     if not os.path.isdir(path):
         try:
-            os.makedirs(path)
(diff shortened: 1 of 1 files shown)


vladak:
This is such a minor change that I will let it go without the usual process scrutiny.

vladak:
Thanks !

vladak:
> thank you @vladak!
> Our companies have OCA signed both ways and I should be added to the opengrok contributor list now as well.

looking forward to more PRs then :-)

PR #4260, opened 2023-03-30
make Tomcat use temporary directory under OpenGrok data root

vladak:
As suggested in https://github.com/oracle/opengrok/discussions/4252#discussioncomment-5479257, Tomcat temporary directory should reside under OpenGrok data root in Docker container.


kulikjak (approved):
+1

Issue #872, opened 2014-12-18
SearchEngine Error - "Too many open files in system"


dev-bloke:
I'm not sure if what I'm doing with OpenGrok is supported but I thought I'd report this anyway. :-)

I wrote a quick utility class to make repeated calls to SearchEngine.search() and I discovered that after 32 calls or so, an IOException was thrown as follows (edited for brevity)
```
java.io.FileNotFoundException: /usr/local/var/opengrok/data/index/api/_1.si (Too many open files in system)
    at java.io.RandomAccessFile.open(Native Method)
        ...
    at org.opensolaris.opengrok.search.SearchEngine.searchMultiDatabase(SearchEngine.java:194)
        ...
```
Trawling through the source distribution, it looks like the IndexReaders opened at the top of this method aren't closed at the end of the search. In my own local build I added a loop at the bottom of the method to close the IndexReaders and this seemed to fix the problem.

vladak:
If I am not mistaken the readers are created in searchMultiDatabase() in this loop:

```
203        for (Project project : root) {
204            IndexReader ireader = (DirectoryReader.open(FSDirectory.open(new File(droot, project.getPath()).toPath())));
205            subreaders[ii++] = ireader;
206        }
```

and `subreaders` is local to the method so should be automatically reaped at the end of the call. Maybe GC just was not quick enough.

czguang:
I also met this problem with opengrok(docker) and aosp code.
And when increase the value RLIMIT_NOFILE from 1024 to 102400 in file docker/sync.yml, it is ok.

code diff:
`
➜  ~/sourcecode/github/opengrok (master) ✗ git diff
diff --git a/docker/sync.yml b/docker/sync.yml
index 51fb6de632c..88894f496d2 100644
--- a/docker/sync.yml
+++ b/docker/sync.yml
@@ -20,7 +20,7 @@ commands:
       --connectTimeout, '%API_TIMEOUT%',
       -r, dirbased, -G, -m, '256', --leadingWildCards, 'on',
       -c, /usr/local/bin/ctags, -U, '%URL%', -H, '%PROJECT%']
-    limits: {RLIMIT_NOFILE: 1024}
+    limits: {RLIMIT_NOFILE: 1024000}
     args_subst: {"%API_TIMEOUT%": "$API_TIMEOUT"}
 - call:
     uri: '%URL%api/v1/messages?tag=%PROJECT%'
`

PR #4356, opened 2023-05-22
chore: fix 'Too many open files in system' error(#872) when indexing AOSP

czguang:
fix 'opengrok java.io.FileNotFoundException Too many open files' when using opengrok docker with large git repo (eg. aosp)

Change-Id: I955be229827a837f6990cb29b095c46be62e8078


vladak:
Rather than bumping the resource limit, maybe it should be removed from the sync configuration altogether.

vladak:
Also, how many CPUs and memory does the container that exhibits the problem have ? Any other tuning was done ? (e.g. Java heap and such)

vladak:
superseded by #4359

PR #4358, opened 2023-05-25
order passages in search results based strictly on file offsets

vladak:
This changes the way how search results are displayed. Previously the passages were scored based on [multiple criteria](https://lucene.apache.org/core/9_4_1/highlighter/org/apache/lucene/search/uhighlight/PassageScorer.html) so that it could happen that trimmed results for given file did not show initial hits within that file which I find confusing.

While I could use `UnifiedHighlighter.Builder` with `withScorer()` in `Context#getContext2()`, I decided to override the `getScorer()` method in the `OGKUnifiedHighlighter` instead as it is more robust - no need to rely on the Builder to be used at all relevant places.

Eventually this could be made tunable per user so that it is possible to choose between the standard Lucene behavior and the new default.


vladak:
For the record, debugging the `Passage` scores is best done in `FieldHighlighter#highlightFieldForDoc()` after the call to `highlightOffsetsEnums()` which returns array of `Passage` objects.

PR #4364, opened 2023-05-31
support pattern matching ignored project names

vladak:
This change allow to match ignored project names, using [fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatchcase), e.g.:

```
opengrok-sync --ignore_project 'foo*' \
    -l debug -c sync-debug.yml -U http://localhost:8080/opengrok_web_war/
```


kulikjak (approved):
LGTM

PR #4381, opened 2023-06-27
lucene 9.7.0

tarzanek:

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/SuggesterSearcher.java
+++ b/suggester/src/main/java/org/opengrok/suggest/SuggesterSearcher.java
@@ -73,12 +73,15 @@ class SuggesterSearcher extends IndexSearcher {
 
     private final int numDocs;
 
+    private final IndexSearcher is;
+
     /**
      * @param reader reader of the index for which to provide suggestions
      * @param resultSize size of the results
      */
     SuggesterSearcher(final IndexReader reader, final int resultSize) {
         super(reader);
+        is = new IndexSearcher(reader);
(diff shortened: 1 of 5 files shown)


tarzanek:
don't merge until I fix tests

tarzanek:
@ahornace hey, do you have any clue why those tests started failing with my patch?

tarzanek:
hmm, it works when I use deprecated methods, so likely we will need to support them, so let me find a way

vladak:
> hmm, it works when I use deprecated methods, so likely we will need to support them, so let me find a way

What deprecated methods specifically ?

suggester/src/main/java/org/opengrok/suggest/SuggesterSearcher.java
```diff
@@ -105,7 +105,7 @@ public List<LookupResultItem> suggest(
 
         try {
             if (query != null) {
-                rewrittenQuery = query.rewrite(getIndexReader());
+                rewrittenQuery = query.rewrite(this);
```
tarzanek, line 108: because of this line the tests fail ...

tarzanek:
ok, ubuntu works, why isn't mac and windows working? flaky tests?

tarzanek:
fwiw no reindex is needed, this is mostly bugfix

vladak:
@tarzanek is this change ready ?

tarzanek:
yep, feel free to merge @vladak

vladak:
Thanks !

PR #4383, opened 2023-07-06
Use Ubuntu Jammy (22.04) as base for Docker image

der-eismann:
Bionic was released in 2018, Jammy is a bit more up-to-date. The later used `tomcat:10.1-jdk11` image is based on Jammy as well, so we update the Perforce repo, too, and make it use HTTPS.


vladak:
Makes sense, thanks !

PR #4386, opened 2023-07-10
Maven wrapper container

der-eismann:
This PR consists of two parts:
- Update Maven wrapper to latest release with `mvn wrapper:wrapper -Dmaven=3.9.3`. Integrity of the JAR can be verified by comparing the SHA sum with the [linked URL](https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.2.0/maven-wrapper-3.2.0.jar) in the maven-wrapper.properties.
- Use the Maven wrapper in the container build. [Ubuntu is still distributing Maven 3.6.3](https://packages.ubuntu.com/jammy/maven), which was released in November 2019. This way we can make use of all the newer features in the container build as well and ensure all environments ([including GitHub Actions](https://github.com/oracle/opengrok/blob/master/.github/workflows/release.yml#L49)) use the same version.


vladak:
Looks good, thanks !

vladak:
Using the Maven wrapper causes problems when building in IDEA, at least in my case (2023.1.4 Ultimate edition, behind a proxy). It reports `Cannot resolve plugin org.apache.maven.plugins:maven-clean-plugin:3.2.0` for all OpenGrok Maven modules:
image
To work around this, one has to use built-in Maven from IDEA:
image

vladak:
Turned out I had the proxy section commented out in `~/.m2/settings.xml`. After uncommenting it, the build in IDEA works even with the bundled wrapper.

der-eismann:
Great! I was a bit confused at first, why this would cause any problems in IDEA.

Issue #3948, opened 2022-05-12
Create opengrok docker image for arm64


santhu3064:
**Is your feature request related to a problem? Please describe.**

Currently all official docker images are of amd64. Can an official docker image for the arm64 arch be published?

**Describe the solution you'd like**

Create docker image to based on Linux  arm64 arch

**Describe alternatives you've considered**

**Additional context**

Looking to run opengrok docker container on arm64 arch machines

vladak:
The [Github Actions virtual environments](Standard_DS2_v2 virtual machines in Microsoft Azure) currently support x64 machines only: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources and the https://github.com/actions/virtual-environments/issues/2552 tracking the addition of arm64 environments remains closed due to Azure limitation.

Although self-hosted arm{,64} runners seem to be [possible](https://github.com/wbond/pi-github-runner), this is not the level of complexity I'd like to get into at this point.

santhu3064:
@vladak  Thank you for the update. Are you going to consider this in the near future?
Can this feature will be open?

vladak:
Sure, this is valid request. I'd like to have arm64 OpenGrok Docker images myself. It's just I don't want to invest the resources into the automated arm/arm64 image builds right now.

der-eismann:
It's actually pretty easy to build multiarch images on GitHub actions, it just might take a while:
https://blog.oddbit.com/post/2020-09-25-building-multi-architecture-im/

vladak:
> It's actually pretty easy to build multiarch images on GitHub actions, it just might take a while: https://blog.oddbit.com/post/2020-09-25-building-multi-architecture-im/

Thanks for the link. Building in Qemu is certainly a possibility.

vladak:
The official Github Linux ARM64 runners are currently in public preview. Will have to [figure out](https://stackoverflow.com/a/66337328) how to push x86 and arm64 Docker images separately under the same set of version tags.

der-eismann:
Ooh nice, didn't know about that. 
So from my PoV the emulation via qemu/buildx is still the easiest way, building the JAR in the native architecture and then creating arch-specific images where you just add the JAR and what else is necessary.

[...]

PR #4387, opened 2023-07-10
Replace docker.sh script with GitHub Actions

der-eismann:
This PR replaces the `docker.sh` script with pure GitHub Actions, which makes it way easier to solve #3948 since we can just add `platforms: linux/arm64` then.

The current behavior is copied without changes:
- Login and push only on pushes to `oracle/opengrok`
- Images will be tagged by branch name, PR name or tag name, where tags will also get the `latest` image tag


vladak:
This looks fine on cursory view; before merging this in I need to check whether the docker Github Actions are allowed for the OpenGrok repository.

vladak:
Confirmed that `docker/*` actions are allowed for this repository so from this standpoint this is okay.

vladak:
Merging, will see how this works when the next OpenGrok release is ready. Thanks !

der-eismann:
I'm very confused by this error. Since the login succeeded and the push failed, it would suggest that the user is actually missing permissions (like this comment said: https://github.com/docker/build-push-action/issues/853#issuecomment-1500309236). But assuming it worked before and you didn't change anything it doesn't really make sense. 

Since this is hard for me to verify and test I'd suggest you revert this PR unfortunately and maybe tackle it another day. However everything else looks fine and the token should be valid for more than 4 minutes. No idea why this is failing, sorry :confused:

vladak:
There are 2 possibilities - either the actions setup wrong username/password or the associated user has insufficient permissions. I changed the username/password recently to use a token. Looking at the token, it has all the permissions. Looking at the error message `pushing oracle/opengrok:master with docker:`, it is not clear to me whether the last part (`docker:`) is the username used.

vladak:
The `docker/build-push-action` uses  https://github.com/docker/buildx. The error message comes from https://github.com/docker/buildx/blob/86ae8ea854e639a3267399fdf3f99e9b900f6e09/build/build.go#L988 and there is no indication which username/password was used. The actual push is done via https://github.com/moby/buildkit. This seems to be using REST API reference on docker.io.

I created new API token on docker.io associated with the username used for the Github action, used a clone of the OpenGrok repository, set the `DOCKER_USERNAME` and `DOCKER_PASSWORD` in the Secrets and variables of that repository clone and let the Docker action run. It failed in the same way.

Using the same username/password with the command line `docker` tool, I was able to successfully push the image to under my user on docker.io.

I don't have the capacity to investigate the inner workings of the above tools, so reverted the change in cf02a2b2c08769fc9c38f83d8ef2953535d42238 for now. It would be nice to try it again later.

vladak:
For the record, the `docker/login-action` is merely a wrapper for the `docker login` command (https://github.com/docker/login-action/blob/553b6f090f15b58451081ce157ff1929a266131d/src/docker.ts#L28C4-L40), at least for Docker hub. So either the authentication credentials stored by this command are not getting picked up by the buildx/moby or there is some difference between how these are used compared to `docker push`.

Issue #4312, opened 2023-04-20
Deprecated warnings cause build failures with JDK 20


WorMzy:
OS:
Arch Linux

Package versions:

- jdk-openjdk: 20.0.1.u9
- maven: 3.8.7
- ctags: 6.0.0
- java-jsvc: 1.3.3
- eclipse-ecj: 4.22

Several deprecated warnings are generated when running mvn:
```
[WARNING] COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java:[385,18] java.security.AccessController in java.security has been deprecated and marked for removal
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Executor.java:[402,43] getId() in java.lang.Thread has been deprecated
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/logger/formatter/LogFormatter.java:[96,23] getThreadID() in java.util.logging.LogRecord has been deprecated
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/OpenGrokThreadFactory.java:[47,59] getId() in java.lang.Thread has been deprecated
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java:[682,19] URL(java.lang.String) in java.net.URL has been deprecated
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java:[1457,19] URL(java.lang.String) in java.net.URL has been deprecated
[WARNING] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java:[1474,19] URL(java.lang.String) in java.net.URL has been deprecated
[INFO] 7 warnings 
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /build/opengrok/src/opengrok/opengrok-indexer/src/main/java/org/opengrok/indexer/authorization/AuthorizationFramework.java: warnings found and -Werror specified
[INFO] 1 error
```

[...]

PR #4414, opened 2023-09-26
[WIP] jdk17 support

ginoaugustine:
JDk 17 support


pom.xml
```diff
@@ -61,10 +61,9 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
 
     <properties>
         <lucene.version>9.7.0</lucene.version>
-        <mavenjavadocplugin.version>3.2.0</mavenjavadocplugin.version>
-        <!-- The following changed syntax from e.g. 1.8 to 11. -->
-        <compileSource>11</compileSource>
-        <compileTarget>11</compileTarget>
+        <mavenjavadocplugin.version>3.6.0</mavenjavadocplugin.version>
+        <compileSource>17</compileSource>
+        <compileTarget>17</compileTarget>
```
vladak, line 66: For the time being please keep the target at 11. I'd like to run the releases on Java 11 even though the compilation uses JDK 17.

opengrok-indexer/src/main/java/org/opengrok/indexer/framework/PluginFramework.java
```diff
@@ -382,7 +380,7 @@ public final void reload() {
 
         // trashing out the old instance of the loaded enables us
```
vladak, line 381: This caught my eye when going through the warnings in #4312. I was wondering why the `doPrivileged()` method was actually needed. Based on the comment I'd assume that it somehow invoked GC to get rid of the old object, based on the *trashing out* sentence to perform some atomic swap, however I think this is not the case.

Rather, it seems to be related to the plugin model as sort of explained on https://stackoverflow.com/a/2243262/11582827 however it is still not clear to me why the constructor would need to be invoked like this. I can see the parent class (`java.lang.ClassLoader`) consulting the `SecurityManager`. Possibly it needs to check the `getClassLoader` permission which might require running the code with higher privileges.

My worry was that the plugin architecture might be dependent on it, to isolate the plugins from the plugin framework.

ginoaugustine: I agree with you regarding the impact on plugin framework, for consumers using java's security manager this change may cause a regression.  
We can undo this modification for the time being, but when we upgrade the minimum required Java version to 17, we must address it and appropriately communicate in release notes.

vladak: I reckon the removal of the `doPrivileged()` call here is fine. I think the PluginFramework tests have sufficient coverage to catch significant problem. It is just not clear to me why it was needed. Building with JDK 11 without the call in place does not cause any warnings as far as I can tell and the `AuthorizationFrameworkReloadTest` tests pass.

ginoaugustine: We are not seeing any warnings in JDK 11 because the doPrivileged() api is only deprecated in JDK 17 and later.

DoPrivileged will ensure that the permission check is limited to the present code and not performed on the entire stack trace. What the doPrivileged method's effects are explained in detail on page 53 of https://apps.dtic.mil/sti/pdfs/AD1146140.pdf . 
Having said that, I agree with you on removing doProvileged because java security manager is rarely used these days, and I don't see a usecase for having callers of the PluginFramework class from a different protection domain.

vladak:
Checked by building Docker image and exercising various pieces of functionality from the UI, in particular the suggester.

vladak:
Thanks @ginoaugustine !

Issue #4333, opened 2023-05-02
FileHistoryCacheTest.testFillLastHistoryEntriesAllOrNothing sometimes fails


vladak:
Github Actions on Windows report this failure:
```
Failures: 
Error:    FileHistoryCacheTest.testFillLastHistoryEntriesAllOrNothing:1079 expected: <0> but was: <3>
```
which is this check:
```java
        assertEquals(0, (int) directoryEntries.stream().filter(e -> e.getDate() != null).count());
```

vladak:
When running the test individually on a Windows, it passes, so I assume this is caused by  unfinished cleanup from some previous test. When running all the tests in `FileHistoryCacheTest` class without Mercurial/SCCS/Subversion installed, the test also passes.

vladak:
Installed Subversion (SilkSVN - had to add it to system `PATH` because the test class executes `svn` and `svnadmin` commands directly, i.e. setting `-Dorg.opengrok.indexer.history.Subversion=...` helps only to enable Subversion run for the tests) and rerun the tests in IDEA, the test still passes.

vladak:
For the record, I am testing this on Windows Server 2022.

vladak:
I saw this failing on both Linux and macOS, so this is not exclusively a Windows environment problem.

vladak:
Running this in IDEA, I accidentally reproduced it, the output was:
```
Oct 09, 2023 2:20:33 PM org.opengrok.indexer.Metrics <clinit>
INFO: configuring PrometheusRegistry
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Oct 09, 2023 2:20:34 PM org.opengrok.indexer.history.PerforceRepository isP4Working
WARNING: Command 'p4' does not work. Some operations with Perforce repositories will fail as a result.
Oct 09, 2023 2:20:34 PM org.opengrok.indexer.history.RepositoryFactory getRepository
WARNING: BazaarRepository not working (missing binaries?): /tmp/source14868807895232430816/bazaar
Oct 09, 2023 2:20:34 PM org.opengrok.indexer.history.RepositoryFactory getRepository
WARNING: Failed to get parent for /tmp/source14868807895232430816/bazaar: java.io.IOException
[...]

PR #4418, opened 2023-10-03
lucene 9.8.0

tarzanek:
no format change
no reindex needed


vladak:
> I assume
> 
> ```
> Error:    FileHistoryCacheTest.testFillLastHistoryEntriesAllOrNothing:1079 expected: <0> but was: <1>
> ```
> 
> is not related to this change at all and is some mac tests flakiness

yep, tracked by #4333 (nice number, eh ?)

vladak:
Thanks !

PR #4425, opened 2023-10-10
upgrade jstl and chronicle-map

ginoaugustine:
Upgrade jstl and chronicle-map
Added new jstl-api dependency which is needed from version 3
Tested by running the web app in tomcat 10 and indexing a repo using the same configuration


vladak:
Checked the UI, suggester. Merging. Thanks for the persistence !

PR #4427, opened 2023-10-11
Junit error in Suggester Controller Test

ginoaugustine:
Junit error was coming up when suggester results are from second Lucene segment(docbase>0).
Document ID map inside Complex data was populated with key as docbase +docID , but documentToPositionsMap inside phrase scorer is considering only doc id. 
Junit was passing mostly since the data will be in one segment and docbase will be zero for this usecase.

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
+++ b/suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
@@ -315,7 +315,7 @@ class SuggesterProjectData implements Closeable {
                 m = new ChronicleMapAdapter(field, conf.getAverageKeySize(), conf.getEntries(), f);
             } catch (IllegalArgumentException e) {
                 logger.log(Level.SEVERE, String.format("Could not create ChronicleMap for field %s in directory " +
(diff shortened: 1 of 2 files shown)


suggester/src/main/java/org/opengrok/suggest/SuggesterProjectData.java
```diff
@@ -324,8 +324,14 @@ private void initSearchCountMap() throws IOException {
                                 +  " , most popular completion disabled, if you are using "
                                 + "JDK9+ make sure to specify: "
                                 + "--add-exports java.base/jdk.internal.ref=ALL-UNNAMED "
-                                + "--add-exports java.base/jdk.internal.misc=ALL-UNNAMED "
-                                + "--add-exports java.base/sun.nio.ch=ALL-UNNAMED", field, suggesterDir), t);
+                                + "--add-exports java.base/sun.nio.ch=ALL-UNNAMED "
```
ginoaugustine, line 327: ChronicleMap JDk17 changes

vladak:
Nice !

PR #4440, opened 2023-10-17
return right away from getHistory() if repository cannot be found

vladak:
Inspired by https://github.com/oracle/opengrok/discussions/4438

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -1188,8 +1188,9 @@ public final class HistoryGuru {
     /**
      * Lookup repository for given file.
      * @param file file object source root
-     * @return repository object
+     * @return repository object or {@code null} if not found
      */
+    @Nullable
     public Repository getRepository(File file) {
         return repositoryLookup.getRepository(file.toPath(), repositoryRoots.keySet(), repositories,
                 PathUtils::getRelativeToCanonical);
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -491,6 +491,11 @@ public History getHistory(File file, boolean withFiles, boolean ui, boolean fall
         final File dir = file.isDirectory() ? file : file.getParentFile();
         final Repository repository = getRepository(dir);
 
+        if (repository == null) {
```
ginoaugustine, line 494: Using [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) class instead of passing around null value is the recommended approach post java 8. 
We can go ahead with this PR and can take up a separate clean up activity later to replace this type of usage with optional.

ginoaugustine:
LGTM

ginoaugustine (approved):

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
vladak: How would the use of `Optional` look like in this case with `getRepository()` ?

ginoaugustine: getHistory should be returning Optional<History> instead of History.
Consumers of getHistory can invoke separate logic based upon optional value.
ex : optionalHistory.map(<logic to execute is history is not null>)
                               .orElse(<logic to execute for null case>)
https://www.youtube.com/watch?v=qrECc0COhtM is a good session about java optionals

PR #4442, opened 2023-10-17
tidy IndexCheck and improve its testing

vladak:
This change primarily converts the `IndexCheck` class from static/utility class to plain class. This allowed to use Mockito for testing normally. While there, I folded the parallelized project checking done previously in `WebappListener` to the class. This resulted in reworking the way how errors are propagated. For index check errors, the result is single exception carrying the list of source root paths corresponding to the indexes that failed the test. For I/O exceptions, just one is propagated.

There are 2 main use cases for the index check:
  1. done per project in the Docker image
  1. done for all indexes (projects enabled or not) during webapp startup

The thread handling in `IndexCheck` uses common per-class-instance pool for the project parallel processing. There is another pool used when index check is done for definitions. The latter pool comes from `IndexParallelizer` so even though multiple projects are checked in parallel, the CPU utilization will be bounded by that.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexCheck.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexCheck.java
@@ -337,10 +337,10 @@ public class IndexCheck implements AutoCloseable {
 
         switch (mode) {
             case VERSION:
-                checkVersion(indexPath);
+                checkVersion(sourcePath, indexPath);
                 break;
             case DOCUMENTS:
-                checkDuplicateDocuments(indexPath);
+                checkDuplicateDocuments(sourcePath, indexPath);
                 break;
             case DEFINITIONS:
                 checkDefinitions(sourcePath, indexPath);
@@ -503,12 +503,13 @@ public class IndexCheck implements AutoCloseable {
     }
 
     /**
+     * @param sourcePath path to the source
      * @param indexPath path to the index directory
      * @throws IOException on I/O error
      * @throws IndexVersionException if the version stored in the document does not match the version
      * used by the running program
      */
-    private void checkVersion(Path indexPath) throws IOException, IndexVersionException {
+    private void checkVersion(Path sourcePath, Path indexPath) throws IOException, IndexVersionException {
         LockFactory lockFactory = NativeFSLockFactory.INSTANCE;
         int segVersion;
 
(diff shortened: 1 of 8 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexCheck.java
```diff
@@ -87,19 +92,41 @@ public enum IndexCheckMode {
         DOCUMENTS
     }
 
+    /**
+     * Common exception for all check modes.
+     */
+    public static class IndexCheckException extends Exception {
```
ginoaugustine, line 98: Moving these exception classes to its own file will help to reduce the complexity of this class

opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexCheckTest.java
```diff
@@ -101,12 +106,16 @@ private void testIndex(boolean projectsEnabled, List<String> subFiles, IndexChec
...
+        try (IndexCheck indexCheck = new IndexCheck(configuration, subFiles)) {
+            indexCheck.check(mode);
+        }
     }
 
     @Test
     void testIndexVersionNoIndex() throws Exception {
-        IndexCheck.isOkay(configuration, IndexCheck.IndexCheckMode.VERSION, new ArrayList<>());
+        try (IndexCheck indexCheck = new IndexCheck(configuration)) {
+            indexCheck.check(IndexCheck.IndexCheckMode.VERSION);
```
ginoaugustine, line 117: Introducing an assert will be better from junit readability perspective.
assertDoesNotThrow(() -> indexCheck.check());

opengrok-indexer/src/main/java/org/opengrok/indexer/index/IndexCheck.java
vladak: fixed

opengrok-indexer/src/test/java/org/opengrok/indexer/index/IndexCheckTest.java
vladak: Thanks ! At one point I had a TODO line saying "assert needed ?" and this definitely helps.

ginoaugustine (approved):

Issue #4170, opened 2023-02-07
upgrade to jflex 1.9.1 and remove the workaround for lex scanners limits


tarzanek:
jflex 1.9.0 is out and
https://github.com/jflex-de/jflex/issues/203
got fixed and jflex now introduces:
(https://github.com/jflex-de/jflex/issues/197 https://github.com/jflex-de/jflex/pull/1045 )
```
%token_size_limit (<identifier>|<number>)
```
which makes the workaround of manually patching generated lexical scanner obsolete, 
though I see @vladak removed it ?!?!?
https://github.com/oracle/opengrok/commit/d051e170ebfdcd5eaca944e0de2228fb2030f164
(I don't see 1.8 fixes it ... )
so it needs to be set where applicable so OpenGrok is in sync with lucene limits and their tokenizers

also https://github.com/jflex-de/jflex/issues/542 is fixed so https://github.com/oracle/opengrok/blob/master/pom.xml#L348
can be cleaned up

sreenath-tm:
Hi I would like to work on this issue can you assign this issue to me

vladak:
> Hi I would like to work on this issue can you assign this issue to me
> 

just submit a PR

PR #4443, opened 2023-10-18
upgrade jflex version and remove workaround

ginoaugustine:
upgrade jflex version and remove workaround added for https://github.com/jflex-de/jflex/issues/542
Removed test api-docs target since we are not pushing api docs for test classes 

Pending : Setting token_size_limit and skipping the entire line if line length is more than the limit

#4170

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -343,12 +343,6 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
                             <value>bar</value>
                         </property>
                     </javaApiLinks>
-                    <sourceFileExcludes>
-                        <!-- workarounds for https://github.com/jflex-de/jflex/issues/542 -->
-                        <exclude>**/HistoryLineTokenizer.java</exclude>
-                        <exclude>**/PlainLineTokenizer.java</exclude>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -329,7 +329,6 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
                   <execution>
                     <goals>
                       <goal>javadoc</goal>
-                      <goal>test-javadoc</goal>
```
vladak, line 332: Does this removal mean that Javadoc within test sources will not be checked ?

ginoaugustine: Yes, Removed this since we were not publishing javadoc for test sources to github pages.
From junit 5 onwards, it is recommended to have default scope for all test classes and methods(instead of public scope), even for this use, test-javadoc will fail since it wont be able to findout any public java classes under test package.
Have we added test-javadoc goal considering any other usecase

vladak: That's a good question about the use case. I am definitely not using the generated javadocs for tests so it's probably okay to say good bye to them. When changing/adding new test cases, the IDE will probably provide some sufficient warnings.

vladak:
Thanks !

PR #4446, opened 2023-10-19
always refresh the list of repositories for projects

vladak:
The Docker startup script should allow refresh of repositories for projects. This is more in line with the appliance like characteristics of the Docker image. This will make the resync slower, depending on the repositories.

While there, I did some small changes to the test.

Also, when log level is set to debug, the list of repositories for given project will be logged.


ginoaugustine (approved):
LGTM

PR #4451, opened 2023-10-23
check commands structure with None values

vladak:
Even though it seems the following is a valid YAML:
```yaml
commands:
- command:
  call:
    uri: 'http://localhost:8080/api/v1/messages'
    method: POST
    data:
      messageLevel: warning
      duration: PT1H
      tags: ['%PROJECT%']
      text: reindex in progress
```
is parsed into invalid `opengrok-sync` configuration, with `command` having `None` value. This change catches this.

Inspired by https://github.com/oracle/opengrok/discussions/4449


kulikjak (approved):
LGTM

PR #4456, opened 2023-10-25
Handle history cache failures

vladak:
This change addresses long standing issue where the 2nd phase of indexing goes through even though history cache creation/update for one of the repositories failed. If projects are enabled and one of the repositories for given project failed history cache creation/update, that project will not be indexed; the projects with all their repositories being fine w.r.t. history cache are allowed to progress to 2nd phase of indexing. If projects are disabled, history cache creation failure of a repository results in the 2nd phase of indexing being skipped altogether.

Obviously, this relies on the individual repository implementations to properly report the history cache operation failure via `HistoryException`.

To suppress the new behavior, one can use the `--ignoreHistoryCacheFailures` indexer option.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
@@ -35,9 +35,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -944,7 +946,7 @@ public final class HistoryGuru {
         }
     }
 
-    private void createHistoryCache(Repository repository, String sinceRevision) {
+    private void createHistoryCache(Repository repository, String sinceRevision) throws CacheException, HistoryException {
         if (!repository.isHistoryEnabled()) {
             LOGGER.log(Level.INFO,
                     "Skipping history cache creation for {0} and its subdirectories", repository);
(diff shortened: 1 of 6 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -1044,12 +1046,13 @@ private void createHistoryCacheReal(Collection<Repository> repositories) {
      * internal map, e.g. via {@code setRepositories()} or {@code addRepositories()}.
      *
      * @param repositories list of repository paths
+     * @return map of repository to optional exception
      */
-    public void createHistoryCache(Collection<String> repositories) {
+    public Map<Repository, Optional<Exception>> createHistoryCache(Collection<String> repositories) {
         if (!useHistoryCache()) {
-            return;
+            return null;
```
ginoaugustine, line 1053: Returning an empty collection is better than null

opengrok-indexer/src/main/java/org/opengrok/indexer/history/HistoryGuru.java
```diff
@@ -955,26 +957,19 @@ private void createHistoryCache(Repository repository, String sinceRevision) {
...
+            repository.createCache(historyCache, sinceRevision);
             elapsed.report(LOGGER, String.format("Done history cache for %s", repository));
         } else {
             LOGGER.log(Level.WARNING,
                     "Skipping creation of history cache for {0}: Missing SCM dependencies?", repository);
         }
     }
 
-    private void createHistoryCacheReal(Collection<Repository> repositories) {
+    private Map<Repository, Optional<Exception>> createHistoryCacheReal(Collection<Repository> repositories) {
```
ginoaugustine, line 968: Can we add repositories with exception only to the map, instead of keeping the value as optional

vladak: This way I like it better because it makes it more apparent what repositories were processed, also good for testing.

ginoaugustine (approved):
LGTM

PR #4457, opened 2023-10-25
opengrok web project code smell fixes

ginoaugustine:


vladak (changes requested):
For the JS changes, it would be good to rename+bump version and do the appropriate change in https://github.com/oracle/opengrok/blob/78ad87a5a965eb7196d253a3b399c6163b6b8656/opengrok-web/src/main/java/org/opengrok/web/Scripts.java#L112-L125

ginoaugustine:
> For the JS changes, it would be good to rename+bump version and do the appropriate change in
> 
> https://github.com/oracle/opengrok/blob/78ad87a5a965eb7196d253a3b399c6163b6b8656/opengrok-web/src/main/java/org/opengrok/web/Scripts.java#L112-L125

Done

vladak (approved):

vladak:
Can you do some casual testing of the UI ? Specifically for the JS libraries that changed.

ginoaugustine:
> Can you do some casual testing of the UI ? Specifically for the JS libraries that changed.

Tested loading of projects in home page and disable of any messages based upon the tag.
Tested table sorting based upon date.
File Type search

PR #4460, opened 2023-10-27
Fix Sonar codesmell issues

ginoaugustine:
Fixing Sonar codesmell issues


vladak (approved):
The property renames in `Configuration` would normally impact serialization, however they do not seem to be documented in the wiki so it seems fine.

vladak:
It would be desirable to test project selection and suggester in the UI by hand, I don't think these are sufficiently covered by tests.

ginoaugustine:
> It would be desirable to test project selection and suggester in the UI by hand, I don't think these are sufficiently covered by tests.

Sanity tested project selection checkboxes and suggestor

PR #4461, opened 2023-10-31
skip inactive history entries when generating history context

vladak:
This change fixes diff link generation for histories that contain inactive history entries.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/search/context/HistoryContextTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/search/context/HistoryContextTest.java
@@ -210,8 +210,8 @@ class HistoryContextTest {
                         "Totoro",
                         "Uaaah\n", true, filePaths),
                 new HistoryEntry("1.2", "1.2", new Date(1485438706000L),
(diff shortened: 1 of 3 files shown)


ginoaugustine (approved):
LGTM

vladak:
> LGTM

Actually, I found more elegant solution - filter the inactive history entries up front.

opengrok-indexer/src/test/java/org/opengrok/indexer/search/context/HistoryContextTest.java
```diff
@@ -190,4 +198,37 @@ void testWriteMatch() throws IOException {
...
                 sb.toString());
     }
+
+    /**
+     * Test that inactive history entries are skipped when generating history context.
+     */
+    @Test
+    void testGetHistoryContextVsInactiveHistoryEntries() {
+        Set<String> filePaths = Set.of(File.separator + Paths.get("teamware", "foo.c"));
+        History history = new History(List.of(
```
vladak, line 208: It would be nice to test some corner cases, like all history entries are inactive or if the next entry is not present etc.,  however not this PR.

PR #4466, opened 2023-11-02
get rid of setup.py

vladak:
This change adapts the Python-based tooling to modern(ish?) times by getting rid of `setup.py` altogether. Nothing complicated: `setuptools` is still used, the tests are run via `tox`.


kulikjak (approved):
+1, Thanks!

PR #4469, opened 2023-11-03
Fix : NPE when displaying a diff

### LINKED ISSUE #4468: NPE when displaying a diff
When displaying a diff between two revisions in the latest master I am getting a NPE:
```
java.lang.NullPointerException
	at org.apache.jsp.diff_jsp._jspService(diff_jsp.java:722)
	at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:71)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:770)
	at org.apache.jasper.servlet.JspServletWrapper.service(JspServletWrapper.java:467)
	at org.apache.jasper.servlet.JspServlet.serviceJspFile(JspServlet.java:379)
...
```
Looking at the `diff_jsp.java` file in the Tomcat's `work` directory I see it happens here:
```java
} else if (data.getRevision().size() == 0) { 
```

ginoaugustine:
Populated the revision data if there is no error message.
Added additional assertions to confirm revision data is populated while created diff for valid versions
closes #4468

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/PageConfig.java
@@ -349,12 +349,11 @@ public final class PageConfig {
                     .forEach(IOUtils::close);
         }
         if (Objects.isNull(data.errorMsg)) {
-            return;
+            populateRevisionData(data);
+            populateRevisionURLDetails(data, filepath);
+            data.full = fullDiff();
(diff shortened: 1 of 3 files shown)


vladak (approved):

vladak:
Thanks for the test improvements. The UI area is generally under-tested so this definitely helps. The fact that the issue was introduced is actually good in the end.

PR #4472, opened 2023-11-05
Change commons-io scope from compile to test.

### LINKED ISSUE #4471: opengrok-indexer seems to set commons-io:commons-io in the wrong scope.
**Describe the bug**
Dependency commons-io:commons-io in opengrok-indexer is set to the default dependency scope "compile". However, it seems that it's only used for tests with class org.apache.commons.io.FileUtils.

**To Reproduce**
It's easy to confirm in latest opengrok-indexer/pom.xml.

**Expected behavior**
The dependency scope of commons-io should be set to test according to maven's doc on dependency scope.

**Additional Context** 
I change the dependency scope to test and recompile the module. The module is successfully compiled, meaning the dependency is indeed not a compile scope dependency. Considering it's used in tests only, test scope would be the appropriate scope.

HappyHacker123:
Fix #4471 .

### DIFF
--- a/opengrok-indexer/pom.xml
+++ b/opengrok-indexer/pom.xml
@@ -61,6 +61,7 @@ Portions Copyright (c) 2020-2020, Lubos Kosco <tarzanek@gmail.com>.
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
             <version>${commons-io.version}</version>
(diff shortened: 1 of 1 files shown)


HappyHacker123:
> PR author: @HappyHacker123

OCA Signed.

vladak:
Thanks !

PR #4475, opened 2023-11-08
use upper limit for output data in Command

vladak:
To avoid exhausting memory when commands produce inordinate amounts of output (think reindex from scratch in verbose mode), this change introduces upper limits for the output lines stored by the `Command` class. The default is 10k lines and 250 characters for each line. The truncation is visible in the output for both cases via `...`.

As a side effect, the newline separators are not longer stored in the list of output lines. They are added only when requested using the `getoutputstr()` or `geterroutputstr()`.

### DIFF
--- a/tools/src/main/python/opengrok_tools/utils/command.py
+++ b/tools/src/main/python/opengrok_tools/utils/command.py
@@ -199,6 +199,7 @@ class Command:
                     # Shorten the list to be one less than the maximum because a line is going to be added.
                     if len(self.out) >= self.max_lines:
                         self.out = self.out[-self.max_lines + 1:]
+                        self.out[0] = "... <truncated>"
 
                     self.out.append(line)
 
--- a/tools/src/main/python/opengrok_tools/utils/command.py
+++ b/tools/src/main/python/opengrok_tools/utils/command.py
@@ -428,7 +428,7 @@ class Command:
     def getoutputstr(self):
         if self.state == Command.FINISHED:
             s = os.linesep.join(self.out)
-            if len(self.out) > 0:
+            if self.out:
                 s += os.linesep
             return s
         else:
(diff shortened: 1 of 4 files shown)


tools/src/test/python/test_command.py
```diff
@@ -199,13 +202,22 @@ def test_long_output():
...
         cmd.execute()
 
         assert cmd.getstate() == Command.FINISHED
         assert cmd.getretcode() == 0
         assert cmd.geterroutput() is None
-        assert len("".join(cmd.getoutput())) == num_bytes
+        assert len(cmd.getoutput()) == num_lines
+        if shorten:
+            # Add 3 for the '...' suffix.
+            assert len([len(x) for x in cmd.getoutput() if len(x) > line_length + 3]) == 0
```
kulikjak, line 218: Isn't the inner `len` unnecessary here?

Something like this might be slightly more descriptive:
```python
assert all(len(x) <= line_length + 3 for x in cmd.getoutput())
```

tools/src/main/python/opengrok_tools/utils/command.py
```diff
@@ -403,11 +427,18 @@ def getretcode(self):
 
     def getoutputstr(self):
         if self.state == Command.FINISHED:
-            return "".join(self.out).strip()
+            s = os.linesep.join(self.out)
+            if len(self.out) > 0:
```
kulikjak, line 431: Just `if self.out:` should work here as well, but it's not a big deal.

vladak: I see, thanks !

tools/src/test/python/test_command.py
vladak: much nicer, thanks !

PR #4476, opened 2023-11-08
avoid annotations for binary files

vladak:
This change prevents getting annotations for data/image files or genre-less files.


ginoaugustine (approved):
LGTM

PR #4485, opened 2023-11-16
Refactoring to reduce sonar code smell fixes

ginoaugustine:
Refactoring to reduce sonar code smell fixes

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -1020,6 +1020,7 @@ public class AnalyzerGuru {
                         throw new WrapperIOException(e);
                     }
                 })
+                .filter(Objects::nonNull)
                 .findFirst();
 
         optionalFactory.ifPresent(factory -> {
@@ -1119,7 +1120,7 @@ public class AnalyzerGuru {
         in.mark(MARK_READ_LIMIT);
 
         String encoding = IOUtils.findBOMEncoding(sig);
-        if (encoding == null) {
+        if (Objects.isNull(encoding)) {
             // SRCROOT is read with UTF-8 as a default.
             encoding = StandardCharsets.UTF_8.name();
         } else {
@@ -1132,6 +1133,7 @@ public class AnalyzerGuru {
 
         int nRead = 0;
         boolean ignoreWhiteSpace = true;
+        boolean breakOnNewLine = false;
         int r;
 
         StringBuilder opening = new StringBuilder();
@@ -1139,7 +1141,7 @@ public class AnalyzerGuru {
 
         while ((r = readr.read()) != -1) {
             char c = (char) r;
-            if (++nRead > OPENING_MAX_CHARS || c == '\n') {
+            if (isNewLineOrMaxReadLimit(c, ++nRead, breakOnNewLine)) {
                 break;
             }
             boolean isWhitespace = Character.isWhitespace(c);
@@ -1153,6 +1155,7 @@ public class AnalyzerGuru {
                 // If the opening starts with "#!", then track so that any
                 // trailing whitespace after the hashbang is ignored.
                 ignoreWhiteSpace = opening.length() == 2 && opening.charAt(0) == '#' && opening.charAt(1) == '!';
+                breakOnNewLine = true;
 
             }
 
@@ -1160,6 +1163,10 @@ public class AnalyzerGuru {
 
         in.reset();
         return opening.toString();
+    }
+
+    private static boolean isNewLineOrMaxReadLimit(char c, int readcount, boolean breakOnNewLine) {
+        return readcount > OPENING_MAX_CHARS || (breakOnNewLine && c == '\n');
     }
 
     private static void addCustomizationKey(String k) {
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -553,7 +553,7 @@ public class AnalyzerGuru {
      */
     public static AbstractAnalyzer getAnalyzer(InputStream in, String file) throws IOException {
         AnalyzerFactory factory = find(in, file);
-        if (factory == null) {
+        if (Objects.isNull(factory)) {
             AbstractAnalyzer defaultAnalyzer = getAnalyzer();
             if (LOGGER.isLoggable(Level.FINEST)) {
                 LOGGER.log(Level.FINEST, "''{0}'': fallback {1}",
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -123,6 +123,7 @@ import org.opengrok.indexer.logger.LoggerFactory;
 import org.opengrok.indexer.search.QueryBuilder;
 import org.opengrok.indexer.util.IOUtils;
 import org.opengrok.indexer.util.Statistics;
+import org.opengrok.indexer.util.WrapperIOException;
 import org.opengrok.indexer.web.Util;
 
 /**
@@ -261,7 +262,7 @@ public class AnalyzerGuru {
                 new IgnorantAnalyzerFactory(),
                 new BZip2AnalyzerFactory(),
                 new XMLAnalyzerFactory(),
-                YamlAnalyzerFactory.DEFAULT_INSTANCE,
+                new YamlAnalyzerFactory(),
                 MandocAnalyzerFactory.DEFAULT_INSTANCE,
                 TroffAnalyzerFactory.DEFAULT_INSTANCE,
                 new ELFAnalyzerFactory(),
@@ -616,7 +617,7 @@ public class AnalyzerGuru {
          */
         File fpath = new File(path);
         String fileParent = fpath.getParent();
-        if (fileParent != null && fileParent.length() > 0) {
+        if (fileParent != null && !fileParent.isEmpty()) {
             String normalizedPath = QueryBuilder.normalizeDirPath(fileParent);
             StringField npstring = new StringField(QueryBuilder.DIRPATH, normalizedPath, Store.NO);
             doc.add(npstring);
@@ -704,7 +705,7 @@ public class AnalyzerGuru {
      * @param file file object, used for logging only
      * @throws java.io.IOException if an error occurs while creating the output
      */
-    @SuppressWarnings("java:S5443")
+    @SuppressWarnings({"java:S5443", "java:S107"})
     public static void writeDumpedXref(String contextPath,
             AnalyzerFactory factory, Reader in, Writer out,
             @Nullable Definitions defs, Annotation annotation, Project project, File file) throws IOException {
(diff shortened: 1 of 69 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -261,7 +262,7 @@ public class AnalyzerGuru {
                 new IgnorantAnalyzerFactory(),
                 new BZip2AnalyzerFactory(),
                 new XMLAnalyzerFactory(),
-                YamlAnalyzerFactory.DEFAULT_INSTANCE,
+                new YamlAnalyzerFactory(),
```
vladak, line 265: why is this needed ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/BazaarHistoryParser.java
```diff
@@ -110,79 +111,81 @@ public void processStream(InputStream input) throws IOException {
...
-                        try {
-                            Date date = repository.parse(s.substring("timestamp:".length()).trim());
-                            entry.setDate(date);
-                        } catch (ParseException e) {
-                            //
-                            // Overriding processStream() thus need to comply with the
-                            // set of exceptions it can throw.
-                            //
-                            throw new IOException("Failed to parse history timestamp:" + s, e);
+            if (Objects.nonNull(entry)) {
```
vladak, line 114: what is the advantage of using `Objects.nonNull` compared to simple `!= null` check ? Also, this is adding extra level of indentation for a giant switch/case block. Can this be avoided by refactoring the block into a method or by `continue` ?

opengrok-indexer/src/main/java/org/opengrok/indexer/history/BazaarHistoryParser.java
```diff
@@ -110,79 +111,81 @@ public void processStream(InputStream input) throws IOException {
...
+                        }
+                        break;
+                    default:
+                        LOGGER.log(Level.WARNING, "Unknown parser state: {0}", state);
+                        break;
                 }
+            }
         }
 
         if (entry != null && state > 2) {
```
vladak, line 191: style-wise, I'd expect `Objects.nonNull()` to be also used here, esp. given this is the same variable.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/ConfigurationHelp.java
```diff
@@ -195,13 +197,13 @@ private static Object getSampleValue(Method setter, Object defaultValue) {
     }
 
     private static Object getSampleListValue(Type genType) {
```
vladak, line 199: add `Nullable` annotation for the parameter ?

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -982,89 +997,84 @@ private static AnalyzerFactory findForStream(InputStream in,
...
-                fac = matcher.isMagic(content, in);
-                if (fac != null) {
-                    if (LOGGER.isLoggable(Level.FINEST)) {
-                        LOGGER.log(Level.FINEST,
-                            "''{0}'': chosen by imprecise magic: {1}",
-                            new Object[]{file, fac.getClass().getSimpleName()});
+    private static Optional<AnalyzerFactory> findFactoryUsingMatcher(byte[] content, InputStream in,
+                                                                 String file,
+                                                              boolean isPrecise) {
+        var optionalFactory = matchers.stream()
```
vladak, line 1014: I wonder if this could benefit from `paralleStream()`

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1110,7 +1120,7 @@ private static String readOpening(InputStream in, byte[] sig)
         in.mark(MARK_READ_LIMIT);
 
         String encoding = IOUtils.findBOMEncoding(sig);
-        if (encoding == null) {
+        if (Objects.isNull(encoding)) {
             // SRCROOT is read with UTF-8 as a default.
             encoding = StandardCharsets.UTF_8.name();
         } else {
```
vladak, line 1126: This branch for skipping the BOM can be probably avoided by using `IOUtils.createBOMStrippedReader()` below. Just an idea (needs some thought w.r.t. buffer limit).

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1122,53 +1132,43 @@ private static String readOpening(InputStream in, byte[] sig)
...
-        boolean postHashbang = false;
+        boolean ignoreWhiteSpace = true;
+        boolean breakOnNewLine = false;
         int r;
 
         StringBuilder opening = new StringBuilder();
         BufferedReader readr = new BufferedReader(new InputStreamReader(in, encoding), OPENING_MAX_CHARS);
+
         while ((r = readr.read()) != -1) {
-            if (++nRead > OPENING_MAX_CHARS) {
```
vladak, line 1133: I'd swear that somewhere I saw a `Reader` that is limited by the amount of characters it can read. It would work well in this case, I think.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1122,53 +1132,43 @@ private static String readOpening(InputStream in, byte[] sig)
...
             char c = (char) r;
-            boolean isWhitespace = Character.isWhitespace(c);
-            if (!sawNonWhitespace) {
-                if (isWhitespace) {
-                    continue;
-                }
-                sawNonWhitespace = true;
-            }
-            if (c == '\n') {
+            if (isNewLineOrMaxReadLimit(c, ++nRead, breakOnNewLine)) {
```
vladak, line 1144: it would be good to have a unit test for this method, esp, if its signature can be simplified.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
```diff
@@ -183,17 +188,11 @@ public List<Tag> getTags() {
      * @return list of tags
      */
     public List<Tag> getTags(int line) {
```
vladak, line 190: add `@Nullable` + javadoc

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: Removed this singleton instances are harder to test, so it is recommended to avoid singleton design pattern. 
https://rules.sonarsource.com/java/RSPEC-6548/

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: Parallel Stream have additional overhead of creating the records, splitting collection etc, so it is recommend to  user sequential streams for most of the usecases.
Parallel Stream have better performance if the collection size is huge or if we are doing many computation for each element.
Considering these factors I think sequential stream is a better fit here.
https://www.baeldung.com/java-when-to-use-parallel-stream
https://stackoverflow.com/questions/23170832/java-8s-streams-why-parallel-stream-is-slower

opengrok-indexer/src/main/java/org/opengrok/indexer/history/BazaarHistoryParser.java
ginoaugustine: Objects.nonNull is same as !=null, AAdvantage is while using in streams.
ex: testlList.stream().filter(Objects::nonNull) vs testlList.stream().filter(elem -> elem!=null)

ginoaugustine: Skip this swtch case to multiple methods

opengrok-indexer/src/main/java/org/opengrok/indexer/history/BazaarHistoryParser.java
ginoaugustine: Found a way to avoid this null check

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: createBOMStrippedReader api is not skipping BOM bytes, here we have the logic to skip the BOM bytes from the stream

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: Removed this method in the latest push

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: I was not able to find a reader with max char limit, changed the loop to for with exit condition based upon characters read.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -1109,66 +1116,68 @@ private static String readOpening(InputStream in, byte[] sig)
...
+     * @return optional of encoding
+     * @throws IOException in case of any read error
+     */
+    private static Optional<String> findAndSkipBomEncoding(InputStream in, byte[] sig) throws IOException {
+        var encoding = IOUtils.findBOMEncoding(sig);
+        if (Objects.isNull(encoding)) {
+            // SRCROOT is read with UTF-8 as a default.
+            encoding = StandardCharsets.UTF_8.name();
+        }
+        var skipForBOM = IOUtils.skipForBOM(sig);
```
vladak, line 1174: `skipForBOM` will call `findBOMEncoding()` once again for the same input. Perhaps it should have a variant where `encoding` can be passed in. Not this PR, though.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
```diff
@@ -180,20 +186,15 @@ public List<Tag> getTags() {
...
-            result = new ArrayList<>();
-            for (Set<Tag> ltags : lineMap.sym_tags.values()) {
-                result.addAll(ltags);
-            }
-        }
-
-        return result;
+    public @Nullable List<Tag> getTags(int line) {
+        return Optional.ofNullable(lineMaps.get(line))
+                .map( lineMap -> lineMap.symTags.values().stream()
```
vladak, line 193:
```suggestion
                .map(lineMap -> lineMap.symTags.values().stream()
```

opengrok-indexer/src/main/java/org/opengrok/indexer/util/PathUtils.java
```diff
@@ -135,14 +136,12 @@ public static String getRelativeToCanonical(Path path, Path canonical,
...
-                    String format = String.format("%1$s is prohibited symlink", iterPath);
-                    LOGGER.finest(format);
-                    throw new ForbiddenSymlinkException(format);
-                }
+            if (Objects.nonNull(allowedSymlinks) && Files.isSymbolicLink(iterPath) &&
+                !isWhitelisted(iterCanon.toString(), canonicalRoots) &&
+                !isAllowedSymlink(iterCanon, allowedSymlinks)) {
+                String format = String.format("%1$s is prohibited symlink", iterPath);
+                LOGGER.finest(format);
+                throw new ForbiddenSymlinkException(format);
```
vladak, line 144:
```suggestion
                    String format = String.format("%1$s is prohibited symlink", iterPath);
                    LOGGER.finest(format);
                    throw new ForbiddenSymlinkException(format);
```

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
```diff
@@ -86,7 +86,7 @@ public abstract class Repository extends RepositoryInfo {
      * List of &lt;revision, tags&gt; pairs for repositories which display tags
      * only for files changed by the tagged commit.
      */
-    protected TreeSet<TagEntry> tagList = null;
+    protected NavigableSet<TagEntry> tagList = null;
```
vladak, line 89: what is the navigable set used for ?

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Laundromat.java
```diff
@@ -78,25 +79,24 @@ public static String launderLog(String value) {
...
+
+        Optional.ofNullable(map)
+                .stream()
+                .map(Map::entrySet)
+                .flatMap(Collection::stream)
+                .sorted(Map.Entry.comparingByKey())
+                .forEach(entry ->
+                        safes.compute(
+                                launderLog(entry.getKey()),
+                                (k, v) -> mergeLogArrays(entry.getValue(), v)
```
vladak, line 92: Could `v` be `null` ? Would it be better to use `safes.merge()` ?

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SettingsHelper.java
```diff
@@ -71,10 +72,9 @@ public Map<String, IndexedSymlink> getSymlinks(String projectName) throws IOExce
         getSettings(projectName);
         String projectKey = projectName != null ? projectName : "";
         Map<String, IndexedSymlink> indexSymlinks = mappedIndexedSymlinks.get(projectKey);
-        if (indexSymlinks != null) {
-            return Collections.unmodifiableMap(indexSymlinks);
-        }
-        return null;
+        return Optional.ofNullable(indexSymlinks)
+                .map(Collections::unmodifiableMap)
+                .orElseGet(Collections::emptyMap );
```
vladak, line 77:
```suggestion
                .orElseGet(Collections::emptyMap);
```

vladak (changes requested):
I went through all the changes, only have a bunch of comments.

opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
ginoaugustine: Will take this up as a separate PR

opengrok-indexer/src/main/java/org/opengrok/indexer/history/Repository.java
ginoaugustine: NavigableSet is the interface implemented by TreeSet. It is recommended to use interfaces instead of implementation.

vladak: I see.

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Laundromat.java
ginoaugustine: Updated to use merge approach using Collectors.toMap

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Laundromat.java
```diff
@@ -78,47 +81,36 @@ public static String launderLog(String value) {
...
-                result[i] = launderLog(values[i]);
-            }
-        }
-        if (safeValues != null) {
-            System.arraycopy(safeValues, 0, result, i, safeValues.length);
-        }
-        return result;
     }
 
     private static String replaceAll(String value, String regex, String replacement) {
```
vladak, line 110: add `@Nullable` ?

ginoaugustine: Done

PR #4490, opened 2023-11-29
Plugins cleanup

vladak:
This change improves coverage of authorization plugins and related classes, either by extending tests or removing dead code.


plugins/src/main/java/opengrok/auth/plugin/decoders/FakeOSSOHeaderDecoder.java
```diff
@@ -1,44 +0,0 @@
...
- * This class should therefore only be used for debugging.
- *
- * @author Krystof Tulinger
- */
-public class FakeOSSOHeaderDecoder extends OSSOHeaderDecoder {
-
-    private static final String PREFIX = "fake-";
-
-    public FakeOSSOHeaderDecoder() {
-        OSSO_COOKIE_TIMESTAMP_HEADER = PREFIX + OSSOHeaderDecoder.OSSO_COOKIE_TIMESTAMP_HEADER;
```
ginoaugustine, line 37: Can we also change these request header names as private final variables in OSSOHeaderDecoder class.

vladak: final yes, private is blocked due to static imports in `OSSODecoderTest.java`

ginoaugustine (approved):
Looks good to me

PR #4500, opened 2023-12-04
update setuptools and pip before installing opengrok-tools

vladak:
This fixes the recently introduced problem in Docker image. Besides upgrading the Python tooling, the change performs some cleanup of the Dockerfile.


kulikjak (approved):
+1

PR #4505, opened 2023-12-05
lucene 9.9.0

tarzanek:

### DIFF
--- a/suggester/src/main/java/org/opengrok/suggest/query/customized/CustomPhraseQuery.java
+++ b/suggester/src/main/java/org/opengrok/suggest/query/customized/CustomPhraseQuery.java
@@ -16,7 +16,6 @@
  */
 package org.opengrok.suggest.query.customized;
 
-import org.apache.lucene.index.IndexReaderContext;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.PostingsEnum;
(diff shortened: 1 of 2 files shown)


tarzanek:
likely reindex needed due to:
```
Change Postings back to using FOR in Lucene99PostingsFormat. Freqs, positions and offset keep using PFOR.
```
https://lucene.apache.org/core/9_9_0/changes/Changes.html#v9.9.0.api_changes
( see https://github.com/apache/lucene/issues/12696 )

suggester/src/main/java/org/opengrok/suggest/query/customized/CustomPhraseQuery.java
```diff
@@ -251,12 +250,10 @@ private static class CustomPhraseWeight extends Weight {
             super(query);
             this.query = query;
 
-            IndexReaderContext context = searcher.getTopReaderContext();
-
             this.states = new TermStates[query.terms.length];
             for(int i = 0; i < query.terms.length; ++i) {
                 Term term = query.terms[i];
-                this.states[i] = TermStates.build(context, term, false);
+                this.states[i] = TermStates.build(searcher, term, false);
```
tarzanek, line 256: @ahornace will above cut it?
or I broke the logic?

tarzanek: hmm, ignore, I checked the implementation and it does exactly above :-)

vladak:
any notable features/improvements in the new Lucene ?

tarzanek:
bugfixes and stability fixes mostly

vladak:
Thanks !

For the record, the change in the index format stems from https://github.com/apache/lucene/pull/12741

PR #4509, opened 2023-12-06
avoid truncation when merging config files

vladak:
This change introduces output file mandatory parameter to the `ConfigMerge` utility. As a result the tools will no longer consume its output in order to produce new configuration file, avoiding truncation.


kulikjak (approved):
+1, (now, I am also a Java reviewer... :))

PR #4511, opened 2023-12-08
add API endpoint for getting file definitions

vladak:
Preliminary PR for getting file definitions via API.


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/FileController.java
```diff
@@ -141,4 +150,21 @@ public String getGenre(@Context HttpServletRequest request,
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<Object> getDefinitions(@Context HttpServletRequest request,
+                                                @Context HttpServletResponse response,
+                                                @QueryParam("path") final String path)
+            throws IOException, NoPathParameterException, ParseException, ClassNotFoundException {
+
+        File file = toFile(path);
+        Definitions defs = IndexDatabase.getDefinitions(file);
+        return Optional.ofNullable(defs).
+                map(Definitions::getTags).orElse(Collections.emptyList()).
```
ginoaugustine, line 167: We can use flatmap here
Optional.ofNullable(defs).
                map(Definitions::getTags).
                flatMap(Collection::stream).
                map(DTOUtil::createDTO).
                collect(Collectors.toList())

               
orElse(Collections.emptyList()).

vladak: That would work if `map()` was applied to a stream, however in this case it is applied to Optional and would return `Optional<List<Definitions.Tag>>` which `flatMap()` does not know how to use.

ginoaugustine: Missed a stream invocation in between
Optional.ofNullable(defs).
map(Definitions::getTags).stream().
flatMap(Collection::stream).
map(DTOUtil::createDTO).
collect(Collectors.toList())

vladak: The `Collections.emptyList()` being static throws an axe into this scheme: _non-static method cannot be referenced from static context_ - the former referring to `Collection::stream`.

ginoaugustine (approved):
Looks Good

PR #4518, opened 2023-12-20
fix finalName warning and update plugins

vladak:
When trying to reproduce a Python tools build issue by running the build through `mvnw`, I got this warning when building distribution:
```
[INFO] --- assembly:3.1.0:single (create-archive) @ opengrok-dist ---
[WARNING]  Parameter 'finalName' is read-only, must not be used in configuration
[INFO] Reading assembly descriptor: assembly.xml
[INFO] Building tar: /home/user/opengrok-vladak-scratch/distribution/target/opengrok-1.13.0.tar.gz
[INFO] 
```
Interestingly this warning is not produced when building with system supplied `mvn`.
 
https://issues.apache.org/jira/browse/MJAR-217 says what needs to be done.

While there, I bumped the version of the Maven plugins used in the distribution module.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -47,11 +47,12 @@
   </dependencies>
 
   <build>
+      <finalName>opengrok-${project.version}</finalName>
       <plugins>
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


ginoaugustine (approved):
Looks good

PR #4520, opened 2023-12-25
Remove helix-p4d from install list

parheliamm:
helix-p4d didn't available on ARM64 platform,
there is no obvious impacts after checking docker image on ARM64

After this change, docker image created successfully on ARM64 platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,7 @@ RUN apt-get update && \
     unzip python3 python3-pip \
     python3-venv python3-setuptools openssh-client libyaml-dev
 
+# hadolint ignore=DL3008,DL3059
 RUN architecture=$(uname -m) && if [[ "$architecture" == "aarch64" ]]; then \
         echo "aarch64: do not install helix-p4d."; else \
         apt-get install --no-install-recommends -y helix-p4d; fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,9 +57,13 @@ RUN echo 'deb https://package.perforce.com/apt/ubuntu jammy release' > /etc/apt/
 # install dependencies and Python tools
 # hadolint ignore=DL3008,DL3009
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


vladak (changes requested):
This will remove Perforce from x86 image,  though. Is there a way how to make it optional ?

parheliamm:
> This will remove Perforce from x86 image, though. Is there a way how to make it optional ?

The  new patches detect the architecture during docker image build stage.
If ARM64 found, skip install helix-p4d only.

vladak:
Ack, need to have your OCA on file before this can go in.

parheliamm:
> Ack, need to have your OCA on file before this can go in.

OCA is under reviewing, it might be passed in coming days.

vladak:
Seems like the Docker image linter complains about the multiple RUNs. The error can be suppressed.

parheliamm:
> Seems like the Docker image linter complains about the multiple RUNs. The error can be suppressed.

Do I need to check the the failure issue and fix it?

vladak:
> > Seems like the Docker image linter complains about the multiple RUNs. The error can be suppressed.
> 
> Do I need to check the the failure issue and fix it?

yes, the build has to be clean for this to get in.

Dockerfile
```diff
@@ -57,10 +57,14 @@ RUN echo 'deb https://package.perforce.com/apt/ubuntu jammy release' > /etc/apt/
 # install dependencies and Python tools
 # hadolint ignore=DL3008,DL3009
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y git subversion mercurial cvs cssc bzr rcs rcs-blame helix-p4d \
+    apt-get install --no-install-recommends -y git subversion mercurial cvs cssc bzr rcs rcs-blame \
     unzip python3 python3-pip \
     python3-venv python3-setuptools openssh-client libyaml-dev
```
vladak, line 63:
```suggestion
# hadolint ignore=DL3008,DL3059
```

vladak:
Thanks !

PR #4525, opened 2024-01-08
pass output file path instead of file object

vladak:
This change should fix the regression with config file merging in Docker.

Currently I am not able to test locally because of 404 errors on security.ubuntu.com while building the image.


MasterAwesome (approved):

MasterAwesome:
The change works for me.

vladak:
Reconfirmed (building the Docker image passes once again).

PR #4532, opened 2024-01-17
handle symlinks for history based reindex

vladak:
This change deals with symlinks properly when performing history based reindex. It makes the code robust enough to avoid having directory entries in `IndexDownArgs`.

When I was writing the code for history based reindex, I remember brushing the need to use `accept()` away for some reason. It seems this was not a good idea as it provides a way how to deal with symlinks.

While at it, I changed some of the log statements to be in line with the style of wrapping paths inside single quotes used throughout the code base.


ginoaugustine (approved):
Looks Good

Issue #4317, opened 2023-04-25
Still dups in search with 1.9.4


ChristopheBordieu:
Hi @vladak 

On one of my PRD OG instance, I have installed 1.9.4 and reindexed everything from scratch.
Users report they still have  dups when searching :-(
Same issue on another UAT instance running 1.9.4 where a user found dup (after he found first a dup on the related (I mean, same code is indexed) PRD one running 1.7.35).

What could I do to try to understand the issue?

vladak:
Firstly, it would help to get more information about the dups, i.e. quantify them and see how they change over time. Ideally track when they started appearing after the reindex from scratch, esp. w.r.t. periodic reindex events.

ChristopheBordieu:
Very difficult for me to quantify the number of dups and when they start to appear, because of the size of the data indexed.

On a case that has been reported on my UAT.
I see no dup on history cache. This is a good point since on PRD (1.7.35) I have plenty of dups for history cache for the file.
```
opengrok@[uat-1.94]:/srv/opengrok-var/historycache/QWERTY/azerty$ zgrep -A 1 '<void property="revision">' Forest.xml.gz  | grep string | sed -e 's/<string>//' -e 's/<\/string>//' | sort | uniq -c | grep -v " 1 " | wc -l
0
```
```
opengrok@[prd-1.7.35]:/srv/opengrok-var/historycache/QWERTY/azerty$ zgrep -A 1 '<void property="revision">' Forest.xml.gz  | grep string | sed -e 's/<string>//' -e 's/<\/string>//' | sort | uniq -c | grep -v " 1 " | wc -l
55
```

What I don't understand is the data regarding number of commits on the file duplicated:
```
opengrok@[uat-1.94]:/srv/opengrok-src/QWERTY/azerty$ git log --oneline --merges -- Forest.xml | wc -l
116
opengrok@[uat-1.94]:/srv/opengrok-src/QWERTY/azerty$ git log --oneline -- Forest.xml | wc -l
1154
opengrok@[uat-1.94]:/srv/opengrok-src/QWERTY/azerty$ cd -
/srv/opengrok-var/historycache/QWERTY/azerty
opengrok@[uat-1.94]:/srv/opengrok-var/historycache/QWERTY/azerty$ zgrep -A 1 '<void property="revision">' Forest.xml.gz  | grep string | sed -e 's/<string>//' -e 's/<\/string>//' | sort | wc -l
1112
```

[...]

PR #4535, opened 2024-01-23
check missing source files for documents in document check

vladak:
As noted on https://github.com/oracle/opengrok/issues/4317#issuecomment-1904361808 , there is another way how the index could be broken - if there are live (i.e. not deleted) documents that are missing the source file. This change augments the document check to report these.


vladak:
The adjusted document check is needed to test the fix for #4317 , so I added the changes here as well.

vladak:
Tested with XXX iterations of the following script:
```bash
#!/bin/bash

#
# attempt to reproduce https://github.com/oracle/opengrok/issues/4317
#
# based on git-commit-hopping.sh but with added randomness

# set -x
set -e

repo_url="https://github.com/oracle/solaris-userland/"
initial_rev=32c0d9faed7b049872ca9bd78f9bf3e901cff482	# from 2022

src_root="/var/tmp/src.opengrok-issue-4317"
data_root="/var/tmp/data.opengrok-issue-4317"

# Assumes built OpenGrok.
function run_indexer()
{
	echo "Indexing $1"

        java -cp '/home/vkotal/opengrok-vladak-scratch/distribution/target/dist/*' \
	    -Djava.util.logging.config.file=/var/tmp/logging.properties-opengrok-FINEST_Console \
            org.opengrok.indexer.index.Indexer \
            -c /usr/local/bin/ctags \
	    --economical \
            -H -S -P -s "$src_root" -d "$data_root" \
	    -W "$data_root/config.xml" \
	    >/var/tmp/opengrok-issue-4317-$1.log 2>&1

	java -cp '/home/vkotal/opengrok-vladak-scratch/distribution/target/dist/*' \
            org.opengrok.indexer.index.Indexer \
	    -H \
	    -R "$data_root/config.xml" \
	    --checkIndex documents \
	    >/var/tmp/opengrok-issue-4317-$1.check.log 2>&1
}

function get_next()
{
	ids=( $(git log --oneline --reverse ..origin/master | awk '{ print $1 }') )
	size="${#ids[@]}"
	modulo=16
	if (( size == 0 )); then
		echo ""
		return
	fi
	if (( modulo > size )); then
		modulo=$size
	fi
	n=`expr $RANDOM % $modulo`
	if (( n == 0 )); then
		n=1
	fi
	echo ${ids[$n]}
}

project_root="$src_root/solaris-userland"
if [[ ! -d $src_root ]]; then
	echo "Cloning $repo_url to source root"
	git clone $repo_url "$project_root"
fi

echo "Removing data root"
rm -rf "$data_root"

echo "Removing logs"
rm -f /var/tmp/opengrok-issue*.log

cd "$project_root"
echo "Checking out base rev $initial_rev"
git checkout -q "$initial_rev"
# Establish a common time base line.
find "$src_root/" -type f -exec touch {} \;

run_indexer $initial_rev

while [[ 1 ]]; do
	rev=$(get_next)
	if [[ -z $rev ]]; then
		break
	fi
	echo "Checking out $rev"
	git checkout -q $rev
	# Git does not preserve/restore file time stamps so simulate a git pull.
	# Ideally this should be done only for the "incoming" files, howerver it suffices for this use case.
	find "$src_root/" -type f -exec touch {} \;

	run_indexer $rev
done
```

ginoaugustine (approved):

PR #4536, opened 2024-01-23
do not set default async API timeout in opengrok-projadm

vladak:
There should be a way to set indefinite async API timeout for `opengrok-projadm` to have the ability to wait for the API call completion no matter how long it takes. This is currently not possible (`None` value is needed for that and there is positive default) and this change fixes that by making it the default which I think makes sense; one rarely wants to let the API result slip away, esp. when doing reindex from scratch (a case of back-to-back index deletion which is done via async API call and project addition).


kulikjak (approved):
+1

PR #4542, opened 2024-02-01
Add null guards

pingpingy1:
While the `@NotNull` decorator informs developers that the correct usage is that the decorated variables should not be `null`, it does not enforce such in any way.
This leads to cases where an incorrectly fed `null` could propagate far away before being detected (such as via `NullPointerException`s).
This commit implements some `null` guards where needed for a fail-fast treatment of `null` inputs.


vladak:
There are some style issues, you need to fix these.

pingpingy1:
Fixed the checkstyle issues, but many untouched tests are failing, indicating coverage loss.
I will investigate the cause and perhaps revert some incorrect patches.

pingpingy1:
@vladak My apologies; it seems the errors were due to my environment setting.
Could you please review my patches after the checkstyle fixes?

vladak:
Looks good. Thanks !

PR #4543, opened 2024-02-14
launder logged file paths

vladak:
This changes logging of file paths to avoid Sonar reported log injection attacks.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileAnnotationCache.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileAnnotationCache.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2021, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2024, Oracle and/or its affiliates. All rights reserved.
  */
 package org.opengrok.indexer.history;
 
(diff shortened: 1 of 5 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/FileAnnotationCache.java
```diff
@@ -169,11 +171,16 @@ public Annotation get(File file, @Nullable String rev) throws CacheException {
              * should be present to catch weird cases of someone not using the store() or general badness.
              */
             if (storedRevision == null) {
-                LOGGER.log(Level.FINER, "no stored revision in annotation cache for ''{0}''", file);
+                if (LOGGER.isLoggable(Level.FINER)) {
```
ginoaugustine, line 174: We can use logger with msg supplier parameter to avoid this if condition. Having supplier type(lambda expression) parameter will make sure that the parameter is only evaluated if corresponding log level is enabled
Ex: LOGGER.finer( () -> String.format("no stored revision in annotation cache for ''%s''",launderLog(file.toString())));

vladak: Good idea, will be more readable that way.

ginoaugustine (approved):
Looks Good

PR #4547, opened 2024-02-20
make the hasAnnotation check more precise

vladak:
This change builds the `hasAnnotation` check on the top of xref-able. The property is part of the analyzer factory classes.


ginoaugustine (approved):
Looks Good

PR #4566, opened 2024-03-25
quote directort paths in logs in suggester

vladak:
This trivial changes unifies logging of paths with the rest of the code base.

This will wrap also cases like this:
```
25-Mar-2024 18:53:31.632 WARNING [Thread-5166] org.opengrok.suggest.SuggesterProjectData.initFields Fields [defs, path, hist, refs, type, full] will be ignored because they were not found in index directory MMapDirectory@/opengrok/data/index/foo lockFactory=org.apache.lucene.store.NativeFSLockFactory@73c140fc
```
however I found no way how to get the file system path out of the `Directory` object.


vladak:
The changes to quote `indexDir` are not exactly quoting the directory path as `indexDir.toString()` will emit some representation of the Lucene's `Directory` object which will contain implementation dependent information and there does not seem to be a way how to extract the actual path, however I think it should still be quoted.

PR #4570, opened 2024-04-23
lucene 10.0.0

tarzanek:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@ Portions Copyright (c) 2018, 2020, Chris Fraire <cfraire@me.com>.
     </scm>
 
     <properties>
-        <lucene.version>10.0.0</lucene.version>
+        <lucene.version>10.3.1</lucene.version>
         <mavenjavadocplugin.version>3.6.0</mavenjavadocplugin.version>
         <java.version>21</java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>


suggester/src/main/java/org/opengrok/suggest/SuggestResultCollector.java
```diff
@@ -62,6 +64,29 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
...
+            @Override
+            public SuggestResultCollector newCollector() {
+                return new SuggestResultCollector(leafReaderContext, data, documentIds);
+            }
+
+            @Override
+            public BitIntsHolder reduce(Collection<SuggestResultCollector> collectors) {
+                BitIntsHolder reduced = documentIds;
+                for (SuggestResultCollector collector : collectors) {
+                    documentIds.or(collector.documentIds); //TODO fix as per https://github.com/apache/lucene/pull/766/files
```
tarzanek, line 83: this needs fixing

tarzanek:
seems SuggestResultCollector.java has no testing coverage ...

tarzanek:
@ahornace any clues for proper fixing of reduce for that collector? Is it even used?

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -205,18 +205,17 @@ private void searchMultiDatabase(List<Project> projectList, boolean paging) thro
     }
 
     private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
-        collector = TopScoreDocCollector.create(hitsPerPage * cachePages, Short.MAX_VALUE);
+        collectorManager = new TopScoreDocCollectorManager(hitsPerPage * cachePages, Short.MAX_VALUE);
```
tarzanek, line 208: perhaps redesign to reuse the object?

tarzanek:
lucene 10 support needs to be rewritten into new PR
and JDK 21 change should be also separate, but a prereq for lucene 10 ...

opengrok-indexer/src/main/java/org/opengrok/indexer/web/Util.java
```diff
@@ -677,13 +677,12 @@ public static void encode(String s, Appendable dest) throws IOException {
...
      * @throws URISyntaxException URI syntax
-     * @throws MalformedURLException URL malformed
      */
-    public static String encodeURL(String urlStr) throws URISyntaxException, MalformedURLException {
-        URL url = new URL(urlStr);
-        URI constructed = new URI(url.getProtocol(), url.getUserInfo(),
-                url.getHost(), url.getPort(),
-                url.getPath(), url.getQuery(), url.getRef());
+    public static String encodeURL(String urlStr) throws URISyntaxException {
+        // URL url = new URL(urlStr); - this call
```
vladak, line 682: I took a longer look at the departure from the `URL()` constructors yesterday. The `URL` class by itself does not support any encoding/decoding, it merely breaks down the URL into pieces. The encoding is then supplied by the `URI` class.

The problem is that `URI(String)` acceots only already encoded URIs and refuses any non-compliant characters. I think the way to go is to propagate the `URL` object all the way where possible, i.e. change the `String url` parameter type of the various `linkify()` and `buildLink()` methods to `URL url` and avoid passing the `URL` in the map of attributes - pass it as an extra argument with the `URL` type.

It is still not clear to me where the unescaped characters could originate from.

tarzanek:
I will split lucene 10 changes away from this PR, get the lucene 9 done (I think only suggester test is missing before it can be merged)
and for lucene 10 we can make the change once jdk 21 is supported

tarzanek:
cool, so I should be ready to revive this draft and ev. let AI do the lucene upgrade :-D, let's try

tarzanek:
will close this in favour of https://github.com/oracle/opengrok/pull/4929

PR #4584, opened 2024-06-05
Jackson update

vladak:
This change updates mainly Jackson but also other outdated dependencies.


vladak:
The macOS build is failing due to an error in pre-requisites install, will be dealt with separately.

ginoaugustine (approved):

Issue #4441, opened 2023-10-17
testDeserializationOfNotWhiteListedClassThrowsError sometimes fails on Windows


vladak:
The `ConfigurationTest.testDeserializationOfNotWhiteListedClassThrowsError` test sometimes fails on Windows, specifically with the 2 parameter:
```
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.657 s <<< FAILURE! - in org.opengrok.indexer.configuration.ConfigurationTest
Error:  org.opengrok.indexer.configuration.ConfigurationTest.testDeserializationOfNotWhiteListedClassThrowsError(String)[2]  Time elapsed: 0.047 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: Unexpected exception type thrown, expected: <java.lang.IllegalAccessError> but was: <java.io.IOException>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:67)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3111)
	at org.opengrok.indexer.configuration.ConfigurationTest.testDeserializationOfNotWhiteListedClassThrowsError(ConfigurationTest.java:223)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
[...]

PR #4599, opened 2024-07-22
convert LDAP authorization plugin configuration to YAML

vladak:
This is a little brother of #2329. For the LDAP authorization plugins I have chosen YAML.

This will also help with the stability of the tests, as I have seen `plugins/src/test/java/opengrok/auth/plugin/configuration/ConfigurationTest.java` fail in the same way as the indexer configuration (#4441).


ginoaugustine (dismissed):

vladak:
https://github.com/oracle/opengrok/wiki/Authorization-plugins#ldap-setup will have to be updated once this is merged in.

PR #4608, opened 2024-07-25
handle empty Git repositories gracefully

vladak:
As mentioned on https://github.com/oracle/opengrok/discussions/4606, empty Git repositories fail history cache creation, making the indexing of related project fail. This change fixes that.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2024, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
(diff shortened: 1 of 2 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/history/GitRepository.java
```diff
@@ -458,6 +459,11 @@ public int getPerPartesCount() {
         return MAX_CHANGESETS;
     }
 
+    private boolean isRepositoryEmpty() {
+        File headsFile = Paths.get(getDirectoryName(), Constants.DOT_GIT, "refs", "heads").toFile();
+        return headsFile.isDirectory() && (Objects.requireNonNull(headsFile.listFiles()).length == 0);
```
vladak, line 464: for better robustness I should also list the repository directory and only consider this empty repository if that listing also comes out empty.

vladak: except it should do the listing via something like `git ls-files` to avoid untracked files.

vladak: The setup required for `ls-files` would require `TreeWalk` with a commit which needs the empty repository check.

ginoaugustine (approved):

PR #4627, opened 2024-08-01
add tests for MercurialRepository.annotate()

vladak:
This change increases coverage of `MercurialRepository`, in particular `MercurialAnnotationParser`. This is a basic test, ideally all the `Annotation` members should be verified in the positive case.


ginoaugustine (approved):
Looks good to me

PR #4635, opened 2024-08-13
add test coverage for GetFile

vladak:
This change adds basic test coverage for the `GetFile` class used in the webapp. I fixed some related nits along the way.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/GetFile.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/GetFile.java
@@ -50,12 +50,12 @@ public class GetFile extends HttpServlet {
         cfg.checkSourceRootExistence();
 
         String redir = cfg.canProcess();
-        if (redir == null || !redir.isEmpty()) {
-            if (redir != null) {
-                response.sendRedirect(redir);
-            } else {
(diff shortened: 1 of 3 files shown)


opengrok-web/src/main/java/org/opengrok/web/GetFile.java
```diff
@@ -29,24 +29,28 @@
...
-    public static final long serialVersionUID = -1;
+    private static final long serialVersionUID = -1;
 
     @Override
     protected void service(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
         PageConfig cfg = PageConfig.get(request);
         cfg.checkSourceRootExistence();
 
         String redir = cfg.canProcess();
-        if (redir == null || redir.length() > 0) {
```
ginoaugustine, line 49: Would it be more efficient to use an early return for both conditions?
if(redir==null){
 response.setnotfound):
 return;
}
if(!redir.isEmpty()){
 response.sendRedirect(redir);
 return;
}

vladak: looks much more readable, thanks !

ginoaugustine (commented):
Looks good to me

PR #4643, opened 2024-08-20
compile Universal ctags with XML support

vladak:
Install `libxml2-dev` to support XML in Universal ctags built for the Docker image.

docker build excerpt:
```
checking for libxml-2.0 >= 2.7.7... yes
checking libxml-2.0 CRLF handling bug... good
```
Previously the answer was `no`.

Checking languages support in the container built with these changes:
```
$ docker exec opengrok-dev /usr/local/bin/ctags --list-languages | grep XML
PlistXML
XML
```


adamhradil:
LGTM 👍

PR #4659, opened 2024-09-09
add tests for Mercurial tag parsing

vladak:
This change introduces test coverage for the `MercurialTagParser` class. While there, I fixed issue with parsing tags containing multiple spaces.


vladak:
Shout-out to @skozina

ginoaugustine (approved):

PR #4671, opened 2024-10-09
do not constrain the output of the Groups command in opengrok-groups

vladak:
Fixes a problem with long lines/output generated by the `Groups` command where previously `opengrok-groups` (the wrapper) would truncat it, leading to invalid XML.

While working on this, the Linux build failed on me due to the Python packages (pip) being externally managed, so this change fixes this (for the time being leaving the script behind).


kulikjak (approved):
LGTM

PR #4677, opened 2024-10-16
enforce Java version

vladak:
Inspired by https://github.com/oracle/opengrok/discussions/4675, I resurrected an old branch to enforce Java version range for the runtime of the indexer and webapp.

Next to the Oracle Java 11, I also tried with Amazon Corretto 11.0.25.


vladak:
updated https://github.com/oracle/opengrok/wiki/How-to-setup-OpenGrok

PR #4679, opened 2024-10-29
fall back to the default temporary directory for ctags check

vladak:
This change relaxes the ctags check so that it works also for non-writable source root which is legit.

In order to test that, I had to bump Mockito in order to be able to use static method mocking. I contemplated a bit about making `CtagsUtil` non-static however turned the idea down as it would be too big change for such small change and generally it seems that the static methods have their own place for this use case.


ginoaugustine (approved):
Looks good to me

PR #4689, opened 2024-11-16
Fortran indexing broken 

navinp0304:
As of now fortran indexing is broken because of my previous change.
```
 Second argument of substring is wrong in my previous commit 

It should be 
cmatch.substring(cmatch.length() - 1, cmatch.length()) instead of  cmatch.substring(cmatch.length() - 1, 1)
```
@vladak  Please review

### DIFF
--- a/opengrok-indexer/src/main/jflex/analysis/fortran/FortranXref.lex
+++ b/opengrok-indexer/src/main/jflex/analysis/fortran/FortranXref.lex
@@ -93,7 +93,7 @@ File = [a-zA-Z]{FNameChar}* "." ("i"|"inc")
     onNonSymbolMatched(cmatch.substring(0, 1), yychar);
     String file = cmatch.substring(1, cmatch.length() - 1);
     onFilelikeMatched(file, yychar + 1);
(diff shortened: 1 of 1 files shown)


opengrok-indexer/src/main/jflex/analysis/fortran/FortranXref.lex
```diff
@@ -93,7 +93,7 @@ File = [a-zA-Z]{FNameChar}* "." ("i"|"inc")
     onNonSymbolMatched(cmatch.substring(0, 1), yychar);
     String file = cmatch.substring(1, cmatch.length() - 1);
     onFilelikeMatched(file, yychar + 1);
-    onNonSymbolMatched(cmatch.substring(cmatch.length() - 1, 1), yychar + 1 + file.length());
+    onNonSymbolMatched(cmatch.substring(cmatch.length() - 1, cmatch.length()), yychar + 1 + file.length());
```
vladak, line 96: I see, the second argument to `substring()` is the end index, not length of the substring.

vladak:
Looks good, although I am surprised there is no change to the test data.

vladak:
> Looks good, although I am surprised there is no change to the test data.

I mean - if the change broke Fortran, there should be some sort of test case to cover this.

PR #4690, opened 2024-11-19
update Micrometer to 1.14.1

vladak:
Updates the Micrometer in order to fix a vulnerability in netty libraries used by the `micrometer-registry-statsd`.


ginoaugustine (approved):

vladak:
Tested with locally running `statsd` and indexer set to send the metrics, verified with `tcpdump -X` that the statsd data is flowing while the indexer is running. Also checked the metrics are recorded by running `wget http://localhost:9123/metrics` in the container.

Issue #3582, opened 2021-05-06
Automatic display of unique searched file


ChristopheBordieu:
Question.

OG 1.6.6.

When full searching, if exactly 1 occurence is found, then the matched file is automatically displayed (/xref/...) instead of getting a search result page (with directory, file and lines).

Is there a way to avoid this automatic display of the file?

vladak:
I don't think so. That would be a global (or perhaps per project) tunable.

ChristopheBordieu:
Ok, thanks.
Should be a tunable per user. Because it is totally user dependant such a behavior.
Closing.

ChristopheBordieu:
Automatic display is not configurable.

vladak:
@ahornace recently mentioned having per user configuration for #3555 so perhaps one day it will be possible.

ChristopheBordieu:
Yep. Remember this. (Users) Would love that.

vladak:
Then keep it open.

vladak:
The foundation for per user settings is now in place via PR #3642.

RetroZelda:
any updates on this?  the redirection with a single result is extremely annoying

vladak:
Not at the moment. PRs are definitely welcome.

PR #4712, opened 2025-01-27
Add option to disable immediate redirect on singular results

edbaunton:
Per #3582 add an option to disable the redirection behaviour when there is only a single file result.


vladak:
Tried to find your OCA on https://oca.opensource.oracle.com/?ojr=contrib-list however could not get a match.

vladak:
Looking into the OCA submission from 2025, it was rejected because of invalid mailing address (that is snail mail address rather than e-mail address). Would you resubmit the OCA, please ?

PR #4727, opened 2025-02-18
HTML/URI encode content/path where possible

vladak:
This change attempts to fix XSS issues found in the webapp, namely the one in the `/history` endpoint handling.

While there, I took the chance to fix neighboring code style/comment issues.


opengrok-web/src/main/webapp/history.jsp
```diff
@@ -259,11 +258,12 @@ document.domReady.push(function() {domReadyHistory();});
             int count=0;
             for (HistoryEntry entry : hist.getHistoryEntries(maxItems, startIndex)) {
                 String dispRev = entry.getDisplayRevision();
-                if (dispRev == null || dispRev.length() == 0) {
+                if (dispRev == null || dispRev.isEmpty()) {
```
ginoaugustine, line 261: Since dispRev.isEmpty() is equivalent to dispRev = "", can we remove the explicit check.

Using the Optional API might simplify this.

Sample:
String dispRev = Optional.ofNullable(entry)
                                        .map(HistoryEntry::getDisplayRevision)
                                        .map(Util::htmlize)
                                        .orElse("")

vladak: That's nicer, thanks.

vladak:
Tested the history UI on locally deployed instance. Specifically the XSS unit test as well as basic functionality (selecting and comparing 2 revisions, displaying historical revision of a file).

ginoaugustine (approved):
Looks good to me

PR #4739, opened 2025-03-10
change max Java version to v22 in wiki

WilliamLeara:
The wiki says Java v11 to v22, but actually v22 is blocked by the installer and won't work.  See this commit, which blocks v22: https://github.com/oracle/opengrok/commit/785896c3c8709731cca53bc6791419743dfcdefe

Therefore, the wiki should be updated to match.


WilliamLeara:
The Oracle OCA is brain-dead stupid.  I'm not signing my life away to fix your wiki.

vladak:
Understood. Fixed the wiki.

PR #4742, opened 2025-03-17
filter out inactive branch tags in Mercurial

vladak:
This change reduces the list of tags shown for Mercurial repository to the tags found on given branch.

Thanks to @vlmarek and @muzikm for supplying the Mercurial command line pieces.


ginoaugustine (approved):
LGTM

PR #4746, opened 2025-03-18
avoid global isHistoryEnabled() checks in favor of per repository check

vladak:
This change allows for real per project history override.

Tested with indexer running with:
```
-P
-S
-R
/var/opengrok/etc/readonly_configuration-per_project_history.xml
-s
/var/opengrok/src.perproj
-d
/var/opengrok/data.perproj
-W
```
where the contents of `/var/opengrok/etc/readonly_configuration-per_project_history.xml` looked like this:
```xml

    <boolean>true</boolean>
  </void>

   <boolean>true</boolean>
  </void>

   <void property="projects">
     <void method="put">
      <string>foo</string>
        <boolean>false</boolean>
       </void>
        <boolean>false</boolean>
       </void>
       <void property="name">
        <string>foo</string>
       </void>
       <void property="path">
        <string>/foo</string>
       </void>
        <boolean>false</boolean>
       </void>
      </object>
     </void>
   </void>

 </object>
</java>
```
The source root had two projects, each with single Git repository, named `foo` and `bar`. I ran two tests, one with global history on and per project/repository off and vice versa. Checked the `list`/`minisearch` JSPs in debugger for each test.

Also ran indexer with and without -H to make sure the history is disabled by default as before the change and also that the history cache is created if history is on (to test the `AnalyzerGuru` changes).

Also tested in project-less mode.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2025, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2021, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.analysis;
@@ -596,7 +596,7 @@ public class AnalyzerGuru {
         doc.add(new SortedDocValuesField(QueryBuilder.FULLPATH,
                 new BytesRef(file.getAbsolutePath())));
 
-        if (RuntimeEnvironment.getInstance().isHistoryEnabled()) {
+        if (HistoryGuru.getInstance().repositorySupportsHistory(file)) {
             populateDocumentHistory(doc, file);
         }
         doc.add(new Field(QueryBuilder.DATE, date, string_ft_stored_nanalyzed_norms));
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2007, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2025, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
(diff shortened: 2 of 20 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/AnalyzerGuru.java
```diff
@@ -596,7 +596,7 @@ public void populateDocument(Document doc, File file, String path, AbstractAnaly
         doc.add(new SortedDocValuesField(QueryBuilder.FULLPATH,
                 new BytesRef(file.getAbsolutePath())));
 
-        if (RuntimeEnvironment.getInstance().isHistoryEnabled()) {
+        if (HistoryGuru.getInstance().repositorySupportsHistory(file)) {
```
vladak, line 599: uses `HistoryGuru#getRepository(file)` which is based on cached repository lookup so this should not incur performance penalty.

opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Configuration.java
```diff
@@ -580,7 +580,7 @@ public Configuration() {
         setHandleHistoryOfRenamedFiles(false);
         setHistoryBasedReindex(true);
         setHistoryCache(true);
-        setHistoryEnabled(true);
+        setHistoryEnabled(false);
```
vladak, line 583: done to avoid `cfg.setHistoryEnabled(false);` in the `Indexer` which causes confusing warning messages about discrepancy between history and history based reindex.

opengrok-indexer/src/main/java/org/opengrok/indexer/index/Indexer.java
```diff
@@ -278,6 +276,8 @@ public static int runMain(String[] argv) {
                 }
             }
 
+            checkConfiguration(cfg);
```
vladak, line 279: Moved to just before the call to `setConfiguration()` so that it could reflect the `cfg` changes done above.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -202,7 +202,7 @@ private void deleteProjectWorkHorse(Project project) {
             group.getProjects().remove(project);
         }
 
-        if (env.isHistoryEnabled()) {
+        if (project.isHistoryEnabled()) {
```
ginoaugustine, line 205: I may have overlooked it, but I am not seeing where isHistoryEnabled is set at project level. Could you clarify how this occurs?

vladak: This happens here in `completeWithDefaults()`: https://github.com/oracle/opengrok/blob/c76efba6f525064aed586fb9a661ded0ef8f9b95/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java#L497-L500

ginoaugustine (approved):

Issue #4459, opened 2023-10-27
jdk #21 support


vladak:
Java SE 21 is a LTS release, we should consider switching to that. Might still want to keep the Java 11 build target to support existing installations. The transition will likely fix the build warnings in #4312.

tarzanek:
I partially did the migration in https://github.com/oracle/opengrok/pull/4570
(which compiles and runs tests with jdk 21 ... just tests fail, since I need to fix URL to URI compat and tests)

PR #4750, opened 2025-04-01
replace URL() constructors

vladak:
This change gets rid of the `URL()` constructors deprecated in Java 20 and addresses a case of unescaped patterns.

Background: https://gist.github.com/vladak/beb43b9056b546869196d09a16542466

Approaches #4459 and takes over the `URL()` changes from @tarzanek's PR #4570 

I had to add a flake8 suppression to the Docker entry point to make the build pass.

Tested with this read-only configuration:
```xml
<java version="1.8.0_121" class="java.beans.XMLDecoder">
 <object class="org.opengrok.indexer.configuration.Configuration">

  <void property="bugPage">
   <string>https://github.com/apache/lucene/issues/</string>
  </void>

  <void property="bugPattern">
   <string>LUCENE-(\d+)</string>
  </void>

 </object>
</java>
```
and The Lucene Git repository.

### DIFF
--- a/docker/start.py
+++ b/docker/start.py
@@ -115,7 +115,7 @@ def index():
     API endpoint for triggering reindex.
     :return: message describing the outcome
     """
-    global periodic_timer   # noqa: F824
+    global periodic_timer  # noqa: F824
 
     if periodic_timer:
         logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


ginoaugustine (approved):
LGTM

PR #4757, opened 2025-04-15
Add support for OCaml.

progoramur:
I am signing Oracle Contributor Agreement. 

This pull request adds support for OCaml. 

Almost full support for OCaml syntax is provided including quoted string literals and extension nodes. The support for docstrings is not implemented. Also, primitive types are added to the list of keywords because most programmers do not use them as variable names (`result` is an exception).


progoramur:
@vladak Vladimir, I've signed OCA. For some reason the check does not pass. Would you please double check.

PR #4759, opened 2025-04-15
switch to Java 21 for build/Docker

vladak:
The bytecode target is still Java 11 to support older deployments.

I had to solve multiple issues:
  - bump Mockito so that it uses newer Byte buddy library which supports Java 21
  - address non-serializable members by either switching them to `transient` or using `Serializable` type
  - avoid `this` leaks from constructors or suppress the warning
  - udpate Chronicle Map so that it does not attempt to get method not present in Java 21

Tested the newly built Docker image with pre-existing data root. Also tested with JRE 11 on Solaris x86 with pre-existing data-root and semi-complex setup (authorization, suggester, annotation cache).

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2025, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2018, Chris Fraire <cfraire@me.com>.
  */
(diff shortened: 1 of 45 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Definitions.java
```diff
@@ -66,24 +66,24 @@ public static class LineTagMap implements Serializable {
 
         private static final long serialVersionUID = 1191703801007779481L;
         @SuppressWarnings("java:S116")
-        private final Map<String, Set<Tag>> sym_tags; //NOPMD
+        private final HashMap<String, Set<Tag>> sym_tags; //NOPMD
```
ginoaugustine, line 69: Could you explain the reasoning behind the shift from using the Map interface to hashmap here?

ginoaugustine (approved):

PR #4764, opened 2025-04-17
use reflection to get thread ID

vladak:
This change gets rid of the remaining deprecation warnings when building with target version of 21. It's a bit of a hack, however removes potential developer barrier.


ginoaugustine (approved):
LGTM

PR #4794, opened 2025-06-04
control history tagging per project

vladak:
This change introduces the `tagsEnabled` per project/repository tunable that allows to control whether tags should be applied to history entries (so that they are visible in history view).

This will require a change to https://github.com/oracle/opengrok/wiki/Per-project-configuration

Tested with the indexer running with:
```
-s
/var/opengrok/src
-d
/var/opengrok/data
-P
-H
-S
-c
/opt/homebrew/bin/ctags
-R
/var/opengrok/etc/ro-config-project.xml
-W
```
and the `/var/opengrok/etc/ro-config-project.xml` overriding the `tagsEnabled`:
```xml
<java version="1.8.0_121" class="java.beans.XMLDecoder">
 <object class="org.opengrok.indexer.configuration.Configuration">

  <void property="projects">
   <void method="put">
    <string>Lucene</string>
      <boolean>true</boolean>
     </void>
      <boolean>true</boolean>
     </void>
     <void property="tagsEnabled">
      <boolean>true</boolean>
     </void>
      <boolean>true</boolean>
     </void>
     <void property="name">
      <string>Lucene</string>
     </void>
     <void property="path">
      <string>/Lucene</string>
     </void>
    </object>
   </void>
  </void>

 </object>
</java>
```
Then after the indexing is done, check that the history view contains list of tags.

The indexer log entry relevant to the repository:
```
INFO: Creating history cache for {dir='/private/var/opengrok/src/Lucene',type=git,history=on,historyCache=on,renamed=true,merge=true,annotationCache=off,tagsEnabled=on}
Jun 04, 2025 3:00:10 PM org.opengrok.indexer.history.HistoryGuru createHistoryCache
```

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/configuration/Project.java
@@ -104,6 +104,11 @@ public class Project implements Comparable<Project>, Nameable, Serializable {
     private Boolean mergeCommitsEnabled = null;
 
     /**
+     * This flag enables/disables per project tagging of history entries.
+     */
+    private Boolean tagsEnabled = null;
+
+    /**
      * Username to used for repository authentication. This is propagated to all repositories of this project.
      */
     private String username = null;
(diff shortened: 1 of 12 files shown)


vladak:
Updated the wiki page.

PR #4800, opened 2025-06-12
avoid marking non-existent project as indexed in the webapp

vladak:
This change gets rid of the warning message for newly discovered projects during indexer run.

I contemplated various solutions:
  - relaxing the indexed API call to ignore non-existent projects (lacks transparency / hides errors)
  - propagating the list of discovered projects from the `prepareIndexer()` (too complex and would not reflect real webapp state)
  - inferring project existence via project property call (clunky)

In the end the code retrieves the current list of projects known to the webapp. While this may seem a bit heavy, because this will be done for each indexed project, the webapp should withstand the traffic just fine and I wanted to avoid stashing the list in `RuntimeEnvironment` or somewhere in the `Indexer` and passing it around to `IndexDatabase`. Of course, I could have introduced new API end point to check for project existence, however deemed this unnecessary at this point.

Naturally, there is a time window between the check and the eventual mark-as-indexed operation, however let's assume the chance of the project to be deleted in that window is negligible.

While there, I refactored the code which performs the API call to mark the project as indexed into `IndexerUtil`.

### DIFF
--- a/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
+++ b/opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
@@ -134,17 +134,17 @@ public class ProjectsController {
                 // This is the goal of this action: if an existing project
                 // is re-added, this means its list of repositories has changed.
                 List<RepositoryInfo> repos = getRepositoriesInDir(projDir);
-                List<RepositoryInfo> allrepos = env.getRepositories();
-                synchronized (allrepos) {
+                List<RepositoryInfo> allRepositories = env.getRepositories();
+                synchronized (allRepositories) {
                     // newly added repository
                     repos.stream()
-                            .filter(repo -> !allrepos.contains(repo))
-                            .forEach(allrepos::add);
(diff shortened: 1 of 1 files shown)


opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/ProjectsController.java
```diff
@@ -420,8 +419,7 @@ public Object get(@PathParam("project") String projectName, @PathParam("field")
 
         Project project = env.getProjects().get(projectName);
         if (project == null) {
-            throw new WebApplicationException(
-                    "cannot find project '" + projectName + "' to get a property", Response.Status.BAD_REQUEST);
+            throw new NotFoundException(String.format("cannot find project '%s' to get a property", projectName));
```
ginoaugustine, line 422: Consider using Optional.ofNullable to handle the potential null value from env.getProjects().get(projectName).
ex : project = Optional.ofNullable(env.getProjects().get(projectName))
                                 .orElseThrow();

vladak: That's nicer, thanks for the suggestion.

PR #4865, opened 2025-10-02
Add sort argument to the search API 🔍

gaborbernat:
Fixes #4740

Adds a `sort` parameter to the search API that allows clients to control result ordering:

- **`relevancy`** (default) - Results ranked by Lucene score
- **`lastmodtime`** - Results ordered by file modification date (newest first)  
- **`fullpath`** - Results ordered alphabetically by file path

The implementation uses Lucene 9's `TopFieldCollectorManager` API for sorted searches and `TopScoreDocCollectorManager` for relevancy-based searches.

📝 API documentation updated in `apiary.apib`
✅ Tests verify correct ordering for both `LASTMODIFIED` and `BY_PATH` sort modes

PR #4867, opened 2025-10-07
Lucene 9

tarzanek:

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -48,7 +48,7 @@ import org.apache.lucene.queryparser.classic.ParseException;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreDoc;
-import org.apache.lucene.search.TopScoreDocCollector;
+import org.apache.lucene.search.TopScoreDocCollectorManager;
 import org.apache.lucene.util.Version;
 import org.opengrok.indexer.analysis.AbstractAnalyzer;
 import org.opengrok.indexer.analysis.CompatibleAnalyser;
@@ -132,7 +132,7 @@ public class SearchEngine {
     int cachePages = RuntimeEnvironment.getInstance().getCachePages();
     int totalHits = 0;
     private ScoreDoc[] hits;
-    private TopScoreDocCollector collector;
+    private TopScoreDocCollectorManager collectorManager;
     private IndexSearcher searcher;
     boolean allCollected;
     private final ArrayList<SuperIndexSearcher> searcherList = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -205,18 +205,17 @@ private void searchMultiDatabase(List<Project> projectList, boolean paging) thro
...
+        hits = searcher.search(query, collectorManager).scoreDocs;
+        totalHits = searcher.count(query);
         stat.report(LOGGER, Level.FINEST, "search via SearchEngine done",
                 "search.latency", new String[]{"category", "engine",
                         "outcome", totalHits > 0 ? "success" : "empty"});
         if (!paging && totalHits > 0) {
-            collector = TopScoreDocCollector.create(totalHits, Short.MAX_VALUE);
-            searcher.search(query, collector);
+            collectorManager = new TopScoreDocCollectorManager(totalHits, Short.MAX_VALUE);
+            hits = searcher.search(query, collectorManager).scoreDocs;
```
idodeclare: Since the new version is compelling this somewhat intrusive revision in order to use `CollectorManager` instead of `Collector`, may we rationalize at this time to call `search()` only once whether `paging` is true or false? That is, avoid a `search()` on line 210 and 217 when `paging` is false?
This way also we can make sure `search()` only happens prior to reporting `stat()` on line 212. As is, the `Statistics` aren't counting the `search()` happening on line 217.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -205,18 +205,17 @@ private void searchMultiDatabase(List<Project> projectList, boolean paging) thro
     }
 
     private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
-        collector = TopScoreDocCollector.create(hitsPerPage * cachePages, Short.MAX_VALUE);
+        collectorManager = new TopScoreDocCollectorManager(hitsPerPage * cachePages, Short.MAX_VALUE);
         Statistics stat = new Statistics();
-        searcher.search(query, collector);
-        totalHits = collector.getTotalHits();
+        hits = searcher.search(query, collectorManager).scoreDocs;
+        totalHits = searcher.count(query);
```
idodeclare: Formerly `totalhits` would have been bounded by the `Short.MAX_VALUE` value of `totalHitsThreshold` (on the `TopScoreDocCollector`). For very large indexes, is there any possible performance penalty getting the full, "fully accurate" `int` value of `totalHits`?

idodeclare (commented):
Hi, Lubos. A couple of comments and questions, mostly concerning `paging` handling

lkosco:
@idodeclare would those changes work?

vladak:
The changes in `SearchEngine` will create conflicts with PR #4865. My gut feel is to merge this PR first, because it is smaller and the collector changes will merely create one more level of abstraction wihch can be easily resolved for the API changes, I believe.

vladak:
Thanks guys !

PR #4872, opened 2025-10-15
Allow to query to index time endpoint without token

gaborbernat:
Resolves https://github.com/oracle/opengrok/issues/4866


opengrok-web/src/test/java/org/opengrok/web/api/v1/filter/IncomingFilterTest.java
```diff
@@ -243,4 +243,9 @@ void systemPathDescWithoutTokenTest() throws Exception {
 
         assertEquals(Response.Status.UNAUTHORIZED.getStatusCode(), captor.getValue().getStatus());
     }
+
+    @Test
+    void systemIndexTimeRemoteWithoutTokenTest() throws Exception {
+        assertFilterDoesNotBlockAddress("10.0.0.1", "system/indextime");
```
vladak: use `SystemController.PATH` plus perhaps add a constant for the endpoint.

vladak (approved):

PR #4905, opened 2026-01-21
Add user-select CSS for line numbers in file viewer

### LINKED ISSUE #4856: Add user-select CSS for line numbers in file viewer
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
Currently line numbers are included in selections on copy, which is not typically useful. Line numbers can be disabled, however many other platforms / file editors exclude these from selections.

add user-select css properties to disable selection of the line numbers.

stescobedo92:
# Add CSS to disable text selection for line numbers
Fixes #4856
### Description
This pull request improves the user experience when copying code from the file viewer and diff views. Currently, selecting text often includes the line numbers, which interferes with copying code snippets.
I have added `user-select: none` (and its vendor-prefixed variants) to the CSS classes responsible for rendering line numbers (`.l`, `.hl`, `del.d`, etc.). This ensures that line numbers are skipped during text selection while ensuring links and other interactions remain functional.
### Changes
- Modified [opengrok-web/src/main/webapp/default/style-1.0.4.css] to apply selection disabling styles to line number classes.
### Verification
- Validated that the modified CSS classes (`.l`, `.hl`, `del.d`) are correctly used for line numbers in [diff.jsp] and other views.
- Verified that `user-select` is the standard approach for this behavior and does not affect the ability to click on line number links (if applicable).

### DIFF
--- a/opengrok-web/src/main/webapp/default/style-1.0.4.css
+++ b/opengrok-web/src/main/webapp/default/style-1.0.4.css
@@ -740,6 +740,10 @@ div[id^='src'] pre {
     background-color: #dddddd;
     color: #666;
     margin-right: .5ex;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
 }
 
 .blame .search {


vladak:
The line numbers being part of the selection is actually intentional. The Line toggle at the top of the page (next to the History and Annotate links) should be used to get contents without line numbers.

PR #4913, opened 2026-01-23
Docker: fix chown missing operand

TadayukiOkada:
I get this error with the latest (1.14.9) docker image:
chown: missing operand after ‘appuser:appgroup’
My change seems to fix the problem.


vladak:
Sorry, did not notice this PR, fixed via #4914.

Thanks for the alternative approach, though.

PR #4921, opened 2026-02-04
Fix zombie ctags processes by adding waitFor() after destroyForcibly()

### LINKED ISSUE #4920: Zombie ctags processes accumulate due to missing waitFor() after destroyForcibly()
When ctags processing times out or is terminated, `destroyForcibly()` is called but `waitFor()` is never called to reap the child process. This causes zombie (defunct) processes to accumulate over long indexing runs.

On Unix-like systems, a zombie process persists until the parent calls `wait()` to collect its exit status. **If the parent never calls `wait()`, zombies remain for the entire lifetime of the parent process** — in this case, potentially 20+ hours during a full index run.

## Affected Code

| Location | Code | Issue |
|----------|------|-------|
| [Ctags.java:145](https://github.com/oracle/opengrok/blob/master/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java#L145) | `ctagsProcess.destroyForcibly();` | No waitFor() |
| [Ctags.java:606](https://github.com/oracle/opengrok/blob/master/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java#L606) | `ctagsProcess.destroyForcibly();` | No waitFor() |
| [Ctags.java:609](https://github.com/oracle/opengrok/blob/master/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java#L609) | `ctagsProcess.destroyForcibly();` | No waitFor() |

## Environment Where Observed

- Long-running indexer (21+ hours)
- High parallelism (`indexingParallelism: 96`)
- Short ctags timeout (`ctagsTimeout: 5s`)
- 232K+ repositories being indexed

## Root Cause

On Unix-like systems, when a child process terminates, it remains as a zombie until the parent calls `wait()` to collect its exit status. Java's `destroyForcibly()` sends SIGKILL but does not reap the process.

From the [Java Process API documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Process.html#destroyForcibly()):

> **API Note:** The process may not terminate immediately. This method may be chained to `waitFor()` if needed.

Since the indexer runs for many hours and spawns thousands of ctags processes (one per file analyzed), zombies accumulate continuously whenever ctags times out or encounters errors.

## Impact

1. **Resource exhaustion**: Hundreds of zombie processes consume kernel process table entries
2. **systemd churn**: On systems using cgroups v2, `systemd --user` continuously polls cgroup membership waiting for zombies to disappear, causing elevated CPU usage
3. **System instability**: In extreme cases, can approach PID limits or trigger monitoring alerts

## Proposed Fix

Add `waitFor()` with a timeout after each `destroyForcibly()` call:

```java
ctagsProcess.destroyForcibly();
try {
    if (!ctagsProcess.waitFor(10, TimeUnit.SECONDS)) {
        LOGGER.log(Level.WARNING, "ctags process did not terminate within timeout after SIGKILL");
    }
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
}
```

Using a bounded `waitFor(timeout)` rather than bare `waitFor()` ensures we:
- Reap the process when it exits normally (preventing zombies)
- Don't block indefinitely if the process is stuck in an uninterruptible kernel state (e.g., blocked on NFS I/O)

## References

- [Java SE 17 Process.destroyForcibly()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Process.html#destroyForcibly())
- [Java SE 17 Process.waitFor(long, TimeUnit)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Process.html#waitFor(long,java.util.concurrent.TimeUnit))
- [Unix zombie process](https://en.wikipedia.org/wiki/Zombie_process) - "A zombie process ... remains in the process table ... This occurs for child processes, where the entry is still needed to allow the parent process to read its child's exit status"

gaborbernat:
When ctags processing times out or is terminated, `destroyForcibly()` sends SIGKILL but the parent never calls `waitFor()` to reap the child process. This causes zombie (defunct) processes to accumulate over long indexing runs.

## Why This Works

Zombies are created the moment a child process exits. Without `waitFor()`, the parent never reaps the child, so zombies persist for the entire lifetime of the indexer (potentially 20+ hours).

With `waitFor()`, the zombie is reaped immediately after the process exits (typically within milliseconds of SIGKILL).

The 10-second timeout is a safety net for rare edge cases where SIGKILL can't complete immediately (e.g., process stuck in uninterruptible kernel I/O).

Fixes #4920

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/Ctags.java
@@ -142,7 +142,19 @@ public class Ctags implements Resettable {
         if (ctagsProcess != null) {
             closing = true;
             LOGGER.log(Level.FINE, "Destroying ctags command");
-            ctagsProcess.destroyForcibly();
+            destroyProcess();
+        }
+    }
+
+    private void destroyProcess() {
+        ctagsProcess.destroyForcibly();
+        try {
+            if (!ctagsProcess.waitFor(10, TimeUnit.SECONDS)) {
+                LOGGER.log(Level.WARNING, "ctags process did not terminate within timeout");
+            }
+        } catch (InterruptedException e) {
+            LOGGER.log(Level.WARNING, "Interrupted while waiting for ctags process to terminate");
+            Thread.currentThread().interrupt();
         }
     }
 
@@ -603,10 +615,10 @@ public class Ctags implements Resettable {
                         }
                     } catch (IllegalThreadStateException e) {
                         LOGGER.log(Level.WARNING, "ctags EOF but did not exit");
-                        ctagsProcess.destroyForcibly();
+                        destroyProcess();
                     } catch (Exception e) {
                         LOGGER.log(Level.WARNING, "ctags problem:", e);
-                        ctagsProcess.destroyForcibly();
+                        destroyProcess();
                     }
                     // Throw the following to indicate non-I/O error for retry.
                     throw new InterruptedException("tagLine == null");


gaborbernat:
Would you be able to cut a release for this?

vladak:
> Would you be able to cut a release for this?

Certainly, done.

vladak:
For the record, I checked the `Executor` class whether it has the same problem - upon timeout it calls `destroy()` on the process in a `Timer` object, however always calls `waitFor()`: https://github.com/oracle/opengrok/blob/14fa06967e607ba9a7ba2ed0c24957ccc93f48d2/opengrok-indexer/src/main/java/org/opengrok/indexer/util/Executor.java#L248-L255

PR #4933, opened 2026-04-05
Add support for Java record definitions in JavaXref

### LINKED ISSUE #4930: OpenGrok records definition
Hello team,
 
We have an OpenGrok deployment where Java records are now widely used, and we need proper support for indexing and navigating record definitions in the UI. Currently, we could not find a way to do so in OpenGrok. 
Is there currently a way, or any plan, to support proper record definitions in OpenGrok?

Please let us know if you need any clarification or if you recommend an alternative approach to handle this use case.
 
Thank you for your efforts.
Regards,
Thalia

NA-V10:
What changes were proposed in this pull request?

This change adds handling for Java record declarations in JavaXref.lex.

Previously, record declarations were treated as normal identifiers, so the record name was not highlighted or indexed similarly to classes, interfaces, and enums.

This update adds a dedicated lexer rule for:

record User(String name, int age) {}

so that User is recognized as a symbol and can participate in navigation and definition lookup.

Why are these changes needed?

Java records are now commonly used in modern Java codebases, but OpenGrok did not recognize record declarations in the same way as other type definitions.

Without this change, record names were not indexed correctly, making navigation less useful for projects using Java records.

How was this patch tested?
Built the opengrok-indexer module successfully with Java 21
Verified that the updated lexer compiles correctly
Added a sample Java record declaration locally to confirm the lexer rule matches record definitions

Example used during testing:

record User(String name, int age) {}
Related Issue

Fixes #4930

### DIFF
--- a/opengrok-indexer/src/main/jflex/analysis/java/JavaXref.lex
+++ b/opengrok-indexer/src/main/jflex/analysis/java/JavaXref.lex
@@ -90,6 +90,18 @@ ParamName = {Identifier} | "<" {Identifier} ">"
  \}     { chkLOC(); onScopeChanged(ScopeAction.DEC, yytext(), yychar); }
  \;     { chkLOC(); onScopeChanged(ScopeAction.END, yytext(), yychar); }
 
+"record" {WhspChar}+ {Identifier} {
+    chkLOC();
+
+    String text = yytext();
+    int idx = text.lastIndexOf(' ');
+    String keyword = text.substring(0, idx);
+    String recordName = text.substring(idx + 1);
+
+    onNonSymbolMatched(keyword + " ", yychar);
+    onFilteredSymbolMatched(recordName, yychar + idx + 1, Consts.kwd);
+}
+
 {Identifier} {
     chkLOC();
     String id = yytext();


NA-V10:
Hello,

I have opened this PR to add support for Java `record` declarations in `JavaXref.lex`.

The change introduces a dedicated lexer rule so record names are recognized similarly to classes, interfaces, and enums.

I verified that the `opengrok-indexer` module builds successfully with Java 21 after the change, and I also tested locally with a simple example such as:

```java id="nh8v5a"
record User(String name, int age) {}
```

Please let me know if you would like me to also add a dedicated test case for record declarations.

vladak (changes requested):
I'd like to see some tests exercising the change. Also, it is suprising that the `record` keyword is missing in the pertaining `Consts.java` file.

vladak:
Also see https://github.com/oracle/opengrok/issues/4930#issuecomment-4572783673

vladak:
For the record, Universtal ctags now support the `record` keyword and consider it as a `class`.

vladak:
Superseded by PR #4977

Issue #3239, opened 2020-09-23
Search API returns different results from Web UI


tony-zhao:
**Describe the bug**
The REST API (api/v1/search) returns different results from Web UI for the same query condition.

Environments:
* OpenGrok 1.3.2
* OpenJDK 1.8.0_222
* OS: Amazon Linux on EC2 (4.14.146-93.123.amzn1.x86_64)
* Apache Tomcat/8.5.42

**To Reproduce**
Searching from GUI, gets "Searched +full:google +refs:google (Results 25801 – 25802 of 25802) sorted by relevance"
But searching from REST API gets
```
curl https://<grok-server>/api/v1/search?full=google&defs=&refs=google&path=&hist=&type=&searchall=true&start=0&maxresults=1 | python -m json.tool
{
    "time": 1170,
    "resultCount": 38082,
    "startDocument": 0,
    "endDocument": 0,
    "results": {
...
```
**Expected behavior**
Web UI and API should return same results for the same search condition.

vladak:
How do you run the indexer ? Do you have projects enabled ?

tony-zhao:
Both web UI and API went to the same OpenGrok instance, and using the same account.
All projects were included in the search.
So, this issue should have nothing to do with indexer.

vladak:
There is #3170, that's why I am asking about projects and indexer.

tony-zhao:
2020-09-25 08:38:15.698+0000 INFO t1 Indexer.parseOptions: Indexer options: [
-v, --displayRepositories, off, --optimize, on, -r, uionly, -H, -S, --depth, 99, --progress, -c, /usr/bin/ctags, -o, /var/opengrok/conf/ctags/config, -m, 256, --leadingWildCards, on, -R, configuration.ro.xml, -W, configuration.xml, -P, -U, http://localhost:9080/vanilla_android, -s, /var/opengrok/stage1/src, -d, /var/opengrok/stage1/data
]

tony-zhao:
I got more results from API than web UI.

vladak:
[...]

PR #4935, opened 2026-04-10
🐛 fix(search-api): stable result order and configurable per-file hit limit

gaborbernat:
Two bugs in /api/v1/search produce results that disagree with the HTML search page for identical queries, making the API unsuitable as a drop-in replacement for clients migrating off HTML scraping.

## Search results change between calls with identical parameters

Running the same search twice with maxresults=20 can return different files each time, even when the index has not changed. A pipeline paging through results risks missing files or processing duplicates. This happens because the API returns files in an arbitrary order that varies between JVM runs, so trimming at maxresults picks a different subset on each call. After this fix, files always appear in the order determined by the sort parameter, making repeated calls with the same parameters return the same results.

## Searches return fewer matching lines per file than expected

When a symbol is referenced many times in a file, the API returns only some of those references. A symbol used 25 times in a file shows up as 10 hits via the API but all 25 via the HTML page, with no indication in the response that lines were dropped. This happens because an internal server-side limit kicks in whenever a query matches more than 100 documents in total, silently capping the per-file hit count. After this fix, the default behavior returns all matching lines per file, consistent with the HTML page.

**Performance implications.** Removing the per-file cap can significantly increase response size for broad queries over large codebases — a query matching thousands of files, each with dozens of hits, will now return all of them. The new maxhitsperfile parameter (default 0 = unlimited) lets callers trade completeness for bounded response size: maxhitsperfile=10 caps each file at 10 hits regardless of total match count. The existing maxresults bounds the number of files returned; maxhitsperfile bounds the lines per file. Using both gives full control over response size with predictable upper bounds.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -647,6 +647,8 @@ The repository path is relative to source root.
 
 ## return search results [GET]
 
+The `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
+
 + Parameters
   + full (optional, string) - full search field value to search for
   + def (optional, string) - definition field value to search for
@@ -667,8 +669,6 @@ The repository path is relative to source root.
          - `relevancy`: Sort by Lucene score (most relevant first).
          - `fullpath`: Sort by file path (alphabetical).
          - `lastmodtime`: Sort by last modification date (newest first).
-
-The `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
 
 + Response 200 (application/json)
   + Body
--- a/apiary.apib
+++ b/apiary.apib
@@ -657,7 +657,6 @@ The repository path is relative to source root.
   + projects (optional, string) - projects to search in
   + maxresults (optional, string) - maximum number of documents whose hits will be returned (default 1000)
   + start (optional, string) - start index from which to return results
-  + Note: the `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
   + maxhitsperfile (optional, number) - maximum number of matching lines returned per file. 0 (default) returns all matching lines. Set to a positive integer to bound response size when a symbol appears many times per file.
   + sort: relevancy (optional, enum[string])
      + Enum
@@ -668,6 +667,8 @@ The repository path is relative to source root.
          - `relevancy`: Sort by Lucene score (most relevant first).
          - `fullpath`: Sort by file path (alphabetical).
          - `lastmodtime`: Sort by last modification date (newest first).
+
+The `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
 
 + Response 200 (application/json)
   + Body
--- a/apiary.apib
+++ b/apiary.apib
@@ -643,7 +643,7 @@ The repository path is relative to source root.
   + repository - repository path with native path separators (of the machine
   running the service) starting with path separator for which to return type
 
-## Search [/search{?full,def,symbol,path,hist,type,projects,maxresults,start,sort}]
+## Search [/search{?full,def,symbol,path,hist,type,projects,maxresults,start,sort,maxhitsperfile}]
 
 ## return search results [GET]
 
@@ -657,6 +657,8 @@ The repository path is relative to source root.
   + projects (optional, string) - projects to search in
   + maxresults (optional, string) - maximum number of documents whose hits will be returned (default 1000)
   + start (optional, string) - start index from which to return results
+  + Note: the `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
+  + maxhitsperfile (optional, number) - maximum number of matching lines returned per file. 0 (default) returns all matching lines. Set to a positive integer to bound response size when a symbol appears many times per file.
   + sort: relevancy (optional, enum[string])
      + Enum
          + relevancy
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -137,6 +137,11 @@ public class SearchEngine {
     private final char[] content = new char[1024 * 8];
     private String source;
     private String data;
+    /**
+     * Holds value of property maxHitsPerFile.
+     * 0 means unlimited (default).
+     */
+    private int maxHitsPerFile;
     int hitsPerPage = RuntimeEnvironment.getInstance().getHitsPerPage();
     int cachePages = RuntimeEnvironment.getInstance().getCachePages();
     int totalHits = 0;
@@ -502,11 +507,15 @@ public class SearchEngine {
                     try {
                         if (AbstractAnalyzer.Genre.PLAIN == genre && (source != null)) {
                             // SRCROOT is read with UTF-8 as a default.
+                            int hitsBefore = ret.size();
                             hasContext = sourceContext.getContext(
                                 new InputStreamReader(new FileInputStream(
                                 source + filename), StandardCharsets.UTF_8),
-                                null, null, null, filename, tags, nhits > 100,
+                                null, null, null, filename, tags, false,
                                 getDefinition() != null, ret, scopes);
+                            if (maxHitsPerFile > 0 && ret.size() - hitsBefore > maxHitsPerFile) {
+                                ret.subList(hitsBefore + maxHitsPerFile, ret.size()).clear();
+                            }
                         } else if (AbstractAnalyzer.Genre.XREFABLE == genre && data != null && summarizer != null) {
                             int l;
                             /*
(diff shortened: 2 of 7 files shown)


gaborbernat:
@vladak this should be ready for review now, thanks!

apiary.apib
```diff
@@ -657,6 +657,8 @@ The repository path is relative to source root.
   + projects (optional, string) - projects to search in
   + maxresults (optional, string) - maximum number of documents whose hits will be returned (default 1000)
   + start (optional, string) - start index from which to return results
+  + Note: the `results` map preserves the order of the `sort` parameter (Lucene scoring order for `relevancy`).
```
vladak, line 660: Apiary parsing will consider the `Note:` line as a parameter, leading to validation error.

vladak:
Is this possibly related to #3239 ?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -86,4 +97,72 @@ void testSearchCors() {
...
+    }
+
+    @Test
+    void testSearchResultsHaveLineNumbers() {
+        // maxhitsperfile defaults to 0 (unlimited) so all per-file hits carry a lineNumber.
+        // Restrict to Java source type to exclude XREFABLE binary files (jar, class, elf)
+        // whose hits go through the Summarizer path and never carry a lineNumber.
+        GenericType<Map<String, Object>> type = new GenericType<>() { };
+        Response response = target(SearchController.PATH)
+                .queryParam("full", "dump")
```
vladak, line 151:
```suggestion
                .queryParam(QueryParameters.FULL_SEARCH_PARAM, "dump")
```

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -86,4 +97,72 @@ void testSearchCors() {
...
+        GenericType<Map<String, Object>> type = new GenericType<>() { };
+        String firstKey1 = getFirstResultKey(type);
+        String firstKey2 = getFirstResultKey(type);
+        assertNotNull(firstKey1);
+        assertEquals(firstKey1, firstKey2);
+    }
+
+    private String getFirstResultKey(GenericType<Map<String, Object>> type) {
+        Response response = target(SearchController.PATH)
+                .queryParam("full", "dump")
```
vladak, line 132:
```suggestion
                .queryParam(QueryParameters.FULL_SEARCH_PARAM, "dump")
```

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -86,4 +97,72 @@ void testSearchCors() {
...
+
+    @Test
+    void testSearchResultsHaveLineNumbers() {
+        // maxhitsperfile defaults to 0 (unlimited) so all per-file hits carry a lineNumber.
+        // Restrict to Java source type to exclude XREFABLE binary files (jar, class, elf)
+        // whose hits go through the Summarizer path and never carry a lineNumber.
+        GenericType<Map<String, Object>> type = new GenericType<>() { };
+        Response response = target(SearchController.PATH)
+                .queryParam("full", "dump")
+                .queryParam("type", "java")
```
vladak, line 152:
```suggestion
                .queryParam(QueryParameters.TYPE_SEARCH_PARAM, "java")
```

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -225,6 +226,61 @@ void testDefaultSortOrder() {
...
+        SearchEngine limited = new SearchEngine();
+        limited.setFile("main.c");
+        limited.setFreetext("arguments");
+        limited.setMaxHitsPerFile(1);
+        hitsCount = limited.search();
+        List<Hit> cappedHits = new ArrayList<>();
+        limited.results(0, hitsCount, cappedHits);
+        limited.destroy();
+
+        assertTrue(allHits.size() >= cappedHits.size());
```
vladak, line 278: should this use strictly greater than ?

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -225,6 +226,61 @@ void testDefaultSortOrder() {
...
+        Map<String, Long> hitsPerFile = cappedHits.stream()
+                .collect(java.util.stream.Collectors.groupingBy(Hit::getPath, java.util.stream.Collectors.counting()));
+        hitsPerFile.values().forEach(count -> assertTrue(count <= 1));
```
vladak, line 281: The value of 1 corresponds to the setting on L272 above I believe, so should probably be a variable.

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -63,14 +70,18 @@ protected TestContainerFactory getTestContainerFactory() throws TestContainerExc
     public static void setUpClass() throws Exception {
         System.setProperty("sun.net.http.allowRestrictedHeaders", "true"); // necessary to test CORS from controllers
         repository = new TestRepository();
-
         repository.create(SearchControllerTest.class.getClassLoader().getResource("sources"));
```
vladak, line 73: While you are there, could you fix the smell warning by checking the URL object is not null like done in other tests ?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -86,4 +97,72 @@ void testSearchCors() {
...
+        // "dump" is a method name defined in java/Main.java — verifies the index is live
+        // and the API returns a well-formed response.
+        GenericType<Map<String, Object>> type = new GenericType<>() { };
+        Response response = target(SearchController.PATH)
+                .queryParam("full", "dump")
+                .request()
+                .get();
+        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+        Map<String, Object> body = response.readEntity(type);
+        assertNotNull(body.get("results"));
```
vladak, line 112: make `body.get("results")` a variable ?

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -86,4 +97,72 @@ void testSearchCors() {
...
+        @SuppressWarnings("unchecked")
+        Map<String, List<Map<String, Object>>> results =
+                (Map<String, List<Map<String, Object>>>) body.get("results");
+        assertFalse(results.isEmpty());
+    }
+
+    @Test
+    void testSearchResultsPreserveHitOrder() {
+        // Results grouped by LinkedHashMap must return the same leading file on repeated calls,
```
vladak, line 121: I'd avoid mentioning the specific implementation and say something like this: "Results presented by the search API must be stable w.r.t. file ordering across repeated calls, while preserving Lucene scoring order."

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -225,6 +226,61 @@ void testDefaultSortOrder() {
...
+    void testMaxHitsPerFileSetterGetter() {
+        SearchEngine instance = new SearchEngine();
+        instance.setMaxHitsPerFile(20);
+        assertEquals(20, instance.getMaxHitsPerFile());
+        instance.setMaxHitsPerFile(0);
+        assertEquals(0, instance.getMaxHitsPerFile());
+    }
+
+    @Test
+    void testMaxHitsPerFileUnlimitedReturnsAllHits() {
```
vladak, line 245: Not sure the method name expresses what is being tested in the method. It appears to be saying that setting unlimited hits per file should lead to all hits to be returned, however that is not verified here - it merely checks that all hits contain line numbers.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -502,11 +507,15 @@ public void results(int start, int end, List<Hit> ret) {
...
+                            int hitsBefore = ret.size();
                             hasContext = sourceContext.getContext(
                                 new InputStreamReader(new FileInputStream(
                                 source + filename), StandardCharsets.UTF_8),
-                                null, null, null, filename, tags, nhits > 100,
+                                null, null, null, filename, tags, false,
                                 getDefinition() != null, ret, scopes);
+                            if (maxHitsPerFile > 0 && ret.size() - hitsBefore > maxHitsPerFile) {
+                                ret.subList(hitsBefore + maxHitsPerFile, ret.size()).clear();
+                            }
```
vladak: I feel that rather than trimming the hit list after it grows too big, this should be pushed down to  `getContext()` to perform the limiting there, thus avoiding the growth in the first place. It is context limit after all.

vladak (changes requested):

vladak (approved):
looks good now

vladak:
Thanks !

PR #4936, opened 2026-04-17
check Github actions with Macaron

vladak:
Due to recent attacks targeting GitHub Actions and CI/CD pipelines, in the context of supply chain attacks, Oracle mandates its Github projects to employ Github actions check performed by the `oracle/macaron` Github action (sic).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,22 +19,22 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
     steps:
     - name: Checkout master branch
-      uses: actions/checkout@v6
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -9,6 +9,9 @@ on:
   schedule:
   - cron: "0 0 * * 0"
 
+permissions:
+  contents: write
```
behnazh-w: Since this workflow runs on `pull_request` events, checking which step requires the `write` requires the `write` permission.

vladak: I just added this mostly because of the `upload-artifact` action. Checking once again, it does not seem to require it as it does not push new commit, create release or update PRs.

PR #4937, opened 2026-04-19
🐛 fix(search): align API result count with web UI

### LINKED ISSUE #3170: UI and API /search return vastly different result sizes
**Describe the bug**
It does not seem like the API and the UI are returning the same results. Maybe I'm missing something, but when I search for the exact same thing from the UI and the API I get 125 results from the API and 9,708 from the UI. Something seems not right.

**To Reproduce**
Try `/search?...` and `/api/v1/search?...` with all the same query parameters. In my test only `full=` was set.

**Expected behavior**
I expect the UI and the REST API to return the same results

**Screenshots**

- UI in Browser

_URL: `http://local-opengrok/search?full=test&defs=&refs=&path=&type=`_

image

- API in terminal

_URL: `http://local-opengrok/api/v1/search?full=test&defs=&refs=&path=&type=`_

```
curl -s 'http://local-opengrok/api/v1/search?full=test&defs=&refs=&path=&type=' | jq .resultCount
125
```

```
curl -s 'http://local-opengrok/api/v1/search?full=test&defs=&refs=&path=&type=' | jq 
{
  "time": 167,
  "resultCount": 125,
  "startDocument": 0,
  "endDocument": 124,
  "results": {
...
```

### LINKED ISSUE #3239: Search API returns different results from Web UI

gaborbernat:
The REST API search endpoint (`/api/v1/search`) reports different result counts than the web UI for identical queries. In project-less mode the API always caps `resultCount` at `hitsPerPage * cachePages` (defaulting to 125) because `SearchEngine.search()` returns the collected `hits.length` rather than Lucene's `totalHits`. 🔍 This is the root cause of both #3239 and #3170.

A second source of inaccuracy comes from using `Short.MAX_VALUE` (32,767) as the Lucene `totalHitsThreshold` in all collector constructions. For queries exceeding that threshold, `totalHits.value` becomes an approximate lower bound. The web UI already uses `Integer.MAX_VALUE` implicitly through the 3-arg `searcher.search(query, n, sort)` overload, so switching the API collectors to match eliminates the discrepancy with zero performance impact for the common case and marginal cost for very large result sets.

The `endDocument` field in the JSON response was also wrong, computed from the grouped result map's key count (unique file paths) instead of the actual document page size. A page of 10 documents spanning 5 files would report `endDocument` as `startDocIndex + 4` instead of `startDocIndex + 9`. ✅ Now derived from the same page-size arithmetic used internally by `SearchEngineWrapper.search()`.

Fixes #3239. Fixes #3170.

### DIFF
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
@@ -305,6 +305,32 @@ class SearchEngineTest {
         instance.destroy();
     }
 
+    /**
+     * Verify that totalHits is exact (not an approximate lower bound) so that results()
+     * can fetch every matching document. With an inaccurate totalHitsThreshold, the
+     * re-query in results() would request too few documents, silently dropping results.
+     */
+    @Test
+    void testTotalHitsIsExactForFullRetrieval() {
+        SearchEngine instance = new SearchEngine();
+        instance.setFile("main.c");
+        instance.setFreetext("arguments");
+        instance.hitsPerPage = 1;
+        instance.cachePages = 1;
+
+        int count = instance.search();
+        assertTrue(count > 1);
+
+        List<Hit> allHits = new ArrayList<>();
+        instance.results(0, count, allHits);
+
+        long uniquePaths = allHits.stream().map(Hit::getPath).distinct().count();
+        assertEquals(count, uniquePaths,
+                "totalHits must be exact so results() retrieves every matching document");
+
+        instance.destroy();
+    }
+
     /* see https://github.com/oracle/opengrok/issues/2030
     @Test
     void testSearch() {
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
@@ -282,6 +282,29 @@ class SearchEngineTest {
         hitsPerFile.values().forEach(count -> assertTrue(count <= maxPerFile));
     }
 
+    /**
+     * Verify that search() returns the total number of matching documents (totalHits),
+     * not the number of collected/cached hits (which is capped at hitsPerPage * cachePages).
+     */
+    @Test
+    void testSearchReturnsTotalHitsNotCachedCount() {
+        SearchEngine instance = new SearchEngine();
+        instance.setFile("main.c");
+        instance.setFreetext("arguments");
+
+        // Restrict the cache to 1 document so that totalHits > hitsPerPage * cachePages.
+        instance.hitsPerPage = 1;
+        instance.cachePages = 1;
+
+        int resultCount = instance.search();
+        assertTrue(resultCount > 1,
+                "Query should match multiple documents across test repositories");
+        assertEquals(instance.totalHits, resultCount,
+                "search() must return totalHits, not the number of cached hits");
+
+        instance.destroy();
+    }
+
     /* see https://github.com/oracle/opengrok/issues/2030
     @Test
     void testSearch() {
--- a/opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
+++ b/opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
@@ -1,0 +1,144 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * See LICENSE.txt included in this distribution for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at LICENSE.txt.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ */
+package org.opengrok.indexer.web;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.opengrok.indexer.configuration.RuntimeEnvironment;
+import org.opengrok.indexer.index.Indexer;
+import org.opengrok.indexer.search.QueryBuilder;
+import org.opengrok.indexer.util.TestRepository;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Verifies that {@link SearchHelper} produces an exact and stable {@code totalHits}
+ * regardless of the {@code maxItems} window, so the UI search total matches the REST
+ * API and does not vary across repeated searches. With the previous implementation
+ * {@code searcher.search(query, n, sort)} uses a {@code totalHitsThreshold} of 1000,
+ * so for queries matching more than 1000 documents {@code totalHits} becomes an
+ * approximate lower bound and can drift between calls.
+ *
+ * <p>To reliably exercise that threshold the test creates a synthetic corpus of
+ * {@value #SYNTHETIC_DOC_COUNT} files, each containing a unique marker token.</p>
+ */
+class SearchHelperStableCountTest {
+
+    private static final String MARKER = "searchhelperbug3239marker";
+    private static final int SYNTHETIC_DOC_COUNT = 1500;
+    private static final String SYNTHETIC_PROJECT = "synthetic_bulk";
+
+    private static TestRepository repository;
+    private static RuntimeEnvironment env;
+
+    @BeforeAll
+    static void setUpClass() throws Exception {
+        repository = new TestRepository();
+        repository.create(SearchHelperStableCountTest.class.getClassLoader().getResource("sources"));
+
+        Path syntheticDir = Path.of(repository.getSourceRoot(), SYNTHETIC_PROJECT);
+        Files.createDirectories(syntheticDir);
+        // Produce strong score variance so Lucene's block-max WAND can skip low-scoring
+        // docs once the heap is full. With the pre-fix implementation that lets the
+        // totalHitsThreshold of 1000 turn the totalHits into a lower-bound estimate.
+        for (int i = 0; i < SYNTHETIC_DOC_COUNT; i++) {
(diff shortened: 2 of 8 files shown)


gaborbernat:
@vladak is now ready

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -225,6 +225,57 @@ void testDefaultSortOrder() {
         assertNull(instance.getSortOrder(), "Default sort should be relevancy (null implies Lucene score ordering)");
     }
 
+    /**
+     * Verify that search() returns the total number of matching documents (totalHits),
+     * not the number of collected/cached hits (which is capped at hitsPerPage * cachePages).
+     * This is the fix for <a href="https://github.com/oracle/opengrok/issues/3239">#3239</a>
+     * and <a href="https://github.com/oracle/opengrok/issues/3170">#3170</a>.
```
vladak: Avoid mentioning issue numbers/links in the comments, please.

vladak:
Needs rebase.

gaborbernat:
@vladak done, thanks!

opengrok-web/src/test/java/org/opengrok/web/api/v1/controller/SearchControllerTest.java
```diff
@@ -170,4 +171,36 @@ void testSearchResultsHaveLineNumbers() {
...
+     * file paths in the grouped results map.
+     */
+    @Test
+    @SuppressWarnings("unchecked")
+    void testResultCountAndEndDocument() {
+        int maxResults = 100;
+        GenericType<Map<String, Object>> type = new GenericType<>() { };
+        Response response = target(SearchController.PATH)
+                .queryParam(QueryParameters.FULL_SEARCH_PARAM, "main")
+                .queryParam("maxresults", maxResults)
```
vladak: time for `QueryParameters.MAXRESULTS_PARAM`, I'd say

vladak:
The webapp is still returning different total hit count on each search when performed via UI, so it is hard to evaluate whether API search is equivalent to UI search. For example with my favorite AOSP test (indexer run with `-s /var/opengrok/src -d /var/opengrok/data -P --economical -c /opt/homebrew/bin/ctags -W /var/opengrok/etc/configuration.xml -U http://localhost:8080/opengrok_web_war`), searching for "google" yields between 20k and 45k hits when done via UI and with your changes always the `resultCount` in the API search http://localhost:8080/opengrok_web_war/api/v1/search?project=AOSP&full=google&searchall=true&start=0&maxresults=1 results is always 200531 in my case.

vladak:
The reason for the non-stable hit count in the UI is in how search totals are produced in web search (thanks Codex): `SearchHelper.executeQuery()` uses `searcher.search(query, start + maxItems, sort)` and then prints `TopFieldDocs.totalHits.value`. In Lucene this value can be an estimated/lower-bound total unless total-hit tracking is explicitly forced, so the displayed “of N” can drift between requests (especially with relevance sorting). 

There is a way how to make the total hit count stable for the UI similarly to how this is proposed for the API:
```patch
diff --git a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
index f8177ef2d..ee9a7bd18 100644
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java
@@ -61,7 +61,9 @@ import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TopFieldCollectorManager;
 import org.apache.lucene.search.TopFieldDocs;
+import org.apache.lucene.search.TopScoreDocCollectorManager;
 import org.apache.lucene.search.spell.DirectSpellChecker;
 import org.apache.lucene.search.spell.SuggestMode;
 import org.apache.lucene.search.spell.SuggestWord;
@@ -475,9 +477,17 @@ public class SearchHelper {
             return this;
         }
         try {
-            TopFieldDocs fdocs = searcher.search(query, start + maxItems, sort);
-            totalHits = fdocs.totalHits.value;
-            hits = fdocs.scoreDocs;
+            int resultWindow = start + maxItems;
+            TopDocs topDocs;
+            if (Sort.RELEVANCE.equals(sort)) {
+                topDocs = searcher.search(query,
+                        new TopScoreDocCollectorManager(resultWindow, Integer.MAX_VALUE));
+            } else {
+                topDocs = searcher.search(query,
+                        new TopFieldCollectorManager(sort, resultWindow, Integer.MAX_VALUE));
+            }
+            totalHits = topDocs.totalHits.value;
+            hits = topDocs.scoreDocs;
 
             /*
              * Determine if possibly a single-result redirect to xref is
```
For my test case that leads to stable 200483 hits in the UI, versus 200531 via the API which is pretty close.

To fix #3239 this patch would be needed.

The cons being:
  - Higher query cost on large result sets because Lucene must track total hits exactly.
  - Potentially increased latency and CPU usage for broad queries (especially very common terms).

Personally I find the varying result count in the UI confusing.

gaborbernat:
Yeah, I agree, I'd consider it a bug that now is unreliable, it communicates the wrong information to the end user.

vladak:
Also, benchmarking the UI search latency before and after the change, the original was actually slighly slower (like 15%) for some reason.

gaborbernat:
@vladak pushed fixes, we should be good now.

gaborbernat:
Also, when we merge this can we get a release cut? Thanks!

vladak:
> Also, when we merge this can we get a release cut? Thanks!

That's my plan. Will take another look at the changes.

opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
```diff
@@ -0,0 +1,144 @@
...
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ */
```
vladak: Technically this is not the correct thing to have here since the code was not authored within the Oracle's realm. Removing the copyright comment altogether would make the style check fail, though.

gaborbernat: So what would you prefer?

vladak: Either add some different copyright or remove the copyright altogether and add the file to the exceptions list https://github.com/oracle/opengrok/blob/master/dev/checkstyle/suppressions.xml

gaborbernat: Done - removed the copyright header and added the file to the Header suppressions in `dev/checkstyle/suppressions.xml`.

gaborbernat: @vladak friendly ping on this, thanks!

opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Verifies that {@link SearchHelper} produces an exact and stable {@code totalHits}
+ * regardless of the {@code maxItems} window, so the UI search total matches the REST
+ * API and does not vary across repeated searches. With the previous implementation
+ * {@code searcher.search(query, n, sort)} uses a {@code totalHitsThreshold} of 1000,
```
vladak: The `totalHitsThreshold` I believe refers to the last argument of the constructor of `TopScoreDocCollectorManager` / `TopFieldCollectorManager`, however it is not clear here.

opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    void testTotalHitsIsExactAndStable() {
+        SortedSet<String> projectNames = new TreeSet<>();
+        projectNames.add(SYNTHETIC_PROJECT);
+
+        // Run with a large maxItems first to capture the true matching document count;
+        // the pre-fix implementation returned an exact count only when the heap was
+        // large enough to disable block-max WAND early termination.
```
vladak: ditto

opengrok-indexer/src/test/java/org/opengrok/indexer/web/SearchHelperStableCountTest.java
```diff
@@ -0,0 +1,122 @@
...
+    @BeforeAll
+    static void setUpClass() throws Exception {
+        repository = new TestRepository();
+        repository.create(SearchHelperStableCountTest.class.getClassLoader().getResource("sources"));
+
+        Path syntheticDir = Path.of(repository.getSourceRoot(), SYNTHETIC_PROJECT);
+        Files.createDirectories(syntheticDir);
+        // Produce strong score variance so Lucene's block-max WAND can skip low-scoring
+        // docs once the heap is full. With the pre-fix implementation that lets the
+        // totalHitsThreshold of 1000 turn the totalHits into a lower-bound estimate.
```
vladak: I feel this is a bit too much detail for the test and arguably it would be also for the relevant part of the indexer. It would be sufficient to say that strong score variance is produced to achieve certain behavior, without going into the Lucene implementation details (which could change over time).

gaborbernat:
I have addressed the change requests.

gaborbernat:
@vladak any update here? Thanks!

vladak:
> @vladak any update here? Thanks!

Still looking. Finally found the time to go through the change one last time. In particular I need to evaluate one finding by Codex to determine how it is relevant to the changes here. Also, I don't like how `SearchEngine` is structured, it definitely needs some work but that will happen separately.

vladak (approved):
There is an issue if resuls() re-runs the search and there are already some documents stored in the `docs` list, where the results of the full re-query are appended, resulting in duplicate documents stored. I will fix this myself with a follow-up PR that will contain cleanup of the SearchEngine code.

PR #4967, opened 2026-05-29
🐛 fix(search-api): bound Lucene collection to prevent OOM

gaborbernat:
The REST search API uses `SearchEngine.search()`, which for multi-project queries always disables paging and [re-runs the Lucene query collecting every matching document](https://github.com/oracle/opengrok/blob/61c43707bf8/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java#L252-L260) before paginating. On large indexes, a broad full-text search triggers this unbounded re-collection path and exhausts the JVM heap.

The HTML search page (`SearchHelper`) does not have this problem because it [calls `searcher.search(query, numHits, sort)` directly](https://github.com/oracle/opengrok/blob/61c43707bf8/opengrok-indexer/src/main/java/org/opengrok/indexer/web/SearchHelper.java#L483-L488), keeping heap usage proportional to the requested page size rather than the total result count.

The fix introduces `SearchEngine.setMaxDocs(int)` to cap collection at the caller-supplied bound. When set, `searchIndex()` uses `maxDocs` as the Lucene collector limit and skips the re-collection block entirely — heap usage stays O(maxDocs) regardless of how many documents match the query. 🔒 `SearchController` sets `maxDocs = startDocIndex + maxResults` before calling `engine.search()`, so collection is always bounded to exactly what the response needs.

`getTotalHits()` is added to preserve the accurate total count in the API response — previously `numResults` was set from `engine.search()` which returns `hits.length` (now capped), losing the true total.

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -139,13 +139,8 @@ public class SearchEngine {
 
     // internal structures to hold the results from Lucene
     private final List<Document> docs;
-    /**
-     * Hard cap on the number of Lucene documents collected during search.
-     * When positive, overrides {@code hitsPerPage * cachePages} and disables
-     * the unbounded re-collection path, keeping heap usage proportional to
-     * this value rather than to the total number of matching documents.
-     * 0 means no cap (default, legacy behavior).
-     */
+    // Caps Lucene collection to prevent unbounded heap use on large result sets.
+    // 0 = no cap (default, preserves legacy behaviour).
     private int maxDocs;
     int totalHits = 0;
     private ScoreDoc[] hits;
@@ -656,20 +651,18 @@ public class SearchEngine {
         this.maxHitsPerFile = maxHitsPerFile;
     }
 
-    /**
-     * Set a hard cap on the number of Lucene documents collected.
-     * When positive, the search collects at most this many documents
-     * and never re-queries for the full result set.
-     *
-     * @param maxDocs maximum documents to collect (0 = unlimited)
-     */
     public void setMaxDocs(int maxDocs) {
         this.maxDocs = maxDocs;
     }
 
+    public int getMaxDocs() {
+        return maxDocs;
+    }
+
     /**
      * Returns the total number of Lucene documents matching the query,
-     * regardless of the {@code maxDocs} cap. Available after {@code search()}.
+     * regardless of the {@code maxDocs} cap. Callers need the true total
+     * to paginate correctly when collection is bounded.
      *
      * @return total matching document count
      */
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -139,6 +139,14 @@ public class SearchEngine {
 
     // internal structures to hold the results from Lucene
     private final List<Document> docs;
+    /**
+     * Hard cap on the number of Lucene documents collected during search.
+     * When positive, overrides {@code hitsPerPage * cachePages} and disables
+     * the unbounded re-collection path, keeping heap usage proportional to
+     * this value rather than to the total number of matching documents.
+     * 0 means no cap (default, legacy behavior).
+     */
+    private int maxDocs;
     int totalHits = 0;
     private ScoreDoc[] hits;
     boolean allCollected;
@@ -230,7 +238,7 @@ public class SearchEngine {
 
     private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
         Statistics stat = new Statistics();
-        final int numHits = hitsPerPage * cachePages;
+        final int numHits = maxDocs > 0 ? maxDocs : hitsPerPage * cachePages;
         TopDocs topDocs;
         Sort luceneSort = getSort();
         if (luceneSort == null) {
@@ -241,7 +249,7 @@ public class SearchEngine {
         hits = topDocs.scoreDocs;
         totalHits = (int) topDocs.totalHits.value;
 
-        if (!paging && totalHits > numHits) {
+        if (maxDocs <= 0 && !paging && totalHits > numHits) {
             if (luceneSort == null) {
                 topDocs = searcher.search(query, new TopScoreDocCollectorManager(totalHits, Integer.MAX_VALUE));
             } else {
@@ -253,7 +261,7 @@ public class SearchEngine {
                 "search.latency", new String[]{"category", "engine",
                         "outcome", totalHits > 0 ? "success" : "empty"});
 
-        allCollected = !paging || totalHits <= numHits;
+        allCollected = maxDocs > 0 || !paging || totalHits <= numHits;
 
         StoredFields storedFields = searcher.storedFields();
         for (ScoreDoc hit : hits) {
@@ -646,6 +654,27 @@ public class SearchEngine {
      */
     public void setMaxHitsPerFile(int maxHitsPerFile) {
         this.maxHitsPerFile = maxHitsPerFile;
+    }
+
+    /**
+     * Set a hard cap on the number of Lucene documents collected.
+     * When positive, the search collects at most this many documents
+     * and never re-queries for the full result set.
+     *
+     * @param maxDocs maximum documents to collect (0 = unlimited)
+     */
+    public void setMaxDocs(int maxDocs) {
+        this.maxDocs = maxDocs;
+    }
+
+    /**
+     * Returns the total number of Lucene documents matching the query,
+     * regardless of the {@code maxDocs} cap. Available after {@code search()}.
+     *
+     * @return total matching document count
+     */
+    public int getTotalHits() {
+        return totalHits;
     }
 
     /**
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -79,6 +79,10 @@ import org.opengrok.indexer.web.SortOrder;
  * <p>
  * Authorization is <b>not</b> enforced here, this has to be done by the caller by filtering out the projects
  * passed to {@link #search(List)}.
+ * <p>
+ * The expected lifecycle is: set the search criteria, call {@code search(...)}, retrieve the hits via
+ * {@link #results(int, int, List)} and finally call {@link #destroy()}. Calling
+ * {@link #results(int, int, List)} without a successful search throws {@link IllegalStateException}.
  *
  * @author Trond Norbye 2005
  * @author Lubos Kosco - upgrade to lucene 3.x, 4.x, 5.x
@@ -168,14 +172,6 @@ public class SearchEngine {
     }
 
     /**
-     * Creates a new instance of SearchEngine collecting at most the configured
-     * hits per page times the number of cache pages.
-     */
-    public SearchEngine() {
-        this(env.getHitsPerPage() * env.getCachePages());
-    }
-
-    /**
      * Create a QueryBuilder using the fields that have been set on this instance.
      *
      * @return a query builder
@@ -235,7 +231,7 @@ public class SearchEngine {
         Statistics stat = new Statistics();
         // The collector managers eagerly allocate a priority queue of the requested size, hence
         // the index-size cap, mirroring IndexSearcher#search(Query, int).
-        final int numHits = Math.min(maxDocs, Math.max(1, searcher.getIndexReader().maxDoc()));
+        final int numHits = Math.clamp(searcher.getIndexReader().maxDoc(), 1, maxDocs);
         TopDocs topDocs;
         Sort luceneSort = getSort();
         if (luceneSort == null) {
@@ -416,19 +412,21 @@ public class SearchEngine {
      * Get results by going through the documents from the search hits and grabbing the context therein.
      * This involves reading various document fields as well as file contents from the respective files
      * under source root.
-     * If no search was started before, no results are returned.
      * Only documents collected by {@code search(...)} are available; {@code endDocIndex} is clamped
      * to the collected count, i.e. the value {@code search(...)} returned.
      *
      * @param startDocIndex start index of the hit list
      * @param endDocIndex end index of the hit list
-     * @param ret output argument that contains list of results from start to end or null/empty if no search was started
+     * @param ret output argument that contains list of results from start to end
+     * @throws IllegalStateException if no search was performed or it did not succeed
      */
     public void results(int startDocIndex, int endDocIndex, @NotNull List<Hit> ret) {
-        ret.clear();
+        if (hits == null) {
+            throw new IllegalStateException("results(...) called without a successful search(...)");
+        }
 
-        // return if no search() was done
-        if (hits == null || (endDocIndex < startDocIndex)) {
+        ret.clear();
+        if (endDocIndex < startDocIndex) {
             return;
         }
 
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
@@ -320,7 +320,8 @@ public class SearchEngine {
      * Execute a search.
      * <p>
      * Before calling this function, you must set the appropriate search criteria with the set-functions.
-     * Note that this search will return the first {@link #cachePages} of {@link #hitsPerPage}.
+     * Note that this search will return the first {@link #cachePages} of {@link #hitsPerPage}, unless a positive
+     * {@code maxDocs} has been set via {@link #setMaxDocs(int)}, in which case collection is bounded to that many hits.
      * <p>
      * Must be eventually followed by call to {@link #destroy()} so that
      * the {@code IndexSearcher} objects are properly freed.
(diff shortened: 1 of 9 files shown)


gaborbernat:
@vladak thanks for approving the CI, if we could cut a new release once this goes in that would be much appreciated (we were trying to switch to use the API, but had to roll back due to this OOM).

vladak:
Sure, will take a look till the end of the week.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -146,23 +146,23 @@ public List<Hit> search(
...
             if (projects == null || projects.isEmpty()) {
-                numResults = engine.search(new ArrayList<>(allProjects));
+                engine.search(new ArrayList<>(allProjects));
             } else {
-                numResults = engine.search(allProjects.stream()
+                engine.search(allProjects.stream()
                         .filter(p -> projects.contains(p.getName()))
                         .collect(Collectors.toList()));
             }
+            numResults = engine.getTotalHits();
```
vladak, line 181: All the `search()` methods still return total number of hits so not sure why this is needed. This does not play well with `results()` where one might expect to pass the actually collected number of hits as the second argument.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -230,7 +233,7 @@ private void searchMultiDatabase(List<Project> projectList, boolean paging) thro
 
     private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
         Statistics stat = new Statistics();
-        final int numHits = hitsPerPage * cachePages;
+        final int numHits = maxDocs > 0 ? maxDocs : hitsPerPage * cachePages;
```
vladak: The public `search()` methods mention in their javadoc that they return the first `cachePages` of `hitsPerPage` and this change makes that doc imprecise.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -648,6 +651,25 @@ public void setMaxHitsPerFile(int maxHitsPerFile) {
         this.maxHitsPerFile = maxHitsPerFile;
     }
 
+    public void setMaxDocs(int maxDocs) {
+        this.maxDocs = maxDocs;
+    }
+
+    public int getMaxDocs() {
```
vladak, line 609: add the `@VisibleForTesting` annotation

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
gaborbernat: Good catch. Reverted to the `search()` return value and dropped `getTotalHits()`. The collected window is bounded by `maxResults` (`maxDocs = startDocIndex + maxResults`), so `results()` never reads past what was collected.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Updated both `search()` javadocs to note that a positive `maxDocs` bounds collection to that many hits.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Added.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
vladak: I think it would be actually more natural to make `search()` return the capped number of hits and have `getTotalHits()` used by the `SearchController`.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -146,6 +146,8 @@ public List<Hit> search(
                 final int startDocIndex,
                 final int maxResults
         ) {
+            engine.setMaxDocs(startDocIndex + maxResults);
```
vladak: The integer expression can overflow. It might be prudent to limit/check.

vladak: Also, it might be wise to reject any negative integer parameters to the API endpoint.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -648,6 +653,15 @@ public void setMaxHitsPerFile(int maxHitsPerFile) {
         this.maxHitsPerFile = maxHitsPerFile;
     }
 
+    public void setMaxDocs(int maxDocs) {
```
vladak: add Javadoc ?

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -241,7 +244,7 @@ private void searchIndex(IndexSearcher searcher, boolean paging) throws IOExcept
         hits = topDocs.scoreDocs;
         totalHits = (int) topDocs.totalHits.value;
 
-        if (!paging && totalHits > numHits) {
+        if (maxDocs <= 0 && !paging && totalHits > numHits) {
```
vladak: with the negative value check mentioned in the `SearchController` comments this could become `maxDocs == 0`

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
gaborbernat: Done as suggested: `search()` now returns the collected count (safe to pass to `results()`), and the controller reads the full match count from `getTotalHits()` for `resultCount`/pagination.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
gaborbernat: Now computed with long arithmetic and `Math.clamp` to `[1, Integer.MAX_VALUE]`. The collector size is additionally bounded by the reader's `maxDoc()` in `searchIndex()`, mirroring what `IndexSearcher#search(Query, int)` does internally, since the collector managers pre-allocate a priority queue of the requested size.

gaborbernat: Negative `maxresults`/`start`/`maxhitsperfile` now get 400. `setMaxDocs()` also rejects negatives with `IllegalArgumentException` so the engine holds the invariant on its own.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Added, for `getMaxDocs()` and `getTotalHits()` too.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Changed to `maxDocs == 0`, backed by the negative check in `setMaxDocs()` rather than only the controller-side validation.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -146,20 +151,24 @@ public List<Hit> search(
                 final int startDocIndex,
                 final int maxResults
         ) {
+            engine.setMaxDocs(Math.clamp((long) startDocIndex + maxResults, 1, Integer.MAX_VALUE));
```
vladak: I'd prefer Math.addExact() as it detects the overflow, rather than enforcing the overflowed value into a range.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -230,7 +233,10 @@ private void searchMultiDatabase(List<Project> projectList, boolean paging) thro
 
     private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
         Statistics stat = new Statistics();
-        final int numHits = hitsPerPage * cachePages;
+        // The collector managers eagerly allocate a priority queue of the requested size, hence
+        // the index-size cap, mirroring IndexSearcher#search(Query, int).
+        final int numHits = Math.min(maxDocs == 0 ? hitsPerPage * cachePages : maxDocs,
```
vladak: I wonder if we can simplify `SearchEngine` by getting rid of the `hitsPerPage` and `cachePages` altogether in favor of the `maxDocs` and just set the value in the `SearchEnginerWrapper`/`SearchController` using single-parameter constructor. E.g. if the maxResults API parameter is set, use `SearchEngine(maxResults)`, otherwise use `SearchEngine()` which would call `this(env.getHitsPerPage() * env.getCachePages())`.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
gaborbernat: Switched to `Math.addExact()`; an overflowing `start + maxresults` is now rejected with 400 instead of being clamped.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Done — `hitsPerPage`/`cachePages` are gone; `SearchEngine(maxDocs)` is the sole bound and `SearchEngine()` defaults to `hitsPerPage * cachePages` from the configuration. This also let the requery in `results()` and the `paging`/`allCollected` machinery go away entirely, so collection is bounded by construction for every caller.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -198,39 +207,35 @@ public boolean isValidQuery() {
...
+        searchIndex(searcher);
     }
 
-    private void searchIndex(IndexSearcher searcher, boolean paging) throws IOException {
+    private void searchIndex(IndexSearcher searcher) throws IOException {
         Statistics stat = new Statistics();
-        final int numHits = hitsPerPage * cachePages;
+        // The collector managers eagerly allocate a priority queue of the requested size, hence
+        // the index-size cap, mirroring IndexSearcher#search(Query, int).
+        final int numHits = Math.min(maxDocs, Math.max(1, searcher.getIndexReader().maxDoc()));
```
vladak: This would be actually a nice place to use `Math.clamp()`.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
```diff
@@ -421,62 +417,22 @@ public Document doc(int docId) throws IOException {
...
      *
      * @param startDocIndex start index of the hit list
      * @param endDocIndex end index of the hit list
      * @param ret output argument that contains list of results from start to end or null/empty if no search was started
      */
     public void results(int startDocIndex, int endDocIndex, @NotNull List<Hit> ret) {
+        ret.clear();
 
-        // return if no start search() was done
+        // return if no search() was done
```
vladak: I'd be probably more stringent for the no-search-done case as this side steps the API contract (+ document it in the class level javadoc) and throw an exception.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
vladak: another use case for `Math.clamp()`

vladak: I think it would make more sense to use `env.getHitsPerPage() * env.getCachePages()` as the default value for the `maxResults` param.

vladak: Also, get rid of the `SearchEngine()` parameterless constructor, it is not correct because the `SearchEngine(int maxDocs)` constructor sets the absolute number of documents rather than the window. It worked with the re-query, now it does not.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Done, `Math.clamp(maxDoc, 1, maxDocs)`.

opengrok-indexer/src/main/java/org/opengrok/indexer/search/SearchEngine.java
gaborbernat: Now throws `IllegalStateException` when called without a successful search, and the expected lifecycle (set criteria, `search(...)`, `results(...)`, `destroy()`) is documented in the class-level javadoc. `SearchController` returns an empty page early when nothing was collected, so a query matching nothing does not trip it.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
gaborbernat: Done.

gaborbernat: Done — `maxresults` now defaults to `hitsPerPage * cachePages`. Since `@DefaultValue` cannot hold a runtime expression, the parameter became a nullable `Integer` with the fallback computed in the method.

gaborbernat: Removed; `SearchEngine(int maxDocs)` is the sole constructor, so every caller states its absolute collection cap. Tests that used the default constructor as "collect everything" now pass `Integer.MAX_VALUE`, which stays safe since the collector size is clamped to the index size.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -147,19 +170,21 @@ public List<Hit> search(
...
-            if (startDocIndex > numResults) {
+            if (startDocIndex >= collected) {
                 return Collections.emptyList();
             }
 
-            int resultSize = numResults - startDocIndex;
+            int resultSize = collected - startDocIndex;
             if (resultSize > maxResults) {
                 resultSize = maxResults;
             }
```
vladak, line 190: I feel this should be rather an assert because it should not be possible to get more than `maxResults` from the `search()`, however let's leave it that way as a extra layer of protection.

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -438,21 +418,112 @@ void testSearch() {
...
+        // search() must report the collected count so that its return value
+        // is safe to pass to results() as the end index.
+        SearchEngine limited = new SearchEngine(1);
+        limited.setFreetext("arguments");
+        assertEquals(1, limited.search());
+        limited.destroy();
+    }
+
+    @Test
+    void testGetTotalHitsReportsFullCountWhenMaxDocsCaps() {
```
vladak, line 481: This method can be combined with the above method as they are nearly identical, however I am willing to live with that.

opengrok-indexer/src/test/java/org/opengrok/indexer/search/SearchEngineTest.java
```diff
@@ -438,21 +418,112 @@ void testSearch() {
...
+        SearchEngine instance = new SearchEngine(1);
+        instance.setFreetext("arguments");
+        instance.search();
+        assertTrue(instance.getTotalHits() > 1, "Total should exceed maxDocs for this test to be meaningful");
+
+        // asking past the collected window must not blow up, only serve what was collected
+        List<Hit> hits = new ArrayList<>();
+        instance.results(0, instance.getTotalHits(), hits);
+        assertEquals(1, hits.size());
+        instance.destroy();
```
vladak, line 527: The destroy call could be moved above the assert to perform cleanup before assert failure, however I am not particularly tied to that, esp. in this case.

vladak (approved):
Looks good now. I am glad this turned into the cleanup which made the code much cleaner/leaner than before.

The request can carry giant values of the `maxResults` parameter so there is still OOM potential, however this is the step in the right direction.

opengrok-web/src/main/java/org/opengrok/web/api/v1/controller/SearchController.java
```diff
@@ -82,13 +82,34 @@ public SearchResult search(
...
+            throw new WebApplicationException("Negative integer parameters are not allowed",
+                    Response.Status.BAD_REQUEST);
+        }
+
+        final int maxResults;
+        if (maxResultsParam != null) {
+            maxResults = maxResultsParam;
+        } else {
+            RuntimeEnvironment env = RuntimeEnvironment.getInstance();
+            maxResults = env.getHitsPerPage() * env.getCachePages();
```
vladak, line 100: Here we assume that the configuration is sane, i.e. the multiplication will not overflow. Probably a sane assumption.

PR #4977, opened 2026-06-16
add Java record definition

vladak:
This change adds the support for Java `record` definition. The matching offset bump in `CtagsReader` might slow ctags output parsing a bit, however given there are multiple tag classes that begin with `r` this is unavoidable.

Requires the latest Universal ctags built from source.

xref of a Lucene source file using `record` before and after:
obrazek

Assisted by Codex (5.5-xhigh)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/CtagsReader.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/CtagsReader.java
@@ -18,7 +18,7 @@
  */
 
 /*
- * Copyright (c) 2005, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2026, Oracle and/or its affiliates. All rights reserved.
  * Portions Copyright (c) 2017, 2020, Chris Fraire <cfraire@me.com>.
  */
 package org.opengrok.indexer.analysis;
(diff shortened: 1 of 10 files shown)


opengrok-indexer/src/main/java/org/opengrok/indexer/analysis/CtagsReader.java
```diff
@@ -122,12 +123,13 @@ public enum tagFields {
...
-         * "find" will be considered "file" and overwrite the value, so if
-         * ctags will send us buggy input, we will output buggy data TOO!
+         * some unknown field with the same prefix as a consumed field will
+         * overwrite the value, so if ctags will send us buggy input, we will
+         * output buggy data TOO!
          * NO VALIDATION happens of input - but then we gain LOTS of speed, due to
          * not comparing the same field names again and again fully.
          */
-        private static final int CHAR_CMP_END_OFFSET = 0;
+        private static final int CHAR_CMP_END_OFFSET = 1;
```
vladak, line 132: Comment from Codex as to why this is needed:

`CHAR_CMP_END_OFFSET` controls how many leading characters `CtagsReader.tagFields.quickValueOf()` compares when deciding whether a ctags extension field is one OpenGrok consumes.

Before the record change, the consumed fields had distinct first letters:

- `class`
- `line`
- `signature`

So `CHAR_CMP_END_OFFSET = 0` was enough because it compared only the first character.

Adding `record` introduced a risky first-letter match for any ctags field beginning with `r`. In Universal Ctags, fields such as `roles:` can also appear. With the old setting, `roles:...` would be treated as `record:...`, because both start with `r`, and could incorrectly populate the record namespace.

Bumping it to `1` makes the comparison use the first two characters:

- `record` -> `re`
- `roles` -> `ro`

That keeps `record:` from colliding with other `r...` fields while preserving the fast prefix-based parsing style already used by `CtagsReader`.

PR #4987, opened 2026-07-01
Only detect Perforce when a P4CONFIG marker is present

SebastienTolron:
## Problem

`PerforceRepository` has no local repository marker (unlike `.git`, `.hg`, `.svn`, …). As a result, `isRepositoryFor()` probes **every scanned directory** by executing `p4 dirs`/`p4 files` whenever the `p4` binary is available.

This has two bad consequences on machines that only use other SCMs:

1. **Misclassification / lost history** — when a `p4` client is configured through the environment (`P4PORT`/`P4CLIENT`/`P4CONFIG`), `p4 files`/`p4 dirs` can return depot paths for directories that are actually Git working trees. Those trees get detected as Perforce, so their real (Git) history is never collected.
2. **Slow and noisy indexing** — `p4` is spawned in every non-repository directory during the discovery walk.

## Fix

Gate Perforce detection on **local evidence** before running any `p4` command: a P4CONFIG marker file (named by the `P4CONFIG` environment variable, defaulting to `.p4config`) present in the directory or one of its ancestors. This mirrors the marker-based detection already used by the other SCM implementations. Real Perforce workspaces that use the standard P4CONFIG convention are still detected; unrelated (e.g. Git) trees are no longer probed or misclassified.

## Tests

Added unit tests in `PerforceRepositoryTest` that do not require a running Perforce server:

- marker absent → not detected
- marker present → detected
- marker in an ancestor directory → detected
- `isRepositoryFor()` short-circuits (no `p4` invocation) without a marker

```
Tests run: 7, Failures: 0, Errors: 0, Skipped: 1
```

(The single skipped test is the pre-existing `@EnabledForRepository(PERFORCE)` integration test.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
+++ b/opengrok-indexer/src/main/java/org/opengrok/indexer/history/PerforceRepository.java
@@ -205,7 +205,39 @@ public class PerforceRepository extends Repository {
 
     @Override
     boolean isRepositoryFor(File file, CommandTimeoutType cmdType) {
+        // Unlike Git/Mercurial/etc., Perforce has no local marker directory, so
+        // historically OpenGrok probed *every* scanned directory by executing p4.
+        // That is expensive and, when a p4 client is configured via the
+        // environment, `p4 files`/`p4 dirs` can return depot paths for
+        // non-Perforce (e.g. Git) working trees, causing them to be misclassified
+        // as Perforce and their real history to be lost. Require local evidence of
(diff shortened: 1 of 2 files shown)


SebastienTolron:
Hi , 

This should fix https://github.com/oracle/opengrok/discussions/4572

Made with Claude Opus 4.8  ( High ).

SebastienTolron:
I  have signed the OCA

PR #4988, opened 2026-07-01
Limit line length in Search API responses

Bipul-programmer:


vladak:
I have several issues with this PR:
  - there is no description (what problem does this solve and why the solution was chosen)
    - further, it introduces implicit change without considering a tunable/parameter
  - there is a lot of line wrapping churn within the diffs, making this a bit more difficitult to comprehend than necessary
  - no OCA

PR #4991, opened 2026-07-07
Fix project classification for projects without repository info

TadayukiOkada:
Projects can be present in the project-repository map without having usable repository information.
The index page classification currently treats map membership as enough to classify a project as
a repository. This can move projects without repository history out of the Projects section.

This change classifies a project as a repository only when the corresponding RepositoryInfo list
is non-null and non-empty, and applies the same rule to grouped projects.


vladak (approved):
Looks good. It might make some sense to put the `hasRepositories()` method to the `Project` class given it already depends on the `RuntimeEnvironment` singleton, and use it in the various `populateGroups()` methods, however it is okay as it is currently.

vladak:
Thanks !

